PR #6, opened 2015-04-03
Fixed-join feature

thomasfla:
Addition of fixed join in the middle of a kinematic chain  (can be several following and/or branching)
- dimension of q is still minimal
- inertias are merged with the closest the movable parent
- placements of fixed join are record in model to send it to gepetto-viewer
- COM and RNEA have been tested locally - unit-test to come.

Now able to parse and display reemC
reemc_in_pinocchio


fvalenza:
Minor changes done and merged from another pull request.
Thanks for this pull request

PR #11, opened 2015-04-16
[Major][Unittest] Unifying unittest framework with boost_test_suite

fvalenza:
Complete switch from raw asserts to BOOST_TEST_SUITE as unittest framework.


fvalenza:
May be we can close this pull request. I will do a second one with - normally - better indentation in a few minuts

PR #13, opened 2015-04-17
[cmake] Add install for libpinocchio.so

flforget:


jcarpent:
This commit is not useful at all because libpinocchio is an empty lib. Shall we revert ?

PR #17, opened 2015-04-27
Add LUA parser

jcarpent:
Add the ability to load LUA files from Pinocchio. This method is more flexible than URDF format allowing the definition of new joint types. 
This code is largely inspired from RBDL (http://rbdl.bitbucket.org). Thanks to Martin Felis to provide this code.


martinfelis:
Looks interesting and I'm very happy to see LuaModels getting adopted in other tools.

In RBDL I am using the string "JointTypeEulerZYX" to denote the Euler ZYX joint. Similarly "JointTypeXYZ", "JointTypeYXZ", "JointTypeTranslationXYZ", and "JointTypeSpherical" to denote the other 3-DoF joints. It would be great if you would use the same strings to ensure compatibility. Personally I would also prefer "JointTypeFloatbase" instead of "JointTypeFreeFlyer".

Fixed joints simply are specified as an empty table.

On a minor note: "Lua" is not an acronym, therefore the spelling "LUA" does not make sense. The Lua community will thank you for spelling it correctly.

jcarpent:
Thanks Martin for your remarks and useful propositions. 

In order to be compatible with RBDL, I will choose the same convention as you did in RBDL. Concerning the fixed joint, I would prefer to mention it explicitly in the model. What do you think about this approach ?

For Lua, I will change the commit message asap. Sorry for my misreading.

Best,

martinfelis:
I would assume that the concept of "no joint specification, no movement" is easy to understand and also easy to implement using LuaTables++.

Another reason is the other use of LuaModels in MeshUp: if one were to add additional frames for visualization meshes only one would have to add joint specifications for every added visualization frame to stay compatible with RBDL.

What do you think?

jcarpent:
I do agree with your proposition. I did some minor modifications in the Lua parser (fb899b4) to fit your suggestions.
Maybe, in order to be compatible in both directions, could you add "JointTypeRX", JointTypePX", etc in RBDL ?
Thank you again Martin for helping us!

martinfelis:
I would rather have the special cases of 1-DoF joints detected from the spatial vector itself. However if you do insist I could imagine adding "JointTypeRevoluteX", "JointTypeRevoluteY", and "JointTypeRevoluteZ". Prismatic joints seldomly are used apart from the first movable body. For now I would leave it out but if there is demand for it I would add it.

PR #18, opened 2015-04-27
Refactoring index, and fixing python "se3ToXYZQuat "

fvalenza:
- Switch index depending on architecture ( std::size_t now).
- Cleaning multiple warning for Clang
- Travis now compiles with Clang and passes
- Added License 
- Fixing pythons utils method " se3ToXYZQuat"


jcarpent:
I think this pull request needs some commit simplifications before being accepted. Indeed commits d1d76e1 and 843ea35 have been written in opposite directions. 
We also need to choose between int type and long type according to the architecture (cf. Eigen).

fvalenza:
The switch of index has been done in three steps because the naive change of int to size_t involved errors. But i can squash these commits together.

jcarpent:
I really think it is better in this way.
In the same way, it is better to use a global Model::Index rather than size_t in 40f3fc6. And don't hesitate to use typedef to avoid long expression (e.g. Index instead of simple Model::Index).

PR #24, opened 2015-05-28
Topic/joint limits

fvalenza:
Added the joint limit parsing and storage.
Joint limits are "statically" stored in the joint model and "dynamically" computed with algorithm jointLimits.
There are a few warning "unused parameters"  due to the fact that for the moment limits do not change. But in the future we could have a limit for each joint depending on the configuration or other.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,6 +114,7 @@ SET(${PROJECT_NAME}_ALGORITHM_HEADERS
   algorithm/kinematics.hpp
   algorithm/center-of-mass.hpp
   algorithm/non-linear-effects.hpp
+  algorithm/joint-limits.hpp
   )
 
 SET(${PROJECT_NAME}_SIMULATION_HEADERS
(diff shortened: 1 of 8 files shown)


jcarpent:
Good job, but I think that the default value for joint limits in addBody could lead to unexpected behaviors because the size of the vector is not the same as the the joint configuration vector.

Moreover, the correction of indentation is only partial. I think we need to postpone this commit to deal with this problem globally.

fvalenza:
In fact in the method jointBase::set<effort/limits> there is a check on the size of the vector. If the vector is valid then we use it. If it is not valid, we implement a default value. ( we fill the limits with default values : +/- inf )
The point on this is because of retrocompatibility : to ensure retrocompatibility with previous calls of addBody, i have put the limits at the end of the function call BUT as there are optional arguments, these arguments must be also optional. That's why i decided to use this default value.

fvalenza:
should be better now

jcarpent:
Just a last point, you might change the type of return for accessors of joint limits in src/multibody/joint/joint-base.hpp by returning a const reference. Just a matter of performances.

PR #25, opened 2015-06-10
Implement explog functions using Eigen.

aelkhour:
PR status: Ready to merge

- [x] exp3
- [x] log3
- [x] exp6
- [x] log6
- [x] explog python bindings
- [x] unit tests


jcarpent:
2 spaces, but there are some incoherent indentations along the code.

jcarpent:
A just a small remark, I think your code could become more efficient by directly computing the exp3 manually, mainly in expo function. Indeed, there is a lot of redundant computations done either in exp3 through Eigen AngleAxis and in the computation of what you call V

aelkhour:
This PR does not include any unittests. It would be a good idea to add some, for both C++ and python.
How do you suggest we do that?

jcarpent:
There is unittest directory in the project where you can easily add some new tests.

PR #29, opened 2015-06-16
[UnitTest] Avoid the external call to BOOST_CHECK macros, to obtain e…

nmansard:
Modify the unit tests to ease the debug discussion with Antonio. 
Tell me if you prefer to merge this commit in the devel branch.


fvalenza:
commit was returning include errors. The changes have been put in a new branch.

PR #36, opened 2015-08-07
Build library only if sources are detected.

aelkhour:


jcarpent:
I do not think that it the proper way to handle such situation where the library is empty.
In this situation, you have to create a cmake variable which will be true as soon as the ${PROJECT_NAME}_SOURCES list is not empty.

Have you got some issues with an empty lib ?

aelkhour:
Compilation fails with qcc when there are no sources, which is why I added the test.

I'll try to propose a cleaner way if it's ok for you.

jcarpent:
Ok, I understand.

So, a cleaner solution may be to add a global variable which indicates if the compilation of the main library is needed or not (i.e. emptiness).

aelkhour:
New proposal.

jcarpent:
Ok. Just to avoid confusions, could you just change BUILD_LIBRARY into BUILD_MAIN_LIBRARY ?

aelkhour:
Done.

aelkhour:
This issue has been fixed in the devel branch. I use on a daily basis, so I would recommend using it, but you might want to double-check with @jcarpent.

fvalenza:
As stated by @aelkhour , there was a hotfix in the devel branhc. If i remember it's commit c8ff00

PR #38, opened 2015-08-26
Topic/crtp

fvalenza:
Spatial classes are now CRTPed.
Modifications have been done to JointRevolute to use CRTP for spatial classes. Once these modifications are validated, other joints will be switched.
The commit about " Renamed macro for spatial traits" could be also implemented in joint-base to avoid the multiple macro definitions for each compiler ( some allow the term "typename" even if we are not in templated class, some raise warning, some raise errors).


fvalenza:
pull request on wrong branch

olivier-stasse:
What is the impact in terms of performances ?

fvalenza:
@olivier-stasse : marginal -> gain of ~3% for rnea

PR #41, opened 2015-09-17
Topic/joint dense

fvalenza:
Added two classes JointModelDense and JointDataDense.
Their aim is to be as generic as possible and to possibly allow conversion to these types for any joints thanks to the methods toDense(). Cf in tests.
They are given as it with a possible use in C++. Unfortunately, the initial goal that was to bind this joint in python as a read-only joint is difficult. Instead, the binding of each joints can be done automatically and will allow the instantiation of such joints. Coming in a future pull request.


jcarpent:
A new pull request is coming soon with dynamic size of the dense joint.

Issue #47, opened 2015-10-05
Enforcement of One Definiton Rule


aelkhour:
While trying to use the C++ API, I encoutered several build errors related to multiple definitions.
This happens mainly:
- in the operators of the joint classes. As pinocchio is header only, all templated functions must be inline to avoid multiple definitions. I can make a PR for that if you agree.
- in model.hpp, with the definition of static const member se3::model::gravity981 in the header file. One solution would be to move the definiton to a .cpp file, forcing the client code to link against libpinocchio. Another solution would be to remove the definition and force the client code to define it when needed, which is not very nice. Other propositions are more than welcome.

jcarpent:
Concerning the model.hpp, it is an ongoing work and it will be released quite soon. The solution of the lib seems to be the best.
For the joint classes, it may really help us if you inline all the functions via a PR.

Thanks for reporting.

aelkhour:
Alright I'll make a PR adding the missing inlines.

jcarpent:
It seems that that this bug has been fixed recently with PR #54 and #49. Thanks to @aelkhour for this fixing.

aelkhour:
I think the issue should be kept opened until se3::model::gravity981 is moved is defined in a .cpp file.

jcarpent:
The gravity vector is now part of the pinocchio library (#56)

Issue #48, opened 2015-10-06
Travis build "not passing" randomly


fvalenza:
Since Pull Request on dynamic dense joints, travis has troubles to build AND check successfully. 

In fact there are two major problems that occur alternatively:
1/ When compiling the unittest urdf, there is an error in the flags ( there is a "urdf_CFLAGS-NOTFOUND " instead of "-I/usr/include/eigen3 -I/opt/openrobots/include" in the travis log).
I do not understand where this comes from* given that:
  - there are no changes in the travis build configuration
  - there are (almost) no changes in the CmakeLists : just one file added at a certain commit
  - urdfdom is found when generating the Makefile

2/ When error #1 does not occur, travis seem to "freeze" at the end of the build and then abort the procedure. The typical error message at the end of log is:
"  No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated "
If you go on https://travis-ci.org/stack-of-tasks/pinocchio/jobs/83854014 , you will see that it occured after all the test passed.

Note that it can happens only on gcc build, or clang build, or both.

* I don't know why the flags are -NOTFOUND but i suppose it happens in the unittest/CMakeLists.txt at line 35
GET_TARGET_PROPERTY(${target}_CFLAGS ${target} COMPILE_FLAGS) 
SET_PROPERTY(TARGET ${target} PROPERTY COMPILE_FLAGS "${${target}_CFLAGS} ${flag}")

Edit : When pushing to my own repository and running travis I have some "strange" behaviour. If you take a look at https://travis-ci.org/fvalenza/pinocchio/builds  the build #7 and #9.
The number 7 passed and number 9 did not whereas the only difference between them is the synchronization of travis and cmake submodule as suggested by @olivier-stasse

fvalenza:
Pull Request #51 should - partially* - solve error 2/
Commit #c8ff00 solve error 1/

* The random failures rate seems to decrease

jcarpent:
It seems that this bug has been fixed via recent PR.

PR #49, opened 2015-10-09
[C++][Bug] Enforce One Definition Rule in template instantiations.

aelkhour:
* Add inline specifier where needed.
  * This is done to avoid multiple definition when using the pinocchio
    library as a dependency.

  * This should **partly** solve #47


aelkhour:
Is the build fail related to #48 ?

PR #50, opened 2015-10-12
[C++][Bug] Avoid confusion between joint and body methods in se3::Model.

aelkhour:
* se3::Model::getBodyId was used to retrieve joint id in parsers.
  * Implement new methods to retrieve joint ids.
  * Fix body methods.

### DIFF
--- a/src/multibody/model.hpp
+++ b/src/multibody/model.hpp
@@ -1,5 +1,6 @@
 //
 // Copyright (c) 2015 CNRS
+// Copyright (c) 2015 Wandercraft
 //
 // This file is part of Pinocchio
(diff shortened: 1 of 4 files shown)


aelkhour:
@jcarpent @fvalenza Could someone please check whether the clang build error is a false positive?
Seems to me like clang crashed during build, but I don't know why.

fvalenza:
@aelkhour 
Hello Antonio, we have some problems with travis currently (see Issue #48 ) . This is due to the build environment running out of RAM when executing paralle builds. I will push a "fix" for it today. Otherwise the use of Travis will become compromised

aelkhour:
OK thanks.

PR #53, opened 2015-10-14
create model using python & (fix) exposure of joints

fvalenza:
- The type of joint used as a root joint can now be specified by the user in C++ and in Python

- Add the possibility to create a model from scratch in python. To do so, first create an empty model using BuildEmptyModel(), then add one by one your joints with addBody(..args..)

- Also fixing a missing functionality : you now have properly access to the joints of the model.

Exemple

```
import pinocchio as se3

from pinocchio.romeo_wrapper import RomeoWrapper

myModel = se3.Model.BuildEmptyModel()

myModel.addBody(0, se3.JointModelFreeFlyer(), se3.SE3.Random(), se3.Inertia.Random(),"j1","b1",True)
myModel.addBody(1, se3.JointModelRX(), se3.SE3.Random(), se3.Inertia.Random(),"j2","b2",True)

joint =  myModel.joints[2]
print joint.idx_q
print joint.nq
```

### DIFF
--- a/benchmark/timings.cpp
+++ b/benchmark/timings.cpp
@@ -43,7 +43,7 @@ int main(int argc, const char ** argv)
   else if( filename == "H2" )
     se3::buildModels::humanoid2d(model);
   else
-    model = se3::urdf::buildModel(filename,true);
+    model = se3::urdf::buildModel(filename,JointModelFreeFlyer());
   std::cout << "nq = " << model.nq << std::endl;
 
   se3::Data data(model);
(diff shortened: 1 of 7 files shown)


fvalenza:
@jcarpent new version of pull request.
parseTree and buildModel have been reworked to allow to be used with or without specifying a root joint. More user friendly.
Again: there are two methods in python to build a model because the macro BOOST_PYTHON_MEMBER_FUNCTION_OVERLOADS seems not be usable because of the different scopes

PR #57, opened 2015-10-19
[CMake][Bug] Fix use of lua sources cmake variable.

aelkhour:


jcarpent:
Thanks for this correction.

Issue #62, opened 2015-11-23
URDF parser: Spatial inertia set to Identity for inertial-less URDF links


aelkhour:
This is partally solved by #60, but the case **(non-fixed joint, inertial-less link)** still needs to be properly handled.

PR #66, opened 2015-12-10
Set default initialization of variables + expose spatial quantities methods + correct bug

jcarpent:
This PR provides different improvements:
1. Initialize with NaN all the kinematics and dynamics quantities in Debug mode (useful for python users) (solves #62) 
2. Expose some methods of spatial quantities for easier modifications
3. Correct a bug for the returning index

### DIFF
--- a/src/python/model.hpp
+++ b/src/python/model.hpp
@@ -201,7 +201,7 @@ namespace se3
       /// \param[in] v The value of to look for in the vector.
       ///
       /// \return The index of the matching element of the vector. If
-      ///         no element is found, return -1.
+      ///         no element is found, return the size of the vector.
       ///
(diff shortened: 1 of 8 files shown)


src/python/model.hpp
```diff
@@ -215,7 +215,7 @@ namespace se3
             return i;
           }
         }
-        return -1;
+        return x.size();
```
aelkhour: Fix documentation accordingly.

jcarpent:
@olivier-stasse Thank you for this remark. It is going to be included in the future version of this PR.
@aelkhour Thank you for pointing this missing element.
@fvalenza I did like in SE3 class for exposing class members.

Issue #69, opened 2015-12-17
Floating joint in URDF file


olivier-stasse:
Hi,
Why the urdf parser is not able to handle floating joint ?

They are inside the pinocchio package.
From http://wiki.ros.org/urdf/XML/joint it looks like this is still valid from the urdf specification viewpoint.

Is it because the case never occurred or is there a rational behind it ?

Best,
Olivier.

jcarpent:
Hi,

With Nicolas, we think that the URDF file must only describe the kinematics chain and not its link to the environment. The idea is then to precise the root joint via the buildModel function specifying the joint type we want.

Do you need the other feature ?

florent-lamiraux:
Some urdf files come with the degrees of freedom corresponding to the floating base.  It is necessary to handle those cases by specifying pinocchio not to add a floating joint.

olivier-stasse:
Like for the fixed joint, it is something inside the URDF specification. I do not believe that asserting without any option is a nice behavior. It is even more bizarre when the joint freeflyer is inside the library. I am fine if you want this to be the default behavior but I do believe that implementing the other policy and make it optional is possible without a strong impact on the code itself.

PR #70, opened 2015-12-17
Fix bug when parsing urdf file model by considering free floating joint

jcarpent:
Solve #69


olivier-stasse:
:+1:
Thanks a lot Justin !
A small remark if you want to close automatically issues you can add the reference to the topic number in the commit. For commit 3c2add6 in addition to your message an example would be:
Fixes issue #70 
But you probably knew already...

PR #72, opened 2016-01-11
Topic/geometry

fvalenza:
Added two new structures to deal with geometry :
- GeometryModel that contains the list of different geometry objects and their { name, id of the parent joint, relative placement wrt parent joint} and the number of collision objects.
Like the Kinematic Model , this should be constant data once it is created and initialized.
- GeometryData in which one can find the absolute position of each collision object, the number and the list of collision pairs ( the first index of the collision pair is always <= second ), and the  data structure for the storage of distance computations between two collision objects ( corresponding to a collision pair).
This structure provides methods to deal with collision pairs/objects such as:
    - bool collide (index, index) -> atomic computation given the index of two geometric objects and that returns whether or not they are in collision
    - bool isColliding () -> if at least one collision pair is effectively colliding, return true; else return false
    - computeDistance(index, index) -> atomic computation that returns the minimal distance and the nearest points for two given collision objects
    - computeDistances() -> compute the distance for all the collision pairs

Added the methods to build a couple <Model, GeometryModel> with or without a specified root joint.
In this method, a second loop is done over the urdf tree to handle and create geometry

Added tests : 
    - 1/ Is the model correctly tested
    - 2/ simple collision test between two boxes
    - 3/ comparisons - when hpp_model is found - of the collide/distance result between pinocchio and hpp_model with hrp2

Added an algorithm UpdateCollisionGeometry, templated by a boolean. It updates the values in GeometryData, recomputing the geometry of the kinematic chain if the template argument is true

Added benchmarks for atomic and non atomic methods.
    - Mean time to determine if two objects are colliding : 3.2 us
    - Mean time to compute Distance between two objects : [200;300] us ( big differences between romeo and hrp2)

In comparison with hpp-model: ( pinocchio time vs hpp time)
    - Compute the geometry of the robot, and test if the robot is in collision : 30.2 us vs 40.8 us
    - Compute the geometry and computeDistances for all collision pairs: 90 ms vs 100 ms


benchmark/timings-geometry.cpp
```diff
@@ -0,0 +1,220 @@
...
+  }
+  double computeDistancesTime = timer.toc(StackTicToc::US)/1000 - (update_col_time + geom_time);
+  std::cout << "Compute distance between two geometry objects (mean time) = \t" << computeDistancesTime / romeo_data_geom.nCollisionPairs
+            << " " << StackTicToc::unitName(StackTicToc::US) << " " << romeo_data_geom.nCollisionPairs << " col pairs" << std::endl;
+
+
+
+#ifdef WITH_HPP_MODEL_URDF
+
+  std::string filenameHRP2 = PINOCCHIO_SOURCE_DIR"/models/hrp2_14_reduced.urdf";
```
jcarpent: HRP2 won't work because nobody has access to the model.

benchmark/timings-geometry.cpp
```diff
@@ -0,0 +1,220 @@
...
+  std::string meshDirHRP2  = "/local/fvalenza/devel/install/share/";
```
jcarpent: You should not use personal path.

src/multibody/geometry.hxx
```diff
@@ -0,0 +1,237 @@
...
+  }
+  
+  inline const std::string& GeometryModel::getGeomName( Index index ) const
+  {
+    assert( index < (Index)collision_objects.size() );
+    return geom_names[index];
+  }
+
+  inline void GeometryModel::addInnerObject(Index joint, Index inner_object)
+  {
```
jcarpent: Maybe, it can be interesting to check if the inner_object is already included.

benchmark/timings-geometry.cpp
olivier-stasse: I would agree that the access to the model is restricted. It usually better to have separate repository and codes because of the psychological aspects. For some reasons a lot of people seems to overlook the preprocessing variables. It is probably much more related to the fact that it is cluttering the code.

src/spatial/se3.hpp
```diff
@@ -102,6 +103,17 @@ namespace se3
       Derived_t act   (const Derived_t& m2) const { return derived().act_impl(m2); }
       Derived_t actInv(const Derived_t& m2) const { return derived().actInv_impl(m2); }
 
+
+      bool operator == (Derived_t other) const
```
jcarpent: You should use const ref here

PR #99, opened 2016-02-09
Set default SINCOS for architecture with no sincos function

jcarpent:
This feature is needed for OS having no sincos function.


jcarpent:
Needed by netBSD and openNa

PR #103, opened 2016-02-15
Enhanced geometric parser, Geometry and use of Model::Index

fvalenza:
- Parse primitive shapes shuch as cylinders, spheres, boxes
- Split Index into JointIndex, GeomIndex, FrameIndex for clarity
- collision_results is now a vector of CollisionResult
- Fixed call to wrong functions ( due to change in API)
- Update doc


jcarpent:
For information, this PR won't compile until hpp-fcl package is not updated.

PR #107, opened 2016-02-18
Handle SRDF file parsing for activation of collision pairs

jcarpent:
A dedicated unit test is coming soon.

### DIFF
--- a/src/multibody/geometry.hpp
+++ b/src/multibody/geometry.hpp
@@ -350,6 +350,16 @@ namespace se3
     
     void desactivateCollisionPairs();
     void initializeListOfCollisionPairs();
+    
+    ///
+    /// \brief Active all possible collision pairs except those mentioned in the SRDF file.
+    ///        It throws if the SRDF file is incorrect.
+    ///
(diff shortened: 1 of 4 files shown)


nmansard:
The parsing code should be moved outside the class. I suggest to add a file src/multibody/parser/srdf.hpp.

jcarpent:
I think what you mentioned needs more justifications. Just because you do not build a Geometry Data from a SRDF file. You just remove some collision pairs between supporting links.

nmansard:
SRDF is only one convention to store a part of the model. A direct dependency on this specific convention should not be added directly in the model header. 

Decoupling the C++ representation of the model and the serialization to text/bin file is a good procedure, that we already applied for the URDF  parser. I simply suggest here to do the same for SRDF.

jcarpent:
I do agree with the first statement. 
The idea is really to avoid the creation of too many algorithms that are not strictly algorithms, just parsers that need to be called after the creation of the GeometryData entity for example. Maybe, we need to define a proper way to deal with such constraints:
- readable code
- also easy to use

Issue #93, opened 2016-02-04
Parser urdf


florent-lamiraux:
In its current version, the urdf parser is not able to understand url starting by "package://"
Moreover, it seems that srdf files are not parsed.

aelkhour:
@florent-lamiraux Do you think there's a solution for the `package://` issue without adding a dependency to ROS?

+1 for srdf.

+1 as well for [sdf](http://sdformat.org/spec), the official format for Gazebo simulator. It offers slightly more options than urdf, such as adding parallel spring-dampers on the joints. Although outside the current scope of pinocchio, the model does not need to be a kinematics tree, it can also be a DAG for parallel robots.

florent-lamiraux:
I think the best solution is to parse ROS_PACKAGE_PATH environment variable in C++ and to search in the corresponding directories for the files. The way ROS proceeds is not very clear and fails sometimes to find new directories because the right setup.sh has not been resourced after installation.

jcarpent:
The final idea is then:
- Add method to directly read a SRDF file to activate collision pairs.
- Add function to locate model and its data (URDF, SRDF, meshes, etc) by only the robot name and seek in $ROS_PACKAGE_PATH 

Have you other requests?

jcarpent:
Solved by #110.

PR #110, opened 2016-03-01
Topic/parse ros path

fvalenza:
Fixed a bug in Geometry parser when parsing twice the urdf tree not in a row.

Added a function to retrieve the list of package directories from the environment variable ROS_PACKAGE_PATH.

The idea is when searching for mesh, replace package:// by one of this path. If no one is found, we try to open a file that doesn't exist in loadPolyhedronFromResource() then it returns an exception ("Could not open ressource")

### DIFF
--- a/src/multibody/parser/urdf-with-geometry.hxx
+++ b/src/multibody/parser/urdf-with-geometry.hxx
@@ -42,7 +42,7 @@ namespace se3
         boost::shared_ptr < ::urdf::Mesh> collisionGeometry = boost::dynamic_pointer_cast< ::urdf::Mesh> (collision->geometry);
         std::string collisionFilename = collisionGeometry->filename;
 
-        std::string full_path = fromURDFMeshPathToAbsolutePathPack(collisionFilename, package_dirs);
+        std::string full_path = convertURDFMeshPathToAbsolutePath(collisionFilename, package_dirs);
 
         ::urdf::Vector3 scale = collisionGeometry->scale;
 
--- a/src/multibody/parser/urdf-with-geometry.hxx
+++ b/src/multibody/parser/urdf-with-geometry.hxx
@@ -133,44 +133,24 @@ namespace se3
 
     inline GeometryModel buildGeom(const Model & model,
                                   const std::string & filename,
-                                  std::vector<std::string> & package_dirs)
+                                  const std::vector<std::string> & package_dirs)
     {
       GeometryModel model_geom(model);
 
-      try
-      {
-        appendRosPackagePaths(package_dirs);
-      }
-      catch (std::runtime_error & e)
-      {
-        std::cout << "ROS_PACKAGE_PATH environment variable is empty. Meshes may not be found when looking for geometry" << std::endl;
-        assert(!package_dirs.empty() && "You did not specify any package directory. Geometric parsing will crash" );
-      }
-      ::urdf::ModelInterfacePtr urdfTree = ::urdf::parseURDFFile (filename);
-      parseTreeForGeom(urdfTree->getRoot(), model, model_geom, package_dirs);
-      return model_geom;
-    }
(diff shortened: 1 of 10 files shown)


fvalenza:
The name of the method fromURDFMeshPathToAbsolutePathPack may be too long

jcarpent:
I do agree, the name seems too long and not explicit.
In addition, you should raised some hint or throws something when the ROS_PACKAGE_PATH is empty.

And you have not updated the robotWrapper class in Python.

jcarpent:
Solve #93

src/multibody/parser/urdf-with-geometry.hxx
```diff
@@ -101,31 +100,38 @@ namespace se3
     inline void parseTreeForGeom(::urdf::LinkConstPtr link,
                                  const Model & model,
                                  GeometryModel & geom_model,
-                                 const std::string & meshRootDir,
+                                 const std::vector<std::string> & package_dirs,
                                  const bool rootJointAdded) throw (std::invalid_argument)
     {
 
       // start with first link that is not empty
-      if(link->collision)
```
jcarpent: I think you should not remove this line, just because if the current link has no collision geometries, you just skip it.

src/python/robot_wrapper.py
```diff
@@ -22,32 +22,28 @@
...
-
-          self.data = self.model.createData()
-          self.geometry_data = se3.GeometryData(self.data, self.geometry_model)
-        else:
-          if(root_joint is None):
+    def __init__(self, filename, package_dirs = None, root_joint = None):
+        if(root_joint is None):
               self.model = se3.buildModelFromUrdf(filename)
-          else:
+              self.root_added = False
```
jcarpent: I don't think that is a good idea to keep the information root_added.

src/multibody/parser/urdf-with-geometry.hxx
```diff
@@ -133,34 +139,37 @@ namespace se3
...
     }
 
-    GeometryModel buildGeom(const Model & model,
-                            const std::string & filename,
-                            const std::string & meshRootDir)
+
+    inline GeometryModel buildGeom(const Model & model,
+                                  const std::string & filename,
+                                  const std::vector<std::string> & package_dirs,
+                                  const bool root_joint)
```
jcarpent: Normally, you do not need to give root_joint information when you build geom model.

fvalenza: How do you know then if you must search for a "root_joint" or not ? as there are two buildModel procedures

src/multibody/parser/urdf-with-geometry.hxx
fvalenza: It was replaced by link->inertial. I thought it had better meaning for the kinematic tree but may be i'm wrong

src/python/robot_wrapper.py
fvalenza: It was for better clarity of the next lines. Instead of testing once again if root_joint is None to build the geometric model

fvalenza:
Also when the ROS_PACKAGE_PATH is empty there is no real check : in fact the absolute path returned will be empty and then an exception raises when trying to open an empty mesh file in loadPolyhedronFromResource.

src/multibody/parser/urdf-with-geometry.hxx
jcarpent: You should first check if the link has a collision information or not. Furthermore, if the link has no inertial information but some collision objects, we must keep this information.

src/multibody/parser/urdf-with-geometry.hxx
jcarpent: You don't need to give this information, as your are just adding meshes to supporting body so supporting joint.

jcarpent:
:+1:

PR #133, opened 2016-03-17
[Python] Some propositions about Format & Tests

nim65s:
Here are three commits. Each one could have been one PR, but this is mainly for discussions.

The first one is about reformatting most python file to comply a little more with some main indications stated in [pep8](https://www.python.org/dev/peps/pep-0008/)

The second makes python tests work again

The third tells to Travis to run those tests.

I will probably change a bit how those tests are implemented, and then work on coverage, if you find this useful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
-language: cpp
+language: python
+python:
+- "2.7"
 sudo: required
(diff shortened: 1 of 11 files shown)


jcarpent:
:+1:

Issue #104, opened 2016-02-16
Add integrate and differentiate methods in JointModel


jcarpent:

jcarpent:
Add also random method for each joint

jmirabel:
For the `differentiate` method of SO3 joints, a convention must be chosen. This convention relates to what is being modelled: the rotation group or the unit quaternions group. Each rotation is mapped to 2 quaternions.
Thus, the "difference" between two quaternions `a` and `b` can be:
- either the angle axis between `a` and `b` (the quaternion representation)
- or the angle axis that has the shortest angle, between `a` and `b` or between `a` and `-b` (the rotation representation)

In HPP, we use the rotation representation. This has several implications. The one I can think of now are:
-  Equality check: `a` and `-a` must be considered as the same rotation for the whole software => two configurations `q1 and `q2` can be equal even if  `q1.isApprox (q2)` is false.
- Difference and distance: use the angle axis that has the shortest angle

See https://github.com/jmirabel/hpp-model/blob/devel/src/joint-configuration.cc#L243

jcarpent:
The second proposition is the rational one. Because, we are working on base orientation, which can be handled by many representations:
- rotation matrix (unique representation)
- quaternion, with a non-infectivity (a and -a leads to the same orientation)
- Euler angles (with a lot of drawbacks - singularities)

fvalenza:
Solved with #141

PR #134, opened 2016-03-18
Topic/hpp joint configuration

fvalenza:
Added methods in joint models to integrate, interpolate, differentiate , compute the distance between two joint configurations. Added corresponding algorithm to perform theses actions on a Model and Data and its python bindings & unittest.


src/math/sincos.hpp
```diff
@@ -29,4 +31,24 @@
   #define SINCOS(a,sa,ca) (*sa) = std::sin(a); (*ca) = std::cos(a)
 #endif
 
+/// Compute quaternion and angle from a SO(3) joint configuration
```
jcarpent: It is not the good doc

src/math/sincos.hpp
```diff
@@ -29,4 +31,24 @@
...
+///
+/// \param q1, q2, robot configurations
+/// \param index index of joint configuration in robot configuration vector
+/// \param unit quaternion corresponding to both joint configuration
+/// \return angle between both joint configuration
+template <typename D>
+static double angleBetweenQuaternions(const Eigen::MatrixBase<D> & q1,
```
jcarpent: why static.
You should return D::Scalar and use it as scalar type all along the function

jcarpent: Why you don't use Eigen::Quaternion ?

src/multibody/joint/joint-base.hpp
```diff
@@ -243,6 +251,21 @@ namespace se3
                   const bool update_I = false) const
     { return static_cast<const JointModel*>(this)->calc_aba(data, I, update_I); }
 
+    const ConfigVector_t integrate(const Eigen::VectorXd & q,const Eigen::VectorXd & v) const
```
jcarpent: Do you want to return a const ref or just a copy of the result?

src/multibody/joint/joint-planar.hpp
```diff
@@ -366,6 +369,41 @@ namespace se3
...
+    const ConfigVector_t interpolate_impl(const Eigen::VectorXd & q1,const Eigen::VectorXd & q2, double u) const
+    { 
+      Eigen::VectorXd::ConstFixedSegmentReturnType<NQ>::Type & q_1 = q1.segment<NQ> (idx_q ());
+      Eigen::VectorXd::ConstFixedSegmentReturnType<NQ>::Type & q_2 = q2.segment<NQ> (idx_q ());
+
+      return ConfigVector_t((1-u) * q_1 + u * q_2);
+    }
+
+    const ConfigVector_t random_impl() const
+    { return ConfigVector_t(); }
```
jcarpent: This solution is not random()

src/multibody/joint/joint-prismatic-unaligned.hpp
```diff
@@ -350,6 +353,49 @@ namespace se3
...
+    { 
+      const Scalar_t & q_1 = q1[idx_q()];
+      const Scalar_t & q_2 = q2[idx_q()];
+
+      ConfigVector_t result;
+      result << ((1-u) * q_1 + u * q_2);
+      return result; 
+    }
+
+    const ConfigVector_t random_impl() const
```
jcarpent: Same here

src/math/quaternion.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+# include <Eigen/Geometry>
+
+
+/// Compute angle between two quaternions
+///
+/// \param q1, q2, unit quaternions
+/// \return angle between both quaternions
+template <typename D>
+static typename D::Scalar angleBetweenQuaternions(const Eigen::MatrixBase<D> & q1,
```
jcarpent: why static ?

src/math/sincos.hpp
```diff
@@ -21,6 +21,8 @@
 
 #include <cmath>
 
+# include <Eigen/Geometry>
```
jcarpent: This line should be removed

src/multibody/joint/joint-free-flyer.hpp
```diff
@@ -226,6 +233,94 @@ namespace se3
...
+      Eigen::VectorXd::ConstFixedSegmentReturnType<NQ>::Type & q = qs.segment<NQ> (idx_q ());
+      Eigen::VectorXd::ConstFixedSegmentReturnType<NV>::Type & q_dot = vs.segment<NV> (idx_v ());
+
+      ConfigVector_t result;
+      // Translational part
+      result.head<3>() =  (q.head<3>() + q_dot.head<3>());
+
+      // Quaternion part
+      Motion_t::Quaternion_t quat(q.tail<4>());
+      Motion_t::Vector3 omega(q_dot.tail<3>());
```
jcarpent: This quantity can be omitted.

src/multibody/joint/joint-free-flyer.hpp
```diff
@@ -226,6 +233,94 @@ namespace se3
...
+      Eigen::VectorXd::ConstFixedSegmentReturnType<NQ>::Type & q_1 = q1.segment<NQ> (idx_q ());
+      Eigen::VectorXd::ConstFixedSegmentReturnType<NQ>::Type & q_2 = q2.segment<NQ> (idx_q ());
+
+      ConfigVector_t result;
+      // Translational part
+      result.head<3>() << ((1-u)*q_1.head<3>() + u * q_2.head<3>());
+
+      //Quaternion part
+      double theta = angleBetweenQuaternions (q_1.segment<4>(3), q_2.segment<4>(3));
+      Motion_t::Vector4 quaternion_result;
```
jcarpent: You can directly work with result vector.

src/multibody/joint/joint-free-flyer.hpp
```diff
@@ -226,6 +233,94 @@ namespace se3
...
+      }
+
+      result.tail<4>() << quaternion_result;
+
+      return result; 
+    }
+
+    const ConfigVector_t random_impl() const
+    { 
+      ConfigVector_t q = ConfigVector_t::Random();
```
jcarpent: You should use the constructor instead of opeartor= for performance reasons.

src/multibody/joint/joint-free-flyer.hpp
```diff
@@ -226,6 +233,94 @@ namespace se3
...
+      q.segment<4>(3) /= q.segment<4>(3).norm();
```
jcarpent: You can just use normalize()

src/python/algorithms.hpp
```diff
@@ -465,6 +508,37 @@ namespace se3
...
+                         "Configuration q2 (size Model::nq)",
+                         "Velocity result (size Model::nv)"),
+                "Difference between two configuration, ie. the constant derivative that must be integrated during unit time"
+                "to go from q2 to q1");
+        bp::def("distance",distance_proxy,
+                bp::args("Model","Data",
+                         "Configuration q1 (size Model::nq)",
+                         "Configuration q2 (size Model::nq)",
+                         "Velocity result (size Model::nv)"),
+                "Distance between two configuration ");
```
jcarpent: configurations

unittest/joint-configurations.cpp
```diff
@@ -0,0 +1,415 @@
...
+//
+// Pinocchio is distributed in the hope that it will be
+// useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+// of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Lesser Public License for more details. You should have
+// received a copy of the GNU Lesser General Public License along with
+// Pinocchio If not, see
+// <http://www.gnu.org/licenses/>.
+
+/*
```
jcarpent: To be removed

unittest/joint-configurations.cpp
```diff
@@ -0,0 +1,415 @@
+//
+// Copyright (c) 2015 CNRS
```
jcarpent: we are in 2016

src/algorithm/joint-configuration.hpp
```diff
@@ -0,0 +1,334 @@
+//
+// Copyright (c) 2015 CNRS
```
jcarpent: 2016

src/algorithm/joint-configuration.hpp
```diff
@@ -0,0 +1,334 @@
...
+                             const Eigen::VectorXd & q,
+                             const Eigen::VectorXd & v,
+                             Eigen::VectorXd & result);
+
+
+  /**
+   * @brief      Interpolate the model between two configurations
+   *
+   * @param[in]  model   Model to be interpolated
+   * @param      data    Corresponding Data to the Model
```
jcarpent: the param is in too

unittest/joint-configurations.cpp
```diff
@@ -46,6 +46,34 @@
 #include <boost/test/unit_test.hpp>
 
 
+Eigen::Quaterniond slerp(const Eigen::Quaterniond & q1, const Eigen::Quaterniond & q2,double t)
```
jmirabel: I realized Eigen has an implementation of SLERP. We should maybe use `return q1.slerp (t, q2)`, no ?

jmirabel: And if you do not use Eigen, could you implement a SLERP interpolating the shortest path ?
`if(q1.dot(q2) < 0) mult2 = - mult2`

src/multibody/joint/joint-free-flyer.hpp
jmirabel: Are you sure of this ? The quaternion copy constructor is implemented as
```
  template<typename Derived>
  explicit inline Quaternion(const MatrixBase<Derived>& other) { *this = other; }
```

fvalenza:
@jcarpent Do not know what is happening with Travis.
See end of log at https://travis-ci.org/fvalenza/pinocchio/builds/117493884. Build stalls after all unittest passed. At the same time, on the stack-of-tasks travis repo, build stalls in the middle of unit testings..

src/math/quaternion.hpp
```diff
@@ -28,16 +28,14 @@
...
 
-  Eigen::Quaternion<typename D::Scalar> p1 (q1);
-  Eigen::Quaternion<typename D::Scalar> p2 (q2);
 
-  Eigen::Quaternion<typename D::Scalar> p (p1*p2.conjugate());
-  Eigen::AngleAxis<typename D::Scalar> angle_axis(p);
+  Eigen::Quaternion<typename D::Scalar> q (q1*q2.conjugate());
+  Eigen::AngleAxis<typename D::Scalar> angle_axis(q);
 
   return (angle_axis.angle() * angle_axis.axis()).norm();
```
jcarpent: This implementation is not efficient at all. Indeed, angle_axis.axis() is already normalised. So the result is just the absolute value of the angle_axis.angle().

But this does not correspond to the minimal distance between two quaternions, just because the solution is between [0, 2pi]. An efficient implementation is provided by HPP. Please refer to this solution.

jmirabel: I updated `hpp-model` (branch devel) and added unit-tests to fix the distance function.

unittest/joint-configurations.cpp
```diff
@@ -42,6 +42,27 @@
 #include <vector>
 
 
+bool quaternion_are_same_rotation(const Eigen::VectorXd & quat1, const Eigen::VectorXd & quat2)
```
jcarpent: To be consistent, please use Eigen::Quaternion as input type.

src/multibody/joint/joint-spherical.hpp
```diff
@@ -346,8 +346,9 @@ namespace se3
 
     TangentVector_t difference_impl(const Eigen::VectorXd & q1,const Eigen::VectorXd & q2) const
     { 
-      // Compute rotation vector between q2 and q1.
-      Motion_t::Quaternion_t p1 (q1.segment<NQ>(idx_q()));
+      // Compute relative rotation between q2 and q1.
+      const int invertor = (q1.segment<4>(3).dot(q2.segment<4>(3)) < 0 ) ? -1: 1 ;
```
jcarpent: Are you sure that you are working with the right segment (cf idx_q) ?

fvalenza: corrected. was an error copying the line from another file

unittest/joint-configurations.cpp
fvalenza: Done. Eigen doesn't provide an operator == that's why I use .coeffs()

unittest/joint-configurations.cpp
```diff
@@ -0,0 +1,399 @@
...
+              q1[11] - q2[11],
+              q1[12] - q2[12],
+              q1[13] - q2[13],
+              q1[14] - q2[14],
+              (q1.segment<3>(15) - q2.segment<3>(15)).norm(),
+              (q1.segment<3>(18) - q2.segment<3>(18)).norm(),
+              (q1.segment<3>(21) - q2.segment<3>(21)).norm();
+  distance(model, data,q1,q2,result);
+
+  assert(result.isApprox(expected) && "Distance between two configs of full model - wrong results");
```
jcarpent: Do not use assert in the test code. Use Boost test function instead.

unittest/joint-configurations.cpp
```diff
@@ -0,0 +1,399 @@
...
+  Eigen::Vector3d quat_spherical_diff( angle_axis_spherical.angle() * angle_axis_spherical.axis());
+
+  expected.head<3>() = q1.head<3>() - q2.head<3>();
+  expected[3] = quat_ff__diff[0];expected[4] = quat_ff__diff[1]; expected[5] = quat_ff__diff[2]; 
+  expected[6] = quat_spherical_diff[0];expected[7] = quat_spherical_diff[1]; expected[8] = quat_spherical_diff[2];
+  expected.tail<13>() = q1.tail<13>() - q2.tail<13>();
+
+  differentiate(model, data,q1,q2,result);
+
+  assert(result.isApprox(expected) && "Differentiation of full model - wrong results");
```
jcarpent: Same here.

src/math/quaternion.hpp
```diff
@@ -35,4 +35,11 @@ D angleBetweenQuaternions(const Eigen::Quaternion< D> & q1,
   return theta;
 }
 
+template <typename D>
```
jcarpent: I think it is not a good idea to put this function in such a file. You can let it in unit test.

fvalenza: Why ? As stated by @jmirabel in
https://github.com/stack-of-tasks/pinocchio/issues/104, it may be usefull to provide this kind of equality check

jcarpent: My concern is first with the naming which is not coherent with the rest of the API. My second concern is about the fact that you are implementing a function and it has rather to be method of quaternion class.

src/algorithm/joint-configuration.hpp
```diff
@@ -0,0 +1,334 @@
...
+   * @brief      Distance between two configuration vectors
+   *
+   * @param[in]  model      Model we want to compute the distance
+   * @param[in]  data       Corresponding Data to the Model
+   * @param[in]  q1         Configuration 1 (size model.nq)
+   * @param[in]  q2         Configuration 2 (size model.nq)
+   * @param      distances  The corresponding distances for each joint (size model.nbody-1 = number of joints)
+   */
+  inline void distance(const Model & model,
+                            Data & data,
```
jcarpent: Remove data. You do not need it here.

src/algorithm/joint-configuration.hpp
```diff
@@ -0,0 +1,334 @@
...
+  inline void interpolate(const Model & model,
+                               Data & data,
+                               const Eigen::VectorXd & q1,
+                               const Eigen::VectorXd & q2,
+                               const double u,
+                               Eigen::VectorXd & result);
+
+
+  /**
+   * @brief      the constant derivative that must be integrated during unit time to go from q2 to q1
```
jcarpent: Please change the API in such way that differentiate goes from q1 to q2 during one unit time to be consistent with the rest of the API.
Please use also q0 instead of q1 and q1 instead of q2.

src/math/fwd.hpp
```diff
@@ -0,0 +1,30 @@
...
+#ifndef __math_fwd_hpp__
+#define __math_fwd_hpp__
+
+#include <cmath>
+
+# include <Eigen/Geometry>
+#include <boost/math/constants/constants.hpp>
+
+
+static const double PI = boost::math::constants::pi<double>();
```
jcarpent: Remove static. AGAIN

src/algorithm/joint-configuration.hpp
```diff
@@ -0,0 +1,334 @@
...
+#define __se3_joint_configuration_hpp__
+
+#include "pinocchio/multibody/visitor.hpp"
+#include "pinocchio/multibody/model.hpp"
+  
+namespace se3
+{
+
+  /**
+   * @brief      Integrate the model for a constant derivative during unit time
```
jcarpent: Yo do not integrate the model. You integrate a configuration with a constant velocity during **one** unit time.

src/math/quaternion.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+/// Compute minimal angle between q1 and q2 or between q1 and -q2
+///
+/// \param q1, q2, unit quaternions
+/// \return angle between both quaternions
+template <typename D>
+D angleBetweenQuaternions(const Eigen::Quaternion< D> & q1,
+                          const Eigen::Quaternion< D> & q2)
+{
+  double innerprod = q1.dot(q2);
```
jcarpent: Instead of double, use D::Scalar

src/algorithm/joint-configuration.hpp
jmirabel: In HPP:
- `differentiate(q1,q2) := q2 - q1` by definition,
- `integrate(q1, differentiate(q1,q2)) = q2`, where `=` takes into account the double representation of the rotations.

fvalenza:
Closing. See #141

Issue #115, opened 2016-03-02
Model joint limits


jcarpent:
Currently, joint limits are implemented as an algorithm and we consider that joint limits may depend on the current configuration. 
But this option is not satisfying. 
One option is to put joint limit quantities in model and to read them when parsing.

What do you think?

olivier-stasse:
Can you develop the rational behind your assumption that $q_min$ is the function of ${\bf q}$ ?

aelkhour:
@olivier-stasse I don't know any example for position limits, but joints which are coupled through a transmission (e.g. a differential train) can have velocity and torque limits which are dependent of the position.

@jcarpent Could you explain more why this option is not satisfying?

jcarpent:
As mentioned by @aelkhour , the velocity and torque limits can depend on the current configuration. But it is not the case for the position limits.

I find this solution a little bit painful. 
But I would suggest to make velocity and torque limits dependant of the configuration, and to keep the other constant regarding the configuration.

aelkhour:
But how can you be sure that position limits are constants for all types of mechanisms?

nmansard:
I am very against the current implementation, that I think complex for 
nothing. Indeed, we have yet no strategy to have limits depending on the 
configuration, and no clear idea about how it may be useful. Taking for 
example the case study reported by Antonio: differential drive: this is 
not a matter of joints, but a matter of actuation (and we have yet not 
actuator models in Pinocchio).

I deeply believe that it is a bad procedure to imagine an API before 
understanding how the API should be used. Here we have no clear idea of 
how it should be implemented: then implement it simply. We will do 
something more complex when we will really need it (at very likely, it 
will be done differently to the current implementation).

More practically. The limits should be implemented in the model: the 
joint limits are element of the configuration space (i.e. represented by 
vector of size nq). The velocity and torque limits are element of the 
joint velocity space (~and its dual, ie size nv). When the joint is 
unbounded, the corresponding value is inf. No other implementation is 
adding value with respect to this simple solution.

Now, more complex limits might be desirable, function of the 
configuration or coupling components of the configuration. But then, 
they are no "bounds", i.e. of the form x <= xsup, but rather of the form 
Ax <= asup, or something similar. If this solution is necessary, we 
should think to a way of storing them (but not in the joint limits).

N

olivier-stasse:
I totally agree with the comment of Nicolas.

aelkhour:
Agreed @nmansard

[...]

PR #135, opened 2016-03-21
Topic/joint limits

fvalenza:
Moved the joint limits in Model. Updated accordingly algorithms and python binding.
Note that within the jointmodel's random() methods, there is no strategy to ensure quaternion is within a determined position limit.  How can we do ? 
This pull request follow #134 . Only 3 last commits are effectives


src/multibody/joint/joint-planar.hpp
```diff
@@ -381,9 +381,17 @@ namespace se3
...
     }
 
-    ConfigVector_t random_impl() const
+    ConfigVector_t random_impl(const ConfigVector_t & lower_pos_limit, const ConfigVector_t & upper_pos_limit ) const
     { 
-      return ConfigVector_t::Random();
+      ConfigVector_t result(ConfigVector_t::Random());
+      for (long i = 0; i < result.size(); i++)
+      {
+        if( *(result.data()+i) < lower_pos_limit[i])
```
jcarpent: Why do you use here the pointer arithmetic?

jcarpent:
I think there might be a bug if the lower and upper bounds are equal. Mainly in the context of the Free Flyer where there is no mentioned bounds in the URDF.

jcarpent:
Related to #115.

src/multibody/joint/joint-planar.hpp
fvalenza: It is faster than using result[i]

src/algorithm/joint-configuration.hpp
```diff
@@ -303,16 +315,14 @@ namespace se3
     template<typename JointModel>
     static void algo(const se3::JointModelBase<JointModel> & jmodel,
                      se3::JointDataBase<typename JointModel::JointData> &,
-                     const se3::Model & model,
+                     const se3::Model & ,
```
jcarpent: Can you remove not useful args, like model, data, jdata, etc

src/algorithm/joint-configuration.hpp
```diff
@@ -100,6 +100,18 @@ namespace se3
...
+  /**
+   * @brief      Generate a random configuration vector.
+   *
+   * @param[in]  model   Model we want to generate a configuration vector of
+   * @param[in]  data    Corresponding Data to the Model
+   * @param      config  The resulted configuration vector (size model.nq)
+   */
+  inline void uniformlySample(const Model & model,
+                                  Data & data,
+                                  Eigen::VectorXd & config);
```
jcarpent: Please use q instead of config, to be coherent with the rest of the API

src/algorithm/joint-configuration.hpp
```diff
@@ -330,6 +340,46 @@ namespace se3
...
+      using namespace se3;
+      
+      jmodel.jointConfigSelector(config) = jmodel.uniformlySample(jmodel.jointConfigSelector(model.lowerPositionLimit),
+                                                                  jmodel.jointConfigSelector(model.upperPositionLimit)
+                                                                  );
+    }
+
+  };
+
+  inline void
```
jcarpent: Can you return a ref to the result?

src/algorithm/joint-configuration.hpp
```diff
@@ -330,6 +340,46 @@ namespace se3
...
+    JOINT_VISITOR_INIT(UniformlySample);
+
+    template<typename JointModel>
+    static void algo(const se3::JointModelBase<JointModel> & jmodel,
+                     se3::JointDataBase<typename JointModel::JointData> &,
+                     const se3::Model & model,
+                     se3::Data &,
+                     Eigen::VectorXd & config) 
+    {
+      using namespace Eigen;
```
jcarpent: Remove two those lines

src/multibody/joint/joint-free-flyer.hpp
```diff
@@ -275,13 +278,42 @@ namespace se3
...
       ConfigVector_t q(ConfigVector_t::Random());
       q.segment<4>(3).normalize();// /= q.segment<4>(3).norm();
       return q;
     } 
 
+    ConfigVector_t uniformlySample_impl(const ConfigVector_t & lower_pos_limit, const ConfigVector_t & upper_pos_limit ) const
+    {
+      ConfigVector_t result;
+      // Translational Part
+      for (int i = 0; i < 3; ++i)
```
jcarpent: Please use correct index type for the vector.

src/multibody/joint/joint-free-flyer.hpp
```diff
@@ -275,13 +278,42 @@ namespace se3
...
 
-    ConfigVector_t random_impl(const ConfigVector_t & , const ConfigVector_t & ) const
+    ConfigVector_t random_impl() const
     { 
       ConfigVector_t q(ConfigVector_t::Random());
       q.segment<4>(3).normalize();// /= q.segment<4>(3).norm();
       return q;
     } 
 
+    ConfigVector_t uniformlySample_impl(const ConfigVector_t & lower_pos_limit, const ConfigVector_t & upper_pos_limit ) const
```
jcarpent: If you throw something, please indicate the throw type like std::invalid_argument

src/multibody/joint/joint-free-flyer.hpp
```diff
@@ -275,13 +278,42 @@ namespace se3
...
+        }
+        result[i] = lower_pos_limit[i] + ( upper_pos_limit[i] - lower_pos_limit[i]) * rand()/RAND_MAX;
+      }
+
+      // Quaternion Part
+      double u1 = (double)rand() / RAND_MAX;
+      double u2 = (double)rand() / RAND_MAX;
+      double u3 = (double)rand() / RAND_MAX;
+      
+      result.segment<4>(3) << sqrt (1-u1)*sin(2*PI*u2),
```
jcarpent: Use temporary for sqrt(1-u1),  to avoid multiple recalculations.

src/multibody/joint/joint-free-flyer.hpp
fvalenza: In the function signature or in the doc ?
http://www.gotw.ca/publications/mill22.htm

src/algorithm/joint-configuration.hpp
fvalenza: jmodel and jdata cannot be removed as it is needed by JOINT_VISITOR_INIT.
model and data will be removed when not useful

src/multibody/joint/joint-free-flyer.hpp
jcarpent: In the signature of course.

src/algorithm/joint-configuration.hpp
jcarpent: This not a problem, you just need to create something like JOINT_MODEL_VISITOR_INIT

fvalenza:
Closing. See #141

PR #136, opened 2016-03-22
[Python] Tests with unittests & coveralls

nim65s:
Python tests has been rewritten with the standard unittest module instead of a list of asserts in a script.
Coverage is then checked, and submitted to coveralls.io together with C++ coverage informations.

As a result, more test have been implemented to improve coverage.


jcarpent:
:+1:

PR #141, opened 2016-03-30
Devel

fvalenza:
New Pull Request for joint limits and joint-configuration algorithms.
Reworked the tree of commits to have something readable. Merged the two pull requests in one as there are dependencies between them.
There is still the procedure in quaternion.hpp


src/multibody/joint/joint-base.hpp
```diff
@@ -243,15 +249,80 @@ namespace se3
...
+     *
+     *
+     * @return     The joint configuration
+     */
+    ConfigVector_t uniformlySample(const ConfigVector_t & lower_pos_limit, const ConfigVector_t & upper_pos_limit) const
+    { return derived().uniformlySample_impl(lower_pos_limit, upper_pos_limit); } 
+
+    
+    /**
+     * @brief      the constant derivative that must be integrated during one unit time to go from q0 to q1
```
jcarpent: Can you **change constant** derivative into **tangent vector**? This sounds much more correct from a mathematical view point.

PR #142, opened 2016-04-01
[Python] fix ViewerConfigurationToXYZQUAT

nim65s:
Not sure if this PR is really useful, but it looks like we had some broken code…


fvalenza:
It is, it is. In addition, it could be nice to change the notation a b c d into something more explicit (like q_x, q_y, q_z, q_w may be. What do you think @jcarpent ?)

jcarpent:
Good. It is good to keep the convention x,y,z,w to denote the quaternion components.

jcarpent:
Just before merging, can you change a,b,c,d into x,y,z,w for labelling of quaternion coefficients ?

nim65s:
Done :)

jcarpent:
:+1:

Issue #106, opened 2016-02-17
Add unit tests on Planar Joint


jcarpent:
Planar Joint is not tested at all and it will be mandatory for the integration of Pinocchio in HPP.
For kinematics checking, one idea is to create a serial chain with same one DOF joints than in the planar one.

For dynamics computation, we can check against RBDL.

jcarpent:
Solved by #148.

Issue #89, opened 2016-02-03
Coverage - Needs of more unit tests


olivier-stasse:
Dear all,
To have a better view of the code state so that we can improve the release process,
I made the link to coveralls. I apologize for the rough process. 
According https://coveralls.io/builds/4952648 they are three files which are not tested at all:
joint_planar.hpp joint_dense.hpp constraint.hpp.
It would be nice to have unit test related to those files.

jcarpent:
Thanks Olivier for this great job!
Concerning those files, they are currently not used at least for joint_dense.hpp and constraint.hpp.
I will do my best to integrate some unit test around joint_planar.hpp asap.

jcarpent:
And also, all the python bindings are not tested at all. But it won't appear in the coverage test, because they are not compiled on travis.

PR #148, opened 2016-04-08
Devel

fvalenza:
Essentially some fixes.
+ added jointPlanar unittest. If the structure of this joint testing is ok, one can change the other joints unittest to uniformize the whole file joints.cpp
The last commit is a proposition to acces the informations of a joint as a dense one ( instead of converting the joint to a DenseJoint and then read the wanted info).

### DIFF
--- a/src/multibody/force-set.hpp
+++ b/src/multibody/force-set.hpp
@@ -123,7 +123,7 @@ namespace se3
       template <typename D>
       Block& operator= (const Eigen::MatrixBase<D> & m)
       {
-        assert(m.rows() == 6);
+        eigen_assert(D::ColsAtCompileTime == 6);
         assert(m.cols() == len);
         linear() = m.template topRows<3>();
         angular() = m.template bottomRows<3>();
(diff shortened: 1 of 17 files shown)


src/multibody/force-set.hpp
```diff
@@ -120,12 +120,22 @@ namespace se3
 	return *this;
       }
 
+      template <typename D>
+      Block& operator= (const Eigen::MatrixBase<D> & m)
+      {
+        assert(m.rows() == 6);
```
jcarpent: Please use Eigen Static assert for the fixed dimension.

src/python/model.hpp
```diff
@@ -97,6 +97,7 @@ namespace se3
       {
 	cl
 	  .def("getBodyId",&ModelPythonVisitor::getBodyId)
+    .def("getJointId",&ModelPythonVisitor::getJointId)
```
jcarpent: Please use proper indent.

jcarpent:
:+1: Minor changes needed. The PR looks nice.

jcarpent:
Related to #89 and #106

jcarpent:
Can you also the missing parsing of joint planar in the URDF parser?

Issue #121, opened 2016-03-04
Handle several collision and visual objects when parsing URDF files


jcarpent:

jcarpent:
solved by #156.

Issue #116, opened 2016-03-03
Add path of meshes file in collision Object


jcarpent:

jcarpent:
solved by #156.

PR #156, opened 2016-04-21
[C++][Python][Major] Replaced the way collisions and visuals are stored

fvalenza:
ed in GeometryModel. Now it is two lists of GeometryObject (one for collisions, one for visuals) aggregating information instead of multiple vectors. Now handle multiple collisions/visuals for one link when parsing urdf file and keep track of the mesh absolute path. Changed binding accordingly.

### DIFF
--- a/src/multibody/geometry.hpp
+++ b/src/multibody/geometry.hpp
@@ -157,6 +157,13 @@ namespace se3
 
   }; // struct CollisionResult
 
+/// \brief Return true if the intrinsic geometry of the two CollisionObject is the same
+bool operator == (const fcl::CollisionObject & lhs, const fcl::CollisionObject & rhs)
+{
+  return lhs.collisionGeometry() == rhs.collisionGeometry()
+          && lhs.getAABB().min_ == rhs.getAABB().min_
+          && lhs.getAABB().max_ == rhs.getAABB().max_;
+}
 enum GeometryType
 {
   VISUAL,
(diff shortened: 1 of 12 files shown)


src/multibody/geometry.hpp
```diff
@@ -156,7 +156,83 @@ namespace se3
...
+  }
+
+  inline std::ostream & operator<< (std::ostream & os, const GeometryObject & geom_object)
+  {
+    os  << "Type: \t" << geom_object.type
+        << "Name: \t" << geom_object.name
+        << "Parent ID: \t" << geom_object.parent
+        // << "collision object: \t" << geom_object.collision_object
+        << "Position in parent frame: \t" << geom_object.placement
+        << "Absolute path to mesh file: \t" << geom_object.mesh_path
```
jcarpent: I think you forgot some `"\n"`

src/python/geometry-object.hpp
```diff
@@ -0,0 +1,95 @@
...
+        cl
+          .add_property("name", &GeometryObjectPythonVisitor::getName, &GeometryObjectPythonVisitor::setName)
+          .add_property("parent_id", &GeometryObjectPythonVisitor::getParentId, &GeometryObjectPythonVisitor::setParentId)
+          .add_property("placement", &GeometryObjectPythonVisitor::getPlacementWrtParentJoint, &GeometryObjectPythonVisitor::setMeshPath)
+          .add_property("mesh_path", &GeometryObjectPythonVisitor::getMeshPath, &GeometryObjectPythonVisitor::setMeshPath)
+          ;
+      }
+
+
+      static std::string getName( const GeometryObject & self) { return self.name; }
```
jcarpent: Maybe you can add  ref for returning type.

jcarpent:
Related to #116.

fvalenza:
Also related to #121

src/python/geometry-object.hpp
fvalenza: It doesn't seem to work :

> binding of
 >     reference to type 'basic_string<[...]>' to a value of type 'const basic_string<[...]>' drops > qualifiers

jcarpent: You should use `boost::python::return_internal_reference<>()` for such exposition.

PR #157, opened 2016-04-29
Python exposure tests

fvalenza:
Added some python tests to check that spatial classes are well exposed in python.
Modified some exposition of class member ( through .def_readwrite instead of .add_attribute for aligned datas).
 + minor correction of the names of Frame members


fvalenza:
The decrease of coverage is, i think, due to the new files for python unittesting ( binding_geometry_object.py, and bindinf_frames.py) that are not tested because they are not imported due to hpp-fcl not installed

Issue #131, opened 2016-03-09
Create a JointGeneric accessor


jcarpent:
Create a struct containing a joint variant and expose joint visitor.

nmansard:
JointAccessor is done, using a JointVariant has input of the construction.

I would be nice to have JointAccessor and JointVariant as a unique class, by inheriting JointVariant both from boost::variant and from JointModel.

@fvalenza will check that this second solution is feasible.

jcarpent:
Solved by #182.

PR #162, opened 2016-05-09
[C++] Added JointAccessor + its unittests.

fvalenza:
Major : Added JointModelAccessor and JointDataAccessor that are an abstraction (more precisely a mapping) on Variant types and that expose all the API of JointBase through the use of visitors. Theses visitors have been moved to another file to facilitate the documentation.
Added related unittest


jcarpent:
According to me, the unit test can be largely improved by using for_each on the joint model and data variants.

jcarpent:
Related to #131.

jcarpent:
I was maybe too fast. 
@fvalenza Do you try to make JointVariant inheriting both from boost::variant and from JointModel ?

fvalenza:
Not yet. It is one of the future things to test. As it is now it maps a
joint as stated previously
Le 23 mai 2016 21:27, "Justin Carpentier" <notifications@github.com> a
écrit :

Issue #168, opened 2016-05-18
Make the python buiding build optionnal (true by default if needed)


MaximilienNaveau:
It would be awesome to be able to compile libpinocchio.so and install it without having tones of memory cache needed by the binding python.

jcarpent:
You're right. Currently, the python bindings are build if eigenpy is found. But it is not necessarily a mandatory feature.This can be added to a future PR.

jcarpent:
Solved by #170.

PR #170, opened 2016-05-19
Add a cmake variable to choose if the pyhton wrapper will be compile or not

MaximilienNaveau:
Allow the user to not compile the python binding if unnecessary


CMakeLists.txt
```diff
@@ -159,6 +160,7 @@ SET(${PROJECT_NAME}_ALGORITHM_HEADERS
   algorithm/compute-all-terms.hpp
   )
 
+IF(${BUILD_PYTHON_INTERFACE} STREQUAL "ON")
```
jcarpent: Minor remark. For future, here you can directly  use IF(BUILD_PYTHON_INTERFACE).

jcarpent:
Related to #168. 
@MaximilienNaveau Thanks for this contribution.

CMakeLists.txt
MaximilienNaveau: In fact I tried and the cmake failed as the value of  `IF(BUILD_PYTHON_INTERFACE)` is a sstring with value 'ON'
Maybe I missed something. Though it should work properly.

jcarpent: But BUILD_PYTHON_INTERFACE is not a string, that's the point.

Issue #165, opened 2016-05-13
EIGEN_MAKE_ALIGNED_OPERATOR_NEW in model and data classes


MaximilienNaveau:
To be able to instantiate a model and data using the `new` operator one should add
`EIGEN_MAKE_ALIGNED_OPERATOR_NEW`
as a public macro in the classes.
This is needed feature if you want to have pointers own and initialized in a single class.

jcarpent:
@fvalenza Can you add this fix in your next PR?

jcarpent:
Solved by #172.

PR #172, opened 2016-05-20
overload the new operator for the model and data classes

MaximilienNaveau:
This allow to use `new se3::Model m` and `new se3::Data(m)`. This problem is specific to eigen.


jcarpent:
Just a little comment on this PR. This PR is mandatory if you want to allocate a structure with aligned Eigen elements inside.

jcarpent:
Solved #165.

PR #177, opened 2016-06-03
Upgraded to new version of jekyll + changed Download and tutorials pages

fvalenza:


jcarpent:
For _includes/install_procedure.md, I would prefer to directly redirect to the wiki page of the installation procedure as it is often updated.

jcarpent:
'There is currently no released binary package. ' can now be removed ;)

fvalenza:
For install_procedure.md it is what I tried to do, importing the wiki Installation.md file but i encountered some problems with the markdown -> html conversion. When there is the command "sudo tee.... << EOF...". the "<<EOF". Half of the page is skipped.

jcarpent:
Ok. Let's continue with this alternative.

jmirabel:
```bash
% sudo tee /etc/apt/sources.list.d/robotpkg.list <<EOF
deb http://robotpkg.openrobots.org/packages/debian/pub xenial robotpkg
EOF
```
or how to make install procedure weird...

If this annoys you, you can replace it be the classical:
```bash
echo "deb http://robotpkg.openrobots.org/packages/debian/pub xenial robotpkg" >> /etc/apt/sources.list.d/robotpkg.list
```

jcarpent:
Indeed, it seems simpler. Changed done on the wiki page.

jmirabel:
The "Getting Started Now" button of the main page points to nothing...

PR #183, opened 2016-06-12
Add exception when loading meshes with lines.

jmirabel:
Triangulated meshes contain faces with 1, 2 or 3 vertex. Faces with 1 and 2 vertex are not supported by FCL.

Refers to humanoid-path-planner/hpp-model-urdf#6


jcarpent:
👍

Issue #161, opened 2016-05-04
Update python viewer


jcarpent:
Since we are now able to save the current visual info, we need to move forward by loading directly the visual geometries inside the gepetto viewer. And then, the URDF module inside the gepetto viewer will become deprecated.

jmirabel:
IMHO, tagging the URDF module of the gepetto-viewer as deprecated is not a good idea as I don't think it is a good idea we can remove it in a near future. It would break the ability of loading data in a remote viewer, with paths to the files being different.*
Keeping the path in ROS format (`package://`) should work though.


PS: if you wonder who is using a remote user: I am using it when I run benchmarks on a faster remote computer

jcarpent:
For sure, what we can do is to use the resource retriever of ROS or Pinocchio for the remote files. But in a near future, I would suggest to remove the strong dependency of URDF parser from Gepetto Viewer and to work only with our own description of kinematics chain for generic and general purposes.

nmansard:
Pinocchio will stop using the URDF interface of Gepetto-Viewer and call directly Gepetto-Viewer with addMesh(), to build single solid objects.

I think that moving URDF_DOM as optional dependency of Gepetto viewer would be nice, as a very optional feature.

jmirabel:
I'm not against removing URDF_DOM dependency but:
- I think it is not worth to spend time making it optional if the goal is to remove it later.
- Pinocchio is not the only software using Gepetto-Viewer, so we cannot design the viewer to fit Pinochio requirements at the cost of losing compatibility with other softwares. I do think this discussion belongs to that repository.
- I don't think this can happen before a merge between Pinocchio and hpp-model (at least, because I do not know who else is using it).

The idea of removing the pain of (poorly handled, at the moment) handling offsets between links and meshes of URDF is very pleasant to me ! It would solve some other dependent bugs. It also means Pinocchio has to handle those offsets.

nmansard:
I fully agree with this last message. Yet, let simply make Pinocchio work without the URDF module of Gepetto-Viewer, without touching Gepetto-Viewer.

jcarpent:
Solved by #184.

Issue #126, opened 2016-03-05
Clean Model


jcarpent:
Currently a model is build from concatenation of bodies that are linked one to each other by joints. We need to rethink this strategy by appending bodies to a joint tree.

This new factorisation should help the merge of two models for example, it should simplify the URDF parsers, etc.

jmirabel:
It could even be a HPP requirement as it is needed for the manipulation stack.

My experience of URDF/SRDF parsing in HPP leads me to thinking :
- the URDF parser should create a joint tree below a root joint provided as input,
- parsing several SRDFs for the same robot makes writing SRDF easier (no need to copy SRDF to add new data)

jcarpent:
Thank you for your feedback. Can you say more about what you need for hpp-manipulation ?

nmansard:
The issue raises two aspects: there is a lot of members std::vector<something> in Joint; and the API is not nice, especially the method addBody (which indeed added both a joint and a body).

It seems that the proper solution is to keep the same members in Model, but modify / rename the API.

@fvalenza will separate the actual "addBody" method into "addJoint" (with zero inertia) and "appendBodyToJoint".

Should we keep a method "addJointAndBody" which calls both? I think yes.

jcarpent:
Solved by #184.

PR #184, opened 2016-06-14
rework model (joints/bodies) + fix RobotWrapper link to viewer

fvalenza:
- Make the separation between bodies and joints and do modifications according to.
- Fixed RobotWrapper display : now correctly get the visuals from the geometrymodel/data, update it and display. The meshes are loaded in the viewer one by one thanks to the mesh_path stored by the GeometryObject itself. Removed some useless informations such as _has_visual_ and _fixedBodies_ that were informations related to visual stuff.

Related to #126 , #161


jcarpent:
:+1:

PR #186, opened 2016-06-15
Topic/hot fix visitors + misc

fvalenza:


src/multibody/joint/joint-basic-visitors.hxx
```diff
@@ -371,6 +361,22 @@ namespace se3
   inline void setIndexes(JointModelVariant & jmodel, JointIndex id, int q,int v) { return JointSetIndexesVisitor::run(jmodel, id, q, v); }
 
 
+  /**
+   * @brief      JointShortnameVisitor visitor
+   */
+  class JointShortnameVisitor: public boost::static_visitor<std::string>
```
jcarpent: Can you use here a const ref for the return type?

fvalenza:
The prototype of the function in each joint is: `static const std::string shortname()` thus we don't return a member of the joint. returning a reference with the visitor leads to
   
     returning reference to local temporary object

One solution would be to give a static member for each joint, but is it worth to have one .hxx file for each joint only for this shortname

jcarpent:
Ok. I understand the point, we cannot create static const member with an header only library.

Issue #179, opened 2016-06-08
Fixed Joints Support


KevinAStein:
Hi,
I am trying to load a urdf model to Pinocchio. Looking at robotModel->bodyNames() I can see that there not all bodies specified in the URDF file are listed.
I have specified some fixed joints in the chain from the root link to the feet. Those seem to be missing. Is it not possible to have fixed joints inside a kinematic chain?
thanks,
Kevin

fvalenza:
Hello Kevin,
In Pinocchio, fixed joints are handled this way :
We don't add a Joint of type "Fixed" in the Model but we add a Frame at the center of the fixed joint ( so that we keep track of the information ) and we merge the inertial information of the child link with the parent of the fixed joint.
If you have Joint1 -> link1 -> fixed Joint -> link2, it will be parsed as Joint1 -> link1 + link2 ( expressing the inertial informations in the good frame).
It is then normal that you can't see it in bodyNames(), you can however find it in robotModel-> fix_bodyNames

KevinAStein:
Thanks for your answer.
I have something like joint1 -> link1 -> fixed joint -> link2 -> fixed joint -> link3 and want to compute kinematics and dynamics for link3. How can I access this link? existBodyName("link3") and getBodyId("link3") lead to an error.

fvalenza:
Sorry for the delay.
It depends on which version of Pinocchio you are running:
Release/Master Branch
----
It can be tricky as the link3 as an object disappear during the parsing. Hence you have to work with the "fixed joint". But Pinocchio doesn't have fixed joints. As stated, what it does is it creates a Frame whose name will be the name of the fixed joint, attach it to the parent joint of the fixed frame with a relative placement. In your case you will have a frame instead of the _fixed joint 2_ ( = fixed joint supporting link2)  and a frame instead of the _fixed joint3_.
- Bodies attached to regular joints just "disappear" because we consider the entry point ( = the frame attached to the body) of the body to be the same as the center of the supporting joint

- So to acces the link you will have to do : getBodyId( getFrameParent( "FixedJoint3" ) )

Devel Branch ( coming in next pull request)
----
Bodies/Joints/Frames are currently being reworked.

- Each body will have a Frame (such as said for bodies attached to fixed joints). There will be redundant informations ( for bodies attached to regular joints, you may access it from an operational_frame or from the joint supporting the body)

- To acces the link, you will have to do: getFrame ("link3") and ensure the Frame is of type "BODY"

[...]

PR #193, opened 2016-06-21
Devel

fvalenza:
Keep track of the bodies as Frames


nmansard:
Thanks for this PR. Unification of names with "frames" is good. Should we start to document body listing as deprecated? 
What is the interest of remembering the type of frames (body vs oppoint vs joint)? Is there any explicit demand for this feature? If so, it would be nice to document this demand somewhere to keep track of it.

jcarpent:
This is related to #179.

src/spatial/frame.hpp
```diff
@@ -29,26 +29,35 @@
 namespace se3
 {
 
+  enum FrameType
+  {
+    OP_POINT,
```
jcarpent: why not to use OP_FRAME instead of OP_POINT?

jcarpent:
I would go a little bit deeper by adding the type Joint inside `FrameType` and registering all the joints inside. The idea is to have everything directly accessible from the vector frames in Model, simplifying the API.

nmansard:
This is an interesting proposition. I am afraid that it may simplify the API a little bit too much (loosing efficiency in runtime). I think we should try to imagine how this "frames" could be used by a third-party.

jcarpent:
The adding of Joint in the frame list is just a redundant information to help the end-user to easily retrieve the objects (SENSOR, BODY, JOINT, ...) created during parsing procedure.
The efficiency can be near the same if in the method `getFrameJacobian` we make the distinction between Joint frame and the other frame types.

jmirabel:
Having a unified way of getting joints regardless of their type (Fixed or not) is a feature that will simplify the work of porting HPP to Pinocchio.

jcarpent:
:+1:

Issue #123, opened 2016-03-04
Add Python interpreter to load directly models from Python script


jcarpent:

nim65s:
I'm on it :)

jcarpent:
Solved by #221.

PR #194, opened 2016-06-21
[C++][Python] python model parser

nim65s:
As it is done in src/multibody/parser/python.cpp, one can create a shared_ptr to a new Model in c++,
send this into a python namespace, update it in python, and continue to work on the Model in c++.

An example to use this as a simple parser is given in unittest/python.cpp and models/simple_model.py


nim65s:
`models/simple_model.py` is mostly copy-pasted from another project ; and maybe the ModelWrapper is overkill if we just want an example.

unittest/CMakeLists.txt
```diff
@@ -82,6 +82,15 @@ IF(LUA5_1_FOUND)
   ADD_TEST_CFLAGS(lua '-DPINOCCHIO_SOURCE_DIR=\\\"${${PROJECT_NAME}_SOURCE_DIR}\\\"')
 ENDIF(LUA5_1_FOUND)
 
+IF(${BUILD_PYTHON_INTERFACE} STREQUAL "ON")
+  IF(EIGENPY_FOUND)
+    FINDPYTHON(2.7 EXACT REQUIRED)
```
jcarpent: This line is not useful

src/multibody/parser/python.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include "pinocchio/multibody/parser/python.hpp"
+
+namespace bp = boost::python;
+
+namespace se3
+{
+  namespace python
+  {
+    Model buildModel (bp::str filename, bool verbose)
```
jcarpent: use `const std::string &` instead of `bp::str`

src/multibody/parser/python.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+#include <boost/python/str.hpp>
+
+#include "pinocchio/multibody/model.hpp"
+
+namespace se3
+{
+  namespace python
+  {
+    Model buildModel (boost::python::str filename, bool verbose = false);
```
jcarpent: Same here, with `const std::string &`

unittest/CMakeLists.txt
```diff
@@ -82,6 +82,15 @@ IF(LUA5_1_FOUND)
...
+    ADD_UNIT_TEST(python "eigen3;eigenpy")
```
jcarpent: I think python is not a good choice for the name to avoid conflict with the python interpreter. python_parser may be better.

nmansard:
It is very strange that you de-refer the smart pointer before returning a simple reference. Is it safe? I guess that you made this choice to comply with Pinocchio classical API of returning references rather than pointers. Could you argument? And if returning a simple reference, document the ownership of the memory (would should release it). Or do you think that it is explicit and does not deserve a comment?

In your simple PY script, could you put all the variable names in French, to be consistent? :p

nim65s:
I think you mean putting variable name in English :)

As discussed earlier, we would like to have the same interface for all the parsers, so yes, I did this to comply with the API of the other existent parsers.
I am not sure about the safety or the ownership, but I can look at that a little if you think this is needed.
At least, valgrind is not insulting me for now…

Anyway, travis is not happy, so I am currently working on understanding its problem.

jmirabel:
IMHO, it is safe as long as copying a `se3::Model` is clean and safe. The instance pointed to by `model` is copied and deleted, the copy being returned.

nmansard:
@jmirabel Indeed you are right. Thx

src/CMakeLists.txt
```diff
@@ -166,5 +137,41 @@ IF(EIGENPY_FOUND)
...
+IF (UNIX)
+    # Create target libpinocchio.so
+    ADD_LIBRARY ( ${PROJECT_NAME} SHARED ${HEADERS} ${${PROJECT_NAME}_SOURCES} )
+		SET_TARGET_PROPERTIES( ${PROJECT_NAME} PROPERTIES LINKER_LANGUAGE CXX)
+    PKG_CONFIG_USE_DEPENDENCY(${PROJECT_NAME} eigen3)
+    TARGET_LINK_LIBRARIES(${PROJECT_NAME} ${Boost_LIBRARIES})
+
+    IF(${BUILD_PYTHON_INTERFACE} STREQUAL "ON")
+      IF(EIGENPY_FOUND)
+        TARGET_LINK_LIBRARIES(${PROJECT_NAME} ${PYTHON_LIBRARIES} boost_python)
```
jcarpent: must be ${Boost_PYTHON_LIBRARIES} instead of boost_python

src/CMakeLists.txt
```diff
@@ -166,5 +137,41 @@ IF(EIGENPY_FOUND)
...
 ENDIF(EIGENPY_FOUND)
```
jcarpent: Can you add missing indents ?

CMakeLists.txt
```diff
@@ -181,6 +181,7 @@ SET(${PROJECT_NAME}_PYTHON_HEADERS
   python/algorithms.hpp
   python/parsers.hpp
   python/explog.hpp
+  multibody/parser/python.hpp
```
jcarpent: This line must appear in ${PROJECT_NAME}_MULTIBODY_PARSER_HEADERS if ${BUILD_PYTHON_INTERFACE}

jcarpent:
Related to #123.

jcarpent:
Reworked in #221.

PR #216, opened 2016-06-28
Moved assimp module in hpp-fcl

fvalenza:
Related to #185

PR #217, opened 2016-06-29
Topic/joint access binding

fvalenza:
A model is now composed of a vector of JointAccessor  instead of JointVariant ( the Boost one). + binded JointModelAccessor.
This PR comes after #216


nmansard:
Excellent.

We discussed of possible renaming but did not yet converged on a name. We should arrest a decision now. I suggest to rename former JointVariant to JointRawVariant and JointAccessor to JointVariant. Alternatively, we may want to explicitely state that (new) JointRawVariant is a typedef with JointRawVariant_t.

Other suggestions?

fvalenza:
JointBoostVariant for the current JointVariant ,  or JointRawVariant are both ok for me
May be JointHandle for JointAccessor ?

jcarpent:
Do we really need to know that it is a variant? Otherwise, we can just use JointModel or JointModelDerived to just counterbalance the JointModelBase.

nmansard:
Why not JointModel short? I am not sure. Maybe.
I doubt JointModelDerived is a good option, because not providing enough additional information with respect to JointModel. It is like a bad trade off to me. If we want it short: JointModel -- if we want it informative: JointModelVariant.

fvalenza:
Just a little remark : in the big macro SE3_JOINT_TYPEDEF (in joint-base) , we define the following types: JointData and JointModel in the trait, which are the type of the derived joint.
Should we rename this ? If we want to use JointModel/JointData instead of JointModelAccessor/JointDataAccessor we will have to.
Possible names are:
- JointModel_t 
- JointModelDerived(_t)

jcarpent:
I would go for JointModelDerived and JointDataDerived.

nmansard:
-1 : JointXXderived is meaningless as explained upper.
JointXX conflicts as nicely observed by @fvalenza . Let s forget it as well

Conclusion: Lets keep JointXXVariant as initially planed, with any name (eg JointXXPureVariant) for the hidden typedef.

fvalenza:
For the CRTP  I don't think that JointXXDerived is meaningless since it is a typedef to the true derived type ( typedef is in the traits)

nmansard:
Agree indeed. Renaming in the joint the current JointXX in 
JointXXDerived is relevant if we want to free the name JointXX for the 
actual JointXXAccessor.

So, JointXXAccessor could be renamed in JointXX or in JointXXVariant. No 
strong opinion between both.

jcarpent:
The two first commits are redundant with #216. I will wait until #221 is merged and the factoring of hpp-fcl into rebotpkg.

fvalenza:
Rework of this PR :
- JointAccessor is now Joint
- The stack of joints in Model is still of type Joint (former JointAccessor)
- Added shortname() in Joint, related to #220 
- Removed duplicate commit with pr #216 , this PR will be reworked or closed and reopened when the refactoring of hpp-fcl will be integrated in robotpkg
-  To free the name Joint, changed the traits typedef traits<D>::Joint into traits<D>::JointDerived

jcarpent:
Looks good for me. I think this PR can be now merged and you must keep #216 opened until update of hpp-fcl on robotpkg.

fvalenza:
#216  is not up-to-date. One commit disables installation of hpp-fcl on travis, and thus the related tests are not compiled
So there are two possibilities:
- Either we merge it  and we do a second pull request to re-enable hpp-fcl on travis
- Or i must rework the PR to enable hpp-fcl and we wait for the integration of the new version of hpp-fcl in robotpkg.

I think it really depends on how much time it can take to have the new version of hpp-fcl in robotpkg

PR #221, opened 2016-07-06
Add python model parsing from #194

jcarpent:
This PR improves the CMake packaging for OS X systems and correct some packaging bugs.
It also integrates the great work done by @nim65s on the integration of the Python interpreter to read python model of robots.


jcarpent:
Related to #123.

PR #224, opened 2016-07-08
[C++] Rename spatial scalar type to follow Eigen convention.

aelkhour:
* Rename Scalar_t to Scalar.
* This allows applying Eigen-compatible template operations to
  se3::spatial objects.


jcarpent:
Looks good. Thanks @aelkhour.

Issue #212, opened 2016-06-27
Reorganize urdf parsers


nmansard:
First, we start to have quite some code in multibody/parsers. It might be a good idea to move this directory one level above, in src/parsers. Let's assume this solution for the following.

The API of URDF parser should only expose two methods: buildModel and buildGeom. The rest of the API are details, and should not be displayed to users and stored in a details namespace (se3::urdf::details).

I suggest that the code about URDF should be organized in 4 files:
parsers/urdf.hpp  /// the API, containing the two functions listed above.
parsers/urdf/model.hxx /// the code for parsing urdf tree building the model.
parsers/urdf/geometry.hxx /// The code for parsing the urdf tree building the geometry model.
parsers/urdf/utils.hpp   /// The utils functions (e.g. convertFromUrdf functions) shared between both hxx)

For now, we must also keep parsers/urdf/assimp.hpp (yet from-collada-to-fcl.hpp), until ### is solved.

All the code in parsers/urdf/ should be put in se3::urdf::details namespace to avoid exposing it in the doxygen.

I do not like to have a magic behavior in buildGeom, by non-explicitly relying on ROS_PATH. I would suggest to expose the ROS_PATH function in src/tools/file-explorer.hpp, while mentionning in the documentation of buildGeom that these two methods should be typically called together.

jcarpent:
I also think that we must moved to the parsers to the src level.

As suggested by some users, I think we must simplify at most the parsing process. So calling the retriever of the ROS_PATH inside the parser is a good option to simplify the parsing process. We have to inform the user through the documentation and maybe add an option to use or not this feature.

In the same spirit as moving the parsers directory, I'd suggest to make a shared library with all the parsers inside libpinocchio.so/.dylib

nmansard:
I would suggest to keep:

buildGeom( Model & model, string urdffilename, vector<string> paths )

with the documentation of paths being:
// @param[in] paths     A vector containing the different directories, 
typically obtained from calling se3::rosPaths()

Alternatively, we can have a second function:
buildGeomUsingROSPath( Model & model, string urdffilename )
or
buildGeom( Model & model, string urdffilename, bool usingROSPath )

nmansard:
Solved by #227

Issue #209, opened 2016-06-27
Remove the duplicata collision / visual from geometry


nmansard:
Syntactically, visual and collision objects are the same. It is only the semantic (brought by the end users) that makes visual and collision different. I therefore suggest to preserve this similarity as long as possible. This will avoid duplication of code. 

In geometry.hpp, remove notions of collision and visual. A ``GeometryModel'' object should then have a list of GeometryObject, without type (COLLISION / VISUAL) and a list of collision pair. In case the GeometryObject will be use to store visuals, this list of pair will simply be empty.

jcarpent:
We can add an enum with type VISUAL and COLLISION just to distinguish between both. It is just a matter of memory. In most of URDF files, COLLISION are lighter than VISUALs and one may prefer to use COLLISION objects for collision detection.
I propose to automatically load COLLISION objects and not VISUAL.

nmansard:
About loading: this is a matter to be handled in the parsers, typically 
in parsers/urdf.hpp. There, an argument should be added in buildGeom to 
specify if we are loading the collision or visual objects. Possibly, we 
may want that this argument is set to a default value "collision", also 
I would rather recommend to force the user to be explicit with this 
argument. No strong opinion there.

To be exhaustive: a complete model of the robot would be composed of 5 
objects:
* model of type Model
* data of type Data
* collisionModel of type GeometryModel
* collisionData of type GeometryData
* visuals of type GeometryModel
The last object "visualData" is useless and would typically not be 
created, also it is armless.

jcarpent:
I totally agree with the way to compose a complete model. This means also that we have to mostly rewrite the URDF geometry parsers. 

The main problem that I see is about the delayed process we have chosen: we load geometries after building the model where some reductions (due to fixed joints) may occur. We must then rely on model.frames to retrieve the near joint parent in the tree.

jcarpent:
We must also fulfil the requirements of #190.

nmansard:
Solved by #227

Issue #210, opened 2016-06-27
Remove dependancy on parsers from model


nmansard:
geometry.hpp should not depend on SRDF. Remove the corresponding code a put it in parsers/srdf.hpp.

jcarpent:
Where is exactly the problem?

nmansard:
Line 30 of geometry.hxx, and consequently line 557 of geometry.hpp (and 
the corresponding implementation in the hxx).

nmansard:
Solved by #227

PR #227, opened 2016-07-12
Devel

fvalenza:
Moved parsers from multibody/parser to parsers
Removed duplicated of collision and visuals in GeometryModel

Related to #209 , #212 

NB: 
The cloud of point for the visuals is created/loaded from the mesh file whereas it seems useless. This is the default behavior and by now I am not sure how one can create a GeometryObject of type VISUAL with no cloud point (i.e the fcl::CollisionObject attribute of GeometryObjects) as there is no default constructor for an fcl::CollisionObject. One solutiion may be to give a null pointer to the constructor ?. Any thoughts ?


src/multibody/geometry.hpp
```diff
@@ -274,100 +268,61 @@ struct GeometryObject
 
     GeometryModel(const se3::Model & model)
       : model(model)
-      , ncollisions(0)
-      , nvisuals(0)
-      , collision_objects()
-      , visual_objects()
+      , ngeoms(0)
+      , geometry_objects()
```
nmansard: Can't we find a better name for geometry_object? 
- fclObject  : I still dont like "object", but with "fcl" attached should be clear enough
- fclCollisionObject : object is not adding meaning, collision is confusing wrt visual
- fclWrapper: better than fclObject, but still not very clear
- fclGeometry: in fcl, they use "geometry" for the 3D volume, i.e. something contain in their CollisionObject.
- fclGeometryWrapper: quite long.

src/multibody/geometry.hpp
```diff
@@ -274,100 +268,61 @@ struct GeometryObject
...
     /**
-     * @brief      Add a geometry object whose type is COLLISION to a GeometryModel
+     * @brief      Add a geometry object of a given type to a GeometryModel
      *
      * @param[in]  parent     Index of the parent joint
      * @param[in]  co         The actual fcl CollisionObject
      * @param[in]  placement  The relative placement regarding to the parent frame
      * @param[in]  geom_name  The name of the Geometry Object
      * @param[in]  mesh_path  The absolute path to the mesh
+     * @param[in]  type       The type of the GeometryObject
```
nmansard: I dont think that keeping the type is relevant. We dont want to have a mix of type in a single GeometryModel object. This data is not used anywhere, and I am afraid that it can be confusing for a user, letting him suppose that he can store in a unique GeometryModel both collisions and visuals. We should remove it.

src/multibody/geometry.hpp
```diff
@@ -408,21 +363,16 @@ struct GeometryObject
     const GeometryModel & model_geom;
 
     ///
-    /// \brief Vector gathering the SE3 placements of the collision objects relative to the world.
+    /// \brief Vector gathering the SE3 placements of the geometry objects relative to the world.
     ///        See updateGeometryPlacements to update the placements.
     ///
-    std::vector<se3::SE3> oMg_collisions;
+    std::vector<se3::SE3> oMg_geometries;
```
nmansard: The _geometry is useless (it is already the meaning of the "g" in oMg. Keep the name simple: oMg.

src/multibody/geometry.hpp
```diff
@@ -408,21 +363,16 @@ struct GeometryObject
...
-    /// \brief Vector gathering the SE3 placements of the visual objects relative to the world.
-    ///        See updateGeometryPlacements to update the placements.
-    ///
-    std::vector<se3::SE3> oMg_visuals;
-    ///
     /// \brief Same as oMg but using fcl::Transform3f to store placement.
     ///        This pre-allocation avoids dynamic allocation during collision checking or distance computations.
     ///
-    std::vector<fcl::Transform3f> oMg_fcl_collisions;
+    std::vector<fcl::Transform3f> oMg_fcl_geometries;
```
nmansard: Same remarke: oMg_fcl

src/multibody/geometry.hpp
```diff
@@ -447,15 +397,14 @@ struct GeometryObject
...
+        , oMg_fcl_geometries(model_geom.ngeoms)
         , collision_pairs()
         , nCollisionPairs(0)
         , distance_results()
         , collision_results()
     {
-      const std::size_t num_max_collision_pairs = (model_geom.ncollisions * (model_geom.ncollisions-1))/2;
+      const std::size_t num_max_collision_pairs = (model_geom.ngeoms * (model_geom.ngeoms-1))/2;
       collision_pairs.reserve(num_max_collision_pairs);
       distance_results.resize(num_max_collision_pairs, DistanceResult( fcl::DistanceResult(), 0, 0) );
```
nmansard: Should be "reserve" instead of "resize", isnt it? (both in :408 and :409)

This remark is not directly about the PR, but maybe it can be included.

src/multibody/geometry.hxx
```diff
@@ -42,82 +42,55 @@
...
+    if (type == NONE)
+    {
+      const std::string exception_message ("You must specify if the geometryobject you want to add is of type VISUAL or COLLISION");
+      throw std::invalid_argument(exception_message);
+    }
+
+    Index idx = (Index) (ngeoms ++);
 
-    collision_objects.push_back(GeometryObject(COLLISION, geom_name, parent, co,
+    geometry_objects.push_back(GeometryObject(type, geom_name, parent, co,
```
nmansard: Same here for the class GeometryObject: type is useless and should be removed.

src/multibody/parser/urdf-with-geometry.hpp
```diff
@@ -1,98 +0,0 @@
-//
```
nmansard: I dont know why git has not been able to track the modification on this file. Maybe you forget to explicitly use git move to move the file to src/parsers/urdf/geometry.hxx ? I think it is good to keep the history of this file. Could you say to git to move it?

src/parsers/urdf.hpp
```diff
@@ -31,6 +31,13 @@
 #include <exception>
 #include <limits>
 
+#ifdef WITH_HPP_FCL
```
nmansard: We discuss yesterday with @jcarpent to remove all #ifdef WITH ... from the header files and only keep then in the cpp files. Not to be done in this PR, just a side remark for later.

nmansard:
Great. Thanks for the work. I added some few remarks. I think the most important is about the history of moved files. The other remarks are optional, but if you believe them to be relevant, I guess they wont take you much time to account for them.

src/multibody/geometry.hpp
fvalenza: As we now have two instances of GeometryModel (one for visual, and one for collision), I totally agree.
(note for me : TODO )

src/multibody/geometry.hpp
fvalenza: (TODO)

src/multibody/geometry.hpp
fvalenza: (TODO)

src/multibody/geometry.hpp
fvalenza: I think I added a resize instead of reserve to not leave uninitialized the vector . So if someone tries to access the distance_results without having called at least once the computeDisances(), he will have Zero Distances.

src/multibody/geometry.hxx
fvalenza: (TODO)

src/multibody/parser/urdf-with-geometry.hpp
fvalenza: I really don't know what happened here ( it's the only file that caused problems).
May be just using git move to geometry.hxx won't work if another file already moved to this one ( i'm thinking of urdf-with-geometry.hxx, not hpp)

src/multibody/geometry.hpp
nmansard: Where is "distance_results" ever resized? Shouldn't it be the same size 
as collision_pairs?

src/multibody/geometry.hpp
fvalenza: @nmansard  Are you talking about the `struct GeometryObject` or the attribute member in GeometryModel `geometry_objects` ?
If we change one, we can change the second accordingly.
Howerver none of the propositions you gave seem totally clear (may be fcl_objects, but it is in fact a wrapper above an fcl::CollisionObject).

GeometryObject/ geometry_objects seems not so bad (as it can be used to represent either collision or visual depending of the use we want to make)

nmansard: What I dont like is that "object" is not meaningfull. Typically, GeometryModel is in a sense a geometry object.
Let's keep it like that for now, then, since no nice idea of name emerges. Shouldnt it be geometryObject to fit with the conventions in Pinocchio?

nmansard:
By the way, could you change the geom index list to a vector:
250: -   typedef std::list<GeomIndex> GeomIndexList;
250: +  typedef std::vector<GeomIndex> GeomIndexList;

Or is there a reason to have the list as a list?

fvalenza:
After discussing with justin, I am adding here the work concerning :
- Removing dependency on srdf parser in GeometryModel/Data ( related to issue #210 )
- Removing references on Model/Data in Geometry Model & GeometryData
- One can now parser mesh file whose path starts with file:// ( thanks to @nmansard )

It can also be noted that when adding a collisionPair, we now add an empty DistanceResult and CollisionResult ( the fcl value is uninitialized in fact) in order to avoid  memory violation access in algorithm/collisions.hpp

src/parsers/utils.hpp
```diff
@@ -52,6 +80,10 @@ namespace se3
     // else return an empty string that will provoke an error in loadPolyhedronFromResource()
     namespace bf = boost::filesystem;
 
+    // In case the mesh_path is such as "file://path/to/mesh", check for /path/to/mesh and return it if it exists
+    if ( bf::exists( bf::path(urdf_mesh_path.substr(6, urdf_mesh_path.size())) ))
+      return std::string(urdf_mesh_path.substr(6, urdf_mesh_path.size()));
```
jmirabel: This seems very weak to me. Why don't you check if the path starts with `file://` or `package://` ?

This code may generate unexpected errors like:
- for file `visual/somepath`, we will first check if `/somepath` exists,
- for `collision/somepath`, we will first check if `packageDir/somepath` exists.

By the way, in the standard UNIX file naming, `file://path/to/mesh` is a relative path (`path/to/mesh`) and  `file:///path/to/mesh` is an absolute path (`/path/to/mesh`) so:
- the comment is wrong,
- the substring should start at 7, not 6.

jcarpent: I do agree with all those remarks. 
@jmirabel Can you provide some URDF model names with those particularities (file instead of package, etc)?

jmirabel: I do not have such examples. The possible errors are just example and I'm sure I can find even more tricky issues but it's not the point. The current code is just not behaving properly because it does not fit with the file naming convention. I'm 99% sure that this code will one day be the cause of some unexpected errors...

Using regex should be a lot more simple and robust. Here is an example to be compile with `-lboost_regex` flag:
```
#include <iostream>
#include <boost/regex.hpp>

using namespace boost;

void run(const std::string str) {
  regex rx ("^(?'scheme'[[:word:]]+://)?(?'path'.+)$");
  cmatch what;
  std::cout << "\nResult for " << str << '\n';
  if(regex_match(str.c_str(), what, rx))
  {
    std::cout << "scheme = " << what["scheme"] << std::endl;
    std::cout << "path = " << what["path"] << std::endl;
  }
}

int main () {
  run("file://path/to/mesh1");
  run("package://robot/path/to/mesh1");
  run("aRepoName/robot/path/to/mesh1");
  run("file:///path/to/mesh1");
  run("package:///path/to/mesh1");
}
```

jmirabel: For completeness of my comment, the output of the above code is
```
Result for file://path/to/mesh1
scheme = file://
path = path/to/mesh1

Result for package://robot/path/to/mesh1
scheme = package://
path = robot/path/to/mesh1

Result for aRepoName/robot/path/to/mesh1
scheme = 
path = aRepoName/robot/path/to/mesh1

Result for file:///path/to/mesh1
scheme = file://
path = /path/to/mesh1

Result for package:///path/to/mesh1
scheme = package://
path = /path/to/mesh1
```

nmansard: Thanks for the comment. I agree that the schema is weak. Lets try to improve it.

However, we should not add boost regex in a header file. The compilation timings of Pinocchio are already way to high. 

One solution could be to move the parser in a cpp file and add it to the library. But one of the terminal objective is to have all the Pinocchio code as pure template (with the Scalar type being templated for floating or double precision, or for automatic derivation). So it might be counterproductive to first introduce regex in a cpp file for then later removing it when the cpp file will become a template header hxx file again.

Could we try to find a solution without adding any further dependancies?

jmirabel: Then, looking for `://` in the file name and analysing what's before should be safe enough.

Boost Regex is not a pure template library. I don't know what it templated and what is not but I know the above code does not work without being linked to the boost regex library.

fvalenza:
I'll ammend the commit to handle such case and do a rebase as the devel branch has changed

jcarpent:
Ok, perfect!

Issue #167, opened 2016-05-13
Joint Model methods need more robust tests


jcarpent:
In Joint Configurations unittest, we need more trivial unit tests. @fvalenza Can you increase the number of trivial tests?

jmirabel:
You may have a look at this (rather simple) test for inspiration: https://github.com/humanoid-path-planner/hpp-model/blob/devel/tests/test-configuration.cc

jcarpent:
Solved by #234.

PR #234, opened 2016-07-19
Topic/joint models more unittests

fvalenza:
Related to #167


unittest/joint-configurations.cpp
```diff
@@ -55,20 +55,12 @@ bool configurations_are_equals(const Eigen::VectorXd & conf1, const Eigen::Vecto
   return true;
 }
 
-#define BOOST_TEST_DYN_LINK
```
jcarpent: You should keep those instructions at the beginning of the file

fvalenza:
Added a fix to distance computations

Issue #239, opened 2016-07-20
Add normalized and isNormalized algorithms


jcarpent:
This issues is linked to humanoid-path-planner/hpp-pinocchio#14.

Sometimes, due to serialization, integration, etc ... the configuration can lose its normalization property. The idea is to add two algorithms: one which checks the normlization of an input config vector, the other one which normlalizes the config vector given as input.

jmirabel:
Do you accept default implementation for method like `normalized_impl` in JointModel ?

jcarpent:
solved by #241.

PR #241, opened 2016-07-20
Add se3::normalized

jmirabel:
Solves #239 and related to https://github.com/humanoid-path-planner/hpp-pinocchio/issues/14

### DIFF
--- a/src/algorithm/joint-configuration.hpp
+++ b/src/algorithm/joint-configuration.hpp
@@ -104,7 +104,15 @@ namespace se3
    */
   inline Eigen::VectorXd randomConfiguration(const Model & model);
 
-
+  /**
+   * @brief      Normalize a configuration
+   *
+   * @param[in]  model      Model
+   * @param[in]  q          Configuration to normalize
+   * @return     The normalized configuration
+   */
+  inline Eigen::VectorXd normalized(const Model & model,
+                                    const Eigen::VectorXd & q);
 } // namespace se3 
 
 /* --- Details -------------------------------------------------------------------- */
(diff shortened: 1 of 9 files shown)


src/multibody/joint/joint-base.hpp
```diff
@@ -336,6 +336,23 @@ namespace se3
...
+     *
+     * @return     The normalized configuration
+     */
+    ConfigVector_t normalized(const Eigen::VectorXd & q) const
+    { return derived().normalized_impl(q); }
+
+    /**
+     * @brief      Default implementation of normalized
+     */
+    ConfigVector_t normalized_impl(const Eigen::VectorXd & q) const
```
fvalenza: This kind of default behavior __can__ be dangerous if someone adds a new type of derived joint that has its particular behavior but do not reimplement the method.
In the current use of the CRTP interface it would lead  to a compilation error as method_impl does not exist in the derived joint, except for this method

It can also be noted that joints with 1 DoF can often be factorized in a default behavior mothod

src/multibody/joint/joint-basic-visitors.hpp
```diff
@@ -155,6 +155,19 @@ namespace se3
...
+  /**
+   * @brief      Visit a JointModelVariant through JointNormalizedVisitor to compute
+   *             the interpolation between two joint's configurations
+   *
+   * @param[in]  jmodel  The JointModelVariant
+   * @param[in]  q       configuration to normalize
+   *
+   * @return     The normalized configuration
+   */
+  inline Eigen::VectorXd interpolate(const JointModelVariant & jmodel,
```
fvalenza: typo error here

src/algorithm/joint-configuration.hpp
```diff
@@ -104,7 +104,15 @@ namespace se3
...
-
+  /**
+   * @brief      Normalize a configuration
+   *
+   * @param[in]  model      Model
+   * @param[in]  q          Configuration to normalize
+   * @return     The normalized configuration
+   */
+  inline Eigen::VectorXd normalized(const Model & model,
+                                    const Eigen::VectorXd & q);
```
jcarpent: I think we must work on the same vector q, so without `const &` but instead `&`

jmirabel: Then it shouldn't be `normalized` but `normalize`.

src/multibody/joint/joint-basic-visitors.hxx
```diff
@@ -232,6 +232,28 @@ namespace se3
...
+  class JointNormalizeVisitor: public boost::static_visitor<>
+  {
+  public:
+    Eigen::VectorXd & q;
+
+    JointNormalizeVisitor(Eigen::VectorXd & q) : q(q) {}
+
+    template<typename D>
+    void operator()(const JointModelBase<D> & jmodel) const
+    { jmodel.normalize(q); }
```
jmirabel: Does any know why method `void operator()(const JointModelBase<D> & jmodel) const` compiles without error ?

The method is const, so variable `q` should be of type `const Eigen::VectorXd &` and JointModelBase<D>::normalize(const  Eigen::VectorXd &)` does not exists.
I must be missing something obvious...

jcarpent:
Thank you!

PR #246, opened 2016-07-22
Added radius algorithm

nmansard:


nmansard:
@fvalenza Could you please add a unittest to validate the values computed by the algorithm against hpp::model? And merge.

src/algorithm/geometry.hxx
```diff
@@ -159,7 +97,53 @@ namespace se3
...
     computeDistances(data_geom);
   }
-  
+
+  /// Given p1..3 being either "min" or "max", return one of the corners of the 
+  /// AABB cub of the FCL object.
+#define SE3_GEOM_AABB(FCL,p1,p2,p3)                                     \
+  SE3::Vector3(                                                         \
+    FCL->aabb_local.p1##_ [0],                                          \
+    FCL->aabb_local.p1##_ [1],                                          \
```
fvalenza: Typo error : p1, p2, p3 instead of p1, p1, p1

fvalenza:
In src/benchmark/timings-geometry.hpp there is also an `#include "pinocchio/algorithm/collisions.hpp" that must be replaced by `#include "pinocchio/algorithm/geometry.hpp"`

nmansard:
Corrected, thanks!

Issue #232, opened 2016-07-18
Possible bug in urdf parser when multiple collision are attached


nmansard:
In parser/urdf/geometry.hxx, line 167
` const std::string & collision_object_name = link_name;`
The new collision object is named as the link. What happen if several collision object are attached to a same body? Shouldnt we append to the collision name a id number? link_name + std::string("_") + std::string(num), with num an integer corresponding to the number of collision objects already attached?

jcarpent:
I'm in favour of this solution.

jmirabel:
This is the solution chosen in HPP though it's not fully safe.

jcarpent:
Solved by #248.

PR #248, opened 2016-07-22
Topic/radius unittest

fvalenza:
- Added unittest of radius algorithm and compared results to hpp-model
- Fixed parsing of multiple collision objects attached to a single link, related to #232 
- Fixed parsing of fixed joint when the child link has no inertial information

### DIFF
--- a/src/parsers/urdf/model.hxx
+++ b/src/parsers/urdf/model.hxx
@@ -392,6 +392,8 @@ namespace se3
               joint_info << "fixed joint";
               if (link->inertial)
                 model.appendBodyToJoint(parent_joint_id, Y, jointPlacement, link->name); //Modify the parent inertia in the model
+              else
+                model.addFrame(link->name, parent_joint_id, nextPlacementOffset, BODY);
               
               //transformation of the current placement offset
               nextPlacementOffset = jointPlacement;
--- a/unittest/geom.cpp
+++ b/unittest/geom.cpp
@@ -238,6 +238,7 @@ BOOST_AUTO_TEST_CASE (radius)
   se3::computeBodyRadius(model, geom, geomData);
   BOOST_FOREACH( double radius, geomData.radius) BOOST_CHECK(radius>=0.);
 
+#ifdef WITH_HPP_MODEL_URDF
   /// *************  HPP  ************* /// 
   /// ********************************* ///
   using hpp::model::JointVector_t;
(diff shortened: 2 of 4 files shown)


src/parsers/urdf/model.hxx
```diff
@@ -365,7 +365,10 @@ inline void parseTree (::urdf::LinkConstPtr link, Model & model, const SE3 & pla
         {
           model.appendBodyToJoint(parent_joint_id, jointPlacement, Y, link->name); //Modify the parent inertia in the model
         }
-        
+        else
+        {
+          model.addFrame(link->name, parent_joint_id, nextPlacementOffset, BODY);
```
jcarpent: Why do you add as BODY and not as FIXED_JOINT?

unittest/geom.cpp
```diff
@@ -139,6 +139,25 @@ std::ostream& operator<<(std::ostream& os, const std::pair < se3::Model, se3::Ge
 } 
 
 
+void loadHumanoidPathPlanerModel (const hpp::model::HumanoidRobotPtr_t& robot,
```
jcarpent: some WITH_HPP_MODEL_URDF are missing

src/parsers/urdf/model.hxx
fvalenza: It is for the link that is attached to the fixed joint in case of no inertia.
In the if statement in the method appendBodyToJoint there is a call to addFrame. For links with no inertial information, the parser did not keep track of them ( was the case for the r_sole, l_sole in our romeo model)

unittest/geom.cpp
fvalenza: exactly

fvalenza: PR updated

src/parsers/urdf/model.hxx
jcarpent: For me, it is still a fixed joint. I would also keep info on both: FIXED_JOINT with joint->name as name and BODY with link->name as body.

fvalenza: line 376 we add the frame corresponding to the fixed_joint.
This line is for the link that follows the fixed_joint

jcarpent: Ok. Perfect. I missed it.

jcarpent:
Seems good to me. Waiting for travis.

jcarpent:
In src/parsers/urdf/geometry.hxx, there is a lot of redundant codes. @fvalenza  Can you remove them by making functions?

nmansard:
What is the status of this PR? A single minor issue was raised but is now solved?

Could we merge?

jcarpent:
@fvalenza is just simplifying the code to avoid redundancy and make it simple.

jcarpent:
Looks nice.

PR #250, opened 2016-07-23
Fix doc of jacobinCenterOfMass

jmirabel:
I lost quite some time looking for a bug caused by this. It may help somebody later.

By the way, I find it a bit confusing that `centerOfMass` and `jacobianCenterOfMass` fills the same data but in different frames.


jcarpent:
Thanks for the fix.

Issue #244, opened 2016-07-21
Model: avoid redundant addJoint methods


nmansard:
In Model, we have several methods addJoint and addJointAndBody. All of them can be achieved equivalently by calling then nominal addJoint(parent,type,placement,inertia,names), followed possibly some few additional methods to complete the call. I would like to suggest to remove this redundancy:
A. either by simply removing the redundant methods, while only keeping the basic ones
B. or by, at least, calling the basic methods in the complex ones, while removing the code duplicate. 

I vote for A.

nmansard:
In one of the methods, neutralConfiguration is assigned to be the middle point of the joint range. If this behavior is important, I suggest to write a dedicated algorithm that force the neutral configuration to be the range median, explicitly and a posteriori.

nmansard:
This becomes a HPP reqiuirement: we do not have the same behavior. I will implement the above-mentioned proposition.

jcarpent:
Partially solved by #251.

jcarpent:
Entirely solved by #251.

PR #251, opened 2016-07-24
Topic/model refactoring

jcarpent:
Related to #244.

This PR cleans the API of class Model. There is now one method addJoint and one method appendBodyToJoint to fill the inertia. For people you directly their own models, they have to update their code accordingly.

I also clean the URDF parser.

This PR also fixes several bugs.

### DIFF
--- a/src/multibody/model.hpp
+++ b/src/multibody/model.hpp
@@ -27,7 +27,6 @@
 #include "pinocchio/multibody/fwd.hpp"
 #include "pinocchio/multibody/frame.hpp"
 #include "pinocchio/multibody/joint/joint.hpp"
-#include "pinocchio/tools/string-generator.hpp"
 #include <iostream>
 #include <Eigen/Cholesky>
 
@@ -99,14 +98,18 @@ namespace se3
 
     /// \brief Vector of operational frames registered on the model.
     std::vector<Frame> frames;
+    
+    /// \brief Vector of subtrees
+    /// subtree[j] corresponds to the subtree supported by the joint j.
+    std::vector<IndexVector> subtrees;
 
-    /// \brief Spatial gravity
+    /// \brief Spatial gravity of the model.
     Motion gravity;
     
     /// \brief Default 3D gravity vector (=(0,0,-9.81)).
     static const Eigen::Vector3d gravity981;
 
-    /// \brief Default constructor
+    /// \brief Default constructor. Builds an empty model with no joints.
     Model()
       : nq(0)
       , nv(0)
--- a/src/multibody/model.hpp
+++ b/src/multibody/model.hpp
@@ -101,6 +101,7 @@ namespace se3
     
     /// \brief Vector of subtrees
     /// subtree[j] corresponds to the subtree supported by the joint j.
+    /// The first element of subtree[j] is the index of the joint j itself.
     std::vector<IndexVector> subtrees;
 
     /// \brief Spatial gravity of the model.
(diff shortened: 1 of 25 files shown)


src/multibody/model.hpp
```diff
@@ -124,94 +128,51 @@ namespace se3
...
-    /// \return The index of the new added joint.
-    ///
-    template<typename D>
-    JointIndex addJoint(JointIndex parent,const JointModelBase<D> & j,const SE3 & jointPlacement,
-                        const std::string & jointName = "");
+    template<typename JointModelDerived>
+    JointIndex addJoint(const JointIndex parent, const JointModelBase<JointModelDerived> & joint_model, const SE3 & joint_placement,
+                        const std::string & joint_name = "",
+                        const Eigen::VectorXd & max_effort = Eigen::VectorXd::Constant(JointModelDerived::NV,std::numeric_limits<double>::max()),
+                        const Eigen::VectorXd & max_velocity = Eigen::VectorXd::Constant(JointModelDerived::NV,std::numeric_limits<double>::max()),
```
nmansard: Is there any reasons why limits cannot be set afterwards? Call addJoint first, set limits next.

src/parsers/sample-models.cpp
```diff
@@ -18,184 +18,153 @@
...
-#include <hpp/fcl/shape/geometric_shapes.h>
-#endif
-
 namespace se3
 {
   namespace buildModels
   {
+    
+    template<typename JointModel>
+    void addJointAndBody(Model & model, const JointModelBase<JointModel> & joint, const std::string & parent_name, const std::string & name)
```
nmansard: This method should be static, isntit?

src/multibody/model.hpp
jcarpent: No, there is no restriction. We just need to add a new dedicated visitor for that.

src/parsers/sample-models.cpp
jcarpent: I think it is not necessary to add such things inside the Model class. I added it just as a shortcut, nothing more.

src/multibody/model.hpp
nmansard: Just a method. Or even just access the public limit vectors to modify 
it. In any case, you would likely have to do so, depending on the 
nondefault argument in the method argument list.

src/parsers/sample-models.cpp
nmansard: Keep it as it is, just add the keyword static to make sure that it is 
not usable outside the .cpp file.

src/multibody/model.hpp
jcarpent: We cannot give a direct access to such vectors, because their are fixed dimensions according to the joint type.

src/parsers/sample-models.cpp
jcarpent: Ok. I see your point now.

src/multibody/model.hpp
jcarpent: So, we need to delete the notion of joint bounds from the model and add visitor either to get access to their value or to set them independently. 

I suggest to do that inside an other PR.

nmansard: All limits are already public attributes:
    Eigen::VectorXd effortLimit;
    Eigen::VectorXd velocityLimit;
    Eigen::VectorXd lowerPositionLimit;
    Eigen::VectorXd upperPositionLimit;

So you could just have:
model.addJoint(JointFreeFlyer() ... blabla);
model.joint[0].jointVelocitySelector(model.effortLimit) = blabla

Maybe it is not convenient to do. Is that your point?
We could add a "limit" method to make it simpler:
model.limit( /*jointId=*/ 0, EFFORT) = blabla

jcarpent: That's is the point. It is not straightforward to write: `model.joint[0].jointVelocitySelector(model.effortLimit) = blabla`

Let's implement the access methods in the new abstract JointModel.

PR #258, opened 2016-07-26
se3::computeDistances return the index of the shortest distance.

jmirabel:
For a rather cheap update, we can have `computeDistances` returning the index corresponding to the shortest distance.

I hesitated between this version and a templatized version like `computeDistances<ReturnShortest>(...)`. As the shortest distance is rather cheap to compute, I think it does not matter do always do it.


nmansard:
You can go for the template version, with a default "true" argument.

PR #260, opened 2016-07-26
Topic/subtree com

jcarpent:
This PR is related to humanoid-path-planner/hpp-pinocchio/issues/17 and recent humanoid-path-planner/hpp-pinocchio/pull/28.

Add the notion of subtrees and supports for the kinematic tree.
Add the subtree jacobian centre of mass algo. This algo can be easily extended to use previous computation.
Remove the `const bool subtree` from centerOfMass algorithm.


src/multibody/model.hpp
```diff
@@ -100,9 +100,13 @@ namespace se3
     std::vector<Frame> frames;
     
     /// \brief Vector of subtrees.
-    /// subtree[j] corresponds to the subtree supported by the joint j.
+    /// subtrees[j] corresponds to the subtree supported by the joint j.
```
nmansard: I do not agree with the choice of modifying (making more complex) model to implement a marginal algorithm. I dont think that it is a good strategy in principle. 
Here, the subtree structure should be given as an argument to the algorithm.

nmansard:
Thanks for the PR. I think it is necessary to discuss it a little bit before merging. I do not understand the following rationale:
* the boolean argument "subtree" has been removed from existing COM algorithms, and the optional computations are now done by default. Why? Is it not more expensive?
* the new algorithm computes the jacobian for a single subtree. For it to be useful, we need it for multiple subtree at the same time. Dont you think?
* As explained in the code comment, I do not support the idea of augmenting Model. We should find a design pattern to avoid it.

In general, I am not fully confident that this algorithm is useful and should be put in Pinocchio. Maybe is it too late to raise this concern. Are you fully in need of it?

src/multibody/model.hpp
jcarpent: I think this is not a marginal feature. It will become necessary in the future for very sparse algo which will deal with only certain links, as the computation of (JM-1Jt). And it is information cheap, easy to compete when building.

It can also be used to make smart update, etc.

jcarpent:
- the additional computation is very small (3ns per joint).
- I think we need to add something like getSubtreeCenterOfMassJacobian which can be called after jacobianCenterOfMass and just extract necessary columns, with small additional computations.
- The additional information in Model are necessary when doing sparse algo with forward pass.

I think is people inside HPP need it, it's certainly means that it can be useful for other people. And this version seems to be almost optimal ;)

jcarpent:
I suggest to merge it and to go back to it later. As I need it for a future PR requested by @jmirabel.

src/multibody/model.hpp
nmansard: Both subtree and support are useless. Subtrees[j] is the range from j to lastChild[j]. Supports can be scanned following the parent map. 

You then can write equivalently
for( int isub=j; isub<=lastChild[j]; isub ++) ; // ... do something with the subtree
for ( int isup = j; isup>0; isup = parents[isup] ) ; // ... do something with the support.

Or
BOOST_FOREACH(int isub, subtrees[j]) ; // .. do something with the subtree
BOOST_FOREACH(int isup, support[j]) ; // .. do something with the support

To validate the use of these extra data, you need to time what is the extra cost of solution 1 wrt solution 2.

nmansard:
I suggest that you divide the pull request in at least 2: the minor refactoring, that can be merged quickly; and the modification of Model API, possibly along with the sub Jcom algorithm, that we must discuss further.

src/multibody/model.hpp
jcarpent: For the support, is only available for backward pass, not forward! And lastChild is in Data, not in Model. So, algo without data will not work !

So I suggest for simplicity and efficiency to keep supports and subtrees.

Issue #231, opened 2016-07-18
Alignment issues


jmirabel:
Hello Pinocchio developpers,

when running the unittest, I get the traditionnal Eigen error:
```
rnea: /local/jmirabel/devel/hpp/install/include/eigen3/Eigen/src/Core/DenseStorage.h:86: Eigen::internal::plain_array<T, Size, MatrixOrArrayOptions, 16>::plain_array() [with T = double; int Size = 6; int MatrixOrArrayOptions = 0]: Assertion `(reinterpret_cast<size_t>(eigen_unaligned_array_assert_workaround_gcc47(array)) & 0xf) == 0 && "this assertion is explained here: " "http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.html" " **** READ THIS WEB PAGE !!! ****"' failed.
```

The error is raised when creating an object of type `JointDataRevolute`.

I have a 32bit computer. I guess there is no alignment issues on 64bit arch. Do you have support for 32bit architecture ?

jcarpent:
Currently the simplest way to solve the problem is to add `-DEIGEN_DONT_ALIGN_STATICALLY` inside the compiler option line.

jcarpent:
To be more precise, it seems that the problem comes from boost::variant which does offer explicitly an aligned structure of the stack. Two solutions exist:

- one is to rewrite from scratch the mechanism of variant
- the other one is to inherit from boost::variant and to create our own aligned allocator

jmirabel:
Thanks for having dug that out !

jcarpent:
Solved by #298.

PR #262, opened 2016-07-26
Topic/32bits handling

jcarpent:
This PR follows #260.

This PR adds flag to unaligned data on 32 bits arch (related to #231).
Move the URDF parser to the main lib. In this way, one can now build a model from a UDRF file from a JointModelVariant directly.


nmansard:
For complex changes like this one (and the previous one) I would advice to i) discuss the changes beforehand with the team ii) properly separate the PR to have one changes only. 

Topic 1: eigen unaligned: I dont see any other solution yet, but there is apparently problems to be expected in downstream libraries whose other dependencies are compiled with Eigen but without the same option. 
from flexible-collision-library/fcl#96
citing http://eigen.tuxfamily.org/bz/show_bug.cgi?id=422 .

>  Some options do break ABI compatibility. EIGEN_DEFAULT_TO_ROW_MAJOR is one of them, but not the only one. Another is EIGEN_DONT_ALIGN_STATICALLY, for example. I think that this is a documentation/communication issue. These features are very useful to some people who know what they're doing but clearly they should not be used in code that needs to link with arbitrary other Eigen-using code.

I guess that @jmirabel requires a patch quickly, so I guess we do not have any choice that merging this part of the PR. An issue "BUG" must be kept opened to investigate the topic further.

Topic 2: adding some code in the library. I am ok with that idea, but we also would like that Pinocchio becomes a pure header library by templating on the scalar type. Both options are incompatible. We must discuss it. If we decide to go for pushing more code in the library, it must be done globally with a single PR, ideally a single commit, that we can revert later if needing to go back to the header-only solution. 
Push this part in a branch, we will take care of it later, when merging pinocchio in HPP will be over.

Topic 3: make URDF parser accept a variant instead of a template. I like the idea, even if I would appreciate that we focus more on urgent work for hpp. I guess the change will also ease the python bindings (is there any other interest?). Please make a dedicated PR for that, so that it is easier to browse the code.

Issue #253, opened 2016-07-25
CollisionObject and CollisionGeometry


nmansard:
We have a CollisionObject (ie a CollisionGeometry plus a FCL placement) in the GeomModel. The placement is never used. We also have a FCL placement in GeomData. 
We should rather have a CollisionGeometry in the model. It would also be nice to have rather a CollisionObject in the GeomData, rather than simply a placement, as it is not more costly and eases the call to collision/distance computations.

jcarpent:
Solved by #270.

PR #270, opened 2016-07-27
[C++] The cloud of points in GeometryObject is now a fcl::CollisionGe…

fvalenza:
Related to #253

### DIFF
--- a/src/parsers/urdf/geometry.hxx
+++ b/src/parsers/urdf/geometry.hxx
@@ -47,15 +47,15 @@ namespace se3
      * @brief      Get a fcl::CollisionObject from an urdf geometry, searching
      *             for it in specified package directories
      *
-     * @param[in]  urdf_geometry  The input urdf geometry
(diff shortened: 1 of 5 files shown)


src/parsers/urdf/geometry.hxx
```diff
@@ -51,11 +51,11 @@ namespace se3
      * @param[in]  package_dirs   A vector containing the different directories where to search for packages
      * @param[out] mesh_path      The Absolute path of the mesh currently read
      *
-     * @return     The geometry converted as a fcl::CollisionObject
+     * @return     The geometry converted as a fcl::CollisionGeometry
```
jcarpent: The doc must specify that it is in fact a pointer !

jcarpent: A shared pointer on ...

fvalenza:
The travis build passed. I will then just amend the change of doc

Issue #271, opened 2016-07-27
Finite differences with joints


jcarpent:
If one wants to do finite differences with revolute joints and similar, one must take care of scalar precisions ... Just because of the presence of cos and sin. So if you want to work with cos for example, with a precision a sensitivity variation around 1e-8, theta must vary around 1e-4 or less to be sensible (it comes from the Taylor expansion).

jcarpent:
This remarks is a reason of #256.

jcarpent:
I suggest that each joint give has a method which returns a scalar increment for finit diff purposes.

nmansard:
Solved by #274

Issue #256, opened 2016-07-25
Unit test : jacobian against integrate


nmansard:
We need to add a unit test to validate the new integrate function against the jacobian methods, by computing an approximation of the jacobians by finite differencing. 

A preliminary test with jacobian COM indicates that there is a mistake somewhere.

jcarpent:
Can you give more details on such a mistake? 
Because jacobian of CoM is already checked against CRBA, which validated against RNEA, which is also validated against RBDL and so on ...

But anyway, it is still interesting to have the finite diff implemented inside the Pinocchio corpus.

nmansard:
Finitie differencing Jcom gives I3 in the 3 first columns, while it should be oR1.

I agree that the current Jcom algo is rock solid. I am rather afraid of a bug in the integrate. It is also possible that I am making a mistake with how to finite-diff on a Lie group.

I will push in the morning the corresponding unit test in hpp-pinocchio (you can already have a look in my branch topic/Jcom of hpp-pinocchio). We may want to add a similar test in pinocchio later.

jmirabel:
In hpp-constraints, class `DifferentiableFunction` has two algorithms for finite difference. It may help.

jcarpent:
After discussing with @jmirabel, it seems the bug comes from the joints with angulars transformation as FreeFlyer or SO3 joint. I will fixe it today.

nmansard:
See #278 . I close this issue as #278 is more generic and cover it.

PR #274, opened 2016-07-27
Topic/finite diff

jcarpent:
This PR solves several bugs that were introduced in joint model classes.
Again, the current unit test inside Pinocchio are not robust at all and must be check twice.

In high priority, we must check differentiate and interpolate. They do NOT return the right results.

Related to #271 and #256.


jcarpent:
In unit test finite-differences.cpp, there is a functional and checked example on how to make good FD with pinocchio.

PR #285, opened 2016-07-29
Topic/visuals without fcl

fvalenza:
GeometryModel and GeometryData are available to one user even if hpp-fcl is not found.
__However__ , only stuff related to geometryplacements is available ( no collision computations).

Moved fcl structures to a dedicated file.

__WARNING: THIS PR IS NOT  TOTALLY FINISHED YET__ 
I still have some problems with the library when compiling python wrapper and linking to libpinocchio(.so). I don't know yet if it comes from an unproper packaging from previous commits or no. I will investigate it on monday morning. Am i the only one in this case ?


nmansard:
Thanks for your work.
I manage to directly compil the lib and the unittest compiled with WITH_FCL. The other tests cannot compile because they do not find fake-fcl.

I think that this reveal a flaw in the initial design of the PR, and maybe of the concept of conditional dependencies (automatic or manually set) in Pinocchio. Here, if we compile the lib with WITH_FCL, all the dependencies linking with the lib (unittest, benchs, third parties) must have the flag on as well, otherwise GeomModel will not have a coherent size in the lib and in the headers. 

So for sure: the set of flags selected for compiling the library must be propagated (forced) for all the binaries compiled in Pinocchio, and added in the pkg .pc file (this is already done).

Minor comments: 
The ifdef's does not bring to much complexity as far as I can observe. The addition of fcl.hpp even tends to simplify some of downstream headers.
I think you should have a fcl.hxx to put the details of fcl.hpp.
I dont like so much the "fake-fcl" header. I think you should merge it with fcl.hpp, even if it makes this header tedious. 
We could think of letting all the methods in GeomModel and GeomData, but simply removing there implementation. As they are inline, this should not lead to compilation errors, except if the user try to call them. Or add a default implementation behind a ifdef leading to an assert. Pro: GeomModel and GeomData would keep a similar API, making it easier to maintain third-party code. E.g. it would simplify bindings. It also avoid ifdef in the main header.  Cons: if not providing implementation at all, compilation errors may be not nice. If we do, then it requires extra work. Possibly, we can simply not provide implementation in a first time, and add asserted implementation later.

benchmark/CMakeLists.txt
```diff
@@ -47,6 +51,10 @@ ELSE(BUILD_BENCHMARK)
   ADD_EXECUTABLE(timings-eigen EXCLUDE_FROM_ALL timings-eigen.cpp)
 ENDIF(BUILD_BENCHMARK)
 PKG_CONFIG_USE_DEPENDENCY(timings-eigen eigen3)
+IF(HPP_FCL_FOUND)
```
jcarpent: Not necessary here

utils/CMakeLists.txt
```diff
@@ -40,6 +40,11 @@ MACRO(ADD_UTIL NAME UTIL_SRC PKGS)
     INSTALL(TARGETS ${NAME} DESTINATION bin)
   ENDIF(BUILD_UTILS)
 
+  IF(HPP_FCL_FOUND)
```
jcarpent: Not needed here

fvalenza:
The last commit is a raw fix for the problem of an empty mesh_path in case  we do not have hpp-fcl.
In case we encounter something else than a urdf::Mesh in the function `addLinkGeometryToGeomModel` the mesh_path member of the GeometryObject will be empty and the Visualisation will not work

jcarpent:
@nmansard Are you still available to complete the review of this PR?

src/multibody/fcl.hxx
```diff
@@ -0,0 +1,87 @@
...
+  inline Eigen::Vector3d DistanceResult::closestPointOuter () const
+  {
+    return toVector3d(fcl_distance_result.nearest_points [1]);
+  }
+    
+  inline bool operator == (const fcl::CollisionObject & lhs, const fcl::CollisionObject & rhs)
+  {
+    return lhs.collisionGeometry() == rhs.collisionGeometry()
+            && lhs.getAABB().min_ == rhs.getAABB().min_
+            && lhs.getAABB().max_ == rhs.getAABB().max_;
```
nmansard: I guess that this test is the only reason for which we have a fake AABB struct defined. But I dont really understand the meaning of this test. If useful, could you comment it? If not, just remove it.
A stupid question over my stupid remark: is it really necessary to have a== operator for geometries?

nmansard:
Seems good to me. I think we can merge. Just see my small remark.
I think we are wrong (from a long time ago) of having conditional define "-D" in the compilation line. Here there is a clear possible mistake that would confuse anybody compiling pinocchio with FCL and forgetting in a third party lib of compiling with -DHPP_FCL. You can expect un-undestandable segfault or even memory leak and wrong pointer access. 
As the mistake is not new, I suggest to merge and to add an issue about this problem, to be corrected later. I would prefer that we can discuss f2f about this issue and decide together of a strategy to avoid it in the future.
@jcarpent please merge.

src/multibody/fcl.hxx
fvalenza: The operator== is necessary for structures/class that we want to bind in boost:python as vectors.
See [this SO](http://stackoverflow.com/a/10682380)

florent-lamiraux:
I think the right solution is to insert -DHPP_FCL in pinocchio.pc file. Thus Dependencies will use it in a transparent way.

jcarpent:
I do agree. The next PR #293 solved the issue.

Issue #247, opened 2016-07-22
Divide the python module in several .o


nmansard:
This would help to reduce the compilation time.

nmansard:
I thought you wrote that trying this direction was leading to increase compilation time (which is indeed reasonable). Am I wrong?

jcarpent:
Solved by #286.

Issue #230, opened 2016-07-18
Move python interfaces to a dedicated directory


jcarpent:
I suggest to move all the bindings to a dedicated directory to simplify the Cmake files and the global organisation of the project.

jcarpent:
Solved by #286.

PR #286, opened 2016-08-01
Move python bindings to a dedicated directory + split bindings in several .o files

jcarpent:
Related to #230 and #247.


jcarpent:
I think the time is disappointing. I tried to create several .o for the python bindings and the compilation reached 6:49 min against 2:30 min. I think the change is not necessary at all.

nmansard:
Sorry, but we cannot track the modifs in this PR because git unproperly tracked the file moves. Could you redo it with proper tracking?

jcarpent:
I do not get your comment. The first commit `[CMake] Move python to a dedicated directory` follows well the `git mv`

nmansard:
Ok. That's the expose algorithm that was split.

jcarpent:
Just updated and it passed. I will now merge.

Issue #259, opened 2016-07-26
Use njoint instead of nbody for joint visitors


jcarpent:
Currently, nbody is used while njoint must be used inside all algorithms. This is a bug which must be solved quickly.

nmansard:
BTW: I dont like this redundancy of information, by explicitly storing the size of the lists (e.g. have njoint which is equal to joints.size()). Dont you think that we should get rid of it?

jcarpent:
Indeed. And it must be the same for nbody.

nmansard:
Ok. Should we remove nbody first and use njoint instead, and then do a second PR (later, less urgent) removing njoint, ngeom etc?

jcarpent:
I agree as njoint is safe.

jmirabel:
This means that, from now on, we should use `model.joints.size()` rather than `model.joints.njoints`, right ?

jcarpent:
Both `model.joints.size()`and `model.njoint` are ok. But to avoid redundant info, it might interesting to only consider the first solution. 
Right, you should avoid to use `model.nbody`which may lead to bugs as nbody >= njoint

nmansard:
For njoint: it rather was a question. Should we keep it or remove it. Maybe keep it is cheap and easy to assert, and simplify downstream use. 
In any case, we should mark nbody as deprecated.

jcarpent:
Solved by #288.

PR #288, opened 2016-08-02
Replace nbody to njoint and activate all collision pairs by default

jmirabel:


src/multibody/model.hxx
```diff
@@ -240,35 +240,35 @@ namespace se3
   inline Data::Data (const Model & ref)
     :model(ref)
```
jmirabel: This has to be double checked as I am not sure of where those values are being used.

src/multibody/model.hxx
```diff
@@ -31,7 +31,7 @@ namespace se3
   inline std::ostream& operator<< (std::ostream & os, const Model & model)
   {
     os << "Nb bodies = " << model.nbody << " (nq="<< model.nq<<",nv="<<model.nv<<")" << std::endl;
-    for(Model::Index i=0;i<(Model::Index)(model.nbody);++i)
```
jmirabel: I guess I should change `"Nb bodies  = " << model.nbody` by  `"Nb joints  = " << model.njoint` ?

jcarpent: Yes, you can operate as you mentioned

jmirabel:
Solve #259

src/multibody/model.hxx
```diff
@@ -238,8 +238,7 @@ namespace se3
 
 
   inline Data::Data (const Model & ref)
```
jcarpent: I think you can replace ref by model, to be more precise

jcarpent:
I think a simple rebase on the current level branch should help travis.

jcarpent:
Seems good to me. I will merge.

PR #290, opened 2016-08-02
Handle recent compilers as gcc5 or gcc6

jcarpent:
Recent compilers use c++11 as default standard. We must inform that we are currently following the c++03 standard.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,9 @@ SET(PROJECT_URL "http://github.com/stack-of-tasks/pinocchio")
 
 SET(DOXYGEN_USE_MATHJAX YES)
 
+# We follows c++03 standard
+ADD_DEFINITIONS(-std=c++03)
+
 IF(APPLE)
   SET(CMAKE_MACOSX_RPATH TRUE)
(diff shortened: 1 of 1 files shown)


nmansard:
I think you posted an issue about this topic, but I cannot find it.

Are you sure that this is the good direction to go? I am afraid that it is misleading to c++11 users, letting them think that we are supporting recent c++ version while we are not. How would this interact with third-party software using pinocchio? Would the pinocchio lib be compatible to other code compiled without the option?

nim65s:
I am using pinocchio compiled with -std=c++03 in a project compile with
-std=c++11 without problems.

Le mer. 3 août 2016 à 17:29, Nicolas Mansard <notifications@github.com> a
écrit :

nmansard:
I do not get how this is possible. Most of Pinocchio is headers. This trick would work only if the problem in 11 is limited to a .cpp file which is compiled and hidden from the third party. Is it the case?

nmansard:
up

nim65s:
I'll dig on this issue and provide more details next week :)

Le mar. 30 août 2016 à 14:17, Nicolas Mansard <notifications@github.com> a
écrit :

PR #295, opened 2016-08-05
[unittest] Fixed precision comparison test.

GaloMALDONADO:
Some tests were using== operator instead of isApprox


jcarpent:
Thanks for the fix.

PR #296, opened 2016-08-05
[C++][Python]Exposed in Python GeometryModel method::addGeometryObjec…

GaloMALDONADO:
One can now add geometry objects to a python GeometryModel one by one. Previously it was only possible parsing a urdf file.

Peer programming with @fvalenza

### DIFF
--- a/src/multibody/geometry.hxx
+++ b/src/multibody/geometry.hxx
@@ -47,6 +47,20 @@ namespace se3
     return idx;
   }
 
+  inline GeomIndex GeometryModel::addGeometryObject(const JointIndex parent,                                                    
+                                                    const SE3 & placement,
+                                                    const std::string & geom_name,
+                                                    const std::string & mesh_path) throw(std::invalid_argument)
+  {
+
+    Index idx = (Index) (ngeoms ++);
+    const boost::shared_ptr<fcl::CollisionGeometry> geometry(new fcl::CollisionGeometry());
+    geometryObjects.push_back(GeometryObject( geom_name, parent, geometry,
+                                               placement, mesh_path));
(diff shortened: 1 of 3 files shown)


jcarpent:
I think this feature is not necessary. You can directly work with the viewer without this feature. Please see with @nim65s for more details.

fvalenza:
It may not be necessary but currently the only way to build a GeometryModel in Python is through the urdf parsing.
One may need to build one by hand and let the RobotWrapper do all the work (i.e link pinocchio Visuals with the viewer, updateGeometryPlacements with a GeometryData  and then manage the positions).

For me it's not a matter of if it is possible to do it without this feature but a matter of simplicity for non advanced users

src/multibody/geometry.hxx
```diff
@@ -47,6 +47,20 @@ namespace se3
...
   }
 
+  inline GeomIndex GeometryModel::addGeometryObject(const JointIndex parent,                                                    
+                                                    const SE3 & placement,
+                                                    const std::string & geom_name,
+                                                    const std::string & mesh_path) throw(std::invalid_argument)
+  {
+
+    Index idx = (Index) (ngeoms ++);
+    const boost::shared_ptr<fcl::CollisionGeometry> geometry(new fcl::CollisionGeometry());
```
fvalenza: Nota: WITH_HPP_FCL  CollisionGeometry is a pure virtual Class. Travis fails because of it

jcarpent:
Ok. In this way, this PR has to be complete: the robotWrapper must be increased with all the methods which make easy the creation and and the addition of new visual or collision objects. 
Otherwise, one can create himself its own data structure with Python and works with it easily.

PR #297, opened 2016-08-08
Update the Frame the part + expose new methods

jcarpent:
Some methods have been set to deprecated inside the model (getFrameParent, getFrameName, etc) has they are directly accessible from the `frames` vector. A new clean of model will be proposed in the very next future.

### DIFF
--- a/bindings/python/model.hpp
+++ b/bindings/python/model.hpp
@@ -135,8 +135,11 @@ namespace se3
 
           .def("getFrameParent", &ModelPythonVisitor::getFrameParent)
           .def("getFramePlacement", &ModelPythonVisitor::getFramePlacement)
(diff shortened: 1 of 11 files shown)


jmirabel:
I do not agree with the rationale of making `getFrame...` deprecated. For instance, `model.frames[frameId].parent` and `model.getFrameParent(frameId)` are to me completely different for the two following reasons:
- `model.getFrameParent` is more robust to API changes.
- `model.getFrameParent` is more robust to bugs. It is good for people who uses the library and do not know the inside. Sanity checks could be added in `getFrameParent` implementation (some other `getFrame...` have sanity checks).

`model.frames[...]` is good for the internal algorithms, not for the users. If you really want a difference between `model.getFrame...(...)` and `model.frames[...]. ... `, add sanity checks to the former.

Issue #299, opened 2016-08-09
Stable API at the current status of devel branch


jmirabel:
Hi Pinocchio developpers,

The API of the devel branch is not stable enough for downstream libraries. Every week, something is changing in the API and I see a lot of issues which involves other API changes.

Would it be possible to have a branch which would be updated with the bug fixes but not all the API changes ? It could be the master branch (updated to the current devel branch) or another branch.

Thanks

jcarpent:
Currently, some methods or functions are set to deprecated, that means that they will be suppress in the near future. So, you have time to adapt your code.
During the meantime, the API is increased with new features. Then, they won't affect your code.

Concerning the Collision part, the API must be adapted to fulfil your requirements. Then it is hard to stabilise everything together.

jcarpent:
Also, if you want to directly discuss with Pinocchio developers, there is the email address pinocchio@laas.fr

jcarpent:
What I propose is to add a devel-stable or hop-stable branch inside the repo, which will be only filled with bug fixes. What do you think?

jmirabel:
Yep, a hpp-stable or devel-stable branch would be nice.

jcarpent:
Ok, I just pushed a new branch dedicated to the merge of Pinocchio inside HPP (https://github.com/stack-of-tasks/pinocchio/tree/hpp-devel). We'll try to maintain it up to date for bug fixes.

jcarpent:
Can we close this Issue?

PR #300, opened 2016-08-09
[C++] GeometryObject parent is a frame of type BODY

jmirabel:
Refers to https://github.com/humanoid-path-planner/hpp-pinocchio/issues/34

### DIFF
--- a/src/algorithm/geometry.hxx
+++ b/src/algorithm/geometry.hxx
@@ -34,7 +34,7 @@ namespace se3
     updateGeometryPlacements(model, data, model_geom, data_geom);
   }
   
-  inline void  updateGeometryPlacements(const Model &,
+  inline void  updateGeometryPlacements(const Model & model,
                                        const Data & data,
                                        const GeometryModel & model_geom,
                                        GeometryData & data_geom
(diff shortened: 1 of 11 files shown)


jmirabel:
I do not know how to update the Python bindings myself. Can someone give me a hint ?

I can as well remove the reference to Model in GeometryModel (and pass it by arguments when needed).

jcarpent:
I think we need to really think twice before adding a const ref on the model inside GeomModel. But I think it is mandatory for your case.

So to sum up the situation, the idea is we have two trees inside the mode:
- a minimal one which contains only joints with at least one DoF and the tree can be retrieve through `model.parents` vector
- a large one, which contains all the frames that a robot can contain: SENSORS, BODIES, JOINTS, FIXED JOINTS, OPERATIONAL_FRAMES, etc. The minimal one can be retrieve from the large one. This large tree is accessible trough `model.frames`vector.

The idea is then to allows frames to index for example body frames as parent and not only joints as it is currently done.

Can we discuss that tomorrow with @florent-lamiraux?

jcarpent:
I can then explain you how to update the Python interface. It is not difficult, but maybe hard to explain through the Github interface.

jmirabel:
> I think we need to really think twice before adding a const ref on the model inside GeomModel. But I think it is mandatory for your case.

I can just add an argument to every function requiring the Model and not add a const ref to the Model, cannot I ?

jmirabel:
I have updated the code to remove the const reference to Model in GeometryModel.

Following #299, I close this pull-request and push to branch hpp-devel.

jmirabel:
Note that some bugs in the URDF parser have been fixed.

PR #302, opened 2016-08-23
Hpp devel

jmirabel:
I open the pull-request to start discussing the changes I did to integrate Pinocchio in HPP.

## Mandatory changes
- URDF parser: allow bodies without inertia. There could be a check somewhere in Pinocchio.
- SRDF parser: disable collisions between link from URDF and not between all pairs of geometry between two joints because fixed joints are merged with their parent.

## Convenient changes (but there exists workaround)
- Geometry object does not necessarily need a `parentFrame`. We could allow `-1` when not set.
- Removing `CollisionResult`
- `GeometraData::computeCollision` and `GeometryData::isColliding` API changes. I found they give more flexibility.


src/parsers/urdf/model.cpp
```diff
@@ -33,13 +33,27 @@ namespace se3
   {
     namespace details
     {
+      void appendBodyToJoint(Model& model, const Model::JointIndex jid,
+                             const boost::shared_ptr< ::urdf::Inertial> Y,
+                             const SE3 & placement,
+                             const std::string & body_name)
```
jmirabel: I hesitated between modifying the method `Model::appendBodyToJoint` and creating this function. I think the check for empty inertias would best fit in `Model::appendBodyToJoint` (other parsers would make use of it).

src/multibody/model.hpp
```diff
@@ -503,7 +503,7 @@ namespace se3
     ///
     /// \param[in] model The model structure of the rigid body system.
     ///
-    Data (const Model & model);
+    explicit Data (const Model & model);
```
jmirabel: This avoids silly typos being interpreted as type conversion. This should be a compilation error:
```cpp
se3::Data data = model
```

jmirabel:
I haven't checked the Python bindings which are generating errors.

jmirabel:
Related to https://github.com/humanoid-path-planner/hpp-pinocchio/issues/34

src/multibody/model.hpp
jcarpent: :+1:
As discussed with @nmansard, the majority of the constructors must be explicit either in multi body or in spatial classes

src/algorithm/geometry.hxx
```diff
@@ -181,6 +181,57 @@ namespace se3
...
+    }
+    BOOST_FOREACH(const Map_t::value_type& outerObject, geomModel2.outerObjects)
+    {
+      GeomIndexList& outerGeoms = geomModel1.outerObjects[outerObject.first];
+      BOOST_FOREACH(const GeomIndex& gid, outerObject.second)
+      {
+        outerGeoms.push_back(nGeom1 + gid);
+      }
+    }
+
```
jcarpent: You can add a reserve at this stage as you already know the number of the added collision pairs

jmirabel: I thought `GeomIndexList` was a `std::list`...

jcarpent: I was talking about `geomModel1.collisionPairs`

jmirabel: It is already done at line 199

src/algorithm/geometry.hpp
```diff
@@ -93,6 +93,14 @@ namespace se3
                                 const GeometryModel & geomModel,
                                 GeometryData &        geomData);
 #endif // WITH_HPP_FCL
+
```
nmansard: Could you describe a little bit further the content of this algo?

src/multibody/fcl.hpp
```diff
@@ -116,43 +116,6 @@ namespace se3
     
   }; // struct DistanceResult
```
nmansard: I think removing CollisionResult is a good idea.

src/multibody/fcl.hpp
```diff
@@ -189,23 +152,27 @@ struct GeometryObject
   /// \brief Name of the geometry object
   std::string name;
 
+  /// \brief Index of the parent frame
+  FrameIndex parentFrame;
```
nmansard: This is quite harmless to add this attribute. I dont like the naming. joint and frame parents seems to have the same level of importance, while parentFrame is only a "decoration", not used in any algorithm. I would suggest to keep the naming parent for the joint, and find something more explicit for remembering the body (bodyFrameId?)
parentFrame should be of type BODY, isnt it? If so, it should be explicit in the doc.

src/multibody/geometry.hpp
```diff
@@ -74,15 +73,25 @@ namespace se3
     /**
      * @brief      Add a geometry object to a GeometryModel
      *
-     * @param[in]  parent     Index of the parent joint
+     * @param[in]  object     Object 
+     *
+     * @return     The index of the new added GeometryObject in geometryObjects
+     */
+    inline GeomIndex addGeometryObject(const GeometryObject& object);
```
nmansard: I guess this design should be propagated everywhere in Pinocchio?

src/multibody/geometry.hpp
```diff
@@ -260,10 +272,11 @@ namespace se3
     /// \brief Compute the collision status between two collision objects of a given collision pair.
     ///
     /// \param[in] pair The collsion pair.
+    /// \param[out] result where the collision result is stored.
     ///
     /// \return Return true is the collision objects are colliding.
     ///
-    CollisionResult computeCollision(const CollisionPair & pair) const;
+    bool computeCollision(const CollisionPair & pair, fcl::CollisionResult& result) const;
```
nmansard: result should be taken from GeometryData, where a vector of collision results has been allocated (it was already a todo for the previous version).

src/multibody/geometry.hpp
```diff
@@ -274,8 +287,15 @@ namespace se3
     
     ///
     /// \brief Check if at least one of the collision pairs has its two collision objects in collision.
+    ///        The results are stored in the vector GeometryData::collision_results.
+    ///
+    /// \param[out] the index of the colliding pair if the function returns true.
     ///
-    bool isColliding() const;
+    bool isColliding(Index& collidingPair);
```
nmansard: I dont like too much to have a redundancy of methods doing very closed stuff. Could we pass the collidingPair results by side effect in the GeoemtryData structure? And then only keep the second method, with explicit description of the side effect in the doc?

src/multibody/geometry.hxx
```diff
@@ -136,8 +143,15 @@ namespace se3
   {
     removeAllCollisionPairs();
     for (Index i = 0; i < ngeoms; ++i)
+    {
```
nmansard: Please make explicit in the doc that "all pairs" means only non stupid ones.

src/multibody/model.hpp
nmansard: +1

src/parsers/urdf/model.cpp
nmansard: we can discuss that later.

src/parsers/urdf/model.cpp
```diff
@@ -450,66 +463,15 @@ namespace se3
...
-          {
-            parseTree(child, model, SE3::Identity(), verbose);
-          }
-        }
-        
-        
+
+        // FIXME: check the inertias
+        // The check in appendBodyToJoint only ensures inertias are not NaN
+        // but pinocchio requires non-zero inertias.
```
nmansard: zeros inertia are a problem for some algorithms only. ABA for sure. CRBA possibly (at least the Chol decomposition). I think it is acceptable to have zero inertias. Similarly, the kine tree should be arranged in a particular (compact) way, but only for the CRBA.
The verification should be done as a post process, maybe in the algorithms, and doublecheck automatically in debug mode when starting the particular algorithm. If we want to make the process explicit for all users, we could make the check when creating Data, with an argument like "checkValidity = ALL_ALGORITHMS", that could be modified by advanced users as "= NO_CHECK" or "= GEOM_ALGORITHMS | RNE_ALGORITHM", for example.

nmansard: Dont bother yet.

src/parsers/urdf/model.cpp
```diff
@@ -524,48 +486,16 @@ namespace se3
...
-            throw std::invalid_argument(exception_message);
-          }
-        
-        }
-        else // otherwise, it is a plain body with inertial info. It processes as usual.
+        addJointAndBody(model,root_joint,
+            0,SE3::Identity(),"root_joint",
+            root_link->inertial,root_link->name);
+
+        BOOST_FOREACH(::urdf::LinkPtr child, root_link->child_links)
```
nmansard: I like this modification! Is it really working?

nmansard:
In brief, you added a "frame" parent in geometry objects, and marginally modified URDF and SRDF parsers to make them more flexible. I like the changes, I think we could work to merge them with the main branch.

src/multibody/fcl.hpp
jmirabel: > parentFrame should be of type BODY
I agree, however, I don't think it should be mandatory to have a parentFrame. I would rather have a condition like `parentFrame == -1 || getFrameType(parentFrame) == BODY`

src/multibody/geometry.hpp
jmirabel: This is mostly used internally like `computeCollision(colPairIndex, collisionResults[colPairIndex])`. Do you see any disadvantage in this ?
I did this way because:
1. I don't think it hurts,
2. it was useful somewhere when I wrote it but I forgot where... (yes, I know it is a bad reason)

src/parsers/urdf/model.cpp
jmirabel: Yes, it seems to. I still have a very few unpushed bug fixes.

src/multibody/geometry.hpp
jmirabel: From a design point of view, adding a field to `GeometryData` for this seems to me a bit too much. But it is definitely possible.

nmansard: I prefer adding an attribute in the data than a method in the model.

src/multibody/geometry.hpp
nmansard: I think this is an initial design mistake. The general idea is that all computations in pinocchio needing buffer memory should use Data and GeomData for storage. Here, we are asking for an external buffer. I think we should stick to the initial desig.

src/multibody/fcl.hpp
nmansard: Fair enough. After reviewing the code entirely, it seems that the second term of the assert is not mandatory, although it is always true this only your parser is creating this. It would be nice to doublecheck if this assertion is needed or not, and add it in the doc if it is.

src/multibody/geometry.hxx
jmirabel: The doc in `src/multibody/geometry.hpp` is
```cpp
     ///
     /// \brief Add all possible collision pairs.
     ///
    /// \note Collision pairs between geometries of having the same parent joint
    ///       are not added.
    ///
    void addAllCollisionPairs();
```

You want it more explicit ?

nmansard: Good enough. Thx

src/multibody/geometry.hxx
```diff
@@ -191,18 +205,17 @@ namespace se3
...
 
-  inline CollisionResult GeometryData::computeCollision(const CollisionPair & pair) const
+  inline bool GeometryData::computeCollision(const CollisionPair & pair, fcl::CollisionResult& collisionResult) const
   {
     const Index & co1 = pair.first;     assert(co1<collisionObjects.size());
     const Index & co2 = pair.second;    assert(co2<collisionObjects.size());
 
-    fcl::CollisionRequest collisionRequest (1, false, false, 1, false, true, fcl::GST_INDEP);
-    fcl::CollisionResult collisionResult;
+    static const fcl::CollisionRequest collisionRequest (1, false, false, 1, false, true, fcl::GST_INDEP);
```
jmirabel: I used `static const fcl::CollisionRequest` but I think it would be better to have a `fcl::CollisionRequest` member in `GeometryData`.

nmansard: You may want to benchmark that. My guess is that, since the object is only locally used in the function, there would be no gain in pre-allocating it in Data. I dont think neither that the keyword static is interesting here, and might be counterproductive if we parallelize the calling algo.

jcarpent:
Just a minor remark, but you have done a lot of modifications inside the geometry module ... and a part of the documentation is not up to date. 

Will it possible to synchronise the doc too?

src/multibody/geometry.hxx
jcarpent: Here const seems to be sufficient and will be handled by the compiler itself.

jcarpent:
In computeBodyRadius algo, you can be more efficient by just expressing the translation of the geometry regarding to the parent frame inside the geometry frame itself. It avoids multiple call to SE3::act.  
But I guess this is not called every time.

src/parsers/urdf/model.cpp
jcarpent: We already discussed that point few times ago. The idea is to add a method inside Model class to check the dynamic validity of the model and other stuff like that. In such a way that dynamic consistency can be checked after building with other parsers (LUA, Python, etc).
I think this is not the job of Data.

src/multibody/geometry.hxx
jmirabel: My concern is not pre-allocation as `const` or `static const` are pre-allocated anyway. The goal is to give to the user the ability to change the request. For instance, one may want to enable contact points computation. I guess @stonneau uses this FCL feature.

I do not see why `static` would be a problem when parallelizing the calling algo.

From here, I'm not sure of my answer. Here is how I see this question of static vs non-static:

I think the main difference between the `static const` and `const` is the place where there are some space allocated to the object.
With `const`, some memory is allocated in the function stack. With `static const`, the memory is allocated within the instruction set. For small objects, I guess allocating in the function stack does not hurt.

src/multibody/fcl.hpp
jmirabel: The same should be done for `DistanceResult` but I keep this for later.

src/multibody/geometry.hpp
jmirabel: You mean you prefer an attribute in `GeometryData` than a method in `GeometryData` ?

src/parsers/urdf/model.cpp
nmansard: Please continue the discussion in #306

src/algorithm/geometry.hpp
jmirabel: 309ae485b05b6065976eccea9e66587ba589742a

src/algorithm/geometry.hpp
```diff
@@ -93,6 +93,21 @@ namespace se3
...
+  /// Append geomModel2 to geomModel1
+  ///
+  /// The steps for appending are:
+  /// \li add GeometryObject of geomModel2 to geomModel1,
+  /// \li add the collision pairs of geomModel2 into geomModel1 (indexes are updated)
+  /// \li add all the collision pairs between geometry objects of geomModel1 and geomModel2.
+  /// \li update the inner objects of geomModel1 with the inner objects of geomModel2
+  /// \li update the outer objects (see TODO)
+  ///
+  /// \warning Radius should be recomputed.
```
nmansard: Maybe all radius of joints whose body list has been increased should be set to nan? at least in debug mode.

src/algorithm/geometry.hxx
```diff
@@ -173,14 +174,67 @@ namespace se3
...
+    Index nColPairs1 = geomModel1.collisionPairs.size();
+    assert (geomModel2.ngeoms == geomModel2.geometryObjects.size());
+    Index nGeom2 = geomModel2.geometryObjects.size();
+    Index nColPairs2 = geomModel2.collisionPairs.size();
+
+    /// Append the geometry objects and geometry positions
+    geomModel1.geometryObjects.insert(geomModel1.geometryObjects.end(),
+        geomModel2.geometryObjects.begin(), geomModel2.geometryObjects.end());
+    geomModel1.ngeoms += nGeom2;
+
```
nmansard: Add here:
BOOST_FOREACH(auto geom, geomModel2.geometryObjects) geomData.radius[geom.jointParent] = NaN;

src/math/quaternion.hpp
```diff
@@ -61,5 +61,20 @@ namespace se3
...
+  /// of the normalization function.
+  ///
+  /// Only additions and multiplications are required. Neither square root nor
+  /// division are used (except a division by 2).
+  ///
+  /// \warning \f$ ||q||^2 - 1 \f$ should already be close to zero.
+  template <typename D> void
+    firstOrderNormalize(const Eigen::QuaternionBase<D> & q)
+  {
+    typedef typename D::Scalar Scalar;
```
nmansard: add a assert for the warning
assert(q.norm()-1<1e-2); //  ||q||^2 - 1 should already be close to zero.

src/multibody/fcl.hpp
```diff
@@ -189,23 +152,27 @@ struct GeometryObject
   /// \brief Name of the geometry object
   std::string name;
 
+  /// \brief Index of the parent frame
```
nmansard: Add here
/// Parent frame may be unset (set to -1) as it is mostly used as a documentation of the tree, or in third-party libraries. 
/// The URDF parser of Pinocchio is setting it to the proper value according to the urdf link-joint tree. 
/// In particular, anchor joints of URDF would cause parent frame to be different to joint frame.

src/multibody/geometry.hxx
nmansard: Finally, you did not add CollisionRequest to geomdata. I was quite convinced by your argumentation. Dont you think it is worth it?

src/multibody/joint/joint-free-flyer.hpp
```diff
@@ -275,21 +276,37 @@ namespace se3
       res.head<3>() = M1.translation();
       QuaternionMap_t res_quat(res.tail<4>().data());
       res_quat = M1.rotation();
+      firstOrderNormalize(res_quat);
```
nmansard: Add here:
// Norm of qs might be epsilon-different to 1, so M1.rotation might be epsilon-different to a rotation matrix.
// It is then safer to re-normalized after converting M1.rotation to quaternion.

src/multibody/joint/joint-free-flyer.hpp
```diff
@@ -275,21 +276,37 @@ namespace se3
...
+      typedef Eigen::Map<Motion_t::Quaternion_t> QuaternionMap_t;
+      
       Eigen::VectorXd::ConstFixedSegmentReturnType<NQ>::Type & q_0 = q0.segment<NQ> (idx_q ());
       Eigen::VectorXd::ConstFixedSegmentReturnType<NQ>::Type & q_1 = q1.segment<NQ> (idx_q ());
 
       if (u == 0.) return q_0;
       else if( u == 1.) return q_1;
       else
       {
+        // TODO: If integrate takes an arguments (ConfigVector_t, TangentVector_t), then we can merely do:
```
nmansard: A possibility here would be to add a static (hidden) function to do the job, with the arguments you like, and that both integrate and interpolate would call. 
Is it worth it?

src/multibody/model.hpp
```diff
@@ -115,22 +115,24 @@ namespace se3
       , njoint(1)
       , nbody(1)
       , nFrames(0)
-      , inertias(1)
-      , jointPlacements(1)
+      , inertias(1, Inertia::Zero())
```
nmansard: Inertia of the world should be nan, and not 0.

src/multibody/model.hpp
```diff
@@ -158,20 +160,46 @@ namespace se3
...
+    ///
+    /// \param[in] body_name Name of the body.
+    /// \param[in] parentJoint Index of the parent joint.
+    /// \param[in] body_placement The relative placement of the body regarding to the parent joint. Set default to the Identity placement.
+    /// \param[in] previousFrame Index of the parent frame. If negative,
+    ///            the parent frame is the frame of the parent joint.
+    ///
+    /// \return The index of the new frame.
+    ///
+    FrameIndex addBodyFrame (const std::string& body_name,
```
nmansard: What is the interest of this method wrt to addFrame?

src/multibody/model.hxx
```diff
@@ -80,27 +80,48 @@ namespace se3
...
+  inline FrameIndex Model::addBodyFrame (const std::string& body_name,
+                                         const JointIndex& parentJoint,
+                                         const SE3 & body_placement,
+                                               int         previousFrame)
+  {
+    if (previousFrame < 0) {
+      previousFrame = getFrameId(names[parentJoint]);
+    }
+    if (previousFrame >= frames.size())
+      throw std::invalid_argument ("Frame not found");
```
nmansard: Pinocchio design is to rather assert invalid argument than throwing errors. We might discuss this design at large, but for now, could you stick to it?
>>>>>>> assert(previousFrame >= frames.size());

src/parsers/urdf/geometry.cpp
```diff
@@ -190,19 +190,19 @@ namespace se3
...
-          assert(link->getParent()!=NULL);
-          if (link->getParent() == NULL)
-          {
-            const std::string exception_message (link->name + " - joint information missing.");
-            throw std::invalid_argument(exception_message);
-          }
           std::vector< boost::shared_ptr< T > > geometries_array = getLinkGeometryArray<T>(link);
 
+          if (!model.existBodyName(link_name))
+            throw std::invalid_argument("No link " + link_name + " in model");
```
nmansard: Here, throw follows the current design.

src/parsers/urdf/model.cpp
nmansard: Please replace by: 
// Inertias might be zero due to the URDF file. This is not a bug. However, it would lead to some
// algorithms of Pinocchio not to work. 
// TODO: add an algorithm or a method in model to check the validity of the value wrt to the algorithms the user want to use.
// See also #306 on GitHub.

nmansard:
Thanks. Nearly good for me. Could you check the latest remarks? 
Then you also have to please Travis.

src/multibody/geometry.hxx
jmirabel: I wasn't sure what was the conclusion. See 4ebb5b9

src/math/quaternion.hpp
jmirabel: 44a8cbb

src/algorithm/geometry.hxx
jmirabel: Actually, I realize now I should not be talking about radius here. Radius is an attribute of `GeometryData`. When modifing `GeometryModel`, users should understand that the corresponding `GeometryData` is not valid anymore, shouldn't they ?

I would rather not add a `GeometryData` as argument.

src/multibody/fcl.hpp
jmirabel: 84b4324

src/multibody/joint/joint-free-flyer.hpp
jmirabel: 570413c

src/algorithm/geometry.hxx
nmansard: I agree.

src/multibody/joint/joint-free-flyer.hpp
jmirabel: We already discussed this with @jcarpent  and I think it should be discussed in a different issue.

src/multibody/model.hpp
jmirabel: NaN or Inf ?

I was told the reason for not initializing the inertia of the world. I agree with the argumentation.

src/multibody/joint/joint-free-flyer.hpp
nmansard: Ok. Then one of you should summarized the discussion in a new issue, so that we can keep track of it.

src/multibody/model.hpp
jmirabel: It is a helper. It keeps the API more consistent (`addJoint` / `addJointFrame`, `appendBodyToJoint` / `addBodyFrame`).

I find it more convenient to do:
```cpp
idx = model.addJoint(...);
model.addJointFrame(idx);
model.appendBodyToJoint(...);
model.addBodyFrame("body_name", idx);
```
instead of:
```cpp
idx = model.addJoint(...);
/* parentFrameId = */ model.addJointFrame(idx);
model.appendBodyToJoint(...);
// compute parentFrameId, something like
// parentFrameId = model.getFrameId("joint_name")
model.addFrame(Frame("body_name", idx, parentFrameId, placement, BODY));
```

src/multibody/model.hpp
jmirabel: e7fa077

src/multibody/model.hxx
jmirabel: 7a9c81f

src/parsers/urdf/model.cpp
jmirabel: 1d0039d

src/multibody/joint/joint-free-flyer.hpp
jmirabel: #312

src/multibody/geometry.hpp
```diff
@@ -220,20 +232,35 @@ namespace se3
     std::vector <DistanceResult> distance_results;
     
     ///
+    /// \brief Defines what information should be computed by collision test.
+    ///
+    fcl::CollisionRequest collisionRequest;
```
nmansard: Should this not be a vector rather than a unique object? It makes the API slightly more complex, but consistant with the two other vectors containing FCL objects. Maybe such a possibility would never be used...

jmirabel: I do not see any use for this but, yes, it could be. Do you see any use for it ?

nmansard: Nope. It just seems very logical to do it. If you dont see any use, lets not do it :)

jmirabel:
Last commits (86675ad, 26ca737, 1d53aec) solves #313

src/multibody/model.hpp
```diff
@@ -124,6 +124,9 @@ namespace se3
       , gravity( gravity981,Eigen::Vector3d::Zero() )
     {
       names[0]     = "universe";     // Should be "universe joint (trivial)"
+      // FIXME Should the universe joint be a FIXED_JOINT even if it is
+      // in the list of joints ? See comment in definition of
+      // Model::addJointFrame and Model::addBodyFrame
       addFrame(Frame("universe", 0, 0, SE3::Identity(), FIXED_JOINT));
```
jmirabel: Naively, I added frame "universe" with type FIXED_JOINT.

I realize now it can also be considered a JOINT because it is in `model.joints`. Any opinion ?

src/multibody/model.hpp
```diff
@@ -275,19 +275,21 @@ namespace se3
     /// (for example to get the id of a parent frame).
     /// 
     /// \param[in] name Name of the frame.
+    /// \param[in] type Type of the frame.
     ///
     /// \return Index of the frame.
     ///
-    FrameIndex getFrameId (const std::string & name) const;
+    FrameIndex getFrameId (const std::string & name, const FrameType& type = (FrameType) (JOINT | FIXED_JOINT | BODY | OP_FRAME | SENSOR )) const;
```
jmirabel: The type of `type` could also be `int`. It would avoid having to cast ORed values.

src/multibody/model.hxx
```diff
@@ -156,18 +168,22 @@ namespace se3
...
+        && "Several frames match the filter");
     return Model::FrameIndex(it - frames.begin());
   }
 
-  inline bool Model::existFrame ( const std::string & name ) const
+  inline bool Model::existFrame ( const std::string & name, const FrameType & type) const
   {
-    return std::find_if( frames.begin(), frames.end(), boost::bind(&Frame::name, _1) == name) != frames.end();
+    return std::find_if( frames.begin(), frames.end(),
+        details::FilterFrame (name, type)) != frames.end();
```
jmirabel: I did not add the same test as in `getFrameId` because I am not sure it is not an error here to have several frames of the same (name + "type mask"). The function must check whether a frame matches a "filter", but I do not know whether it should check it is unique.

src/multibody/model.hpp
```diff
@@ -170,10 +170,10 @@ namespace se3
     /// \param[in] frameIndex Index of the parent frame. If negative,
     ///            the parent frame is the frame of the parent joint.
     ///
-    /// \return The index of the new frame.
+    /// \return The index of the new frame or -1 in case of error.
     ///
-    FrameIndex addJointFrame (const JointIndex& jointIndex,
-                                    int         frameIndex = -1);
+    int addJointFrame (const JointIndex& jointIndex,
```
jmirabel: A typedef for signed indexes would be useful.

jmirabel:
Congrats to all the participant of the pull-requests, I think most of the requirements for HPP are now achieved !

nmansard:
Any chance that we could merge this branch today?

jcarpent:
It will be good in few seconds. We need to rebase first hpp-devel on top of devel.

PR #304, opened 2016-08-26
[C++] Added methods & algo to test if 2 configs are equivalent

fvalenza:


src/multibody/joint/joint-free-flyer.hpp
```diff
@@ -362,6 +362,15 @@ namespace se3
       q.segment<4>(idx_q()+3).normalize();
     }
 
+    bool isSameConfiguration_impl(const Eigen::VectorXd& q1, const Eigen::VectorXd& q2) const
+    {
+      Eigen::VectorXd::ConstFixedSegmentReturnType<NQ>::Type & q_1 = q1.segment<NQ> (idx_q ());
+      Eigen::VectorXd::ConstFixedSegmentReturnType<NQ>::Type & q_2 = q2.segment<NQ> (idx_q ());
+
+      return ( q_1.head<3>().isApprox(q_2.head<3>()) 
+              && defineSameRotation(Eigen::Quaterniond(q_1.tail<4>()), Eigen::Quaterniond(q_2.tail<4>())) );
```
jcarpent: You must use maps on quaternions here

src/multibody/joint/joint-base.hpp
```diff
@@ -356,6 +357,15 @@ namespace se3
      */
     void normalize_impl(Eigen::VectorXd &) const { }
 
+    /**
+     * @brief      Check if the DoFs of the joint are equivalents for two configurations
+     *
+     * @param[in]  q1     Configuration 1 (size full model.nq)
+     * @param[in]  q2    Configuration 2 (size full model.nq)
+     */
+    bool isSameConfiguration(const Eigen::VectorXd & q1, const Eigen::VectorXd & q2) const
```
jcarpent: This is a static method of the JointModel class.

src/multibody/joint/joint-revolute.hpp
```diff
@@ -522,6 +522,14 @@ namespace se3
       return result; 
     } 
 
+    bool isSameConfiguration_impl(const Eigen::VectorXd& q1, const Eigen::VectorXd& q2) const
+    {
+      const Scalar & q_1 = q1[idx_q()];
+      const Scalar & q_2 = q2[idx_q()];
+
+      return (fabs(q_1 - q_2) < 1e-12);
```
jcarpent: You must use Eigen::NumTraits<Scalar>::dummy_precision()

jcarpent:
Thanks for the PR, but there is some minor corrections to add:
- you must add the default argument `prec` as in the isApprox method of all the spatial classes.
- the `isSameConfiguration`must  be a static method of the class

src/multibody/joint/joint-basic-visitors.hpp
```diff
@@ -174,6 +174,20 @@ namespace se3
                         Eigen::VectorXd & q);
 
   /**
+   * @brief      Visit a JointModelVariant through JointIsSameConfigurationVisitor to determine
+   *             if the two given configurations are equivalent or not.
+   *
+   * @param[in]      jmodel  The JointModelVariant
```
jcarpent: Please correct the indent here

src/multibody/joint/joint-base.hpp
```diff
@@ -356,6 +357,15 @@ namespace se3
      */
     void normalize_impl(Eigen::VectorXd &) const { }
 
+    /**
+     * @brief      Check if the DoFs of the joint are equivalents for two configurations
```
jcarpent: It should be rather: `Check if two configurations are equivalent within the given precision`

jmirabel:
> the `isSameConfiguration` must be a static method of the class

Why static ? I think it cannot be because it depends on `idx_q` for each joint.

src/multibody/joint/joint-dense.hpp
```diff
@@ -30,6 +30,7 @@ namespace se3
       NQ = _NQ, // pb
       NV = _NV
     };
+    typedef double Scalar;
```
nmansard: Scalar (if really needed) should not be declared in such a deep source file. 
Is Scalar already introduced otherwhere in pinocchio? If not, I would prefer to have a dedicated PR to introduce it everywhere (and we must discuss how to). If yes, it is very unfortunate.

src/multibody/joint/joint-base.hpp
```diff
@@ -25,6 +25,7 @@
 #include "pinocchio/spatial/inertia.hpp"
 #include "pinocchio/multibody/constraint.hpp"
 #include "pinocchio/multibody/fwd.hpp"
+#include "pinocchio/math/quaternion.hpp"
```
nmansard: why?

src/multibody/joint/joint-dense.hpp
fvalenza: Currently in all traits ( for Derived Motions, Constraints, and Joints), Scalar is redefined (always a typedef double Scalar).
We once talked about introducing a typedef float/double Scalar in a file such as math/fwd.h but it remained a hypothesis

jcarpent: In fact, the Scalar type was missing in the JointDense traits.

jcarpent:
Seems good to me. Concerning the Scalar type, it will be reworked globally in the future.

jmirabel:
Thanks

PR #310, opened 2016-09-06
Bug correction in joint visitor

nmansard:
[C++] Corrected a bug in joint visitor preventing proper access to attribute of joint models.


jcarpent:
Thanks for this PR and the major fix.
The next things to do is to correct the JointRevolute and JointPrimastic unaligned and avoid unnecessary axis copy. I'll open a new Issue for that.

Issue #281, opened 2016-07-28
Refactoring Geometry


nmansard:
I would like to implement the following modification in multibody/geometry.hpp

1. Remove classes DistanceResults and CollisionResults and directly work with FCL native types. For distance results, possibly add the 3 member functions as static functions (although I do not think they are vital). Keep operator== outside of the classes.

2. Remove NONE from the enum.

3. Move all the definitions before struct GeometryModel in a geometry-fwd.hpp

4. move inner/outer objects from GeomModel to GeomData. Stop automatically feeling them at model construction time. Possibly fill them automatically at data construction time, or explicitly with a dedicated method.

5. rename all xxx_yyy as xxxYyy, except oMg_fcl

6. It was suggested by @stonneau to rather store in data a CollisionObject that a fcl::Transform, which makes the call to fcl computations easier. 

7. Remove ref to GeomModel from GeomData. Consequently change the signature of the algorithms using GeomData, to explicitly add GeomModel as an argument.

8. Remove methods computeAllCollisions and computeAllDistance, as they are redundant with the algorithm geometry.hpp

9. Possibly also remove computeCollision and computeDistance, and put the two as external functions in algorithm/geometry. Change their internal process to avoid (stack) memory allocation of a Collision/DistanceResult object.

10. Remove displayCollisionPair.

11. Add a PairIndex type (like JointIndex and GeomIndex) to make arguments of some methods more explicit.

12. Modify addGeometryObject so that it simply takes a GeometryObject in argument.

Opinion, in particular from @fvalenza, welcome. Please help.

nmansard:
@jcarpent agreed on 6. I start implementing 6 that I need for hpp::pinocchio

fvalenza:
1/ If we work directly with fcl type, we lose the info concerning what are the collision pair ( the two GeomIndex) involved in this particular result.
2/ No particular opinion.  It was to force user to understand what is loaded by specifying it explicitly ( i.e using the GeometryModel as a CollisionModel or a VisualModel)
3/ Agreed. Some structs are even binded in python, so we could even put them in their own file may be?
4/ If we want to add/remove outerobjects or innerobjects on the fly, it may be better. Do we need to consider updated the list of collision pairs then ?
[...]

PR #314, opened 2016-09-07
Refactorize geometry module

nmansard:
Follow #281

### DIFF
--- a/src/algorithm/geometry.hxx
+++ b/src/algorithm/geometry.hxx
@@ -68,6 +68,7 @@ namespace se3
     const PairIndex & co1 = pair.first;     assert(co1<collisionObjects.size());
     const PairIndex & co2 = pair.second;    assert(co2<collisionObjects.size());
 
+    collisionResult.clear();
     fcl::collide (&geomData.collisionObjects[co1],&geomData.collisionObjects[co2],
                   geomData.collisionRequest,
                   collisionResult);
--- a/src/algorithm/geometry.hxx
+++ b/src/algorithm/geometry.hxx
@@ -27,28 +27,28 @@ namespace se3
   /* --- GEOMETRY PLACEMENTS -------------------------------------------------------- */
   inline void updateGeometryPlacements(const Model & model,
                                       Data & data,
-                                      const GeometryModel & model_geom,
-                                      GeometryData & data_geom,
+                                      const GeometryModel & geomModel,
+                                      GeometryData & geomData,
                                       const Eigen::VectorXd & q
                                       )
   {
     forwardKinematics(model, data, q);
-    updateGeometryPlacements(model, data, model_geom, data_geom);
+    updateGeometryPlacements(model, data, geomModel, geomData);
   }
   
   inline void  updateGeometryPlacements(const Model &,
                                        const Data & data,
-                                       const GeometryModel & model_geom,
-                                       GeometryData & data_geom
+                                       const GeometryModel & geomModel,
+                                       GeometryData & geomData
                                        )
   {
-    for (GeomIndex i=0; i < (GeomIndex) model_geom.ngeoms; ++i)
+    for (GeomIndex i=0; i < (GeomIndex) geomModel.ngeoms; ++i)
     {
-      const Model::JointIndex & joint = model_geom.geometryObjects[i].parentJoint;
-      if (joint>0) data_geom.oMg[i] =  (data.oMi[joint] * model_geom.geometryObjects[i].placement);
-      else         data_geom.oMg[i] =  model_geom.geometryObjects[i].placement;
+      const Model::JointIndex & joint = geomModel.geometryObjects[i].parentJoint;
+      if (joint>0) geomData.oMg[i] =  (data.oMi[joint] * geomModel.geometryObjects[i].placement);
+      else         geomData.oMg[i] =  geomModel.geometryObjects[i].placement;
 #ifdef WITH_HPP_FCL  
-      data_geom.collisionObjects[i].setTransform( toFclTransform3f(data_geom.oMg[i]) );
+      geomData.collisionObjects[i].setTransform( toFclTransform3f(geomData.oMg[i]) );
 #endif // WITH_HPP_FCL
     }
   }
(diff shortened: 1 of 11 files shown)


jmirabel:
I guess there is an issue with the filenames:
- unittest/#com.cpp#
- unittest/#tspatial.cpp#
- unittest/urdg.cpp
- unittest/test_matrix_prod.cpp (this may not be an error, I'm just surprised to see it appearing now.)

src/algorithm/geometry.hxx
```diff
@@ -206,34 +261,48 @@ namespace se3
...
     }
 
-    /// 3. add the collision pairs between geomModel1 and geomModel2.
-    for (Index i = 0; i < nGeom1; ++i) {
-      for (Index j = 0; j < nGeom2; ++j) {
-        geomModel1.collisionPairs.push_back(CollisionPair(i, nGeom1 + j));
-      }
-    }
+    // FIXME: I copy here the previous algo. Maybe it would be safer to simply use the
+    // fillInnerOuterObjectMaps() method (however marginally less efficient: is it important?).
```
jmirabel: Efficiency is not a very important matter here.

For a record of our discussion: there are a few additional attributes of `geomData1` that should be updated (`activeCollisionPairs`...).

src/multibody/geometry.hxx
```diff
@@ -31,30 +31,48 @@
...
-    Index idx = (Index) (ngeoms ++);
+    assert( (object.parentFrame != -1) || (object.parentJoint != -1) );
+
+    if( object.parentJoint == -1 )  // if no parent joint, autofill the attribute.
+      object.parentJoint = model.frames[object.parentFrame].parent; 
+
+    assert( (object.parentFrame == -1) 
+            || (model.frames[object.parentFrame].type == se3::BODY)  );
+    assert( (object.parentFrame == -1) 
+            || (model.frames[object.parentFrame].parentJoint == object.parentJoint) );
```
jmirabel: `parentFrame` and `parentJoint` are of type `std::size_t` so they will never be `-1`.

jmirabel: `model.frames[object.parentFrame].parentJoint` should be `model.frames[object.parentFrame].parent`

src/algorithm/geometry.hxx
nmansard: Should I simply remove this second function?

src/multibody/geometry.hxx
nmansard: -1: right. I dont like unsigned int, and here is a new reason for that. 
I thought that we said that parentFrame would be set to -1 when unset. Did I misunderstood your proposition? I was here simply following the same idea. 
For this particular place, we could simply remove this automatic behaviour (set parentJoint if the user set it to -1). Right now, the automatic behaviour is only used in the unittest, and is easy to modify.

Thx for the second bug. I indeed did not compile in debug mode before pushing.

src/algorithm/geometry.hxx
```diff
@@ -22,48 +22,72 @@
...
+                               GeometryData & geomData,
+                               const PairIndex& pairId)
+  {
+    const CollisionPair & pair = geomModel.collisionPairs[pairId];
+    fcl::CollisionResult& collisionResult = geomData.collisionResults[pairId];
+
+    const PairIndex & co1 = pair.first;     assert(co1<collisionObjects.size());
+    const PairIndex & co2 = pair.second;    assert(co2<collisionObjects.size());
+
+    collisionResult.clear();
```
nmansard: Should we have the same line in computeDistance?

src/algorithm/geometry.hxx
```diff
@@ -75,67 +99,96 @@ namespace se3
...
+                                               const PairIndex & pairId )
   {
-    return computeDistances<true>(data_geom);
+    assert( pairId < geomModel.collisionPairs.size() );
+    const CollisionPair & pair = geomModel.collisionPairs[pairId];
+
+    assert( pair        < geomData.distanceResults.size() );
+    assert( pair.first  < geomData.collisionObjects.size() );
+    assert( pair.second < geomData.collisionObjects.size() );
+
```
nmansard: Should we add here
geomData.distanceResults[pairId].clear()
???

src/multibody/geometry.hxx
```diff
@@ -173,100 +217,30 @@ namespace se3
...
-    {
-      if (activeCollisionPairs[i])
-        distance_results[i] = computeDistance(model_geom.collisionPairs[i]);
-    }
-  }
-
   inline void GeometryData::resetDistances()
   {
-    std::fill(distance_results.begin(), distance_results.end(), DistanceResult( fcl::DistanceResult(), 0, 0) );
+    std::fill(distanceResults.begin(), distanceResults.end(), fcl::DistanceResult() );
```
nmansard: For collisions, we do the reset during the collision computation (in computeCollision(), collisionResult.clear()).
For distances, we do the reset with this method. 

I guess we should have it with an homogeneous manner. I suggest to remove resetDistances and do the clear in computeDistance(). Would you aggree?

src/multibody/geometry.hxx
jmirabel: I remember I said it could be -1 as it is not a necessary information. I
guess I forgot it is not signed. Sorry about that.

There are a few places where I used int instead of FrameIndex because of
this. It's ugly. I think we should add a typedef SignedIndex. Let's keep
this for later.

For now, we can simply comment the automatic detection with a todo to
remind us to re enable it when relevant.

src/algorithm/geometry.hxx
jmirabel: I would say it's rather dangerous to keep a function that doesn't do the
job it's is meant to do. Creating a new data from geomModel1 solves the
issue, right? If yes, removing is better.

Joseph Mirabel

Le 8 sept. 2016 09:50, "Nicolas Mansard" <notifications@github.com> a
écrit :

src/multibody/geometry.hxx
jmirabel: I agree.

jmirabel:
I think it is ready to be merged.

PR #318, opened 2016-09-08
Topic/joint composite

fvalenza:
This PR adds a JointComposite as a specialised Joint (recursive variant)

Please note that some computations are still wrong ( see ./unittest/joint-composite , finite-differences), help is very welcome for such computations ( bias for example ).


unittest/joint-configurations.cpp
```diff
@@ -186,48 +216,9 @@ struct TestDifferentiationJoint
...
     TV qdot = jmodel.difference(q0,q1);
 
-    BOOST_CHECK_MESSAGE( jmodel.integrate(q0, qdot).isApprox(q1), std::string("Error in difference for joint " + jmodel.shortname()));
+    BOOST_CHECK_MESSAGE( jmodel.isSameConfiguration(jmodel.integrate(q0, qdot), q1), std::string("Error in difference for joint " + jmodel.shortname()));
 
-    BOOST_CHECK_MESSAGE( jmodel.integrate(q1, -qdot).isApprox(q0), std::string("Error in difference for joint " + jmodel.shortname()));
-
-  }
-
-  void operator()(JointModelBase<JointModelFreeFlyer> & jmodel)
```
nmansard: Why did you remove this test?

unittest/joint-configurations.cpp
```diff
@@ -186,48 +216,9 @@ struct TestDifferentiationJoint
...
-    
-    TV qdot = jmodel.difference(q0,q1);
-    
-    BOOST_CHECK_MESSAGE( jmodel.integrate(q0, qdot).head<3>().isApprox(q1.head<3>()), std::string("Error in difference for joint " + jmodel.shortname()));
-    BOOST_CHECK_MESSAGE( defineSameRotation(Eigen::Quaterniond(jmodel.integrate(q0, qdot).tail<4>()), Eigen::Quaterniond(q1.tail<4>()))
-                      , std::string("Error in difference for joint " + jmodel.shortname()));
-
-  }
-
-  void operator()(JointModelBase<JointModelSpherical> & jmodel)
```
nmansard: why ? same below

src/multibody/model.hpp
```diff
@@ -141,22 +141,44 @@ namespace se3
...
+    /// \param[in] min_config Lower joint configuration.
+    /// \param[in] max_config Upper joint configuration.
     ///
     /// \return The index of the new joint.
     ///
     /// \sa Model::appendBodyToJoint
     ///
     template<typename JointModelDerived>
     JointIndex addJoint(const JointIndex parent, const JointModelBase<JointModelDerived> & joint_model, const SE3 & joint_placement,
-                        const std::string & joint_name = "",
```
nmansard: please dont change the API here, rather remove the default argument for jointName (keep it for the new signature you added below.

Nota: this has nothing to do with the actual PR, should be pushed separately. Lets keep it like this for once.

src/algorithm/crba.hxx
```diff
@@ -161,27 +161,41 @@ namespace se3
...
-      typedef typename SizeDepType<JointModel::NV>::template ColsReturn<Matrix6x>::Type ColsBlock;
+      typedef typename SizeDepType<JointModel::NV>::template ColsReturn<Data::Matrix6x>::Type ColsBlock;
       
       const Model::JointIndex & i = (Model::JointIndex) jmodel.id();
       const Model::Index & parent = model.parents[i];
       
       data.Ycrb[parent] += data.liMi[i].act(data.Ycrb[i]);
       
       jdata.U() = data.Ycrb[i] * jdata.S();
-      if (JointModel::NV == -1)
```
nmansard: I agree this ==-1 should never occur now JointDense is about to disappear.

src/algorithm/crba.hxx
```diff
@@ -161,27 +161,41 @@ namespace se3
...
+      typedef SizeDepType<JointModel::NV>::ColsReturn<Data::Matrix6x>::Type ColsBlock;
+      
+      const Model::JointIndex & i = (Model::JointIndex) jmodel.id();
+      const Model::Index & parent = model.parents[i];
+      
+      data.Ycrb[parent] += data.liMi[i].act(data.Ycrb[i]);
+      
+      jdata.U() = data.Ycrb[i] * jdata.S();
+      
+      ColsBlock jF
```
nmansard: However, why is it necessary to reintroduce it here for Composite. It is a very bad design that algorithm should be specific to some joints. Dont you have a NV in joint composite? You should, it is possible to have a constant (known a compil time) from both included joints.

Please remove this specific code.

src/multibody/model.hpp
```diff
@@ -26,7 +26,7 @@
 #include "pinocchio/spatial/inertia.hpp"
 #include "pinocchio/multibody/fwd.hpp"
 #include "pinocchio/multibody/frame.hpp"
-#include "pinocchio/multibody/joint/joint.hpp"
+// #include "pinocchio/multibody/joint/joint.hpp"
```
nmansard: this should be ncessary.

src/multibody/model.hpp
```diff
@@ -26,7 +26,7 @@
...
 #include "pinocchio/multibody/joint/joint-composite.hpp"
```
nmansard: this should not be necessary

src/multibody/model.hxx
```diff
@@ -72,12 +72,12 @@ namespace se3
     nq += joint_model.nq();
     nv += joint_model.nv();
 
-    if (D::NV == -1)
+    if (D::NV == Eigen::Dynamic)
```
nmansard: If this is necessary here, then it also is in CRBA. You should have consistent design.
I think it is never necessary.

src/multibody/visitor.hpp
```diff
@@ -61,6 +61,16 @@ namespace se3
 			       static_cast<const Visitor*>(this)->args));
       }
 
+  //     void operator() (const JointModelBase<JointModelComposite> & jmodel) const
```
nmansard: Please clean this.

unittest/joint-composite.cpp
```diff
@@ -17,7 +17,7 @@
 
 #include "pinocchio/tools/timer.hpp"
 
-#include "pinocchio/multibody/joint/joint-accessor.hpp"
+// #include "pinocchio/multibody/joint/joint-accessor.hpp"
```
nmansard: idem

src/multibody/joint/joint-variant.hpp
```diff
@@ -33,14 +33,27 @@
...
-  
+  struct JointComposite;
+  struct JointModelComposite;
+  struct JointDataComposite;
+
+  typedef boost::variant< JointModelRX, JointModelRY, JointModelRZ, JointModelRevoluteUnaligned, JointModelSpherical,
+                          JointModelSphericalZYX, JointModelPX, JointModelPY, JointModelPZ,
+                          JointModelPrismaticUnaligned, JointModelFreeFlyer, JointModelPlanar, JointModelTranslation,
+                          JointModelDense<-1,-1> ,JointModelRUBX, JointModelRUBY, JointModelRUBZ,
+                          boost::recursive_wrapper<JointModelComposite> >JointModelVariant;
```
nmansard: please comment

src/algorithm/crba.hxx
fvalenza: The sizes NQ, NV for a JointComposite are not known at compilation as it can hold any type of joint. Hence these values are set to Eigen::Dynamic ( that's why there was previous tests like ' if NV == -1'.
This algo has been specialized specificly  for Jointcomposite in order not to loose the templated version of middleCols for other type of joints. I agree that duplicating the code for only this line is surely too much (I didn't succeed to have the 'ColsBlock jF' line compiling and properly set for both cases of JointComposite or other joints in only one function. One easy solution would be to use the dynamic version of middleCols for both cases. Should i change to this ?

src/multibody/model.hpp
fvalenza: It is included in joint-composite.hpp

src/multibody/model.hpp
fvalenza: Ok

unittest/joint-configurations.cpp
fvalenza: This was a specialization because of quaternions. Now that isSameConfiguration algo is available, this can be removed.

unittest/joint-configurations.cpp
fvalenza: Same here ( isSameConfiguration)

src/algorithm/crba.hxx
fvalenza: Please see commit 3af876e and the call to `middleCols` 
According to http://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#aa257ee10f662780e48cd9cec53bbbb86 it seems legit and the trade-off seems acceptable (for crba i have the following bench: from 4,35 us 90% of time & 10% : 4.5 us  to 4.4 us 60% of time, and 4.7 us otherwise )

fvalenza:
I am putting here a remark done on an outdatted commit:

According to (this)[http://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#aa257ee10f662780e48cd9cec53bbbb86] and (this)[http://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#ac0d50068b3e133e938202f4e23445aba] the changes made on commit 3af876e and commit 2c92049 compile on my computer. It doesn't on Travis.

nmansard:
I dont get your remark.

src/algorithm/crba.hxx
nmansard: Could you give the time without the change for comparison?

fvalenza: no changes : _4.35_ / 4.4 us
with changes: _4.55_ / 4.6 us

nmansard:
0.2us is indeed relevant as a score. So let's keep it separated.

The proper way of doing it would be:

template <D>
void algo(JointModelBase<D> jmodel)
{
     function_depending_on_NV<D::NV>();
     function2_depending_on_NV_and_joint<D::NV>::run(jmodel);
}

template <int NV>
void function_depending_on_NV() { ... }

template<int NV>
struct function2_depending_on_NV_and_joint
{
     template<D>
     static void run(JointModelBase<D> jmodel) { ... }
};

You can then specialize both functions for NV==Eigen::Dynamic.

If you feel confident, you could implement the corresponding pattern in 
CRBA. Alternatively, leave an open issue with this message.

fvalenza:
Using  `middleCols<JointModel::NV>(joint_model.idx_v())` is not possible with the JointComposite ( NV is Eigen::Dynamic).

Using `middleColls(joint_model.idx_v(), joint_model.nv())` is possible for JointComposite but not possible for other specialized joints because of the typedef for `ColsBlock` (`NColsBlockXpr` if JointModel::NV is fixed, `ColsBlockXpr` if JointModel::NV is Dynamic) and the return type of middleCols.
`middleCols<NV>` returns `NColsBlockXpr` and `middleCols(start, dim)` returns `ColsBlockXpr`. 
The conversion from `ColsBlockXpr` to `NColsBlockXpr` seems not possible. 

One possible solution ( see doc linked above) is to use `middleCols<JointModel::NV>(idx_v(), nv()) `even if JointModel::NV is Eigen::Dynamic,, and passing the real dimension even if it's not the same as JointModel::NV. 
This does not compile however on Travis ( but does on my computer)

nmansard:
Ok, that s another reason to keep the code separated.

src/multibody/model.hxx
```diff
@@ -72,20 +72,11 @@ namespace se3
...
-    }
-    else
-    {
-      effortLimit.conservativeResize(nv);effortLimit.bottomRows<D::NV>() = max_effort;
-      velocityLimit.conservativeResize(nv);velocityLimit.bottomRows<D::NV>() = max_velocity;
-      lowerPositionLimit.conservativeResize(nq);lowerPositionLimit.bottomRows<D::NQ>() = min_config;
-      upperPositionLimit.conservativeResize(nq);upperPositionLimit.bottomRows<D::NQ>() = max_config;
-    }
+
+    effortLimit.conservativeResize(nv);effortLimit.bottomRows<D::NV>(joint_model.nv()) = max_effort;
```
nmansard: replace by
// Optimal efficiency here would be using the static-dim bottomRows<D::NV>, while specifying the dimension in argument in the case where D::NV is Eigen::Dynamic. However, this option is not compiling in Travis (why?). 
// As efficiency of Model::addJoint is not critical, the dynamic bottomRows is used here.
velocityLimit.conservativeResize(nv);velocityLimit.bottomRows(joint_model.nv()) = max_velocity;

fvalenza: Done

fvalenza:
I close this PR as I have done a rebase on hpp-devel and added 2 commits. New PR to come

PR #322, opened 2016-09-14
[Python][hpp-fcl] Provide temporary operator== for DistanceResult.

fvalenza:
Previous PR has been merged before i pushed this commit.
As we don't know yet if hpp-fcl will integrate the operator== for DistanceResult, i've put it where it is needed in Pinocchio


nmansard:
You can merge asa travis agrees

Issue #311, opened 2016-09-06
Correct Joint{Revolute,Prismatic} Unaligned


jcarpent:
#310 solves a major issue for the access of class properties.
The next thing to do is to update the Joint{Revolute,Prismatic} Unaligned classes to remove unnecessary copies.

nmansard:
I suggest to keep the green "help wanted" for urgent issues.

fvalenza:
solved by #324

PR #324, opened 2016-09-15
removing deprecated methods + change unaligned joints with boost fusion fix

fvalenza:
Related to #311

### DIFF
--- a/src/multibody/joint/joint-revolute-unaligned.hpp
+++ b/src/multibody/joint/joint-revolute-unaligned.hpp
@@ -289,7 +289,6 @@ namespace se3
     Bias_t c;
 
     F_t F;
-    Eigen::AngleAxisd angleaxis;
     
     // [ABA] specific data
     U_t U;
@@ -298,13 +297,11 @@ namespace se3
 
     JointDataRevoluteUnaligned() 
       : M(1),S(Eigen::Vector3d::Constant(NAN)),v(Eigen::Vector3d::Constant(NAN),NAN)
-      , angleaxis( NAN,Eigen::Vector3d::Constant(NAN))
       , U(), Dinv(), UDinv()
     {}
     
     JointDataRevoluteUnaligned(const Motion::Vector3 & axis) 
       : M(1),S(axis),v(axis,NAN)
-      , angleaxis(NAN,axis)
       , U(), Dinv(), UDinv()
     {}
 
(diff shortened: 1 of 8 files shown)


src/multibody/joint/joint-revolute-unaligned.hpp
```diff
@@ -342,12 +339,8 @@ namespace se3
...
     {
       const double & q = qs[idx_q()];
-
-      /* It should not be necessary to copy axis in jdata, however a current bug
-       * in the fusion visitor prevents a proper access to jmodel::axis. A
-       * by-pass is to access to a copy of it in jdata. */
-      data.angleaxis.angle() = q;
-      data.M.rotation(data.angleaxis.toRotationMatrix());
+      
+      data.M.rotation(Eigen::AngleAxisd(q, axis).toRotationMatrix());
```
fvalenza: Is it ok to create an angleaxis each time ?

nmansard:
Less efficient.

src/multibody/joint/joint-revolute-unaligned.hpp
nmansard: I see no reasons not to keep an initialized AngleAxis in JData. 
Possibly add an assert to double check that angleaxis.axis == jmodel.axis (I guess not an isApprox but an exact ==).

src/multibody/model.hpp
```diff
@@ -292,43 +292,8 @@ namespace se3
...
     bool existFrame (const std::string & name, const FrameType& type = (FrameType) (JOINT | FIXED_JOINT | BODY | OP_FRAME | SENSOR )) const;
     
-    ///
-    /// \brief Get the name of a frame given by its index.
-    ///
-    /// \param[in] index Index of the frame.
-    ///
-    /// \return The name of the frame.
-    ///
-    PINOCCHIO_DEPRECATED const std::string & getFrameName (const FrameIndex index) const;
```
nmansard: I think it is relevant to remove that. It is a quite recent feature, and nobody should complain.

src/parsers/urdf.hpp
```diff
@@ -58,21 +58,6 @@ namespace se3
...
-    ///
-    /// \brief Build the model from a URDF file with a particular joint as root of the model tree.
-    ///
-    /// \param[in] filemane The URDF complete file path.
-    /// \param[in] rootJoint The joint at the root of the model tree.
-    /// \param[in] verbose Print parsing info.
-    ///
-    /// \return The se3::Model of the URDF file.
-    ///
-    PINOCCHIO_DEPRECATED
```
nmansard: Maybe here we should keep the deprecated method until v1.3. I am not sure. We are not very rigorous with that. I certainly will not complain.

jmirabel:
> Less efficient.

Really ? I would have said the current implementation is at least as efficient...
On one side, you marginally increase the memory footprint. On the other, you have (not even sure) a marginally bigger call stack... With good compilation options, I suspect the temporary `Eigen::AngleAxisd` is not even created.

Did somebody (@jcarpent) run some benchmark for such things ?

src/multibody/joint/joint-revolute-unaligned.hpp
nmansard: Ok, let's keep it as you proposed, with a temporary.

Issue #305, opened 2016-08-30
Bug when python wrapper built


jcarpent:
There is currently one bug in the packaging of Pinocchio.
When one compiles with Python bindings, eigenpy is added as a Pinocchio dependency in the .pc file.
This addition of such a dependancy leads to crashes when calling Python Interpreter inside an application depending on Pinocchio. This bug is recently appeared in the SoT application.

To solve this bug, I suggest to split the Python bindings from the pinocchio package and create an independent python_pinocchio package (depending on pinocchio, boost python, Python and eigenpy only).

olivier-stasse:
:+1:

nim65s:
You'll need something like
`git filter-branch --prune-empty --subdirectory-filter src/python devel`

nmansard:
Yes! And it would make exactly 100 packages maintained by the team (we currently have 99 packages and desperately looking for the 100th). 
Could we possibly  think of an alternative?

jcarpent:
I know and I understand your pessimism. But, to say the truth, we'll never reach HPP ;)

Another solution is to remove eigenpy inside the packaging procedure. So, it may be necessary to add a cmake macro as `ADD_REQUIRED_BUILD_DEPENDENCY`which will not add the package to the .pc file.

jcarpent:
Any feedback?

olivier-stasse:
Another point why we *SHOULD* fix this: Working on our pattern generator I found out that with the current status of pc file we have to bind python to be able to use pinocchio. This does not make sense.

olivier-stasse:
A possible solution would be to remove from pinocchio.pc the boost python component, the required dependency to eigenpy and generate by other means than the cmake modules an other .pc file.

olivier-stasse:
I have a possible solution implementing the above solution here:
https://github.com/olivier-stasse/pinocchio/tree/devel
One problem was coming the parser-python which IMHO should be move to libpinocchio_pywrap.
I still have a problem when doing make tests (parser-python is failing)

fvalenza:
@olivier-stasse
For the make test fails did you install the python bindings before (ie make install then make test)?

jcarpent:
I already implemented a fix for that issue. I will sync the devel branch accordingly ASAP.

jcarpent:
[...]

PR #333, opened 2016-09-21
Improve packaging

jcarpent:
It follows #305 and the code proposition of https://github.com/olivier-stasse/pinocchio/tree/devel.

- The PYTHON_PASER can be build only if the PYTHON_INTERFACE is built
- EigenPy is no more added in the .pc file


olivier-stasse:
Thanks this is in the good direction.
But keeping python.cpp in parser keeps links with python.
Then it creates python link error when linking with libpinocchio.
I do not see any alternative to put the file python.cpp in the libpinocchio_pywrap.

jcarpent:
I agree with @olivier-stasse. I will follow your proposition to move the python parser to the python lib. In this way, people who needs the python wrapper would need to link against pinocchiopy

jcarpent:
@olivier-stasse I propose to merge first this PR. I will open soon a new one concerning the shift of the Python parser to the Python lib.

olivier-stasse:
@jcarpent : I may have a solution for python_parser.cpp which does not imply to modify the code.
The key is to not bind python_parser to eigenpy, pinocchio and pinocchio_pywrap but simply to libboost-python and python. 
A small customization of python_parser cmake instructions is needed.
You can use the 4 lines of file 
unittest/CMakeLists.txt
in commit d264bf6d6d7e22339753dd09567dee270d0eafba
Of course this implies also to bind the python parser with libpinocchio_pywrap
This would be a great addition to your current PR.

jcarpent:
Thanks @olivier-stasse for your remarks. I'll take some inspiration from your proposal  to move the parser into the python lib. I'll do it in a new PR for clarity reasons.

jcarpent:
I solved the problem by fixing a bug in EigenPy. I just open a new PR fixing the linkage problem (https://github.com/stack-of-tasks/eigenpy/pull/12). 
I will then open a new PR to fix the bug in Pinocchio after making a new release of EigenPy available on robotpkg.

PR #360, opened 2016-11-28
Fix bug in computeBodyRadius

florent-lamiraux:
There was a confusion between radius and radius squared.


jcarpent:
Thanks for the fix.

PR #364, opened 2016-12-20
Planar joint uses cosine and sine of the angle.

jmirabel:
The python bindings do not compile. The error is:
```
fatal error: pinocchio/parsers/python.hpp: No such file or directory
```
I guess it is suppose to find `bindings/python/parsers/python.hpp`...

### DIFF
--- a/src/multibody/joint/joint-planar.hpp
+++ b/src/multibody/joint/joint-planar.hpp
@@ -424,25 +424,25 @@ namespace se3
         // M0 * Mu = ( R0 * Ru, R0 * tu + t0 )
 
         Eigen::VectorXd::ConstFixedSegmentReturnType<2>::Type v = vs.segment<2>(idx_v());
-        double ct,st; SINCOS (theta, &st, &ct);
+        Eigen::Matrix<Scalar,2,1> cst;
+        SINCOS (t, &cst[1], &cst[0]);
         const double inv_theta = 1/theta;
-        const double c_coeff = (1.-ct) * inv_theta;
-        const double s_coeff = st * inv_theta;
+        const double c_coeff = (1.-cst[0]) * inv_theta;
+        const double s_coeff = std::fabs(cst[1]) * inv_theta;
         const Vector2 Sp_v (-v[1], v[0]);
 
         if (t > 0) res.head<2>() = q.head<2>() + R0 * (s_coeff * v + c_coeff * Sp_v);
         else       res.head<2>() = q.head<2>() + R0 * (s_coeff * v - c_coeff * Sp_v);
-        res(2) = c0 * ct - s0 * st;
-        res(3) = s0 * ct + c0 * st;
+        res.tail<2>() = R0 * cst;
         return res;
       }
       else
       {
+        // cos(t) ~ 1 - t^2 / 2
+        // sin(t) ~ t
         res.head<2>() = q.head<2>() + R0*q_dot.head<2>();
-        // TODO: theta is small. Use a 1st order approx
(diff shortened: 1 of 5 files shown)


jmirabel:
Forget about the compilation error. The symbolic link was not updated by CMake.

src/multibody/joint/joint-planar.hpp
```diff
@@ -459,14 +466,17 @@ namespace se3
 
     ConfigVector_t random_impl() const
     { 
-      ConfigVector_t result(ConfigVector_t::Random());
+      Scalar angle = PI * (-1 + 2 * rand() / RAND_MAX);
+      ConfigVector_t result;
+      result.head<2>() = Eigen::Matrix<Scalar, 2, 1>::Random();
```
jcarpent: Just use setRandom here to be simple.

src/multibody/joint/joint-planar.hpp
```diff
@@ -478,6 +488,8 @@ namespace se3
         }
         result[i] = lower_pos_limit[i] + ( upper_pos_limit[i] - lower_pos_limit[i]) * rand()/RAND_MAX;
       }
+      Scalar angle = PI * (-1 + 2 * rand() / RAND_MAX);
```
jcarpent: You should cast your division to avoid wrong results.

src/multibody/joint/joint-planar.hpp
```diff
@@ -412,30 +415,34 @@ namespace se3
...
+        //   =      ( 1 - ct) / |t| * S +       st / |t|   * I2
         //
         // Ru = exp3 (w)
         // tu = R * v = (1 - ct) / |t| * S * v + st / t * v
         //
         // M0 * Mu = ( R0 * Ru, R0 * tu + t0 )
 
         Eigen::VectorXd::ConstFixedSegmentReturnType<2>::Type v = vs.segment<2>(idx_v());
-        double ct,st; SINCOS (theta, &st, &ct);
+        Eigen::Matrix<Scalar,2,1> cst;
```
jcarpent: I don't understand why you changed from `double ct, st`to an Eigen object.

src/multibody/joint/joint-planar.hpp
```diff
@@ -459,14 +466,17 @@ namespace se3
 
     ConfigVector_t random_impl() const
     { 
-      ConfigVector_t result(ConfigVector_t::Random());
+      Scalar angle = PI * (-1 + 2 * rand() / RAND_MAX);
```
jcarpent: Use const for efficiency reasons + cast division

src/multibody/joint/joint-planar.hpp
jmirabel: For `res.tail<2>() = R0 * cst;` which makes more sense to me than writing trigonometric relations (`cst` is a element of R^2 that we rotate with `R0`)

src/multibody/joint/joint-planar.hpp
```diff
@@ -459,14 +466,17 @@ namespace se3
 
     ConfigVector_t random_impl() const
     { 
-      ConfigVector_t result(ConfigVector_t::Random());
+      ConfigVector_t result;
+      result.head<2>().setRandom();
+      const Scalar angle = PI * ((Scalar)(-1 + 2 * rand()) / (Scalar) RAND_MAX);
```
jmirabel: I realize now I made a mistake here. It should be
```cpp
const Scalar angle = PI * (-1 + 2 * ( (Scalar) rand() / (Scalar) RAND_MAX ) );
```
otherwise `2 * rand()` overflows...

Issue #344, opened 2016-09-29
Remove integrate, distance, differentiate from JointModel classes and add TopologyBase class


jcarpent:
To be flexible and extensible, those methods need to be modifiable. The idea is to move them to a dedicated classes templated by the JointType having those methods.

jmirabel:
I propose to decide the name of the base class and of the file where it should go. I propose `LieGroupOperationBase` with a one file `src/multibody/liegroup/name` per file in file `src/multibody/joint/name`.

nmansard:
See also the same discussion in #312 

I think "manifold" would fit better than "Lie group"  (less idiomatic and at least as accurate, possibly more). I think Justin mentioned some name that seemed more suitable (ie easier to understand) like differential-topology, or something similar. 
I would prefer to have it as a subdirectory of joint, ie src/multibody/joint/manifold/manifold-operation-base.hpp. 

In any case, namings is easy to change. If you are in a good mood, go ahead with writing the code with the name you like. We can refine it afterwards if we later manage to all agree on a better name.

jcarpent:
LieGroup seems a complex term, certainly without real meaning for most users. Few weeks ago, I proposed something like Diff{efferential}Calculus but others propositions look nice too.

florent-lamiraux:
LieGroup is more restrictive than manifold, but as pinocchio inheritence is static, it is unlikely that we try to represent more general manifolds. I think the most general structure that we can represent with the current architecture is Riemanian Manifold, but it is a bit long.

jcarpent:
@florent-lamiraux. Use classname identifiable for any user -> we must avoid complex names as `Riemanian Manifold`even they have the right meaning.

jcarpent:
Solved by #365. Nevertheless, a refactoring of the code of #365 is needed to avoid duplicate codes between Lie Group operations and Spatial classes.

jmirabel:
The code could be improved with the missing `inline` keyword.

PR #365, opened 2017-01-03
[C++] Move space operation to new class LieGroup and inherited classes.

jmirabel:
The spaces taken into account are SO1, SO3, SE1, SE3, vector spaces and the Cartesian product of two of them.
The association `JointModel` <-> `LieGroupOperationBase` derived classes is made by the class `LieGroup` in file `src/multubody/liegroup/liegroup.hpp`.

Comments are welcome. Please focus first on important matters, such as names and architectures.
I emphasize the following points:
- name `SpecialOrthogonal1` and  `SpecialEuclidean1` are both very ugly and should be 2 instead of 1.
- `SpecialEuclidean1::forwardKinematics` seems weird to me (naming or architecture problem?).

Solves #344

### DIFF
--- a/src/spatial/explog.hpp
+++ b/src/spatial/explog.hpp
@@ -58,7 +58,7 @@ namespace se3
   template <typename D> Eigen::Matrix<typename D::Scalar,3,1,Eigen::internal::traits<D>::Options>
   log3(const Eigen::MatrixBase<D> & R)
   {
-    EIGEN_STATIC_ASSERT_MATRIX_SPECIFIC_SIZE(D, 3, 3);
+    EIGEN_STATIC_ASSERT_SAME_MATRIX_SIZE(D, Eigen::Matrix3d);
     Eigen::AngleAxis<typename D::Scalar> angleAxis(R);
     assert(0 <= angleAxis.angle() && angleAxis.angle() <= 2 * M_PI);
     if (angleAxis.angle() > M_PI)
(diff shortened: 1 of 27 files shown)


jmirabel:
Related to #344

src/spatial/explog.hpp
```diff
@@ -58,7 +58,7 @@ namespace se3
   template <typename D> Eigen::Matrix<typename D::Scalar,3,1,Eigen::internal::traits<D>::Options>
   log3(const Eigen::MatrixBase<D> & R)
   {
-    EIGEN_STATIC_ASSERT_MATRIX_SPECIFIC_SIZE(D, 3, 3);
+    EIGEN_STATIC_ASSERT_SAME_MATRIX_SIZE(D, Eigen::Matrix3d);
```
jcarpent: I do not know why do you prefer to relate to Eigen::Matrix3d?

jmirabel: `EIGEN_STATIC_ASSERT_MATRIX_SPECIFIC_SIZE` is strict while `EIGEN_STATIC_ASSERT_SAME_MATRIX_SIZE` is permissive. See http://eigen.tuxfamily.org/dox/TopicAssertions.html

`log3(R0.transpose() * R1)` was not compiling because of `EIGEN_STATIC_ASSERT_MATRIX_SPECIFIC_SIZE`. The work around is `log3((R0.transpose() * R1).eval())`.
With `EIGEN_STATIC_ASSERT_MATRIX_SPECIFIC_SIZE`, it compiles correctly.
I think this is a bug in some traits somewhere and it has been corrected lately.

jcarpent:
Thank you Joseph for your contribution.

Issue #385, opened 2017-04-21
Pinocchio does not compile with Eigen 3.3.3


andreadelprete:
Pinocchio doesn't seem to compile with Eigen 3.3.3, which is the last stable release of Eigen. This doesn't allow me to benefit from the last developments of Eigen, such as the implementation of the [Complete Orthogonal Decomposition](https://eigen.tuxfamily.org/dox/classEigen_1_1CompleteOrthogonalDecomposition.html).

jcarpent:
I checked a bit, but it seems that the problem comes from humanoid-path-planner/hpp-fcl#10 where @jmirabel added by default the use of Eigen vectors inside FCL instead of internal linear algebra library.

jcarpent:
To tackle this issue, you can set the flag FCL_EIGEN to OFF when installing hpp-fcl.

andreadelprete:
Actually after some tests I found out that the problem was that I had not recompiled pinocchio with Eigen 3.3.3, so I in my library I was mixing Eigen 3.2 (used to compile pinocchio) with Eigen 3.3.3. After recompiling pinocchio everything seems to work fine.

jcarpent:
Ok. Perfect. I close the issue then.

PR #387, opened 2017-05-22
Update to last changes in hpp-fcl

jmirabel:
hpp-fcl will use only Eigen types in future version.

The FCL_HAVE_EIGEN makes it compatible with older version of hpp-fcl. See comment in src/spatial/fcl-pinocchio-conversions.hpp for more details.

Note that at the moment, the copy is temporarily needed but can be removed when hpp-fcl is upgraded.

### DIFF
--- a/src/spatial/fcl-pinocchio-conversions.hpp
+++ b/src/spatial/fcl-pinocchio-conversions.hpp
@@ -24,6 +24,18 @@
 namespace se3
 {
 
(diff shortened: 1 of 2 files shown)


jmirabel:
Related to #385

jcarpent:
Thanks for the contribution.
I need to activate the use of eigen on robotpkg. I will do it later.
In the meantime, we still rely on FCL data structures.

PR #388, opened 2017-05-24
Add name to model and set it when loading a urdf file.

olivier-stasse:
Useful when one does not known which robot will be loaded.


jcarpent:
Thanks for the contribution.

PR #399, opened 2017-09-05
Lie group operation struct are aware of the dimension of the element they manipulate

florent-lamiraux:
In the previous implementation, vector space of dynamic size were ill defined and random sampling was not correct.

With this pull request, random sampling is possible for vector spaces of dynamic size.


jmirabel:
:+1:

PR #402, opened 2017-09-25
Add algo to compute [inefficiently] the time variation of the joint jacobian

jcarpent:


jcarpent:
@aelkhour It may be helpful to you

PR #410, opened 2017-10-06
Fix bug in distance computation

jcarpent:


unittest/joint-configurations.cpp
```diff
@@ -329,6 +329,18 @@ BOOST_AUTO_TEST_CASE ( neutral_configuration_test )
...
+BOOST_AUTO_TEST_CASE ( distance_configuration_test )
+{
+  Model model; buildModel(model);
+  
+  Eigen::VectorXd q0(model.neutralConfiguration);
+  Eigen::VectorXd q1(q0 + Eigen::VectorXd::Ones(model.nq));
+
+  double dist = distance(model,q0,q1);
+  
+  BOOST_CHECK_MESSAGE(dist > 0., "distance - wrong results");
```
jmirabel: The current implementation should satisfy (with some numerical tolerance):
```cpp
difference(model, q1, q2, v);
distance(model, q1, q2) == v.norm();
```

jmirabel: By the way, the naming `differentiate` is not very good IMHO. In some places, `difference` is used instead, which makes more sense.

PR #414, opened 2017-11-10
[bindings][robot_wrapper]  add frameAcceleration to robot_wrapper.

proyan:
Currently only joint accelerations are supported


jcarpent:
It could be great if you can add frameVelocity too.

jcarpent:
Thanks

PR #426, opened 2017-12-07
Fix typo in model string

jviereck:
Not sure if this fix makes sense. Saw it while reading over the code. Feel free to reject in case it should be written like this :)


jcarpent:
Thanks for this PR.
I already fixed this issue in a PR which will come soon. 
Then, I prefer to not merge your PR. And thanks for contributing to Pinocchio.

PR #433, opened 2018-01-18
Add static regressor

jcarpent:
@andreadelprete @thomasfla 
I've just implemented the static regressor that links the total system CoM to the links' inertial parameters


jcarpent:
I will implement the dynamic regressor in the following weeks.

PR #435, opened 2018-02-06
Add algo to compute the Jacobian of the integration operation + some code cleaning.

jmirabel:


jmirabel:
Note that this PR introduces code in `SpecialEuclideanOperation<2>` which is redundant with `JointModelPlanar`.

After discussing with @jcarpent, solutions to avoid this redundancy are:
- call `SpecialEuclideanOperation<2>` from `JointModelPlanar` or vice-versa,
- put this code in a child class of `SE3Base`

We have to decide the order of dependencies between theses classes. I guess `JointModel` should depend on one of the two others. But between `LieGroupOperation` and `SE3`, which should depend on which ?

src/multibody/liegroup/special-euclidean.hpp
```diff
@@ -168,60 +168,18 @@ namespace se3
...
-      }
+      ConfigOut_t& out = const_cast< ConfigOut_t& >(qout.derived());
+
+      Matrix2 R0, R;
+      Vector2 t0, t;
+      forwardKinematics(R0, t0, q);
+      exp(v, R, t);
+
+      out.template head<2>() = R0 * t + t0;
+      out.template tail<2>() = (R0 * R).col(0);
```
jcarpent: Here it would be better to direct call:
`R0*R.col(0)` to be sure that useless computations are not performed.
`noalias()`could also be helpful.

jcarpent:
Seems good to me.

Issue #440, opened 2018-02-22
Actuation, transmission and mimic joints


jcarpent:
@olivier-stasse together with @flforget have recently opened a PR #434 concerning the adding of DC motor models into Pinocchio in order to to tack also into account the dynamical properties of these components into rigid body dynamics algorithms (inertia effects, etc).

To be effective, we also need to consider the transmission parts, namely gear reducers, pulley-belt systems. It is then important to notice that such transmission model can also handle mimic joints, like the joints that appear in the gripper of HRP-2 or Pyrene humanoids.

The idea of this issue is to collect all the concerns about adding actuation and transmission into Pinocchio.

aelkhour:
It can be useful to handle complex transmissions such as differential drives, where N joints are actuated by M motors.

This seems to be well handled by the ros transmission_interface API.

jcarpent:
#442 is a temporary inclusion of a simple transmission and motor inertia parameters.
This will be extended through #434.

PR #434, opened 2018-01-22
Add DC motor actuator models.

olivier-stasse:
First trial to implement actuator model.
This pull request implements DC motor actuator models
Missing: Transmission model + Pneumatic model 
Details: the documentation is available [here](https://olivier-stasse.github.io/pinocchio/doxygen/HEAD/group__actuators__group.html)

PR #442, opened 2018-03-05
Topic/srdfrotorparams

proyan:
Parses the rotor mass and gear ratios from the srdf file.
1) This pull request only stores the vectors in the model class, and does not change the crba/rnea or forward dynamics algorithms.

2) On discussion with @olivier-stasse, this PR is separate from #434 as they don't have an intention of immediately adding the rotor transmission parameters in pinocchio.

3) Regarding #440, the current PR only addresses the gear transmission. In order to be generic, the nomenclature `rotorGearRatio` in `model` class can be changed to something more suitable, and a visitor can be added to access the transmission parameters of a joint. @jcarpent what are your thoughts on that?

4) Regarding the addition of the transmission effects in the Mass matrix, the easiest way in my opinion would be to add to the diagonal term after full crba is complete. Since default rotor params are zero, in the case of unspecified rotor parameters, this would result in the addition of a zero vector. @jcarpent if you are okay with that, I can go ahead and implement it further.

### DIFF
--- a/bindings/python/parsers/parsers.hpp
+++ b/bindings/python/parsers/parsers.hpp
@@ -130,6 +130,15 @@ namespace se3
         return se3::srdf::getNeutralConfigurationFromSrdf(model, filename, verbose);
       }
 
+      static bool loadRotorParamsFromSrdf(Model & model,
+                                          const std::string & filename,
+                                          bool verbose
+                                          )
+      {
+        return se3::srdf::loadRotorParamsFromSrdf(model, filename, verbose);
+      }
+
+      
       /* --- Expose --------------------------------------------------------- */
       static void expose();
     }; // struct ParsersPythonVisitor
@@ -191,6 +200,9 @@ namespace se3
                        ),
               "Get the neutral configuration of a given model associated to a SRDF file");
 
+      bp::def("loadRotorParamsFromSrdf",loadRotorParamsFromSrdf,
+              bp::args("Model for which we are loading the rotor parameters","srdf filename (string)", "verbosity"),
+              "Load the rotor parameters of a given model from an SRDF file. Results are stored in model.rotorMass and model.rotorGearRatio.");
     }
     
   }
--- a/src/multibody/model.hxx
+++ b/src/multibody/model.hxx
@@ -94,6 +94,12 @@ namespace se3
     
     neutralConfiguration.conservativeResize(nq);
     neutralConfiguration.tail(joint_model.nq()) = joint_model.neutralConfiguration();
+
+    rotorMass.conservativeResize(nv);
+    rotorGearRatio.conservativeResize(nv);
+
+    rotorMass.tail(joint_model.nv()).setZero();
+    rotorGearRatio.tail(joint_model.nv()).setZero();
     
     // Init and add joint index to its parent subtrees.
     subtrees.push_back(IndexVector(1));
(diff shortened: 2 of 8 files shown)


src/multibody/model.hxx
```diff
@@ -93,7 +93,10 @@ namespace se3
     upperPositionLimit.conservativeResize(nq);upperPositionLimit.bottomRows(joint_model.nq()) = max_config;
     
     neutralConfiguration.conservativeResize(nq);
-    neutralConfiguration.tail(joint_model.nq()) = joint_model.neutralConfiguration();
+    rotorMass.conservativeResize(nv);
+    rotorGearRatio.conservativeResize(nv);
+    rotorMass.conservativeResize(nv);
```
jcarpent: double call

src/parsers/srdf.cpp
```diff
@@ -141,10 +141,75 @@ namespace se3
     }
     
 #endif // ifdef WITH_HPP_FCL
+    Eigen::VectorXd getRotorParamsFromSrdf(Model & model,
```
jcarpent: Why returning Eigen::VectorXd while you are in fact returning some bool values.

src/parsers/srdf.cpp
```diff
@@ -141,10 +141,75 @@ namespace se3
...
+                  rotor_mass << " , " << rotor_gr << ")" << std::endl;
+              }
+              // Search in model the joint and its config id
+              Model::JointIndex joint_id = model.getJointId(joint_name);
 
+              if (joint_id != model.joints.size()) // != model.njoints
+              {
+                const JointModel & joint = model.joints[joint_id];
+                model.rotorMass(joint.idx_v()) = rotor_mass;
+                model.rotorGearRatio(joint.idx_v()) = rotor_gr;  // joint with 1 dof
```
jcarpent: You should add an assert to check if jmodel.nv is indeed of dimension one.

src/parsers/srdf.hpp
```diff
@@ -71,6 +71,21 @@ namespace se3
     Eigen::VectorXd getNeutralConfigurationFromSrdf(Model & model,
                                                     const std::string & filename,
                                                     const bool verbose = false) throw (std::invalid_argument);
+
+
+    ///
+    /// \brief Get the rotor params of a given model associated to a SRDF file.
+    ///        It throws if the SRDF file is incorrect.
+    ///
+    /// \param[in] model The Model for which we want the neutral config
```
jcarpent: Wrong doc

bindings/python/parsers/parsers.hpp
```diff
@@ -130,6 +130,15 @@ namespace se3
         return se3::srdf::getNeutralConfigurationFromSrdf(model, filename, verbose);
       }
 
+      static bool loadRotorParamsFromSrdf(Model & model,
```
jcarpent: This function is not required.

jcarpent:
Please avoid merge commits. Use rebased instead to have a clean commit tree.

jcarpent:
You should add a unit test to be validate your parsers.

src/parsers/srdf.cpp
proyan: This has been corrected already

src/multibody/model.hxx
proyan: This has been corrected already

src/parsers/srdf.hpp
proyan: This has been corrected already

jcarpent:
PERFECT. Thanks for your contribution.

Issue #424, opened 2017-12-05
Romeo's models


nim65s:
Hi,

I noticed that https://github.com/stack-of-tasks/pinocchio/tree/master/models/romeo/meshes/romeo is the same as https://github.com/laas/romeo/tree/master/romeo_description/meshes/V1.

The package robotpkg-romeo-description installs the version in laas/romeo, and there is no packaging of the folder in this package, so maybe we should remove it ?

jcarpent:
You're right. The first idea was to have a single repo without too many submodules.
We may just check that the two models are identical at the URDF level.

nim65s:
I clone both repos, and launch meld on those folders. It said they are no differences.

PR #452, opened 2018-04-16
Use ROMEO description as model

jcarpent:
This is related to #424


nim65s:
You have 2 .gitmodule

nim65s:
And you are adding 5k lines, when we want to remove stuff… this might be a mistake :)

jcarpent:
Right. Just a problem from my side. I'm fixing it!

PR #454, opened 2018-04-19
Fix TYPEDEF for GNUC 442

gabrielebndn:
I am pretty sure this is a bug. There was a \ missing, thereby breaking the definition of the macro.
This should only happen when compiling with GCC 4.4.2 . Haven't tested it, but the problem seems obvious.

Speaking about it, it seems very weird to me that the compiler version is apparently checked for equality, instead of using >= or <= . Is there a reason for it?


jcarpent:
@gabrielebndn Thank you for your contribution. Before merging, can you set this PR on the devel branch?

PR #456, opened 2018-04-23
CMakeLists.txt: Add HPP_FCL_INCLUDE_DIRS to include directories

wxmerkt:
In order to build with HPP-FCL (current devel/master) in a non-standard location, the include directories need to be explicitly added.


jcarpent:
Dear Wolfgang,
Thanks for contributing to Pinocchio.
The `INCLUDE_DIRECTORIES` instruction is already achieved in https://github.com/stack-of-tasks/pinocchio/blob/3af5a209109fc025c6f882b2fc7bda9d3b277223/src/CMakeLists.txt#L72
through `PKG_CONFIG_USE_DEPENDENCY` which belongs to https://github.com/jrl-umi3218/jrl-cmakemodules
What you have proposed then appears redundant. Do you encounter any trouble to compile or install Pinocchio without your modification?

jcarpent:
Thanks for this insight.
The problem was coming from the CMakeLists of the utils directory.
I fixed the issue in https://github.com/jcarpent/pinocchio/tree/topic/cmake.
Can you confirm that the problem is fixed?

Best,
Justin

wxmerkt:
Yes, this works - thank you very much for looking into it and fixing it. I will close this PR.

Best,
Wolfgang

jcarpent:
Your welcome. Thanks for reporting this bug.
Do no hesitate to ask us more questions if needed.
Best,
Justin

PR #457, opened 2018-04-26
Topic/joint composite

gabrielebndn:
Re-implemented calc for joint composite.
I am still checking calc_aba.
Also, although I am sure they are buggy, I didn't touch all sub-methods that will be superseded by LieGroups, except for the neutral configuration method which was necessary for the kinematic tests to run correctly (otherwise the program would segfault when the JointComposite is added to a model).

### DIFF
--- a/src/multibody/joint/joint-composite.hxx
+++ b/src/multibody/joint/joint-composite.hxx
@@ -1,0 +1,141 @@
+//
+// Copyright (c) 2016 CNRS
+//
+// This file is part of Pinocchio
+// Pinocchio is free software: you can redistribute it
+// and/or modify it under the terljMj of the GNU Lesser General Public
+// License as published by the Free Software Foundation, either version
+// 3 of the License, or (at your option) any later version.
+//
+// Pinocchio is distributed in the hope that it will be
+// useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+// of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Lesser Public License for more details. You should have
+// received a copy of the GNU Lesser General Public License along with
+// Pinocchio If not, see
+// <http://www.gnu.org/licenses/>.
+
+#ifndef __se3_joint_composite_hxx__
+#define __se3_joint_composite_hxx__
+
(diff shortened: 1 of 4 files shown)


src/multibody/joint/joint-composite.hxx
```diff
@@ -0,0 +1,141 @@
...
+        data.S.matrix().middleCols(idx_v,model.m_nvs[i]) = data.iMlast[succ].inverse().act(jdata.S()); // TODO: avoid computing inverse
+      }
+
+    }
+    
+  };
+
+  inline void jointCompositeCalcZeroOrder(const JointModelComposite & model, JointDataComposite & data, const Eigen::VectorXd & qs)
+  {
+    for (int i=(int)(model.joints.size()-1); i >= 0; --i)
```
jcarpent: Use size_t instead of int

gabrielebndn:
int is necessary to check non-negativity in the loop

unittest/joint-composite.cpp
```diff
@@ -207,5 +212,64 @@ BOOST_AUTO_TEST_CASE(test_copy)
...
+  Model model;
+  JointModelComposite jmodel_composite;
+
+  SE3 config=SE3::Random();
+  //JointModelVariant joint_model;
+  JointIndex parent=0;
+
+  for(int i=0; i<10; i++)
+  {
+    //joint_model = JointModelPX();
```
jcarpent: Remove these lines if they are not needed.

unittest/joint-composite.cpp
```diff
@@ -207,5 +212,64 @@ BOOST_AUTO_TEST_CASE(test_copy)
...
+
+  Model model_c;
+  model_c.addJoint(0,jmodel_composite,SE3::Identity(),"joint");
+
+  Data data_c(model_c);
+
+  BOOST_CHECK(model.nv == model_c.nv);
+  BOOST_CHECK(model.nq == model_c.nq);
+  
+//  VectorXd ql(ConfigVector_t::Constant(model.nq,-M_PI));
```
jcarpent: Remove these lines if they are not needed.

jcarpent:
@gabrielebndn Thank you for your contribution.

PR #466, opened 2018-05-29
[Minor] Set lower position limit to -infinity, not +infinity

jmirabel:


jcarpent:
I was not able to see if the two versions are returning exactly the same value.

jmirabel:
I wasn't sure of what the doc means by *generally*:

> Minimum finite value. (since C++11)
For integral types: the same as min().
For floating-point types: implementation-dependent; generally, the negative of max().

but I'll do as you suggest.

jcarpent:
Apparently the tests are not passing.can you come to your previous solution? Sorry for this last modification.

jcarpent:
@jmirabel Just a pop up to know if you can come back to your previous commit?
The current one is not passing on Travis.

jcarpent:
Perfect!

Issue #471, opened 2018-06-06
eigen 3.3.4 compatibility & Eigen::internal


nim65s:
Hi,

I'm forwarding some comments from https://git.openrobots.org/issues/164:

> In https://github.com/stack-of-tasks/pinocchio/blob/master/src/spatial/explog.hpp#L41
> I don't see why [Eigen::internal] is necessary, it should probably be dropped since it's used only to create an identity matrix. At least it breaks with eigen-3.3.4, which says:
>> error: 'Options' is not a member of 'Eigen::internal::traits<blah blah
>
> See this for reference: http://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html

jcarpent:
This bug has been solved by #472.

PR #472, opened 2018-06-06
[LieGroup] Fix bug related to the Scalar product return type

jcarpent:
This bug is appearing with Eigen3 >= 3.3.4.
The scalar product return type does not have an Options which is required by exp3.


jcarpent:
Related to #471

PR #477, opened 2018-06-20
In SE(n) and SO(n), difference is exactly zero if operands are equal.

florent-lamiraux:
Otherwise, distance between equal configurations may not be equal to 0.


jcarpent:
Thanks for this contribution.
I think it should be great if you can also add a test in unittest/liegroups.cpp to check this specific feature.

florent-lamiraux:
I have added a test in ligroups.

jcarpent:
Perfect!

PR #481, opened 2018-06-29
[python][robot_wrapper] Discrepancy in the API for frameJacobian.

proyan:
python:(model, data, q, index).
cpp:(model, data, index, q).

This commit is meant to correct it with the minimal change. Ideally, cpp should be updated.


proyan:
I refrain from harmonizing the API for the fear of breaking downstream packages

proyan:
It looks like I missed a boolean in the argument list. Sorry about the noise, I will recreate a new PR.

PR #500, opened 2018-08-08
[algorithm/dynamics] add option to take damped inverse when doing cholesky factorization of JMinvJt

proyan:


nim65s:
I think you should add an unit test, when you add a new option.

jcarpent:
+1

jcarpent:
@proyan Indeed, it would be great if you can add a unit test with damping.
It should failed with classic damping = 0. but succeeds with damping = 1e-8 for instance.

By the way, I also realized that you forgot to update document of forwardDynamics.
Can you update your PR accordingly?

PR #501, opened 2018-08-09
Documentation and unittest for forward dynamics with damped cholesky inverse

proyan:
This PR is continuation of PR #500 

In test_fd_with_damping, the constraints (J) are repeated twice. So, while the dimension of the constraints is 12, the rank is 6.

Notes:
When trying to find the undamped solution to the test given in test_fd_with_damping, I discovered that cholesky decomposition was still able to find a solution. This is not what I observed when using a reduced HRP2 model and constraints on the two feet (which were rank deficient).

### DIFF
--- a/src/algorithm/dynamics.hpp
+++ b/src/algorithm/dynamics.hpp
@@ -37,6 +37,7 @@ namespace se3
   ///           \text{s.t.} & & J (q) \ddot{q} + \gamma (q, \dot{q}) = 0 \end{eqnarray} \f$ </CENTER> <BR>
   ///       where \f$ \ddot{q}_{\text{free}} \f$ is the free acceleration (i.e. without constraints),
   ///       \f$ M \f$ is the mass matrix, \f$ J \f$ the constraint Jacobian and \f$ \gamma \f$ is the constraint drift.
+  ///  By default, the constraint Jacobian is assumed to be full rank, and undamped cholesky inverse is performed.
   ///
   /// \param[in] model The model structure of the rigid body system.
   /// \param[in] data The data structure of the rigid body system.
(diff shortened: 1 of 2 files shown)


src/algorithm/dynamics.hpp
```diff
@@ -37,6 +37,7 @@ namespace se3
   ///           \text{s.t.} & & J (q) \ddot{q} + \gamma (q, \dot{q}) = 0 \end{eqnarray} \f$ </CENTER> <BR>
   ///       where \f$ \ddot{q}_{\text{free}} \f$ is the free acceleration (i.e. without constraints),
   ///       \f$ M \f$ is the mass matrix, \f$ J \f$ the constraint Jacobian and \f$ \gamma \f$ is the constraint drift.
+  ///  By default, the constraint Jacobian is assumed to be full rank, and undamped cholesky inverse is performed.
```
jcarpent: Cholesky and not chloesky ;)

src/algorithm/dynamics.hpp
```diff
@@ -45,6 +46,7 @@ namespace se3
   /// \param[in] tau The joint torque vector (dim model.nv).
   /// \param[in] J The Jacobian of the constraints (dim nb_constraints*model.nv).
   /// \param[in] gamma The drift of the constraints (dim nb_constraints).
+  /// \param[in] inv_damping Damping factor for cholesky decomposition of JMinvJt. Set to zero if constraints are full rank.
```
jcarpent: You can add a typical value that works fine.

src/algorithm/dynamics.hpp
proyan: I assume you meant cholesky, not chloesky :trollface: :stuck_out_tongue_winking_eye:

jcarpent:
It seems good to me. I'm gonna merge!

PR #510, opened 2018-09-10
[Doc] Revision of outline after discussion with Nicolas.

gabrielebndn:
After discussing with Nicolas, I changed the outline of the doc.
I think the doc should be modeled on Eigen's "long tutorial": a series of pages, each covering a topic from the user point of view, with worked examples in the form of short snippets intertwined with the code.
Then the proper tutorials will contain larger in-depth examples.
The technical part, containing implementation details, should be separate.
So, the main user doc in a part I called  "Features". As "List of features" should closely reflect the content of "Features", I removed all content from the former part and put in into the latter.
Also, I moved the math to later in the doc, as discussed

### DIFF
--- a/doc/outline.txt
+++ b/doc/outline.txt
@@ -2,44 +2,32 @@
 - ABSTRACT (taken from SII paper)
 - Installation link
 - Simplest example with compilation command (load urdf and run rnea)
-- More complete example with C++ and Python (load model and run a loop for an IK with 3D task)
-- How to cite Pinocchios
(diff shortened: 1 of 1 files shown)


jcarpent:
I suggest to open a project in GitHub to work on this topic. This is the project address: https://github.com/stack-of-tasks/pinocchio/projects/1

@nim65s @gabrielebndn  Can you complete the project with the different items of `doc/outline.txt`

PR #515, opened 2018-09-11
[script/robot_wrapper] add momentum function

proyan:


proyan:
I am following the template of robot.com() which either returns com, or (com,vcom) or (com,vcom,acom) depending on the input arguments

proyan:
Current commit changes the com into 3 functions, and changes function name for am.

bindings/python/scripts/robot_wrapper.py
```diff
@@ -66,17 +66,27 @@ def nq(self):
...
-            return data.com[0]
-        if v is not None:
-            if a is None:
-                se3.centerOfMass(self.model, self.data, q, v)
-                return self.data.com[0], self.data.vcom[0]
-            se3.centerOfMass(self.model, self.data, q, v, a)
-            return self.data.com[0], self.data.vcom[0], self.data.acom[0]
-        return se3.centerOfMass(self.model, self.data, q)
+    def com(self, q=None):
+      if q=None:
```
jcarpent: It seems that it should be` q==None`

proyan:
I am not sure about whether (and how) to add the deprecated tag. The only part of robot.com() which is deprecated is when used with v and a. I don't know if we can make only one part of a function deprecated in python

PR #516, opened 2018-09-11
[frames] Helper functions for frames

gabrielebndn:
framesForwardKinematics(model, data, frame_id) computes oMf[frame_id], stores it and returns it. Assumes forwardKinematics has already been run. Thought it might be useful if the model contains many frames but we are only interested in one or two of them.

getFrameVelocity(model, data, frame_id, v_frame) computes the velocity of frame frame_id and puts it in v_frame. Assumes first-order forwardKinematics has already been run.

getFrameAcceleration(model, data, frame_id, a_frame) computes the acceleration of frame frame_id and puts it in a_frame. Assumes second-order forwardKinematics has already been run.

### DIFF
--- a/src/algorithm/frames.hpp
+++ b/src/algorithm/frames.hpp
@@ -25,7 +25,7 @@ namespace se3
 {
 
   /**
-   * @brief      Updates the position of each frame contained in the model
+   * @brief      Updates the placement of each frame contained in the model
    *
    * @param[in]  model  The kinematic model.
    * @param      data   Data associated to model.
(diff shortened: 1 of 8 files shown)


proyan:
I think you should also implement the python bindings, and update the robot_wrapper API which does the same thing.

gabrielebndn:
I agree. I was planning to do it if this pull request gets accepted

src/algorithm/frames.hpp
```diff
@@ -48,22 +48,70 @@ namespace se3
                                       const Eigen::VectorXd & q);
 
   /**
-   * @brief      Returns the jacobian of the frame expresssed either expressed in the LOCAL frame coordinate system or in the WORLD coordinate system,
+   * @brief      Updates the position of the given frame.
```
jcarpent: placement instead position

jcarpent:
Thanks for these feedbacks.

I would go for updateFramePlacement instead of framesForwardKinematics by making this last function deprecated.

jcarpent:
@gabrielebndn Can you make the modifications?

gabrielebndn:
Alright

gabrielebndn:
Ok, to summarize:
- framesForwardKinematics(model,data,q) stays the same
- framesForwardKinematics(model,data) is deprecated and replaced by updateFramePlacements
- updateFramePlacements(model,data) replaces framesForwardKinematics(model,data)
- updateFramePlacement(model,data,frame_id) [without final s] updates and returns a single frame
- getFrameVelocity(model, data, frame_id, v_frame) returns the velocity of a single frame
- getFrameAcceleration(model, data, frame_id, v_frame) returns the acceleration of a single frame

So, are we sure we want to split framesForwardKinematics into two different methods depending on the arguments? It might be confusing for those who are already using these methods. Maybe it's just easier to stay with framesForwardKinematics.

In Python, I also did the following
- framesKinematics(model,data[,q]) is deprecated. It should be replaced either by framesForwardKinematics(model,data,q) or updateFramePlacements(model,data). I deprecated this function because the name was not consistent with C++
- created the bindings for the new methods
- updated RobotWrapper to use the bindings
- other changes to RobotWrapper. Notably, deprecated "position" and "framePosition" to replace them with "placement" and "framePlacement"

gabrielebndn:
Update: I slightly modified deprecation.py so that the instructions are automatically used as docstring in case none is provided

jcarpent:
Looks fine.
I'm going to merge!

PR #530, opened 2018-10-03
use coverage from gitlab, fix #525

### LINKED ISSUE #525: Coverage is buggy
It seems that the coverage of Pinocchio is not well updated at each master branch pushes.
@nim65s Can you fix this issue?

nim65s:


jcarpent:
Very good. I will merge because just Reame.md has been changed.

nmansard:
Just for me to understand: now, when a PR is made on Pinocchio, it is automatically run on gepgitlab to compute the coverage. Is that correct?

jcarpent:
Not really. It costs a lot to run coverage on Travis.
So, with @nim65s, we decided to put this part on gitlab from LAAS.
Normally, gitlab is run much often than travis. And the idea should be to run it only for release. Or pre-release should be better.

nim65s:
The coverage badge gives the coverage of the master branch, but on each push to gitlab we can look at the coverage in the CI logs.

I think that if we used gitlab's merge request system instead of github's pull request system, it would show if a MR changes the coverage or not. We will probably check this in other projects soon.

PR #531, opened 2018-10-03
[bindings] Implement bindings, solve name inconsistencies, fix #529

### LINKED ISSUE #529: [bindings] Function naming inconsistencies between Python and C++
The following methods have a Python name which is different from the C++ name:
- impulseDynamics (impactDynamics in C++)
- nle (nonLinearEffects in C++)

gabrielebndn:
- turned impactDynamics to impulseDynamics in python, deprecated impactDynamics
- exposed nonLinearEffects, but left nle available as a shorthand
- exposed computeGeneralizedGravity
- exposed computeCoriolisMatrix
- exposed computeGeneralizedGravityDerivatives
- exposed all available overloads of impulseDynamics
- exposed all available overloads of forwardDynamics
- turned RobotWrapper.bias to RobotWrapper.nle, deprecated RobotWrapper.bias
- used newly exposed computeGeneralizedGravity instead of rnea in RobotWrapper.gravity

### DIFF
--- a/bindings/python/algorithm/expose-rnea.cpp
+++ b/bindings/python/algorithm/expose-rnea.cpp
@@ -45,31 +45,28 @@ namespace se3
                        "Vector of external forces expressed in the local frame of each joint (size Model::njoints)"),
               "Compute the RNEA with external forces, put the result in Data and return it.",
               bp::return_value_policy<bp::return_by_value>());
-      
 
-      bp::def("nle",
+      bp::def("nonLinearEffects",
               &nonLinearEffects<double,0,JointCollectionDefaultTpl,VectorXd,VectorXd>,
               bp::args("Model","Data",
                        "Configuration q (size Model::nq)",
                        "Velocity v (size Model::nv)"),
               "Compute the Non Linear Effects (coriolis, centrifugal and gravitational effects), put the result in Data and return it.",
               bp::return_value_policy<bp::return_by_value>());
-      
-      
+
       bp::def("computeGeneralizedGravity",
               &computeGeneralizedGravity<double,0,JointCollectionDefaultTpl,VectorXd>,
               bp::args("Model","Data",
                        "Configuration q (size Model::nq)"),
-              "Computes the generalized gravity contribution g(q) of the Lagrangian dynamics.",
+              "Compute the generalized gravity contribution g(q) of the Lagrangian dynamics, put the result in data.g and return it.",
               bp::return_value_policy<bp::return_by_value>());
-      
-      
+
       bp::def("computeCoriolisMatrix",
               &computeCoriolisMatrix<double,0,JointCollectionDefaultTpl,VectorXd,VectorXd>,
               bp::args("Model","Data",
                        "Configuration q (size Model::nq)",
                        "Velocity v (size Model::nv)"),
-              "Computes the Coriolis Matrix C(q,v) of the Lagrangian dynamics.",
+              "Compute the Coriolis Matrix C(q,v) of the Lagrangian dynamics, put the result in data.C and return it.",
               bp::return_value_policy<bp::return_by_value>());
       
     }
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -195,6 +195,7 @@ SET(PYTHON_FILES
   romeo_wrapper.py
   rpy.py
   explog.py
+  shortcuts.py
   )
 
 FOREACH(python ${PYTHON_FILES})
--- a/bindings/python/algorithm/expose-com.cpp
+++ b/bindings/python/algorithm/expose-com.cpp
@@ -80,7 +80,9 @@ namespace se3
       
       bp::def("jacobianCenterOfMass",&jacobianCenterOfMass<double,0,JointCollectionDefaultTpl,VectorXd>,
               jacobianCenterOfMass_overload(bp::args("Model","Data",
-                       "Joint configuration q (size Model::nq)"),
+                       "Joint configuration q (size Model::nq)",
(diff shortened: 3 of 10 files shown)


gabrielebndn:
In order to expose all available bindings of impulseDynamics and impactDynamics, I used the mechanism provided by BOOST_PYTHON_FUNCTION_OVERLOADS, instead of doing it manually. I think it is accepable

gabrielebndn:
One question: when I overloaded computeGeneralizedGravityDerivatives, I created an auxiliary function computeGeneralizedGravityDerivatives_proxy. Does it have to be static? I see computeRNEADerivativesDefault, in the same file, is not static, but all other examples, such as those in expose-jacobian.cpp, are static.

jcarpent:
I would prefer to not merge this PR now as it is in conflict with the future release of Pinocchio (which will come soon).

jcarpent:
It is better to set them static, but with the mechanism of visitor, this does not matter.

bindings/python/algorithm/expose-rnea.cpp
```diff
@@ -48,13 +48,36 @@ namespace se3
...
-      bp::def("nle",
+      bp::def("nonLinearEffects",
               (const VectorXd & (*)(const Model &, Data &, const VectorXd &, const VectorXd &))&nonLinearEffects,
               bp::args("Model","Data",
                        "Configuration q (size Model::nq)",
                        "Velocity v (size Model::nv)"),
               "Compute the Non Linear Effects (coriolis, centrifugal and gravitational effects), put the result in Data and return it.",
               bp::return_value_policy<bp::return_by_value>());
+
+      bp::def("nle",
```
jcarpent: By the way, if nle is a shorthand, it has to be handled like that. Which means that pinocchio.nle = pinocchio.nonLinearEffects.
As it is done now, `nle` and `nonLinearEffects`are two different functions doing the same job.

jcarpent:
The mechanism of overload of Python is sometimes buggy.
Can you add a unit test in Python to check that all the exposed functions are indeed working.

gabrielebndn:
I checked the overloads on my computer and they were working, but of course I can provide a unit test, I'll do it. If you do not trust boost python on this, I can still do it manually, but it seems fine

gabrielebndn:
As for `pinocchio.nle = pinocchio.nonLinearEffects`, I totally agree it's a much better way to do it. Only, I did not know where to put it. Maybe in bindings/python/scripts/__init__.py ?

gabrielebndn:
As I only touched the bindings, leaving the Pinocchio core intact, I think the conflicts will be very easy to solve, no more than a dozen of lines. I think it would be nice to have the "definitive" version of Pinocchio 1.0 with all important bindings in place and consistent naming. However, if you do not want to deal with them, just leave the pull request open and I will think about it myself.

proyan:
I think you can put it in global __init__.py

jcarpent:
Yes, this is the precise idea behind a single `shortcuts.py` included by `__init__.py`.

gabrielebndn:
It seems to me kind of an overkill to create a dedicated file for all shortcuts  when we only have one shortcut yet and no plans of making others. Unless you do have plans of making others, in which case I understand

jcarpent:
You know, it is like the deprecated.py file which contains a lot of deprecated functions, but with the purpose of being removed.
This would really simplify the management of the project!

gabrielebndn:
Ok no problem. I added the shortcuts file. I will soon add the unit tests for python

gabrielebndn:
Ok. I've added a unit test for all overloads of forwardDynamics and impulseDynamics. It works for me. You can find them in python/bindings_dynamics

gabrielebndn:
I shall add here that the whole python folder is a mess. Most tests seem to be broken.
- One test did not even parse properly because of a typo. That's the only thing I corrected, because it was too bad.
- Still, tests.py does not even run to completion because the the paths of romeo are wrong. I corrected those errors, but I did not include the correction in this pull request.
- Once corrected the above errors, lots of tests fail. I did not check the errors, nor I mean to

By the way, I don't like the name "python", it's confusing, it makes me think that's were I can find the bindings, it is not clear that it only contains tests

gabrielebndn:
Ok, so I guess I'll keep the changes on my devel branch and I will follow the central devel branch in another way.
I  rolled back the correction in the python test folder, as that will be  the object of a future pull request, either from me or from Guilhem.

jcarpent:
@gabrielebndn Can you come back on this?

gabrielebndn:
Right now, the changes are up-to-date with Pinocchio 1.3.3.
Do you want me to update the changes so that they work for Pinocchio 2.0? No problem.
I think some bindings have already been implemented in Pinocchio 2.0

bindings/python/algorithm/expose-rnea.cpp
```diff
@@ -45,31 +45,28 @@ namespace se3
...
               bp::return_value_policy<bp::return_by_value>());
-      
-      
+
       bp::def("computeCoriolisMatrix",
               &computeCoriolisMatrix<double,0,JointCollectionDefaultTpl,VectorXd,VectorXd>,
               bp::args("Model","Data",
                        "Configuration q (size Model::nq)",
                        "Velocity v (size Model::nv)"),
-              "Computes the Coriolis Matrix C(q,v) of the Lagrangian dynamics.",
```
jcarpent: I think computes is fine too.

gabrielebndn: Yes, "computes" is probably better, but we use "compute [...] put" everywhere, it was for consistency. The main change was adding that the result is stored in data

jcarpent:
Thanks!

PR #537, opened 2018-10-04
[All] Removed METAPOD dependency

gabrielebndn:
Cmake was still looking for metapod. Apparently the dependency should have been removed with pull request #490, but the removal was not complete.


jcarpent:
I've already fixed this issue in the future release. I will close this PR.
Anyway, thank you for the contribution.

PR #541, opened 2018-10-05
[doc] Overview section, fix #540

### LINKED ISSUE #540: [doc] Create the overview section
The outline of the overview is 
- ABSTRACT (taken from SII paper)
- Installation link
- Simplest example with compilation command (load urdf and run rnea)
- More complete example with C++ and Python (load model and run a loop for an IK with 3D task)
- About Python wrapings (philosphy and example)
- How to cite Pinocchio
- Presentation of the content of the documentation.

I think all this subsection can be wrote at once by a single dev. Otherwise, please create subtask for concurrent allocation.

nmansard:


gabrielebndn:
Very nice. I think the file ur5.urdf should also be included within pinocchio (I would say in the folder models), especially if the example is meant to be automatically built and tested.

nim65s:
@gabrielebndn we had romeo's urdf included within pinocchio before, and I suggested to remove it in #424.
It doesn't change anything for CI automation, as it can be installed with all the other dependencies.

gabrielebndn:
I like all changes. My comments follow. I've already mentioned some to Nicolas

gabrielebndn:
- I'm ok with not including ur5.urdf within pinocchio, but then a better link should be provided to the file. First of all, I was forced to dig inside the repo until I found the one file I was looking for, which is not nice. Most importantly, I could not find ur5.urdf. Instead, I found ur5.urdf.xacro, which I needed to convert to urdf with ROS. I do not think we should expect the user to that, nor should pinocchio testing routines depend on ROS

gabrielebndn:
- concerning the examples inverse-kinematics: the python example employs `jointJacobian`, while the one in C++ employs `computeJointJacobians` followed by `getJointJacobian`. Nicolas told me he could not make the two examples to be identical because of problems with deprecated methods. So I dug deep into the code, and I found some critical points.
First of all, the easy part: in  o use `jointJacobian` in C++, the correct procedure is the following:
`
se3::Data::Matrix6x J(6,model.nv); J.setZero();
se3::jointJacobian(model,data,q,JOINT_ID,J);
`
while the following procedure
`se3::Data::Matrix6x J = se3::jointJacobian(model,data,q,JOINT_ID);`
is deprecated.
Then there is a bigger problem: if you try to use the above line in Nicolas' example, it *won't work*. You *also* need to call `forwardKinematics` (before or after, it does not make a difference). You do not need it in python. Why? Because the python method `jointJacobian` DOES NOT call its corresponding C++ function. Instead, it internally calls `computeJointJacobians` followed by `getJointJacobian`, and `computeJointJacobians` internally does the equivalent of `forwardKinematics`, which C++ `jointJacobian` does not. In Nicolas' example, `forwardKinematics` is needed later in order to correctly retrieve `data.oMi[i]`. I opened an issue on the point, you can check #545

gabrielebndn:
@nmansard ok

examples/i-inverse-kinematics.cpp
```diff
@@ -1 +1,33 @@
...
+#include "pinocchio/algorithm/joint-configuration.hpp"
+
+int main(int argc, char ** argv)
+{
+  std::string filename = (argc<=1) ? "ur5.urdf" : argv[1];
+  se3::Model model;
+  se3::urdf::buildModel(filename,model); 
+  se3::Data data(model);
+
+  const int    JOINT_ID = 6;
```
jcarpent: For easy use, I would look the joint_id by its name in the kinematic tree. 
This is the standard way of doing it.

examples/i-inverse-kinematics.cpp
```diff
@@ -1 +1,33 @@
...
+  se3::Data::Matrix6x J(6,model.nv); J.fill(0);
+  unsigned int svdOptions = Eigen::ComputeThinU | Eigen::ComputeThinV;
+
+  for (int i=0 ; i<100 ; ++i)
+    {
+      se3::computeJointJacobians(model,data,q);
+      se3::getJointJacobian<se3::LOCAL>(model,data,JOINT_ID,J);
+      const Eigen::Vector3d & x   = data.oMi[JOINT_ID].translation();
+      const Eigen::Matrix3d & R   = data.oMi[JOINT_ID].rotation();
+      const Eigen::Vector3d & err = R.transpose()*(x-xdes);
```
jcarpent: Why to not use directly the operations permitted by the SE3 class?

examples/i-inverse-kinematics.cpp
```diff
@@ -1 +1,33 @@
...
+  Eigen::Vector3d xdes;        xdes << 0, -0.5, 0.5;
+
+  se3::Data::Matrix6x J(6,model.nv); J.fill(0);
+  unsigned int svdOptions = Eigen::ComputeThinU | Eigen::ComputeThinV;
+
+  for (int i=0 ; i<100 ; ++i)
+    {
+      se3::computeJointJacobians(model,data,q);
+      se3::getJointJacobian<se3::LOCAL>(model,data,JOINT_ID,J);
+      const Eigen::Vector3d & x   = data.oMi[JOINT_ID].translation();
```
jcarpent: SE3::Linear_t might be preferred.

examples/i-inverse-kinematics.cpp
```diff
@@ -1 +1,33 @@
...
+      const Eigen::Matrix3d & R   = data.oMi[JOINT_ID].rotation();
+      const Eigen::Vector3d & err = R.transpose()*(x-xdes);
+      const Eigen::VectorXd v     = -J.topRows<3>().bdcSvd(svdOptions).solve(err);
+      q   = se3::integrate(model,q,v*DT);
```
jcarpent: remove one space here

examples/overview-simple.cpp
```diff
@@ -0,0 +1,16 @@
...
+{
+  se3::Model model;
+  se3::buildModels::humanoidSimple(model);
+  se3::Data data(model);
+
+  Eigen::VectorXd q = se3::neutral(model);
+  Eigen::VectorXd v = Eigen::VectorXd::Zero(model.nv);
+  Eigen::VectorXd a = Eigen::VectorXd::Zero(model.nv);
+
+  const Eigen::VectorXd & rnea = se3::rnea(model,data,q,v,a);
```
jcarpent: tau should replace rnea here

examples/overview-simple.cpp
```diff
@@ -0,0 +1,16 @@
...
+  std::cout << rnea.transpose() << std::endl;
```
jcarpent: Maybe a line like 
```suggestion
  std::cout << "tau: " << rnea.transpose() << std::endl;
```
is clearer during execution.

examples/overview-simple.py
```diff
@@ -0,0 +1,11 @@
+import pinocchio
+
+model = pinocchio.Model.BuildHumanoidSimple()
+data = model.createData()
+
+q = pinocchio.neutral(model)
+v = pinocchio.utils.zero(model.nv)
```
jcarpent: I think here we should called numpy matrix directly. Just to let people aware that we can directly NumPy with Pinocchio in Python.

A simple remark, I think pinocchio.utils.matrixStuff should be made deprecated, as it is not the standard way in Python to do that. But this is only a personal viewpoint that can be discussed later.

examples/overview-urdf.cpp
```diff
@@ -0,0 +1,18 @@
...
+#include "pinocchio/parsers/urdf.hpp"
+#include "pinocchio/algorithm/kinematics.hpp"
+
+int main(int argc, char ** argv)
+{
+  std::string filename = (argc<=1) ? "ur5.urdf" : argv[1];
+  se3::Model model;
+  se3::urdf::buildModel(filename,model); 
+  se3::Data data(model);
+  Eigen::VectorXd q(model.nq); q    << 2, -.5, 1.8, 1.8, 2.6, -2;
```
jcarpent: remove trailing white space

jcarpent: What happens if the model is not the same as ur5 here? I think we should initialize q with the neutral element.

doc/Overview.md
```diff
@@ -19,37 +19,153 @@
...
+\@article{carpentier-sii19, \n
+&nbsp;&nbsp;  title={The Pinocchio C++ library--A fast and flexible implementation of rigid body dynamics algorithms and their analytical derivatives},\n
+&nbsp;&nbsp; author={J. Carpentier and G. Saurel and G. Buondonno and J. Mirabel and F. Lamiraux and O. Stasse and N. Mansard},\n
+&nbsp;&nbsp; year={2018}\n
+}
+
+
+### Citing the software package
+
+Pinocchio: fast forward and inverse dynamics for poly-articulated systems, 2015–2018, by J. Carpentier, J. Mirabel, N. Mansard et al, at https://stack-of-tasks. github.io/pinocchio
```
jcarpent: This is not the correct citation. The correct one is inside the Readme.md

examples/i-inverse-kinematics.cpp
```diff
@@ -1 +1,33 @@
...
+  Eigen::VectorXd q(model.nq); q    << 2, -.5, 1.8, 1.8, 2.6, -2;
+  Eigen::Vector3d xdes;        xdes << 0, -0.5, 0.5;
+
+  se3::Data::Matrix6x J(6,model.nv); J.fill(0);
+  unsigned int svdOptions = Eigen::ComputeThinU | Eigen::ComputeThinV;
+
+  for (int i=0 ; i<100 ; ++i)
+    {
+      se3::computeJointJacobians(model,data,q);
+      se3::getJointJacobian<se3::LOCAL>(model,data,JOINT_ID,J);
```
jcarpent: Maybe some comments should be added just to mention for instance: 'we extract the Jacobian of JOINT_ID and expressed in the LOCAL frame of the joint'

doc/Overview.md
```diff
@@ -19,37 +19,153 @@
...
+
+\code ./sample-rnea \endcode
+
+In Python, just run it:
+
+\code python sample-rnea.py \endcode
+
+\subsection OverviewSimpleExplain Explaination of the program
+
+This program load a robot model, create a data structure for algorithm buffering, and run the Recursive-Newton-Euler Algorithm (rnea) to compute the robot inverse dynamics.
```
jcarpent: RNEA in place of rnea

doc/Overview.md
```diff
@@ -19,37 +19,153 @@
...
+
+The first paragraph defines the model. In C++, we first define the object, then allocate it to be our sample model (a human-like kinematic tree with random parameters, very useful for small tests). In Python, the model is created while calling the sample humanoid model.
+The model class contains the constant parameters of the robot (masses, segment length, body names, etc), i.e. the parameters that are not to be modified by the algorithms. As most algorithms need extra room for storage of internal values, we also have to allocate a Data structure, dedicated to the model. Pinocchio relies on a strict separation between constant parameters, in Model, and calculus buffer, in Data.
+
+Inverse dynamics computes the needed torque to track a trajectory defined by position, velocity and acceleration in the configuration space.
```
jcarpent: I would add here to be explicit: 'a trajectory defined by the joint position, velocity and acceleration' and remove in the configuration space.

doc/Overview.md
```diff
@@ -19,37 +19,153 @@
...
+\code g++ -I /usr/include/eigen3/  -I  /path/to/pinocchio/include -L /path/to/pinocchio/lib -DURDFDOM_TYPEDEF_SHARED_PTR -DWITH_URDFDOM  ../examples/sample-urdf.cpp  -lpinocchio -o sample-urdf \endcode
+
+The program typically runs with a UR5 URDF description, that can be found for example in this repository https://github.com/humanoid-path-planner/universal_robot
+
+Launch the program from the directory containing the ur5.urdf file
+\code cd /path/to/ur5.urdf && ./sample-urdf \endcode
+
+In Python, just run it:
+
+\code python sample-urdf \endcode
```
jcarpent: I would rename sample-urdf by something like loading-urdf

doc/Overview.md
jcarpent: Or maybe, you want to refer to the name of the algo in Pinocchio.

doc/Overview.md
```diff
@@ -19,37 +19,153 @@
...
 
-## Installation
+Pinocchio is open-source, mostly written in C++ with Python bindings, and distributed under LGPL-v3 licence.
+Contributions are welcome.
+
+In this doc, we will find the usual description of the library functionalities, a quick tutorial to catch over the mathematics behind the implementation, a bunch of examples about how to implement classical applications (inverse kinematics, contact dynamics, collision detection, etc) and a set of practical exercices for beginners.
+
+\section OverviewInstall How to install Pinocchio?
+
+Pinocchio is best installed from APT packaging on Ubuntu 14.04 and 16.04, from our repository.
```
jcarpent: I think this section should replicate what is written on the website.
Because there is also binairies available for OSX which is not mentioned.
We also support 18.04, Fedore, BSD and so on.

doc/Overview.md
```diff
@@ -19,37 +19,153 @@
...
+</td>
+<td>
+\include overview-simple.py
+</td></tr></table>
+
+\subsection OverviewSimpleCompile Compiling and running your program
+
+You would compile the C++ version by including Pinocchio and Eigen header directories.
+
+\code g++ -I /usr/include/eigen -I /path/to/pinocchio/include sample-rnea.cpp -L /path/to/pinocchio/lib -lpinocchio -o sample-rnea \endcode
```
jcarpent: The path '/usr/include/eigen' is not correct.

gabrielebndn: Since the installation path may vary for different users, I suggest to write something generic like /path/to/eigen

jcarpent: Or simpler than that, you should use pkg-config for that:
`pkg-config --cflags eigen3`

gabrielebndn:
I think we have discussed enough about this. My suggestion:
- Rebase topic/doc-v2 to bring it on par with the latest progress on devel
- Address all comments made by me and @jcarpent 
- For the Jacobians in the inverse kinematics examples: let's freeze the discussion on the API. Here we have four options, listed from the ones I deem most to least desirable:
    - switch to computeJointJacobian + getJointJacobian in python
    - keep the two examples as they are, with a slight inconsistency between the two. In the text, just present the two methods as two alternative ways of achieving the same result.
    - use jointJacobian in C++, but add a call to forwardKinematics() in C++. Explain why this is needed.
    - use jointJacobian in C++, add a call to forwardKinematics() in C++, and also add a (useless) call to forwardKinematics() in python (notice that this would still not fully eliminate all inconsistencies).
- merge in devel and close.
I would happily do the job myself, but the changes are in Nicolas' repo

gabrielebndn:
Update: I will think about it by doing a PR on nicolas' repo.

examples/i-inverse-kinematics.cpp
nmansard: IK example is not mature. I just kept it for future edition.

doc/Overview.md
nmansard: We should also update the README.md. 
I do not think the contribution of Florian put him in the top 3 dev of the current release of Pinocchio. Any suggestion of fairly deciding the contributor list is welcome. We should maybe remove direct naming of contributors in such a reference, as it would always be unfair, and have a page detailing the contribution in the doc.
I open an issue to discuss it.

doc/Overview.md
nmansard: In the example dir, I yet named the 2 tests "overview-*" and then choose arbitrary names in the doc. I guess we should unified, and name the tests overview-simple and overview-urdf .

doc/Overview.md
nmansard: I tried here to give a fair opinion about what are the typical system using pinocchio. I think we have 90% of the user/dev of Pinocchio in u 14 and 16. Hence I am very confident that Pinocchio "works" on these 2 configs. Are you similarly confident with eg OSX and BSD? I advise we find a wording that render this opinion.

doc/Overview.md
nmansard: No, keep it simple. We should put either /usr/include/eigen3 or /path/to/eigen/include . I think the first solution is clear enough and is easier to use in 90% of the cases, but would agree to use the second one if you believe it is misleading. We should avoid more complex procedure like pkg-config.

Although we might refer the reader to the example dir, where the example will finally be compiled (compilation of example dir is not yet set up).

gabrielebndn: Ok to keep it simple. I support /path/to/eigen/include. Additionally, maybe we can give /usr/include/eigen3 as a possible suggestion the text, or hint to pkg-config as a way to retrieve the correct path.

doc/Overview.md
gabrielebndn: For OS X yes, I use it on my Macbook, and Justin also routinely works on OS X as far as I know. Can't tell anything about brew the brew installation though, never tried it for pinocchio.

gabrielebndn: According to @nim65s, Pinocchio works correctly on ubuntu 18, and it also builds correctly on Fedora and BSD (don't know about the tests though). I will add a reference to ubuntu 18 and to OSX

gabrielebndn:
Ping. I've done my part.You can find the changes in Nicolas repo, PR#4.

nmansard:
@gabrielebndn I marked some of the request changes of @jcarpent as resolved, but I am not sure of the other ones. Could you double check then merge?

examples/i-inverse-kinematics.cpp
gabrielebndn: I think that for a 6D fixed-base manipulator saying "joint 6" is much easier and clearer than "wrist_3_joint"

examples/i-inverse-kinematics.cpp
gabrielebndn: You mean something like `err = - data.oMi[JOINT_ID].actInv(xdes);` ? I thought Nicolas' code would be easier to understand

examples/i-inverse-kinematics.cpp
gabrielebndn: Did not do it. You may be right, but I do not think it is important. Those are Eigen objects, this is not going to change, and I think it is easier for the user to think of the as Eigen 3D vectors rather than some abstract thing.

examples/overview-simple.py
gabrielebndn: I think that as for now, `utils.zero` is a functionality of Pinocchio, a functionality I happen to like and use a lot because numpy's expressions can be very clumsy. If you want to move it to the main `pinocchio` namespace, why not, but it can be discussed later. I made it clear in the Overview text that we are using numpy

doc/Overview.md
gabrielebndn: I followed Nicolas' advice

doc/Overview.md
gabrielebndn: See related issue. As a side note, I put each citation in a separate tiny bib file, you can find them under doc/bib. If you want to modify this one, it is doc/bib/pinocchioweb.bib

doc/Overview.md
gabrielebndn: I included 18.04 and Mac OS X. For Ubuntu 14.04, 16.04 and 18.04 mentioned APT, for Mac Homebrew. I said that it can be installed from source for other systems.
I am thinking now that in between we could add a sentence like "Pinocchio should also work on Fedora, BSD [etc.], but this possibility is not fully tested". Feel free to add this sentence, including any OS you like

examples/i-inverse-kinematics.cpp
gabrielebndn: I did not add this comment

gabrielebndn:
Ok, so I marked as solved most of Justin's comments, except for 6:
- 4 in the code of i-inverse-kinematics.cpp . I preferred not to follow them. I think we can agree these are secondary and we should not focus on this piece of code
- 1 in overview-simple.py . I preferred not to follow it
- 1 About the supported systems. I mostly followed Justin's advice, mentioning Ubuntu 18.04 and Mac OS X, but I did not explicitly mention other systems. Probably I should have. What else do we support? Fedora, BSD and ... ? Maybe Nicolas can add a sentence when he makes his last modifications

gabrielebndn:
Justin, if you agree with things as they are now, you can mark your unresolved changes as resolved yourself.

jcarpent:
It seems fine for me now.

PR #543, opened 2018-10-15
[spatial] fixed bug in random force and motion

gabrielebndn:
Motion.setRandom() and Force.setRandom() were setting to zero instead of random


jcarpent:
I already fixed this issue in the new version of Pinocchio. But, I will accept it as it is a big mistake.

PR #550, opened 2018-10-17
[python] Fixed python tests and moved them into unit test, fix #533

### LINKED ISSUE #533: [python] python test folder is broken
The whole python folder is a mess. Most tests seem to be broken.
- One test did not even parse properly because of a typo. I discovered it while working on pull request #531, the correction is there, but that's the only thing I corrected
- Still, once the typo is corrected, tests.py does not even run to completion because some tests load the romeo model and the path is wrong, raising an uncaught exception. I corrected those errors, but I did not include the correction in the pull request, because I don't think romeo should be used in the tests anyway
- Once the above problems are corrected, lots of tests fail. I did not check the errors

By the way, I don't like the name "python", it's confusing, it is not clear that it only contains tests.

gabrielebndn:
I fixed all python tests and moved them to unittest/python (plus one to examples/python).
I do not like some of them, but at least now they are up to date and they work correctly.
Running tests.py imports them and runs them all, but now the tests can be also run independently from each other.
I got rid of loading URDF where I could, but test bindings_geometry_object.py still loads romeo; this should be changed when we are able to do without it.
I also removed some stuff from a previous example and made a tiny correction to se3.log


jcarpent:
Good. Thanks for the contribution.
I will merge when travis is fine.

PR #554, opened 2018-10-19
[parser] Add two sample models 

nmansard:
Add sample manipulator and humanoid models, with kinematic tree and geometries.
Add python bindings for the two models (4 functions)
Add unittest.

### DIFF
--- a/bindings/python/parsers/sample-models.hpp
+++ b/bindings/python/parsers/sample-models.hpp
@@ -18,9 +18,6 @@
 #ifndef __se3_python_sample_models_hpp__
 #define __se3_python_sample_models_hpp__
 
-#include <eigenpy/exception.hpp>
-#include <eigenpy/eigenpy.hpp>
-
 #include "pinocchio/parsers/sample-models.hpp"
 #include "pinocchio/bindings/python/multibody/data.hpp"
 #include "pinocchio/bindings/python/multibody/geometry-model.hpp"
--- a/bindings/python/parsers/sample-models.hpp
+++ b/bindings/python/parsers/sample-models.hpp
@@ -1,0 +1,85 @@
+//
+// Copyright (c) 2015-2018 CNRS
+//
+// This file is part of Pinocchio
+// Pinocchio is free software: you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation, either version
+// 3 of the License, or (at your option) any later version.
+//
+// Pinocchio is distributed in the hope that it will be
+// useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+// of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Lesser Public License for more details. You should have
+// received a copy of the GNU Lesser General Public License along with
+// Pinocchio If not, see
+// <http://www.gnu.org/licenses/>.
+
+#ifndef __se3_python_sample_models_hpp__
+#define __se3_python_sample_models_hpp__
+
+#include <eigenpy/exception.hpp>
+#include <eigenpy/eigenpy.hpp>
+
+#include "pinocchio/parsers/sample-models.hpp"
+#include "pinocchio/bindings/python/multibody/data.hpp"
+#include "pinocchio/bindings/python/multibody/geometry-model.hpp"
+#include "pinocchio/bindings/python/multibody/geometry-data.hpp"
+
+namespace se3
+{
+  namespace python
+  {
+    struct SampleModelsPythonVisitor
+    {
+
+      static Model buildSampleModelHumanoidSimple(bool usingFF)
+      {
+        Model model;
+        buildModels::humanoidSimple(model,usingFF);
+        return model;
+      }
+
+      static Model buildSampleModelManipulator()
+      {
+        Model model;
+        buildModels::manipulator(model);
+        return model;
+      }
+      
+      static GeometryModel buildSampleGeometryModelManipulator(const Model& model)
+      {
+        GeometryModel geom;
+        buildModels::manipulatorGeometries(model,geom);
+        return geom;
+      }
+      
+      /* --- Expose --------------------------------------------------------- */
+      static void expose();
+    }; // struct SampleModelsPythonVisitor
+
+    inline void SampleModelsPythonVisitor::expose()
+    {
+      bp::def("buildSampleModelHumanoidSimple",
+              static_cast <Model (*) (bool)> (&SampleModelsPythonVisitor::buildSampleModelHumanoidSimple),
+              bp::args("bool (usingFreeFlyer)"),
+              "Generate a (hard-coded) model of a simple humanoid robot."
+              );
+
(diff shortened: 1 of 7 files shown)


src/parsers/sample-models.cpp
```diff
@@ -137,6 +137,207 @@ namespace se3
...
+                                 SE3(Eigen::Matrix3d::Identity(), Eigen::Vector3d(0,0,0.1)) );
+      geom.addGeometryObject(effectorArm,model,true);
+    }
+
+
+    void manipulator(Model& model) { addManipulator(model); }
+    void manipulatorGeometries(const Model& model, GeometryModel & geom)
+    { addManipulatorGeometries(model,geom); }
+
+    static const Eigen::Vector3d AX_X(1,0,0);
```
jcarpent: Shortcut for that is Eigen::Vector3::Unit{X,Y,Z}

src/parsers/sample-models.cpp
```diff
@@ -137,6 +137,207 @@ namespace se3
...
+    static const Eigen::Vector3d AX_Y(0,1,0);
+    static const Eigen::Vector3d AX_Z(0,0,1);
+    
+    static Eigen::Matrix3d rotate(const double angle, const Eigen::Vector3d & axis)
+    { return Eigen::AngleAxisd(angle,axis).toRotationMatrix(); }
+    
+    void humanoid(Model & model, bool usingFF)
```
jcarpent: This code is duplicate with respect to buildHumanoidSimple

nmansard: What is duplicated: humanoid function? Yes and no: humanoidSimple is a random kinematic chain with the same tree structure as many humanoids. Here, humanoid is a realistic chain, with realistic masses, no randomization and a set of relevant geometries. I guess that we can mark humanoidSimple as deprecated. 
I have chosen to make a new function rather than overwriting the existing humanoidSimple function, because there is no unittest on top of humanoidSimple, and I am not sure how humanoidSimple is used in Pinocchio or HPP unittest. I did not want to break those tests hence preferred to had something concurrent in the API.

@jmirabel if you are interested by this model for HPP test, you may want to decide some namings of the frames, or add some frames to the tree.

gabrielebndn:
I like this a lot! One observation: I would not use a .cpp in src. I am talking about src/parsers/sample-models.cpp. Only headers would be nicer. Pinocchio is going to be header-only anyway. To keep the implemetation separate from the interface, better to use the .hpp + .hxx pattern we employ everywhere else in src

src/parsers/sample-models.cpp
jcarpent: humanoidSimple is mostly used for unit tests purposes.

gabrielebndn:
In the python bindings for the humanoid you have: `bp::args("bool (usingFreeFlyer)")`.
I would add some more explanations here:
`bp::args("bool (usingFreeFlyer). If true, a Freeflyer is used as root joint. Otherwise a Composite Translation + SphericalZYX is used")`.
If you could add some doxygen comments for this in the main header too

src/parsers/sample-models.cpp
jmirabel: Same joint names as in buildHumanoidSimple would be perfect for HPP.

nmansard: Hum, ok, I should not have let my question so open. Your request would be tedious to implement. Could you look at the current model and modify the names as you want. Or even better, add a unittest in pinocchio/unittest/sample-models.cpp about humanoidSimple that is checking the properties that you are expecting from the model. I would then modify the new model to fit your test.

jmirabel: I don't think we are expecting any special properties from the model. Joint names are not of uttermost importance as long as they are meaningful and follow a coherent logic.

src/parsers/sample-models.hpp
```diff
@@ -20,6 +20,7 @@
 #define __se3_sample_models_hpp__
```
jcarpent: The two new models are never sampled, but there is the name sample in the title which may lead to some confusions. Can you change it for something like {basic, simple} models?

jcarpent: The two new models are never sampled, but there is the name sample in the title which may lead to some confusions. Can you change it for something like {basic, simple} models?

gabrielebndn: I guess "sample" is used in the sense "specimen serving as example"

gabrielebndn: I guess "sample" is used in the sense "specimen serving as example"

gabrielebndn:
@jcarpent, I think github was having some problems, it was not refreshing correctly, now you have sent 10 copies of the same message. It was doing the same to me

gabrielebndn:
@jcarpent there's problems with github. We keep receiving your same comment. Sometime it shows up on github, sometimes it does not

jcarpent:
Just for information, I've sent the request only once. Github is currently facing some troubles.

src/parsers/sample-models.hpp
jcarpent: But, I would make things clear and use simple or basic terms compared to buildHumanoidSimple which in fact a sampled model.

jcarpent: But, I would make things clear and use simple or basic terms compared to buildHumanoidSimple which in fact a sampled model.

jcarpent:
Just for information, I've sent the request only once. Github is currently facing some troubles.

src/parsers/sample-models.cpp
jmirabel: Very good idea !

src/parsers/sample-models.cpp
```diff
@@ -137,6 +137,203 @@ namespace se3
 
     }
 
+    static void addManipulator(Model & model,
+                               Model::JointIndex rootJoint = 0,
+                               const SE3 & Mroot = SE3::Identity(),
+                               const std::string pre = "")
```
jcarpent: Use reference here

bindings/python/parsers/sample-models.hpp
```diff
@@ -0,0 +1,110 @@
...
+// of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Lesser Public License for more details. You should have
+// received a copy of the GNU Lesser General Public License along with
+// Pinocchio If not, see
+// <http://www.gnu.org/licenses/>.
+
+#ifndef __se3_python_sample_models_hpp__
+#define __se3_python_sample_models_hpp__
+
+#include <eigenpy/exception.hpp>
```
jcarpent: Not needed

bindings/python/parsers/sample-models.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include <eigenpy/eigenpy.hpp>
```
jcarpent: Not needed

nmansard:
Ready to merge once travis is happy. The reviews of Justin are still in red also all resolved, I guess because of the bug with multiple mail copies of github the other day.

PR #557, opened 2018-10-24
Devel sync with master

jcarpent:


gabrielebndn:
I suggest waiting until #554 is also merged

jcarpent:
It would change nothing.

gabrielebndn:
As you like, it was just a suggestion

PR #558, opened 2018-10-24
[bindings] [sample models] Manage deprecation and overloads in python

gabrielebndn:
Correct the bindings for the newly-introduced models.
- expose `buildSampleModelHumanoidRandom` in python
- deprecate `buildSampleModelHumanoidSimple` in python
- deprecate static method `Model.BuildHumanoidSimple` in python
- add overloads for `buildSampleModelHumanoid` and `buildSampleModelHumanoidRandom` in python (for `usingFF`)

Two observations:
- `buildSampleModelHumanoidSimple` was introduced by @nmansard in his last commits. Since it calls a function which is deprecated in C++, I thought it should be deprecated in python too. Until now, the way to create a simple humanoid in python was to call static method `Model.BuildHumanoidSimple`, which was calling the deprecated C++ function. It needs to be deprecated too if the warning has to appear in existing python code. Also, I think the pattern `se3.buildSampleModelXXX` is much better than `se3.Model.BuildXXX`. @nmansard's latest commit was going in this direction
- If you agree, I would also deprecate `Model.BuildEmptyModel`, replacing it with `se3.buildEmptyModel`. Or, better, providing no replacement, since the empty constructor `Model()` is enough

### DIFF
--- a/bindings/python/multibody/model.hpp
+++ b/bindings/python/multibody/model.hpp
@@ -154,10 +154,6 @@ namespace se3
         .def("addFrame",(bool (Model::*)(const Frame &)) &Model::addFrame,bp::args("frame"),"Add a frame to the vector of frames.")
         
         .def("createData",&ModelPythonVisitor::createData)
-        .def("BuildEmptyModel",&ModelPythonVisitor::maker_empty)
-        .staticmethod("BuildEmptyModel")
-        .def("BuildHumanoidSimple",&ModelPythonVisitor::maker_humanoidSimple)
-        .staticmethod("BuildHumanoidSimple")
         
         .def("check",(bool (Model::*)(const Data &) const) &Model::check,bp::arg("data"),"Check consistency of data wrt model.")
         ;
@@ -175,14 +171,6 @@ namespace se3
       }
       
       static Data createData(const Model & model) { return Data(model); }
-      static Model maker_empty() { return Model(); }
-      
-      static Model maker_humanoidSimple()
-      {
-        Model model;
-        buildModels::humanoidSimple(model);
-        return model;
-      }
 
       ///
       /// \brief Provide equivalent to python list index function for
--- a/bindings/python/parsers/sample-models.hpp
+++ b/bindings/python/parsers/sample-models.hpp
@@ -30,10 +30,17 @@ namespace se3
     struct SampleModelsPythonVisitor
     {
 
-      static Model buildSampleModelHumanoidSimple(bool usingFF)
+      static Model buildSampleModelHumanoidRandom()
       {
         Model model;
-        buildModels::humanoidSimple(model,usingFF);
+        buildModels::humanoidRandom(model);
+        return model;
+      }
+
+      static Model buildSampleModelHumanoidRandom(bool usingFF)
+      {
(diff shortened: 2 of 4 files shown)


bindings/python/parsers/sample-models.hpp
```diff
@@ -71,10 +85,15 @@ namespace se3
...
+      bp::def("buildSampleModelHumanoidRandom",
+              static_cast <Model (*) ()> (&SampleModelsPythonVisitor::buildSampleModelHumanoidRandom),
+              "Generate a (hard-coded) model of a humanoid robot with 6D limbs and random joint placement.\nOnly meant for unit tests."
+              );
+
+      bp::def("buildSampleModelHumanoidRandom",
+              static_cast <Model (*) (bool)> (&SampleModelsPythonVisitor::buildSampleModelHumanoidRandom),
               bp::args("bool (usingFreeFlyer)"),
-              "Generate a (hard-coded) model of a simple humanoid robot."
+              "Generate a (hard-coded) model of a humanoid robot with 6D limbs and random joint placement.\nOnly meant for unit tests."
```
jcarpent: "6 DoF" and "random joint placements".

bindings/python/multibody/model.hpp
```diff
@@ -156,8 +156,6 @@ namespace se3
         .def("createData",&ModelPythonVisitor::createData)
         .def("BuildEmptyModel",&ModelPythonVisitor::maker_empty)
```
jcarpent: You can also remove this lines and put them as deprecated.
They are indeed redundant with respect to `pinocchio.Model()`

gabrielebndn: You mean, I should also remove `model.createData()`? Or only `Model.BuildEmptyModel()`?

bindings/python/parsers/sample-models.hpp
gabrielebndn: This comment was copied from the original C++ code, I'll modify that too

bindings/python/multibody/model.hpp
jcarpent: `Model.BuildEmptyModel()` is the static method to remove.

bindings/python/parsers/sample-models.hpp
jcarpent: Perfect!

bindings/python/multibody/model.hpp
gabrielebndn: Perfect

gabrielebndn:
Done. I also fixed a couple of other doxygen comments in C++, as they were not correct.
Other observation: right now, `buildSampleModelHumanoidRandom(false)` uses 3 prismatic and 3 revolute joints. This changes not only the joint configuration, but also the total number of joints, the name of the "root" joint, and the inertial properties of the model. If we want to keep a random humanoid, I think it should do as the newly-introduced `buildSampleModelHumanoid(false)`, with a JointModelComposite (`JointModelTranslation`+`JointModelSphericalZYX`). However, I think it will be removed soon, therefore I did not go through the hassle of changing it and checking all tests still work (they should though).

PR #560, opened 2018-10-25
Remove warnings

jcarpent:
This PR removes the warning recently introduced in #554.


gabrielebndn:
So you want to use `humanoidRandom` ? Why not switching to `humanoid`?
Whether you use one or the other, can you do the same in `unittest/python` ?

gabrielebndn:
I see it has already been merged

gabrielebndn:
Should I fix the python unit tests too?

Issue #547, opened 2018-10-17
Pinocchio 2.0


nmansard:
@jcarpent is preparing a major change of API. Here are some suggestions to prepare the release.
- initial candidate release should be proposed as 1.99.x
- propose a guideline to convert developments based on 1.3 to 2.0
- don't move master to a 2.0 version until the main dependencies of Pinocchio have given their agreement.
- during the transition phase, pinocchio/devel will be in 2.0, while pinocchio/master will be in 1.3. Bug fixes can be pushed to pinocchio/master during the transition phases. Parallel developments should enforce that the same bug fixes are pushed in pinocchio/devel. 
- as the master branch will diverge from the devel branch, we will have to revert possible commits in master before merging it to devel.
- we shoul wait for a consensus among the users of Pinocchio before merging master to devel.

It is suggested to developers using Pinocchio to start moving their devel branch to API 2.0 when it is officially released. If possible, we would like to set up a dead line for the transition phase. Pinocchio/master will be merged to version 2.0 after this deadline.

In summary, here would be the agenda:
0. update pinocchio/master to the latest commit.
1. merge new API in pinocchio/devel and tag it 1.99.x
2.a validate pinocchio/devel and release it as version 2.0.0 when solid enough.
2.b during this period, master is staying in 1.3.x but can be added with bug fixes
3. once 2.0 is officially released, the software depending on pinocchio should move their devel branches to pinocchio/devel
4. at a given date (consensually defined), pinocchio/master is merged to pinocchio/devel (after reverting possible head commits). Pinocchio 2.0 becomes the official version.

jcarpent:
From my side, I will merge the current devel branch into 2.0.0 to have an aligned history.
It would be nice to not modify the current devel branch in order to not add more stuff in my todo list.
The PR can remain open and would be merged in the new version by their authors.

nim65s:
I can take care of the robotpkg part

nmansard:
When merging devel, please follow the guidelines I described upper (or discuss it beforehand).

jcarpent:
Yes, this is pretty standard. No worry!

jcarpent:
By the way, I will release a new version of Pinocchio before proposing 2.0.0-beta.
This will add the last fixes already present in the current devel branch.

nmansard:
It was propose to release in 1.99.0 for the candidate.

jcarpent:
[...]

PR #561, opened 2018-10-25
Split Jdifference in dDifference_dq<0> and dDifference_dq<1>

jmirabel:


jcarpent:
Thanks for this contribution. I think I will merge it after version 2.0.0 because I don't how much this PR is in conflict with the next release.

src/multibody/liegroup/operation-base.hpp
```diff
@@ -165,20 +166,26 @@ SE3_LIE_GROUP_PUBLIC_INTERFACE_GENERIC(Derived,typename)
...
+                     const Eigen::MatrixBase<JacobianLOut_t>& J0,
+                     const Eigen::MatrixBase<JacobianROut_t>& J1) const
+    PINOCCHIO_DEPRECATED;
+
     /**
      * @brief      Computes the Jacobian of the difference operation.
      *
      * @param[in]  q0    the initial configuration vector.
      * @param[in]  q1    the terminal configuration vector.
+     * @tparam[in] iVar  0 (resp. 1) to get the Jacobian with respect to q0 (resp. q1).
```
jmirabel: The template parameter is already documented.

jmirabel:
But I have always found `difference(q0,q1)` unclear and I always have to check the doc or the code.

jmirabel:
> I think I will merge it after version 2.0.0 because I don't how much this PR is in conflict with the next release.

I would be happy to put it myself within the beta version. Why isn't there a branch in the main repo containing the content of this release ?
This way, we can actually contribute to the latest point rather than being replied that either:
- you already added it in the next release so the same work was done twice,
- we should wait for your release and someone must make the fix compatible...
There are 4-5 people actively working on Pinocchio and this work would gain from more cooperation.

By that, I do not ask you to merge it if it makes things more complicated.

jcarpent:
I do understand, but I'm not full time on it, and the current version is not passing all the tests, and I don't think that someone else would like to handle it.

src/multibody/liegroup/operation-base.hpp
jcarpent: Ok. I did not see it when I was reading your code.

jcarpent:
Concerning the choice of template naming, I do not think that putting an int is helpful neither.
You will have to in the documentation too, so far and so on.

I was just suggesting a convenient naming, and I suggest to think about it now.
Why not to use something like WITH_RESPECT_TO_FIRST_ARGUMENT, WITH_RESPECT_TO_SECOND_ARGUMENT, etc.

jmirabel:
> I'm not full time on it

I understand this and you already do a lot.

> the current version is not passing all the tests

This is not important for a branch made for elaborating a new release. It is still in beta version so, yes, it is not ready. An appropriate branch name is sufficient to make people understand they should not take it as a stable version. But if you want a patch to be included in the next release, it is the branch you must use.

jmirabel:
> You will have to in the documentation too, so far and so on.

I think I documented it.

> I suggest to think about it now.

Agreed. For clarity, it should be rather short. ARG0, ARG1... ?

jcarpent:
The other more important issue is that this branch is not yet aligned with Pinocchio 1.3.0 and further.
This is a tedious work. I will try to do my best to make it available by the end of the next week as I mentioned in #547.

jcarpent:
Yes, it should be short, but also explaining the argument. WRT_ARG0?

jmirabel:
Is there other places where this problem occurs ?
Otherwise, I can make two functions (like `dDifference_darg0` and  `dDifference_darg1` or  `dDifference_dq0` and  `dDifference_dq1`).

gabrielebndn:
I like dDifference_dq0 and dDifference_dq1. Much better than having a silly enum which is only used for this single method.

jmirabel:
Cherry-picked ont `topic/release-2.0.0` branch. See https://github.com/jcarpent/pinocchio/pull/2

PR #574, opened 2018-10-30
[bug] Fix sample geometry model and python geometry unit test

gabrielebndn:
For some reason, the corrections I had made to  unittest/python/bindings_geometry_object.py seem to have been eaten up by last merge, so I restored them.
While correcting, I discovered a bug which was introduced with Pinocchio 2.0 in the sample geometry models, so I fixed it. Basically the parent joint of the geometry objects was not set correctly.

This is due to the fact that, before, `geommodel.addGeometryObject(object, model, bool)` was employed and this method was taking care of setting the parent. But in Pinocchio 2.0, this method was deprecated and replaced by `geommodel.addGeometryObject(object)`, leaving the parent joint unset.

### DIFF
--- a/src/parsers/sample-models.cpp
+++ b/src/parsers/sample-models.cpp
@@ -193,38 +193,46 @@ namespace se3
                                          GeometryModel & geom,
                                          const std::string & pre = "")
     {
+      FrameIndex parentFrame;
+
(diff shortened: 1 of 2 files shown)


jcarpent:
Yes, I have changed the API of it because the default argument for autofill was not satisfying.
I will recover the previous in a future PR.

Thanks for your contribution.

PR #580, opened 2018-11-02
Increase performances of Jlog6 computations

jcarpent:


src/spatial/explog.hpp
```diff
@@ -502,17 +510,17 @@ namespace se3
...
-          (beta_dot_over_theta*wTp)*w*w.transpose()
-          - (t2*beta_dot_over_theta+Scalar(2)*beta)*p*w.transpose()
-          + wTp * beta * Matrix3::Identity()
-          + beta * w*p.transpose()
-          ) * J3);
-
-    value << J3             , J,
-             Matrix3::Zero(), J3;
+    Vector3 v3_tmp((beta_dot_over_theta*wTp)*w - (t2*beta_dot_over_theta+Scalar(2)*beta)*p);
+    // C can be treated as a temporary variable
```
jmirabel: Did you benchmark this ?
I would say using C as a temporary variable is less efficient than creating a fixed size temporary.

Issue #581, opened 2018-11-02
Taylor expansion threshold


jcarpent:
It appears that Jlog6 and Jlog3 do not have the same threshold for the Taylor expansion.
1e-4 for [Jlog6](https://github.com/stack-of-tasks/pinocchio/blob/912418bd2d1b707800162f131e2013d6acb3bfc6/src/spatial/explog.hpp#L488) and 1e-6 for [Jlog3](https://github.com/stack-of-tasks/pinocchio/blob/912418bd2d1b707800162f131e2013d6acb3bfc6/src/spatial/explog.hpp#L209).
It would be nice to make uniform the behavior of explog Taylor expansions.

@jmirabel and @florent-lamiraux What do you think?

florent-lamiraux:
I agree.

jmirabel:
I agree too.

jcarpent:
Fixed by #584.

PR #584, opened 2018-11-06
Handle precision for Taylor expansions according to the scalar type

jcarpent:
Solves issue #581

### DIFF
--- a/src/math/taylor-expansion.hpp
+++ b/src/math/taylor-expansion.hpp
@@ -1,0 +1,52 @@
+//
+// Copyright (c) 2018 INRIA
+//
+// This file is part of Pinocchio
+// Pinocchio is free software: you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation, either version
+// 3 of the License, or (at your option) any later version.
+//
+// Pinocchio is distributed in the hope that it will be
+// useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+// of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Lesser Public License for more details. You should have
+// received a copy of the GNU Lesser General Public License along with
+// Pinocchio If not, see
+// <http://www.gnu.org/licenses/>.
+
+#ifndef __se3_math_taylor_expansion_hpp__
+#define __se3_math_taylor_expansion_hpp__
+
+#include <cmath>
+#include <limits>
+
+namespace se3
+{
+  
+  ///
+  /// \brief Helper struct to retrieve some useful information for a Taylor series
+  ///        expansion according to the a given Scalar type.
+  ///
+  /// \tparam Scalar the Scalar type of the Taylor series expansion.
+  ///
+  template<typename Scalar>
+  struct TaylorSeriesExpansion
+  {
+    ///
+    /// \brief Computes the expected tolerance of the argument of a Taylor series expansion for a certain degree
+    ///        according to the machine precision of the given input Scalar.
+    ///
+    /// \tparam degree the degree of the Taylor series expansion.
+    ///
+    template<int degree>
+    static Scalar precision()
+    {
+      static Scalar value = std::pow(std::numeric_limits<Scalar>::epsilon(),Scalar(1.)/Scalar(degree));
+      return value;
+    }
+  }; // struct TaylorSeriesExpansion
+  
+}
+
+#endif // ifndef __se3_math_taylor_expansion_hpp__
--- a/src/spatial/explog.hpp
+++ b/src/spatial/explog.hpp
@@ -108,7 +108,9 @@ namespace se3
     // From runs of hpp-constraints/tests/logarithm.cc: 1e-6 is too small.
     if (theta < PI_value - 1e-2)
     {
-      const Scalar t = ((theta > TaylorSeriesExpansion<Scalar>::template precision<4>())? theta / sin(theta) : Scalar(1)) / Scalar(2);
+      const Scalar t = ((theta > TaylorSeriesExpansion<Scalar>::template precision<4>())
+                        ? theta / sin(theta)
+                        : Scalar(1)) / Scalar(2);
       res(0) = t * (R (2, 1) - R (1, 2));
       res(1) = t * (R (0, 2) - R (2, 0));
       res(2) = t * (R (1, 0) - R (0, 1));
@@ -171,10 +173,9 @@ namespace se3
     
     if (n < TaylorSeriesExpansion<Scalar>::template precision<4>())
     {
-
-      a =   Scalar(1)           - n/Scalar(6)   + n2/Scalar(120);
-      b = - Scalar(1)/Scalar(2) + n/Scalar(24)  - n2/Scalar(720);
-      c =   Scalar(1)/Scalar(6) - n/Scalar(120) + n2/Scalar(5040);
+      a =   Scalar(1)           - n2/Scalar(6);
+      b = - Scalar(1)/Scalar(2) - n2/Scalar(24);
+      c =   Scalar(1)/Scalar(6) - n2/Scalar(120);
     }
     else
     {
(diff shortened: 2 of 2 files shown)


src/spatial/explog.hpp
```diff
@@ -51,7 +52,7 @@ namespace se3
     const Scalar t2 = v.squaredNorm();
     
     const Scalar t = math::sqrt(t2);
-    if(t > 1e-4)
+    if(t > TaylorSeriesExpansion<Scalar>::template precision<4>())
     {
```
jmirabel: Minor comment:
Isn't 4 misleading here. The code does a taylor expansion at the 2nd order only...

jcarpent: I may adjust the code to go up to order 4th. What do you think?

jmirabel: I don't know what is better between a higher order Taylor expansion and a narrower interval of definition of it.
The following code:
```cpp
#include <iostream>
#include <cmath>
#include <iomanip>

int main()
{
    std::cout << std::setprecision(17) << std::fixed;
    std::cout << "sin(t) / t\n";
    for (double d = 0.1; d > 1e-15; d /= 10) {
        double sinc = std::sin(d)/d;
        double sinc_t = 1 - d*d/6;
      std::cout << d << '\t' << sinc << '\t' << sinc_t << '\t' << sinc - sinc_t << std::endl;
    }
    
    std::cout << "\n(1-cos(t)) / t^2\n";
    for (double d = 0.1; d > 1e-15; d /= 10) {
        double cosc = (1-std::cos(d))/(d*d);
        double cosc_t = 0.5 - d*d/24;
      std::cout << d << '\t' << cosc << '\t' << cosc_t << '\t' << cosc - cosc_t << std::endl;
    }
}
```
outputs
```
sin(t) / t

0.10000000000000001	0.99833416646828155	0.99833333333333329	0.00000083313494825
0.01000000000000000	0.99998333341666645	0.99998333333333334	0.00000000008333312
0.00100000000000000	0.99999983333334164	0.99999983333333331	0.00000000000000833
0.00010000000000000	0.99999999833333342	0.99999999833333331	0.00000000000000011
0.00001000000000000	0.99999999998333322	0.99999999998333333	-0.00000000000000011
0.00000100000000000	0.99999999999983336	0.99999999999983336	0.00000000000000000
0.00000010000000000	0.99999999999999833	0.99999999999999833	0.00000000000000000
0.00000001000000000	1.00000000000000000	1.00000000000000000	0.00000000000000000
0.00000000100000000	1.00000000000000000	1.00000000000000000	0.00000000000000000
0.00000000010000000	1.00000000000000000	1.00000000000000000	0.00000000000000000
0.00000000001000000	1.00000000000000000	1.00000000000000000	0.00000000000000000
0.00000000000100000	1.00000000000000000	1.00000000000000000	0.00000000000000000
0.00000000000010000	1.00000000000000000	1.00000000000000000	0.00000000000000000
0.00000000000001000	1.00000000000000000	1.00000000000000000	0.00000000000000000

(1-cos(t)) / t^2

0.10000000000000001	0.49958347219741783	0.49958333333333332	0.00000013886408451
0.01000000000000000	0.49999583334736641	0.49999583333333331	0.00000000001403311
0.00100000000000000	0.49999995832550326	0.49999995833333333	-0.00000000000783007
0.00010000000000000	0.49999999696126451	0.49999999958333335	-0.00000000262206884
0.00001000000000000	0.50000004137018539	0.49999999999583333	0.00000004137435206
0.00000100000000000	0.50004445029117028	0.49999999999995831	0.00004445029121197
0.00000010000000000	0.49960036108132022	0.49999999999999956	-0.00039963891867933
0.00000001000000000	0.00000000000000000	0.50000000000000000	-0.50000000000000000
0.00000000100000000	0.00000000000000000	0.50000000000000000	-0.50000000000000000
0.00000000010000000	0.00000000000000000	0.50000000000000000	-0.50000000000000000
0.00000000001000000	0.00000000000000000	0.50000000000000000	-0.50000000000000000
0.00000000000100000	0.00000000000000000	0.50000000000000000	-0.50000000000000000
0.00000000000010000	0.00000000000000000	0.50000000000000000	-0.50000000000000000
0.00000000000001000	0.00000000000000000	0.50000000000000000	-0.50000000000000000
```

src/spatial/explog.hpp
```diff
@@ -168,7 +169,7 @@ namespace se3
     Scalar n2 = r.squaredNorm(),a,b,c;
     Scalar n = math::sqrt(n2);
     
-    if (n < 1e-6)
+    if (n < TaylorSeriesExpansion<Scalar>::template precision<4>())
     {
 
       a =   Scalar(1)           - n/Scalar(6)   + n2/Scalar(120);
```
jmirabel: Btw, aren't this Taylor expansion is wrong ? I think they should be:
```cpp
a =   Scalar(1)           - n2/Scalar(6);
b = - Scalar(1)/Scalar(2) - n2/Scalar(24);
c =   Scalar(1)/Scalar(6) - n2/Scalar(120);
```

src/spatial/explog.hpp
```diff
@@ -206,7 +207,7 @@ namespace se3
     Matrix3Like & Jout = const_cast<Matrix3Like &>(Jlog.derived());
     typedef typename Matrix3Like::Scalar Scalar3;
     
-    if (theta < 1e-6)
+    if (theta < TaylorSeriesExpansion<Scalar>::template precision<4>())
       Jout.setIdentity();
```
jmirabel: Here, shouldn't it be ?
```cpp
Jout.diagonal().setConstant(2 - theta*theta / Scalar(6));
Jout(0,1) = -log(2); Jout(1,0) =  log(2);
Jout(0,2) =  log(1); Jout(2,0) = -log(1);
Jout(1,2) = -log(0); Jout(2,1) =  log(0);
Jout /= Scalar3(2);
double alpha = Scalar(1)/Scalar(12) + theta*theta / Scalar(720);
Jout += alpha * log * log.transpose();
```

jcarpent: Yes, I agree. I don't know why is there. It seems you did that (git blame).
Anyway, I will add this line to my PR.

src/spatial/explog.hpp
jcarpent: You're right too. I will apply this change.

src/spatial/explog.hpp
jcarpent: I would say here that we are making a Taylor expansion at order 3 instead, and then the error is of order M*|x|^4 / 4! , and this error must be evaluated at the numerical precision of the scalar type.

jcarpent:
@jmirabel I have done the merge. Feel free to update the discussion with new remarks. I will add them to a future PR if needed.

src/spatial/explog.hpp
jmirabel: I think the decision should be driven by the numerical precision of the expression we want to evaluate. In this case, the limitation comes from `(1-cos(t)) / t^2`. It seems it starts to be numerically unstable somewhere between `t = 0.001` and `t = 0.0001`. Lets call this limit `T`.

This means that we must find the order `k` of the Taylor expansion such that `M * T^{k+1} / (k+1)!` is close to the machine precision. In this case, `k=3` gives `2e-15` which is good enough. So we need the 3rd order Taylor expansion.

Fortunately, that's what the code does and `TaylorSeriesExpansion<Scalar>::template precision<4>()` is the good value.

src/math/taylor-expansion.hpp
```diff
@@ -0,0 +1,52 @@
...
+    ///
+    /// \brief Computes the expected tolerance of the argument of a Taylor series expansion for a certain degree
+    ///        according to the machine precision of the given input Scalar.
+    ///
+    /// \tparam degree the degree of the Taylor series expansion.
+    ///
+    template<int degree>
+    static Scalar precision()
+    {
+      static Scalar value = std::pow(std::numeric_limits<Scalar>::epsilon(),Scalar(1.)/Scalar(degree));
```
jmirabel: Assuming the user does a Taylor expansion at order `k = degree`, the precision he/she wants is `pow(epsilon(),1/(k+1))`
So I would change this line into:
```cpp
static Scalar value = std::pow(std::numeric_limits<Scalar>::epsilon(),Scalar(1.)/Scalar(degree+1));
```

So that in the code,
```cpp
if (t < TaylorSeriesExpansion<double>::precision<3>()) {
  // Use 3rd order Taylor series.
} else {
  // Use original expression.
}
```

jcarpent: I do agree!

Issue #570, opened 2018-10-29
From the pre-release 2.0.0-alpha to the official release 2.0.0


jcarpent:
What to do next?

Before releasing the new version of Pinocchio 2.0.0, I would suggest fixing the following bullet points:

- [x] fix the issues related to 2.0.0-alpha on dependencies;
- [x] adjust the current open PRs to the current pre-release;
- [x] change the main namespace from se3 to pinocchio and make se3 namespace deprecated;
- [x] change macro naming to follow the standard PINOCCHIO_"MACRO_NAME";
- [x] change the license from LGPL to BSD-2;
- [x] remove some 'boring' warnings from GCC; 
- [x] make the library 100% header only (this assumes that sample-models are now header-only);

Feel free to add additional points.

gabrielebndn:
Completely agree, especially on warnings, header-only and macro-naming

jmirabel:
A CMake option to generate a library containing a default implementation would be nice as well and should not be hard to obtain. The compilation time was already an issue and keep increasing a lot. It would save time to developers and energy...

The method on how to do it and the content of the library should not be discussed here.

jmirabel:
And a *porting from 1.3 to 2* note.

gabrielebndn:
I believe we should study a "rationalization" of Pinocchio headers, as @nmansard put it in the Overview of the doc. That is to say, super-headers that include smaller headers, grouped by theme, the way they do in Eigen with `#include <Eigen/Core>`. This will definitely make Pinocchio easier to use in C++, but we should keep in mind that, if done incorrectly, this may non-negligibly further increase compilation time (I prepared an example). I already have a grouping in mind. If you think it is a good idea, you can open an issue on the subject

gabrielebndn:
I also think that when Pinocchio 2.0.0 is officially out, a "minimal" doc should be available in the master branch. By minimal, I mean what we have now in topic/doc-v2 + clean doxygen comments

jcarpent:
This is already in the pipe ;)
But in any case, by gathering headers, it won't improve compilation times.

jcarpent:
When I said in the pipe, I meant that I have already implemented such a structure.
But this one should be added carefully to the repository, to avoid breaking of inclusion for current projects using Pinocchio.

gabrielebndn:
[...]

PR #586, opened 2018-11-07
all: add PINOCCHIO_ prefix at the begining of each MACRO naming

jcarpent:
Following discussions in #570


src/deprecated-macros.hpp
```diff
@@ -0,0 +1,48 @@
...
+#ifndef __se3_deprecated_macros_hpp__
+#define __se3_deprecated_macros_hpp__
+
+#ifdef PINOCCHIO_WITH_HPP_FCL
+  #ifndef WITH_HPP_FCL
+    #define WITH_HPP_FCL // for backward compatibility
+  #endif
+  #ifndef PINOCCHIO_DISABLE_DEPRECATED_MACRO_WARNING
+    #pragma message("'WITH_HPP_FCL' is deprecated. Please use PINOCCHIO_WITH_HPP_FCL instead.")
+  #endif
```
gabrielebndn: Is this way of handling the deprecation correct? It seems the warning will be issued even if `WITH_HPP_FCL` is not defined. Same for the other macros

jcarpent: It is driven by PINOCCHIO_WITH_HPP_FCL which defined and propagated by the Pinocchio project.

gabrielebndn: It seems really weird. So, if `PINOCCHIO_WITH_HPP_FCL` is defined, the `WITH_HPP_FCL` is issued, and at the same time a warning is given that `WITH_HPP_FCL` is deprecated? This does not seem correct

gabrielebndn: Shouln't `PINOCCHIO_WITH_HPP_FCL` and `WITH_HPP_FCL` be switched here?
```
#ifdef WITH_HPP_FCL
  #ifndef PINOCCHIO_WITH_HPP_FCL
    #define PINOCCHIO_WITH_HPP_FCL // for backward compatibility
  #endif
  #ifndef PINOCCHIO_DISABLE_DEPRECATED_MACRO_WARNING
    #pragma message("'WITH_HPP_FCL' is deprecated. Please use PINOCCHIO_WITH_HPP_FCL instead.")
  #endif
```
With the same for all other deprecated macros?

jmirabel: Another option that the `PINOCCHIO_DISABLE_DEPRECATED_MACRO_WARNING` could be `PINOCCHIO_ENABLE_COMPATIBILITY_WITH_VERSION_1` and define `WITH_HPP_FCL` if the backward compatibility was requested, without compilation warning.

jcarpent: It is Pinocchio that manages PINOCCHIO_WITH_HPP_FCL or WITH_HPP_FCL.
So, if the user has used WITH_HPP_FCL in his project, it has to be defined. This is done here for backward compatibility.

In fact, there is no clean way of making a deprecated Macro as far as I know.

jmirabel: Cannot this be handled in CMake, i.e. define WITH_HPP_FCL is backward compatibility is requested ?

gabrielebndn: I see the intended use case now. It is not really clean that the user is forced to issue a macro in order to silence a warning which is not caused by his own code. My feeling is that if a user wants to keep using deprecated macros, it is him that should handle them, not those who switch to the updated version

gabrielebndn: Also, one reason of adding the `PINOCCHIO` prefix was to avoid name clashes. This solution does not solve this problem. Every time `PINOCCHIO_WITH_HPP_FCL` is defined, then `WITH_HPP_FCL` is also defined.

jcarpent: The solution from @jmirabel solves this.

gabrielebndn: I also agree with @jmirabel

jcarpent: @jmirabel Do you think that `PINOCCHIO_ENABLE_COMPATIBILITY_WITH_VERSION_1` should be defined ON by default?

jmirabel: I don't think so. I assumed that people moving from Pinocchio 1 to Pinocchio 2 are ready to dig further into encountered problems. It should be mentioned in the porting notes.

Anyway, for WITH_HPP_FCL, I don't think there will be much problem.

PR #588, opened 2018-11-08
Change from namespace se3 to pinocchio

jcarpent:
Related to discussions in #570.

The former se3 is still valid. The deprecated message is not shown if `PINOCCHIO_ENABLE_COMPATIBILITY_WITH_VERSION_1` is defined.

### DIFF
--- a/src/deprecated-namespaces.hpp
+++ b/src/deprecated-namespaces.hpp
@@ -15,11 +15,11 @@
 // Pinocchio If not, see
 // <http://www.gnu.org/licenses/>.
 
-#ifndef __se3_deprecated_namespaces_hpp__
-#define __se3_deprecated_namespaces_hpp__
+#ifndef __pinocchio_deprecated_namespaces_hpp__
+#define __pinocchio_deprecated_namespaces_hpp__
 
 #ifdef PINOCCHIO_ENABLE_COMPATIBILITY_WITH_VERSION_1
 #define se3 PINOCCHIO_PRAGMA_MESSAGE_CALL("The se3 namespace has been set to deprecated since version 2.0.0. Please use namespace pinocchio instead") pinocchio
 #endif
 
-#endif // ifndef __se3_deprecated_namespaces_hpp__
+#endif // ifndef __pinocchio_deprecated_namespaces_hpp__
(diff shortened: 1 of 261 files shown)


src/deprecated-namespaces.hpp
```diff
@@ -0,0 +1,27 @@
...
+// Pinocchio If not, see
+// <http://www.gnu.org/licenses/>.
+
+#ifndef __pinocchio_deprecated_namespaces_hpp__
+#define __pinocchio_deprecated_namespaces_hpp__
+
+#if PINOCCHIO_ENABLE_COMPATIBILITY_WITH_VERSION_1 // do not warn
+  #define se3 ::pinocchio
+#else
+  #define se3 PINOCCHIO_PRAGMA_MESSAGE_CALL("The se3 namespace has been deprecated since version 2.0.0. Please use namespace pinocchio instead.") ::pinocchio
```
jmirabel: I think this is a very bad idea. Using a macro does not make the code compatible with Pinocchio 1.

I would merely do:
```cpp
#if PINOCCHIO_ENABLE_COMPATIBILITY_WITH_VERSION_1 // do not warn
  namespace se3 = ::pinocchio;
#endif
```

unittest/liegroups.cpp
```diff
@@ -544,13 +544,13 @@ BOOST_AUTO_TEST_CASE ( test_size )
   BOOST_CHECK (se2.name () == "SE(2)");
   BOOST_CHECK (se2.neutral () == neutral);
   // SE(3): neutral = [0, 0, 0, 0, 0, 0, 1]
-  SpecialEuclideanOperationTpl <3,double> se3;
+  SpecialEuclideanOperationTpl <3,double> se3_group;
   neutral.resize (7); neutral.setZero ();
```
jmirabel: To emphasize the reason why the `#define se3 pinocchio` is bad: you had to modify a unit test, which can be considered as user code.

This is not what I call compatible. I think you should remove this change.

src/deprecated-namespaces.hpp
jcarpent: I understand. This also means that users would be only informed by reading release notes.

unittest/liegroups.cpp
jcarpent: I have the same opinion. Let's go for that then.

jcarpent:
@jmirabel I have made the change. Can you approve?

jmirabel:
Thank you.

jcarpent:
Thanks to you too.

PR #590, opened 2018-11-09
[geometry] fix addGeometryObject

gabrielebndn:
Method `addGeometryObject(object,model,bool);` is now deprecated, but it contained a small bug, so I fixed it (it was causing a warning)


jcarpent:
Thanks!

Issue #549, opened 2018-10-17
Switch to BSD v2.0


olivier-stasse:
Dear all,
We will soon switch to BSD License v2.0 to ease our collaborations.
Let me know if there is a problem with this.
Cheers,
Olivier.

jcarpent:
This has to be discussed with all the partners of the project: Wandercraft, INRIA, ... and also the Authors.
From my side, this change of license has to be discussed and justified further.

jcarpent:
@olivier-stasse The first package that should be modified for the license is the jrl-cmake-modules, which is LGPL.
Please consider this package as your first priority.

nmansard:
I will contact WAN to check. I don't think it is a problem.
Sorry, but INRIA is not yet a partner of this project. We can revert any commit by INRIA if you believe that some of your recent commit is more than just release your previous work done when under CNRS contract.

jcarpent:
The code I'm writing now (with new features) and I'm pushing now is INRIA.

jcarpent:
Just to be clear, release 2.0.0 has been made (also) during my stay at INRIA, under the LGPL license. I won't revert any commit from it as I'm about to release my code.

aelkhour:
> I will contact WAN to check. I don't think it is a problem.

@nmansard @olivier-stasse Wandercraft agrees with switching to the standard three-clause BSD license (BSD License v2.0).

jcarpent:
@aelkhour Thanks for your answer.

jcarpent:
@aelkhour The same applies for eigenpy. Does Wandercraft agree?

aelkhour:
> @aelkhour The same applies for eigenpy. Does Wandercraft agree?

Yes Wandercraft also agree to switch eigenpy to the standard three-clause BSD license (BSD License v2.0).

I'm not sure it's the case for eigenpy, but Wandercraft has authored a number of commits in pinocchio. @olivier-stasse Could you ensure that Wandercraft still appears as author after making the license switch ?

olivier-stasse:
@aelkhour. We will pay attention with respect to eigenpy.
@jcarpentier. LGPL is problematic for code which are embedded. This has
happen for a lot of projects where embedded application are being
developped.
The code is statically embedded in the application and thus become
contaminent.
I think this is not nice for our industrial partner.

The LGPL licence has further problem: most of our code is not
industrialized and very often need investment to be robustify and
integrated into real world application.
[...]

PR #592, opened 2018-11-09
Change from LGPL license to BSD-2-Clause

jcarpent:
Solved issue #549


jcarpent:
Also related to #570

jcarpent:
For information, I only kept the copyright holder in the files. I removed the license from them because the one at the root of the project is sufficient.

PR #594, opened 2018-11-12
Set of minor fixes

jmirabel:
I am finalizing the tests in HPP using Pinocchio v2.

I played a bit with JointCollection. I have one minor question. Is there a reason to use:
```cpp
template<typename Scalar, int Options=0,
                template<typename,int> class JointCollectionTpl = JointCollectionDefaultTpl>
struct Model
```
rather than this:
```cpp
template<typename Scalar, int Options=0,
               class JointCollection = JointCollectionDefaultTpl<Scalar, Options> >
struct Model
```
?

With the current option, it is not possible to do:
```cpp
template <template<typename,int> class JointCollectionTpl>
void some_function_I_want_to_call (...)

some_function_I_want_to_call <Model::JointCollection> (...)
```

### DIFF
--- a/doc/Doxyfile.extra.in
+++ b/doc/Doxyfile.extra.in
@@ -524,7 +524,7 @@ FORMULA_FONTSIZE       = 12
 # If the GENERATE_LATEX tag is set to YES (the default) Doxygen will
 # generate Latex output.
 
-GENERATE_LATEX         = YES
+GENERATE_LATEX         = NO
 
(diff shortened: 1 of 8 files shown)


jcarpent:
I'm going to merge. Thanks for the contribution.

jmirabel:
> With the current prototype, I want to force the class to be templated by a Scalar and an Options.

I see. It isn't blocking but I think it would be good to add an access to the template parameter.

PR #596, opened 2018-11-12
Joint bindings + fix explog

gabrielebndn:
- exposed constructors and axis for JointModelPrismaticUnaligned and JointModelRevoluteUnaligned
- exposed some constructors and addJoint for JointModelComposite
- fixed bugs in python's explog
- related minor stuff

While working on the unit test for JointModelComposite, I found the unit tests using explog were not working any more, so I fixed them.
As a rule of thumb, If we have unit tests, better use them

### DIFF
--- a/bindings/python/multibody/joint/joints-models.hpp
+++ b/bindings/python/multibody/joint/joints-models.hpp
@@ -8,6 +8,7 @@
 #include <eigenpy/exception.hpp>
 #include <eigenpy/eigenpy.hpp>
 #include "pinocchio/multibody/joint/joint-collection.hpp"
+#include "pinocchio/multibody/joint/joint-composite.hpp"
 
 namespace pinocchio
 {
@@ -23,6 +24,7 @@ namespace pinocchio
       return cl;
     }
 
+    // specialization for JointModelRevoluteUnaligned
     template<>
     inline bp::class_<JointModelRevoluteUnaligned>& expose_model<JointModelRevoluteUnaligned> (bp::class_<JointModelRevoluteUnaligned> & cl)
     {
@@ -36,6 +38,7 @@ namespace pinocchio
                ;
     }
 
+    // specialization for JointModelPrismaticUnaligned
     template<>
     inline bp::class_<JointModelPrismaticUnaligned>& expose_model<JointModelPrismaticUnaligned> (bp::class_<JointModelPrismaticUnaligned> & cl)
     {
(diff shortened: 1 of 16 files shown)


bindings/python/spatial/expose-explog.cpp
```diff
@@ -25,7 +25,7 @@ namespace pinocchio
               "Jacobian of exp(R) which maps from the tangent of SO(3) at exp(v) to"
               " the tangent of SO(3) at Identity.");
       
-      bp::def("log3",(Eigen::Vector3d (*)(const Eigen::MatrixBase<Eigen::Matrix3d> &))&log3<Eigen::Matrix3d>,
+      bp::def("log3",&log3_proxy,
```
jcarpent: Why? I prefer to use cast here because I will then expose new signature for types other than double.

bindings/python/multibody/joint/joints-models.hpp
```diff
@@ -16,17 +17,79 @@ namespace pinocchio
     namespace bp = boost::python;
 
 
-    // generic expose_constructor : do nothing special
+    // generic expose_model : do nothing special
     template <class T>
-    inline bp::class_<T>& expose_constructors(bp::class_<T>& cl)
+    inline bp::class_<T>& expose_model(bp::class_<T>& cl)
```
jcarpent: expose_joint_model is better naming

jcarpent:
>     * exposed constructors and axis for JointModelPrismaticUnaligned and JointModelRevoluteUnaligned
> 
>     * exposed some constructors and addJoint for JointModelComposite
> 
>     * fixed bugs in python's explog
> 
>     * related minor stuff
> 
> 
> While working on the unit test for JointModelComposite, I found the unit tests using explog were not working any more, so I fixed them.
> As a rule of thumb, If we have unit tests, better use them

I don't think that Python are run on Travis.

bindings/python/spatial/expose-explog.cpp
gabrielebndn: It was crashing, I could not understand why. This was the quickest way to fix it. But I can come back on this later

bindings/python/multibody/joint/joints-models.hpp
gabrielebndn: Ok

gabrielebndn:
> > As a rule of thumb, If we have unit tests, better use them
> 
> I don't think that Python are run on Travis.

I mean, during the development process, better to run them manually to check everything still works

bindings/python/spatial/explog.hpp
```diff
@@ -53,6 +53,11 @@ namespace pinocchio
       return res;
     }
     
+    Eigen::Vector3d log3_proxy(Eigen::Matrix3d R)
```
jcarpent: Use a reference here then.

gabrielebndn: Oh! Don't know how I did that! Will fix it immediately

PR #597, opened 2018-11-13
[bindings] fix explog

gabrielebndn:
The bindings for explog were not working.
I obtained segmentation faults and/or wrong results when using fixed-size types as template parameters, so I had to switch to dynamic size (tested on my computer and on @jmirabel's laptop).

Apparently, this only happens when Pinocchio is compiled in *Release* mode. It does not happen in Debug mode, and it does not happen when neither Debug nor Release is specified.

It seems the problem is related to eigenpy. We only tested for eigenpy in Release mode.
This pull request patches the problem for explog in Pinocchio, but the problem in eigenpy should be investigated.

### DIFF
--- a/bindings/python/scripts/deprecated.py
+++ b/bindings/python/scripts/deprecated.py
@@ -101,9 +101,6 @@ def exp6FromVector(vector6):
 def log6FromSE3(transform):
   return pin.log6(transform)
 
-@deprecated("This function will be removed in future releases of Pinocchio. You can build a SE3 transform from a rotation matrix and a translation vector and use the standard log function to recover the same behavior.")
+@deprecated("This function will be removed in future releases of Pinocchio. You can use log or log6.")
 def log6FromMatrix(matrix4):
-  M = pin.SE3()
-  M.rotation = matrix4[:3,:3]
-  M.translation = np.matrix(matrix4[:3,-1]).T
-  return pin.log6(M)
+  return pin.log6(matrix4)
--- a/bindings/python/scripts/explog.py
+++ b/bindings/python/scripts/explog.py
@@ -30,10 +30,7 @@ def log(x):
         return math.log(x)
     if isinstance(x, np.ndarray):
         if x.shape == (4, 4):
-            M = pin.SE3()
-            M.rotation = x[:3,:3]
-            M.translation = np.matrix(x[:3,-1]).T
-            return pin.log6(M)
+            return pin.log6(x)
         if x.shape == (3, 3):
             return pin.log3(x)
         raise ValueError('Error only 3 and 4 matrices are allowed.')
--- a/src/spatial/explog.hpp
+++ b/src/spatial/explog.hpp
@@ -373,12 +373,7 @@ namespace pinocchio
   {
     PINOCCHIO_ASSERT_MATRIX_SPECIFIC_SIZE (Matrix4Like, M, 4, 4);
 
-    typedef typename SE3Tpl<typename Matrix4Like::Scalar,Matrix4Like::Options>::Vector3 Vector3;
-    typedef typename SE3Tpl<typename Matrix4Like::Scalar,Matrix4Like::Options>::Matrix3 Matrix3;
-
-    Matrix3 rot(M.template block<3,3>(0,0));
-    Vector3 trans(M.template block<3,1>(0,3));
-    SE3Tpl<typename Matrix4Like::Scalar,Eigen::internal::traits<Matrix4Like>::Options> m(rot, trans);
+    SE3Tpl<typename Matrix4Like::Scalar,Eigen::internal::traits<Matrix4Like>::Options> m(M);
     return log6(m);
   }
 
--- a/unittest/explog.cpp
+++ b/unittest/explog.cpp
@@ -157,6 +157,10 @@ BOOST_AUTO_TEST_CASE(Jexp3_fd)
   Motion::Vector3 v = log3(R);
 
   SE3::Matrix3 Jexp_fd, Jexp;
+
+  Jexp3(Motion::Vector3::Zero(), Jexp);
+  BOOST_CHECK(Jexp.isIdentity());
+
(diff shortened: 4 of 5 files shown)


bindings/python/spatial/expose-explog.cpp
```diff
@@ -15,22 +15,22 @@ namespace pinocchio
...
     {
       
-      bp::def("exp3",&exp3<Eigen::Vector3d>,
+      bp::def("exp3",&exp3<Eigen::VectorXd>,
               bp::arg("Angular velocity (vector of size 3)"),
               "Exp: so3 -> SO3. Return the integral of the input"
               " angular velocity during time 1.");
       
-      bp::def("Jexp3",&Jexp3_proxy<Eigen::Vector3d>,
+      bp::def("Jexp3",&Jexp3_proxy<Eigen::VectorXd>,
```
jcarpent: Why are you doing that: changing from Vector3d to VectorXd?

bindings/python/spatial/explog.hpp
```diff
@@ -23,9 +23,10 @@ namespace pinocchio
     }
     
     template<typename Matrix3Like>
-    typename EIGEN_PLAIN_TYPE(Matrix3Like) Jlog3_proxy(const Eigen::MatrixBase<Matrix3Like> & M)
+    typename Eigen::Matrix<typename EIGEN_PLAIN_TYPE(Matrix3Like)::Scalar,3,3,EIGEN_PLAIN_TYPE(Matrix3Like)::Options>
```
jcarpent: I don't know why you have to change here?

bindings/python/spatial/expose-explog.cpp
```diff
@@ -50,7 +50,7 @@ namespace pinocchio
               "Log: SE3 -> se3. Pseudo-inverse of exp from SE3"
               " -> { v,w in se3, ||w|| < 2pi }.");
       
-      bp::def("log6",&log6<Eigen::Matrix4d>,
+      bp::def("log6",&log6<Eigen::MatrixXd>,
```
jcarpent: I don't think that Matrix4d is exposed. But it should be.

bindings/python/spatial/expose-explog.cpp
gabrielebndn: That's what I explain above: it was giving wrong results with `Vector3d,` so I had to switch to `VectorXd`

bindings/python/spatial/explog.hpp
gabrielebndn: I had to do this in order to be able to use `MatrixXd` as template parameter instead of `Matrix3d`, which I did for the reasons explained above

gabrielebndn:
Baiscally, with pinocchio and eigenpy both in Release mode, do this simple test:
```
v = pinocchio.utils.zero(3)
J = pinocchio.Jexp3(v)
```
The result should be the identity. Instead, I obtain something like diag([1,0.5,1.6]). This only in Release mode, because in Debug and with unspecified mode it gives the correct result

bindings/python/spatial/expose-explog.cpp
jcarpent: It is really weird. It is better to investigate first eigenpy rather than putting some fix here.

gabrielebndn: I would rather have a working system now and think about solving this problem later. The transition on Pinocchio 2.0 is already tough enough. The correction as it stands now does not do a lot of harm, and it will be very easy to roll back.
Anyhow, you can take the newly-added lines in unittest/python/explog as a reference.
In this PR I also include a correction to python's log function.

jcarpent:
I don't have any trouble from my side with `J = pinocchio.Jexp3(v)`.
Your issue seems to come from another place. Are you up to date with eigenpy?

gabrielebndn:
I am using the up-to-data installation from source. @jmirabel was also using the source installation. Are you testing in Release mode? Both Pinocchio and Eigenpy? Did you check the result is correct? @jmirabel and I were both getting the same wrong result, [[1.,0.5,0],[-0.5,0.5,0],[0,0,1.6]] if I recall.

jcarpent:
Yes, the result is identity.
I'm indeed using both of them in release mode.

jcarpent:
I got it. It comes from EigenPy and the support of `Eigen::MatrixBase<MatrixDerived>. The right destructor is not correctly invoked. I think this is related to stack-of-tasks/eigenpy#20.

gabrielebndn:
Then it's really weird. It's surprising @jmirabel and I are experiencing the same behavior. You may try to test the other methods too, most of them were segmenting, but at this point I expect they will work for you.

It is probably better to discard my fix. I will change the PR so to remove most fixes to the bindings, but keep the modified unit test for future reference. Do you agree?

gabrielebndn:
Saw your last comment. Then why is working for you, but not for us?

jcarpent:
It is mostly due to Eigen in fact.
Depending on your compilation flags, Eigen chooses one of the constructors (they are several for aligned data or not for instance). And the destructor should be in adequation with it. But when you compile EigenPy and Pinocchio with different options, bad things may happen. 
I already discussed this point with Eigen's main developers, but the fix I proposed was not satisfying for them.

I have suggested a while ago, to remove the shared library from eigenpy to avoid such segmentation fault. I think this is time to implement this strategy.

bindings/python/spatial/expose-explog.cpp
gabrielebndn: It seems to be exposed

gabrielebndn:
Voilà. As #598 solved my problem, I've updated the PR.
Now basically all I'm doing is adding and fixing unit tests.
I've also fixed a little problem is python's `log` function, and slightly simplified the implementation of `log6(Matrix4d)` in C++

jcarpent:
looks good now! 
@gabrielebndn Thanks for this PR.

PR #598, opened 2018-11-14
bindings/python: fix bug related to eigenpy

jcarpent:
It seems that the desctructor invoked by eigenpy does not fit with the constructor invoked in Pinocchio.

This PR is related to #597

PR #600, opened 2018-11-16
[bindings] Fix Jcom + remove se3 namespace

gabrielebndn:


gabrielebndn:
I also fixed the bindings for `exp3` because a unit test was failing on my Macbook (related to #597  and #598)

PR #602, opened 2018-11-20
Add missing /opt/openrobots/lib/plugin folder

jviereck:
When installing `sot-core`, the `/opt/openrobots/lib/plugin` was missing in the `LD_LIBRARY_PATH`. This PR fixes it.


jcarpent:
The notion of `plugin` is something very related to sot-core, not to Pinocchio.
For this reason, I think those lines should go there.

jviereck:
Okay, thanks for the explanation! Going to close this then.

jcarpent:
No problem. Thank you for your contributions.

PR #603, opened 2018-11-21
[python] [JointModelComposite] Various fixes

gabrielebndn:
Mostly python fixes
- Fixed a few things in `RobotWrapper` and `RomeoWrapper`
- Finished exposing `JointModelComposite` constructors and data
- Additionally, I slightly modified `addJoint` in `JointModelComposite`.
It now returns a reference to the same joint being modified, so that the expressions can be easily chained, as in
`joint_composite.addJoint(joint1).addJoint(joint2).addJoint(joint3);`

### DIFF
--- a/bindings/python/multibody/joint/joints-models.hpp
+++ b/bindings/python/multibody/joint/joints-models.hpp
@@ -54,7 +54,7 @@ namespace pinocchio
 
     // specialization for JointModelComposite
 
-    struct JointModelCompositeAddJointVisitor : public boost::static_visitor<void>
+    struct JointModelCompositeAddJointVisitor : public boost::static_visitor<JointModelComposite &>
     {
       JointModelComposite & m_joint_composite;
       const SE3 & m_joint_placement;
@@ -66,22 +66,22 @@ namespace pinocchio
       {}
 
       template <typename JointModelDerived>
-      void operator()(JointModelDerived & jmodel) const
+      JointModelComposite & operator()(JointModelDerived & jmodel) const
       {
         return m_joint_composite.addJoint(jmodel,m_joint_placement);
       }
     }; // struct JointModelCompositeAddJointVisitor
 
-    static void addJoint_proxy(JointModelComposite & joint_composite,
-                               const JointModelVariant & jmodel_variant,
-                               const SE3 & joint_placement = SE3::Identity())
+    static JointModelComposite & addJoint_proxy(JointModelComposite & joint_composite,
+                                                const JointModelVariant & jmodel_variant,
+                                                const SE3 & joint_placement = SE3::Identity())
     {
       return boost::apply_visitor(JointModelCompositeAddJointVisitor(joint_composite,joint_placement), jmodel_variant);
     }
 
     BOOST_PYTHON_FUNCTION_OVERLOADS(addJoint_proxy_overloads,addJoint_proxy,2,3)
 
-    struct JointModelCompositeConstructorVisitor : public boost::static_visitor<boost::shared_ptr<JointModelComposite> >
+    struct JointModelCompositeConstructorVisitor : public boost::static_visitor<JointModelComposite* >
     {
       const SE3 & m_joint_placement;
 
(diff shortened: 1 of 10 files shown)


bindings/python/multibody/joint/joints-models.hpp
```diff
@@ -54,41 +54,87 @@ namespace pinocchio
...
+    struct JointModelCompositeConstructorVisitor : public boost::static_visitor<boost::shared_ptr<JointModelComposite> >
+    {
+      const SE3 & m_joint_placement;
+
+      JointModelCompositeConstructorVisitor(const SE3 & joint_placement)
+      : m_joint_placement(joint_placement)
+      {}
+
+      template <typename JointModelDerived>
+      boost::shared_ptr<JointModelComposite> operator()(JointModelDerived & jmodel) const
```
jcarpent: Why shared_ptr?
We must avoid them because they are inefficient, and certainly not useful here.

gabrielebndn: I had problems exposing this constructor.
The C++ signature is the following
```
template<typename JointModel>
JointModelCompositeTpl(const JointModelBase<JointModel> & jmodel,
                                        const SE3 & placement = SE3::Identity())
```
The problem is that to work in python it should accept a `JointModelVariant`, therefore I need a proxy, and this is the way I found to use a proxy for the constructor.
I basically followed the example here
[](https://github.com/boostorg/python/blob/master/test/injected.cpp)
Do you know of any alternatives? I don't want to use a raw pointer, I fear it might cause leaks and all sort of problems

bindings/python/multibody/joint/joints-models.hpp
```diff
@@ -54,41 +54,87 @@ namespace pinocchio
...
+                   )
                .add_property("joints",&JointModelComposite::joints)
-               .def("addJoint",&addJoint_proxy,addJoint_proxy_overloads(bp::args("joint_model","joint_placement"),"Add a joint to the vector of joints."))
+               .add_property("jointPlacements",&JointModelComposite::jointPlacements)
+               .add_property("njoints",&JointModelComposite::njoints)
+               .def("addJoint",
+                    &addJoint_proxy,
+                    addJoint_proxy_overloads(bp::args("joint_model","joint_placement"),
+                                             "Add a joint to the vector of joints."
+                                            )[bp::return_value_policy<bp::return_by_value>()]
```
jcarpent: You should not make a return by value here. It will duplicate with a copy

bindings/python/multibody/joint/joints-models.hpp
jcarpent: The pointer is the right solution

bindings/python/multibody/joint/joints-models.hpp
gabrielebndn: You are right, I've just forced-pushed a fix, with an additional unit test to check it behaves well.
I'm now using `return_internal_reference`

src/multibody/joint/joint-composite.hpp
```diff
@@ -200,9 +200,11 @@ namespace pinocchio
     /// \param jmodel Model of the joint to add.
     /// \param placement Placement of the joint relatively to its predecessor.
     ///
+    /// \return A reference to this JointModelComposite itself
```
jcarpent: `A reference to  *this` is the way Eigen write their documentation. I think we should follow their example.

src/multibody/joint/joint-composite.hpp
```diff
@@ -200,9 +200,11 @@ namespace pinocchio
...
+    ///
     template<typename JointModel>
-    void addJoint(const JointModelBase<JointModel> & jmodel,
-                  const SE3 & placement = SE3::Identity())
+    JointModelDerived& addJoint(const JointModelBase<JointModel> & jmodel,
```
jcarpent: `JointModelDerived &` is more readable I think. And it follows your other modifications.

bindings/python/multibody/joint/joints-models.hpp
jcarpent: Good!

src/multibody/joint/joint-composite.hpp
gabrielebndn: Thanks for the suggestion

src/multibody/joint/joint-composite.hpp
gabrielebndn: I don't know if it is more readable or not, but I am willing to sick to any coding convention we might have, so I fixed it

bindings/python/multibody/joint/joints-models.hpp
gabrielebndn: I changed to the pointer

gabrielebndn:
I pushed some changes (modifying the last commit). I think I've addressed all problems

jcarpent:
Seems good to me. Just waiting for Travis result.

jcarpent:
Thanks for the contribution.

PR #605, opened 2018-11-23
[doc] Port doc to 2.0

gabrielebndn:
Hi,
I successfully merged devel into topic/doc-v2 on my repo, it was straightforward, and I ported the examples to 2.0.

This PR updates the central topic/doc-v2 to 2.0.

I think we will soon be ready to move the first chunk of documentation into devel.

PR #606, opened 2018-11-23
[doc] Update porting notes

gabrielebndn:
Update the porting notes.
Mostly cosmetic changes, but I added a subsection explaining Pinocchio is now header-only, plus additional clarifications on the deprecated macros. @jcarpent may you check it is correct?

I'm directly updating the devel branch, rather than topic/doc-v2 (see PR #605). This is to make the changes immediately available to anybody working with devel.

### DIFF
--- a/doc/_porting.md
+++ b/doc/_porting.md
@@ -18,10 +18,6 @@ Relevant changes are listed below.
 \subsection PortingCHeaderonly Header-only
 Pinocchio is now fully header-only. This means you do not have to link to the Pinocchio library when compiling your code.
 On the other hand, you might need to link to additional system libraries.
-For instance, when using URDF, you now need to add
-```
--lboost_system -lurdfdom_model
-```
 
 \subsection PortingCNamespace Namespace
 The most important change is the namespace.
@@ -48,14 +44,12 @@ The following marcos are not employed anymore
 WITH_HPP_FCL
 WITH_URDFDOM
 WITH_LUA5
-URDFDOM_TYPEDEF_SHARED_PTR
 ```
 in favor of
 ```
 PINOCCHIO_WITH_HPP_FCL
 PINOCCHIO_WITH_URDFDOM
 PINOCCHIO_WITH_LUA5
-PINOCCHIO_URDFDOM_TYPEDEF_SHARED_PTR
 ```
 Therefore, you now need to issue the new macros in your compilation commands.
 
--- a/doc/_porting.md
+++ b/doc/_porting.md
@@ -1,5 +1,7 @@
 # Porting from Pinocchio 1.3.3 to 2.0.0
 
+\section PortingIntro What is included
+
 This section describes how to port your code from the latest Pinocchio 1 release (1.3.3) to 2.0.
 
 Note that this section does not cover API changes that were made *before* Pinocchio 1.3.3.
@@ -9,11 +11,19 @@ In particular, remove all calls to deprecated methods and replace them appropria
 
 The vast majority of the changes took place in C++.
 
-# Changes in C++
+\section PortingC Changes in C++
 Although the class system was heavily re-worked, it should not make a lot of difference from the user's point of view.
 Relevant changes are listed below.
 
-## Namespace
+\subsection PortingCHeaderonly Header-only
+Pinocchio is now fully header-only. This means you do not have to link to the Pinocchio library when compiling your code.
+On the other hand, you might need to link to additional system libraries.
+For instance, when using URDF, you now need to add
+```
+-lboost_system -lurdfdom_model
+```
+
+\subsection PortingCNamespace Namespace
 The most important change is the namespace.
 Now, the top-level Pinocchio namespace is not `se3` anymore, but `pinocchio`.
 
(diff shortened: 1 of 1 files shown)


doc/_porting.md
```diff
@@ -9,11 +11,19 @@ In particular, remove all calls to deprecated methods and replace them appropria
...
-# Changes in C++
+\section PortingC Changes in C++
 Although the class system was heavily re-worked, it should not make a lot of difference from the user's point of view.
 Relevant changes are listed below.
 
-## Namespace
+\subsection PortingCHeaderonly Header-only
+Pinocchio is now fully header-only. This means you do not have to link to the Pinocchio library when compiling your code.
+On the other hand, you might need to link to additional system libraries.
+For instance, when using URDF, you now need to add
```
jcarpent: We do not have to detail here the link against other libraries.
I think we should keep the description as simple as possible.
The link against the dependencies has to be detailed through CMake or Pkg-config usage.
The way you present here is not safe and generic at all.

doc/_porting.md
```diff
@@ -31,26 +41,30 @@ In order to make it work, you need to compile it with the following flag
...
 WITH_URDFDOM
 WITH_LUA5
+URDFDOM_TYPEDEF_SHARED_PTR
 ```
 in favor of
 ```
 PINOCCHIO_WITH_HPP_FCL
 PINOCCHIO_WITH_URDFDOM
 PINOCCHIO_WITH_LUA5
+PINOCCHIO_URDFDOM_TYPEDEF_SHARED_PTR
```
jcarpent: This is normally transparent to the users. So do not name it here.

doc/_porting.md
gabrielebndn: This was not meant at all to be generic and exhaustive. It was just an example based on what I found when porting the examples in doc-v2. If you think the reference to `libboost_system` and `liburdfdom_model` is potentially confusing, I can certainly remove it, but I would leave a generic warning about linking (I assume the Pinocchio library was thinking about it before).

doc/_porting.md
gabrielebndn: I disagree. Users might need to pass this macro as a compilation flag for their code to compile. That's how I discovered it, while porting the `overview-urdf` example in the `doc-v2` branch (see the compilation command). The very fact I discovered it proves it is *not* 100% user-transparent. In fact, it might be a good idea to list here all other macros that were changed, it will not do harm, if you could provide a full list I would be grateful

jcarpent: This information is contained inside the pkg file.
This is where you have to find the information for linking Pinocchio.

jcarpent: And this is how it has been working for a while with Pinocchio!

jcarpent: For instance, `PINOCCHIO_URDFDOM_TYPEDEF_SHARED_PTR`depends on your version of urdfdom package.

gabrielebndn: Then you night want to check the URDF example in topic/doc-v2, because in that example PINOCCHIO_URDFDOM_TYPEDEF_SHARED_PTR is passed explicitly without comments.
Still, I do not see why not documenting the change. These are porting notes meant to ease the migration, the more complete, the better. I repeat, I had to discover it myself in order for that example to work.

gabrielebndn: If the configuration files are meant to be the privileged (if not the only) way of compiling Pinocchio, then they should be properly documented and used throughout the doc, which is not the case now.
Otherwise, if we want to leave open the possibility of compiling without them, and if we are explicitly providing compilation commands, I think these "internal" macros should be documented, including in the porting notes.

jcarpent: Managing all the dependencies by hand is a nightmare. 
That's why Cmake or pkg-config are making our lives easier.
It will become a nightmare again to need to document all the examples with precise compilation options which depend on the system used and also the dependencies.
So, I feel against providing this information in the documentation or porting notes.

nmansard: My opinion:
* in the porting notes, we would like to keep track of this changes. They are for expert users, so don't be afraid of it.
* in the doc: the minimal examples are provided with their explicit compilation instruction, following the example of Eigen doc. The remark of @jcarpent justifies the addition in the doc of compilation instruction, in particular explaining how Pinocchio could be easily integrated in a third-party project.

jcarpent: But what I meant is that there is no a unique compilation line that will work for sure on all the systems.
So, we should add documentation of them here, we are wasting time and we will never be able to maintain it!

jcarpent: If you want to do it correctly, you should use `pkg-config` command line to retrieve the dependencies and the compilation flags. This is the only way to be generic!!!

jcarpent: By the way, @nmansard and @gabrielebndn you should check this package here: https://github.com/stack-of-tasks/pinocchio-minimal
That has been done to show how to use Pinocchio in an external project!!!

gabrielebndn: I knew that package, thanks. But in the doc the example just gave the compilation command and that's how I stumbled on the problem.

jcarpent: But they are not generic. This means that they will work for your configuration but not on my computer for instance.

gabrielebndn:
I really do not agree with this, but your choice. I removed the additional portions from the porting notes. You can now merge this pull request or close it without merging.

jcarpent:
Thanks to having taken my remarks into account.

PR #612, opened 2018-11-30
[doc] Work on Doxygen

gabrielebndn:
Hi,
I've started working on cleaning the Doxygen-generated documentation, turns out it will require more work than I expected.

As for now, I have hidden a few irrelevant things and added the typedefs to their groups, and fixed some latex-related errors. Also, I've added a custom CSS style sheet. I used it to break ling template names inside tables which were making the whole thing unreadable. It is a bit less messy now.

Notice the documentation for /d-practical-exercises/src/ was generated too. I explicitly ignored it, but an alternative would be to move it to a top-level directory, called, e.g., exercises or labs.


jcarpent:
Thanks for the doc contribution.

Issue #555, opened 2018-10-19
[bindings] Expose C++ vectors as python lists


jcarpent:
This issue is a follow-up from a previous project note called: **Fix inconsistent naming: StdVec vs StdVect**

*Low priority. The templated classes of* std::vector *and of its subclass* se3::container::aligned_vector *are exposed in Python as, for instance,* StdVec_Inertia *or* StdVect_Force. *This naming scheme is inconsistent. The python classes should be either all* StdVec *or all* StdVect

@jcarpent reply: std::vector should not be exposed, and list should be preferred. 
This is a bad practice that has been followed then.

@gabrielebndn reply: I do agree that using lists would be much better. Right now, I think there's three Python methods which take a `StdVec[t]` as input, and at least one which returns a `StdVec[t]` I think. These can be changed easily, I have already written a note on that. However, then
there's also all the instance `std::vector`s of `Model` and `Data`: it will require a massive amount of work to expose them all as python lists. Worse, I think it would break some of the current interface

jcarpent:
To just continue this discussion in a more formal way.
You can bind a std::vector to a list without any allocation. This how we must handle std::vector in Python, not the reverse.
Currently, we use the Boost.Python wrapper of std::vector. But this one is just a way to expose vector as a list (in terms of API).

For Model and Data this will behave the same. The std::vector will appear as a Python list.
This requires some works, but this is not hard.

jcarpent:
This issue is solved.

PR #626, opened 2019-01-08
Added bindings to derivative algorithms

nmansard:
I worked a bit with the derivative algos that I never tried before. Very nice work. It like them a lot!

I added a direct and naive binding to the derivative of the joint integration. I believe that it follows the naive joint algorithm implemented in src/algo. Should we had some comment in the documentation (both of the python and the c++ function) to specify that this algorithm is best used at the joint level, using the sparsity of the kinematic tree?
I added a copy of the current ABA derivative to account for external forces, following the pattern of the rnea algorithm. 

NB: 
Meanwhile, I noticed that some StdVector bindings are named StdVec while others are named StdVect. I guess we should unified. I would be interest to now the best way for creating a filled StdVec. I was not able to imagine something convenient from a generator, similar to StdVect_Force([ Force.Random() for j in model.joints]). I rather used fs=StdVect_Force(); for j in model.joints: fs.append(Force.Random()), which is less convenient.

I tested both python bindings, and validate them (again) against finite differences. I also validated that ddq_dq = -M^-1 dtau_dq. I would be wanted to add these unit-tests as python examples. What is the current policy? Should I add them naively in pinocchio/python?

### DIFF
--- a/bindings/python/algorithm/expose-aba-derivatives.cpp
+++ b/bindings/python/algorithm/expose-aba-derivatives.cpp
@@ -17,7 +17,7 @@ namespace pinocchio
       computeABADerivatives(model,data,q,v,tau);
     }
     typedef container::aligned_vector<Force> ForceAlignedVector;
-   void computeABADerivatives_fext(const Model & model, Data & data,
+    void computeABADerivatives_fext(const Model & model, Data & data,
                                     const Eigen::VectorXd & q,
                                     const Eigen::VectorXd & v,
                                     const Eigen::VectorXd & tau,
@@ -25,7 +25,7 @@ namespace pinocchio
     {
       computeABADerivatives(model,data,q,v,tau,fext);
     }
-    
+
     void exposeABADerivatives()
     {
       using namespace Eigen;
@@ -53,6 +53,5 @@ namespace pinocchio
               "velocity and torque vectors.\n"
               "The forces are of type StdVect_Force.");
     }
-    
   } // namespace python
 } // namespace pinocchio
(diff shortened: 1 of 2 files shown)


gabrielebndn:
I agree `StdVec` vs `StdVect` should be unified. I had already raised this issue, see #555 . Then the idea became to directly expose the vectors as python lists, but I do not see how it can be done without copying information

gabrielebndn:
A slightly more convenient way for creating a filled StdVec is
`fs=StdVect_Force(); fs.extend(Force.Random() for j in model.joints)`

jcarpent:
I think StdVec should removed from the API and be convertible directly as a Python list. It will be transparent for the user. See #555 for further comments.

jcarpent:
> I tested both python bindings, and validate them (again) against finite differences. I also validated that ddq_dq = -M^-1 dtau_dq. I would be wanted to add these unit-tests as python examples. What is the current policy? Should I add them naively in pinocchio/python?

I would say that Python should not validate the results. This is the role of core unit tests that are implemented in C++ with intensive testings.
But I imagine that we should only with Python that the API are correctly binding, which take the right input arguments and return the right type of output.

jcarpent:
Concerning the content of this PR, it seems good to me, except for some additional blanks that should not be here.
@nmansard Do you have time to correct them?

nmansard:
@jcarpent about python test/examples: I agree validation should be done in C++. The tests I did in python validate the API binding and can be used as example, e.g. to build the examples that we decided to put in the doc.

@jcarpent about trailing spaces: I quickly checked and corrected some, but am not sure about the convention you are preferring to use in indentation. Please dont hesite to do your job of Pinocchio guru and edit my commit.

jcarpent:
Ok. It sounds good to me. I'm going to merge!
@nmansard Thank for this proposal.

PR #631, opened 2019-01-09
[Python] Update to changes in hpp-fcl.

jmirabel:
There are some work being done in hpp-fcl currently. It should be stabilized by the end of the week. It should not impact Pinocchio except for the following changes.

The following three functions are now marked as deprecated.
```
          .def("isOccupied",&CollisionGeometry::isOccupied)
          .def("isFree",&CollisionGeometry::isFree)
          .def("isUncertain",&CollisionGeometry::isUncertain)
```
Should I remove them from the bindings ?

The reason why only `isOccupied` is removed for now is historical. I will change this once we agree on the strategy.


jcarpent:
Sounds good to me.
I was not using these methods.

jcarpent:
May I make the release afterward?

jmirabel:
Yes, you can.

PR #633, opened 2019-01-09
Update CI, fix #629

### LINKED ISSUE #629: Good news, Travis now supports 16.04
It would then be nice to make a smooth transition to this version of Ubuntu, 14.04 will be deprecated soon.
@nim65s Do you think that you can make it?

See https://docs.travis-ci.com/user/reference/xenial/ for further details.

nim65s:
Hi,

This PR fixes #629 by switching Travis CI builds from trusty to xenial.

While here, I am also pushing a way to fix gitlab-ci cloning process, and therefore version detection, which was required for the build since https://github.com/jrl-umi3218/jrl-cmakemodules/pull/169

Unfortunately, this only makes pinocchio tests working on Gitlab-CI, and not py-pinocchio ones, because of a separate issue.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -18,7 +18,7 @@ cache:
     - cd /root/robotpkg/math
     - git pull
     - cd pinocchio
-    - make checkout MASTER_REPOSITORY="dir ${CI_PROJECT_DIR}"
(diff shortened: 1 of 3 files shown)


jcarpent:
@nim65s I think to make a smooth transition, it would be nice to keep both xenial and trusty. Do you agree?

jcarpent:
Seems to work. @nim65s Thank you for this fast contribution.

PR #636, opened 2019-01-10
all: fix additional bug related to recent versions of boost

jcarpent:
Related to discussions in #511.


jcarpent:
@nim65s Could you try this additional fix on fedora and report any other left issues?

nim65s:
I can confirm that it is working in Fedora 28 :)

PR #645, opened 2019-01-18
[Parser] Provide API to avoid creating twice the same FCL geometry. 

jmirabel:
Needs https://github.com/humanoid-path-planner/hpp-fcl/pull/50

### DIFF
--- a/bindings/python/parsers/parsers.hpp
+++ b/bindings/python/parsers/parsers.hpp
@@ -115,6 +115,34 @@ namespace pinocchio
         return geometry_model;
       }
 
+      static GeometryModel
+      buildGeomFromUrdf(const Model & model,
+                        const std::string & filename,
+                        const GeometryType type,
+                        const fcl::MeshLoaderPtr& meshLoader
+                        )
+      {
+        std::vector<std::string> hints;
+        GeometryModel geometry_model;
+        pinocchio::urdf::buildGeom(model,filename,type,geometry_model,hints,meshLoader);
+        
+        return geometry_model;
+      }
+
+      static GeometryModel
+      buildGeomFromUrdf(const Model & model,
+                        const std::string & filename,
+                        const std::vector<std::string> & package_dirs,
+                        const GeometryType type,
+                        const fcl::MeshLoaderPtr& meshLoader
+                        )
+      {
+        GeometryModel geometry_model;
+        pinocchio::urdf::buildGeom(model,filename,type,geometry_model,package_dirs,meshLoader);
+        
+        return geometry_model;
+      }
+
 #endif // #ifdef PINOCCHIO_WITH_URDFDOM
 
 #ifdef PINOCCHIO_WITH_LUA5
@@ -182,6 +210,17 @@ namespace pinocchio
       
       bp::def("buildGeomFromUrdf",
               static_cast <GeometryModel (*) (const Model &, const std::string &, const GeometryType)> (&ParsersPythonVisitor::buildGeomFromUrdf),
+              bp::args("Model to assosiate the Geometry","filename (string)"),
+              "Parse the URDF file given in input looking for the geometry of the given Model and return a proper pinocchio  geometry model ");
+
+      bp::def("buildGeomFromUrdf",
+              static_cast <GeometryModel (*) (const Model &, const std::string &, const std::vector<std::string> &, const GeometryType, const fcl::MeshLoaderPtr&)> (&ParsersPythonVisitor::buildGeomFromUrdf),
+              bp::args("Model to assosiate the Geometry","filename (string)", "package_dirs (vector of strings)"
+                       ),
+              "Parse the URDF file given in input looking for the geometry of the given Model and return a proper pinocchio geometry model ");
+      
+      bp::def("buildGeomFromUrdf",
+              static_cast <GeometryModel (*) (const Model &, const std::string &, const GeometryType, const fcl::MeshLoaderPtr&)> (&ParsersPythonVisitor::buildGeomFromUrdf),
               bp::args("Model to assosiate the Geometry","filename (string)"),
               "Parse the URDF file given in input looking for the geometry of the given Model and return a proper pinocchio  geometry model ");
       
--- a/bindings/python/parsers/parsers.hpp
+++ b/bindings/python/parsers/parsers.hpp
@@ -220,13 +220,12 @@ namespace pinocchio
 
       bp::def("buildGeomFromUrdf",
               static_cast <GeometryModel (*) (const Model &, const std::string &, const std::vector<std::string> &, const GeometryType, const fcl::MeshLoaderPtr&)> (&ParsersPythonVisitor::buildGeomFromUrdf),
-              bp::args("Model to assosiate the Geometry","filename (string)", "package_dirs (vector of strings)"
-                       ),
+              bp::args("Model to assosiate the Geometry","filename (string)", "package_dirs (vector of strings)","Geometry type", "Mesh loader"),
               "Parse the URDF file given in input looking for the geometry of the given Model and return a proper pinocchio geometry model ");
       
       bp::def("buildGeomFromUrdf",
               static_cast <GeometryModel (*) (const Model &, const std::string &, const GeometryType, const fcl::MeshLoaderPtr&)> (&ParsersPythonVisitor::buildGeomFromUrdf),
-              bp::args("Model to assosiate the Geometry","filename (string)"),
+              bp::args("Model to assosiate the Geometry","filename (string)","Geometry type", "Mesh loader"),
               "Parse the URDF file given in input looking for the geometry of the given Model and return a proper pinocchio  geometry model ");
       
       bp::def("removeCollisionPairs",
--- a/bindings/python/parsers/parsers.hpp
+++ b/bindings/python/parsers/parsers.hpp
@@ -115,6 +115,7 @@ namespace pinocchio
         return geometry_model;
       }
 
+#ifdef PINOCCHIO_WITH_HPP_FCL
       static GeometryModel
       buildGeomFromUrdf(const Model & model,
                         const std::string & filename,
@@ -142,6 +143,8 @@ namespace pinocchio
         
         return geometry_model;
       }
+
+#endif // #ifdef PINOCCHIO_WITH_HPP_FCL
 
 #endif // #ifdef PINOCCHIO_WITH_URDFDOM
 
@@ -213,6 +216,8 @@ namespace pinocchio
               bp::args("Model to assosiate the Geometry","filename (string)"),
               "Parse the URDF file given in input looking for the geometry of the given Model and return a proper pinocchio  geometry model ");
 
+#ifdef PINOCCHIO_WITH_HPP_FCL
+
       bp::def("buildGeomFromUrdf",
               static_cast <GeometryModel (*) (const Model &, const std::string &, const std::vector<std::string> &, const GeometryType, const fcl::MeshLoaderPtr&)> (&ParsersPythonVisitor::buildGeomFromUrdf),
               bp::args("Model to assosiate the Geometry","filename (string)", "package_dirs (vector of strings)"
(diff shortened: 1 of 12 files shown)


jmirabel:
> @jmirabel to ensue compatibility with older versions of hpp-fcl it would be nice to add some ifdef statement according to the versions of hpp-fcl.

I can do it but:
- I don't see any scenario where Pinocchio would be updated and not FCL,
- it makes the code less readable.

bindings/python/parsers/parsers.hpp
```diff
@@ -115,6 +115,34 @@ namespace pinocchio
         return geometry_model;
       }
 
+      static GeometryModel
```
jcarpent: Some `ifdef PINOCCHIO_WITH_HPP_FCL`are missing here

bindings/python/scripts/robot_wrapper.py
```diff
@@ -18,7 +18,7 @@ def BuildFromURDF(filename, package_dirs=None, root_joint=None, verbose=False):
         robot.initFromURDF(filename, package_dirs, root_joint, verbose)
         return robot
 
-    def initFromURDF(self,filename, package_dirs=None, root_joint=None, verbose=False):
+    def initFromURDF(self,filename, package_dirs=None, root_joint=None, verbose=False, meshLoader=None):
```
jcarpent: You must handle the case where Pinocchio is no built with `hpp-fcl`support.

bindings/python/parsers/parsers.hpp
```diff
@@ -184,6 +212,17 @@ namespace pinocchio
               static_cast <GeometryModel (*) (const Model &, const std::string &, const GeometryType)> (&ParsersPythonVisitor::buildGeomFromUrdf),
               bp::args("Model to assosiate the Geometry","filename (string)"),
               "Parse the URDF file given in input looking for the geometry of the given Model and return a proper pinocchio  geometry model ");
+
```
jcarpent: The same applies here.

src/parsers/urdf.hpp
```diff
@@ -10,6 +10,17 @@
 #include "pinocchio/multibody/geometry.hpp"
 #include "pinocchio/parsers/urdf/types.hpp"
 
+/// \cond
+// Commented because this is needed in function prototypes.
+//#ifdef PINOCCHIO_WITH_HPP_FCL
+namespace fcl
+{
+  class MeshLoader;
+  typedef boost::shared_ptr<fcl::MeshLoader> MeshLoaderPtr_t;
```
jcarpent: I would prefer `MeshLoaderPtr` instead of `MeshLoaderPtr_t`in order to make uniform the naming conventions in Pinocchio (still in progress yet).

jcarpent:
We need to first make a new release of `hpp-fcl`, then modifiy the minimal version requested in the `CMakeLists.txt`before merging this PR.

nim65s:
the next release of hpp-fcl will be here by the end of the day

bindings/python/scripts/robot_wrapper.py
jmirabel: What do you expect me to do here. Raise an exception is meshLoader is provided and not built with hpp-fcl ?
If so, how do you get the information about hpp-fcl support in Python ?

jcarpent: One way is to check if hpp-fcl is provided, I was planning to do that by adding a global variable called maybe `__PINOCCHIO_WITH_HPP_FCL__`.
In the RobotWrapper, we have a similar strategy but checking if https://github.com/stack-of-tasks/pinocchio/blob/36aeeebadfe1c55675f2265bdb8fe534d50e95cb/bindings/python/scripts/robot_wrapper.py#L27 if a function is included inside the namespace `pinocchio`.

jmirabel: As far as I understand, "buildGeomFromUrdf" is always part of pinocchio in Python, regardless the use of hpp-fcl. (even before this PR). So I guess this test is always false.

jmirabel:
It should be fine now.

jcarpent:
@nim65s What is the current status of the `hpp-fcl` release?

jmirabel:
The release will wait for https://github.com/humanoid-path-planner/hpp-fcl/pull/56.

jmirabel:
There is a snake biting its own tail here:
- on travis, Pinocchio uses hpp-fcl release of robotpkg so a new release of hpp-fcl is required to make it work,
- to release hpp-fcl, we need to update all packages downstream, including Pinocchio.

I see three workarounds:
1. pushing to a branch like `hpp-devel` in Pinocchio, validate the whole HPP framework, release hpp-fcl, merge this and release the rest of HPP.
2. make Pinocchio devel branch use hpp-fcl devel branch.
3. merge this and all HPP waiting PRs and releasing them.

We need to decide on one of the above solutions to move on.

jmirabel:
My favorite solution is 3 for its convenience. I suppose people using the devel branch of Pinocchio can also use hpp-fcl from source.

jcarpent:
I think we should release first `hpp-fcl` and then update all the other packages. This action is not risky here. 
Another solution is to introduce hpp-fcl-devel into robotpkg inside the wip branch. Easy to do and much more flexible than the three other solutions!

jmirabel:
> Another solution is to introduce hpp-fcl-devel into robotpkg inside the wip branch. Easy to do and much more flexible than the three other solutions!

Can you do it ?

jcarpent:
I think for this tack @nim65s will be faster and much more efficient than me.

jmirabel:
I will let @nim65s <https://github.com/nim65s> reply but I think his
stack of packages to release is already rather big.

nim65s:
No, sorry, but I will not duplicate a package in robotpkg just have it available in multiple versions, and even more maintain multiple times the same package.

Yes, it would be really fast and easy to do so, but I am pretty sure that it will bring me nightmares in the future, and that at the end we will all loose time by trying to use shortcuts.

nim65s:
I see 2 solutions:
1. we update all robotpkg releases at once (@jmirabel 's solution n.3)
2. we release hpp-fcl, and I add patches to all packages dependending on hpp-fcl to be sure their old versions still work with the new hpp-fcl, and then we can proceed to release every other packages. eg.: https://github.com/nim65s/robotpkg/commit/554d0f326d73b0e8c11989d5e345c1fe5c836a95

I find the solution 2. a bit more elegant and less error-prone and stressful, but it require extra work. This work might be worth it or not, I can't tell yet.

jcarpent:
I totally understand your reasons to not duplicate the package. But, we are very limited by the fact that all the packages are inter-dependent, and any break in the compilation procedure will affect all the users and all the robots together. I continue to think that we should adopt a fast release solution like the one that I've just suggested.

I would also prefer also the second version, which seems to be more elegant indeed.

nim65s:
I totally agree with the "Release early, release often" strategy :)

jmirabel:
I don't call safe a method that pushes to robotpkg changes, every now and then, so that it is synchronized with the devel branch, i.e. that deploys a development version.

I don't call elegant a method which mixes deployment and development. Deployed versions must be synchronized together. Development versions must be synchronized together. To continue development on Pinocchio, you must deploy a development version of hpp-fcl and this is not elegant.

We have to deploy hpp-fcl and Pinocchio at the same time. It doesn't make sense to make Pinocchio devel branch wait for hpp-fcl devel branch to be deployed because then the Pinocchio deployed version will be broken.

nim65s:
We already discussed that. I think that development versions should be based on the released versions of their dependencies. This gives the safety you are looking for, and making the releases is then easier and requires less work.
I don't see why you would call a release of hpp-fcl a "development version".

But as you want, let's release hpp-fcl, pinocchio, hpp, the sot, and rbprm at once. It will be long and painful, but in the end it should provide a polished release of everything.

nim65s:
Looking at this once again, releasing everything at once will break CI/CD everywhere. Is everybody OK with this ?
If so, what are we waiting for ?
Or are you still wanting me to provide a parallel CI/CD using devel branches of some dependencies ?

jcarpent:
You can start with the head of the dependency at least.

stonneau:
I guess we are waiting for us, since rbprm has yet to be updated to pinocchiov2 before integrating these changes (ongoing but not so smooth)

jmirabel:
I let you choose the method you prefer.

olivier-stasse:
I am not ok to break everything everywhere I have at least three remote collaborations which are based on the consistency of robotpkg.

We have tools to check the consistency of candidate releases inside robotpkg.

robotpkg cannot handle all the combinaison of possible releases.
For CI on devel, a solution could be to use a custom docker.
It could be created by using robotpkg, make checkout and a list of branches validated by the responsible of the repository.

nim65s:
Ok, therefore, I propose to start by merging PR in hpp-fcl and make a release of this package, but not to push it on robotpkg. 
This way, no project from the team will suffer from it, and @jcarpent  will be able to manually check that we can do the same with pinocchio, using the .tar.gz of hpp-fcl, as the CI won't work.
Once this will be done, @jmirabel will be able to finish the release of HPP while I finish the release the SoT in the same way (@jmirabel will probably prefer to use the tag instead of the tar.gz, but it doesn't change anythnig)
This lets time to @stonneau to finish with RBPRM (maybe it would have been better not to merge https://github.com/humanoid-path-planner/hpp-doc/pull/37 )

Then, when every release is done, I will put everything at once on robotpkg, and CI will work again.
I let you choose individually if you want to put the release tag on the master branch or not before the CI works again.

nim65s:
Also, @wxmerkt will be happy to have a new github release of hpp-fcl soon :)

nim65s:
Oh, I forgot to sync also with @andreadelprete about TSID…

jmirabel:
> I propose to start by merging PR in hpp-fcl

I agree.

> and make a release of this package

this could surely wait for validation, but if you want to do it, go ahead. 

> but not to push it on robotpkg.

I agree.

jcarpent:
@nim65s @olivier-stasse All the difficulties to have some stable releases during the validation of the next releases let me think that we really need to put in the WIP of robotpkg the clone of all the existing packages with the prefix -dev or -wip in order to allow smooth maintenance and the release process.

nim65s:
hpp-fcl v1.0.0 is out on github.
@jcarpent : can you merge this, update https://github.com/stack-of-tasks/pinocchio/blob/master/CMakeLists.txt#L76 and create a new github release ?

nim65s:
The idea of having simultaneously multiple versions for the same package in the same repo, and that any user or any automated test could be using one or the other at any time terrifies me, this is why I do not want to do this myself, nor I want to maintain it, nor I want to help anybody who use it.
Sorry, but this is really too much work for me.

jcarpent:
@nim65s We cannot simply do what you suggest.
The current version of Pinocchio is not compatible with `hpp-fcl` because of the namespace change.

nim65s:
That's right. I will have a look at it.

jcarpent:
@jmirabel @nim65s I have created another branch on the main repo to allow to use it as a sandbox, i.e. we don't care if travis is happy or not.

jmirabel:
Thank for the merge.

olivier-stasse:
Dear Justin,
Maybe the problem comes from the fact that all CI is done against robotpkg binaries.

Looking at the devel branch the CMakeLists.txt specifies that the current code is consistent with hpp-fck 0.5.1 where robotpkg provides 0.7.0.

If you provides a set of branches of your gepetto dependencies on which we can build the devel branch it is possible to build a docker file from the source of robotpkg (and not the binaries).

A more coherent way will be to release indenpendtly from robotpkg-binaries each package.

Wrt to wip, it already exists in robotpkg. The recurrent problem is that you cannot release package in robotpkg for binaries if they are not consistent together.

But we use robotpkg *from source* to make deployment test with specific release candidate.
This could be use to build the appropriate dockerfile.

We could try provide script to build the docker file, but that would be possible only if the CMakeLists.txt is consistent.

jcarpent:
> A more coherent way will be to release indenpendtly from robotpkg-binaries each package.

I totally agree. Robotpkg is now becoming the limit of our approach and the docker approach seems to be a nice solution.

nim65s:
Agreed too.

olivier-stasse:
Just one precision, right now, if we provide the script to build the dockerfile for a specific branch, it will be the duty of the developer to:

-  use the provided script to build the dockerfile and 
- put it for the specific branch.

jcarpent:
@olivier-stasse Yes, I totally agree on these two points.

jmirabel:
For the update, I will merge devel into fcl-release, add a few minor changes and validate HPP (including hpp-fcl next release) in the next few days. I will then open a PR of fcl-release into devel.
Any objections ?

jcarpent:
It sounds great to me. You can proceed as you've suggested.

PR #649, opened 2019-01-24
[doc] Inverse kinematics worked example

gabrielebndn:
I have slightly modified the example and code (changed termination condition) and added comments.

### DIFF
--- a/doc/b-examples/i-inverse-kinematics.md
+++ b/doc/b-examples/i-inverse-kinematics.md
@@ -1,7 +1,84 @@
 # Inverse kinematics (clik)
 
+This example shows how to position the end effector of a manipulator robot to a given position.
+The example employs a simple Jacobian-based iterative algorithm, which is called closed-loop inverse kinematics (CLIK).
+
 ## Python
-\include i-inverse-kinematics.py
+\includelineno i-inverse-kinematics.py
+
+### Explanation of the code
+First of all, we import the necessary libraries and we create the `Model` and `Data` objects:
+\until data
+
+The end effector corresponds to the 6th joint
+\skipline JOINT
+
+and its desired position is given as
+\skipline xdes
+
+Next, we define an initial configuration
+\skipline q
+
+This is the starting point of the algorithm. *A priori*, any valid configuration would work.
+We decided to use the value returned by `neutralConfiguration`.
+For a simple manipulator such as the one in this case, it simply corresponds to an all-zero vector,
+but using this method generalizes well to more complex kinds of robots, ensuring validity.
+
+Next, we set some computation-related values
+\until DT
+corresponding to the desired position precision (a tenth of a millimeter will do), a maximum number of iterations (to avoid infinite looping in case the position is not reachable) and a positive "time step" defining the convergence rate.
+
+Then, we begin the iterative process.
+At each iteration, we begin by computing the end-effector pose for the current configuration value:
+\skip forwardKinematics
+\until R
+
+Next, we compute the error between the desired position and the current one. Notice we chose to express it in the local joint frame:
+\skipline err
+
+If the error norm is below the previously-defined threshold, we have found the solution and we break out of the loop
+\until break
+
+Otherwise, we search for another configuration trying to reduce the error.
+
+We start by computing the Jacobian, also in the local joint frame. Since we are only interested in the position, and not in the orientation, we select the first three lines, corresponding to the translation part
+\skipline J
+
+Next, we can compute the evolution of the configuration by taking the pseudo-inverse of the Jacobian:
+\skipline v
+
+Finally, we can add the obtained tangent vector to the current configuration
+\skipline q
+
+where `integrate` in our case amounts to a simple sum.
+
+If the loop terminates without ever breaking, it means a solution has not been found.
+We just warn the user
+\skip else
+\until print
+
+Finally, we display the result:
+\skip print
+\until final
 
 ## C++
-\include i-inverse-kinematics.cpp
+The equivalent C++ implemetation is given below
+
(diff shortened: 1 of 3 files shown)


jcarpent:
@gabrielebndn Very nice. I will wait for Travis and then merge. Thanks for the contribution.

doc/b-examples/i-inverse-kinematics.md
```diff
@@ -1,7 +1,84 @@
...
+The end effector corresponds to the 6th joint
+\skipline JOINT
+
+and its desired position is given as
+\skipline xdes
+
+Next, we define an initial configuration
+\skipline q
+
+This is the starting point of the algorithm. *A priori*, any valid configuration would do.
```
jcarpent: would work.

doc/b-examples/i-inverse-kinematics.md
```diff
@@ -1,7 +1,84 @@
...
+
+Next, we compute the error between the desired position and the current one. Notice we chose to express it in the local joint frame:
+\skipline err
+
+If the error norm is below the previously-defined threshold, we have found the solution and we break out of the loop
+\until break
+
+Otherwise, we search for another configuration trying to reduce the error.
+
+We start by computing the Jacobian, also in the local joint frame. Since we are only interested in the position, and not in the orientation, we select the first three lines
```
jcarpent: we select only the first three lines corresponding to the translation part

doc/b-examples/i-inverse-kinematics.md
```diff
@@ -1,7 +1,84 @@
...
+\skipline J
+
+Next, we can compute the evolution of the configuration by taking the pseudo-inverse of the Jacobian:
+\skipline v
+
+Finally, we can add the obtained pseudo-velocity to the current configuration
```
jcarpent: I would remove the term pseudo-velocity and would replace it with a tangent vector of the configuration space.

doc/b-examples/i-inverse-kinematics.md
```diff
@@ -1,7 +1,84 @@
...
+\skipline q
+
+where `integrate` in our case amounts to a simple sum.
+
+If the loop terminates without ever breaking, it means the a solution has not been found.
```
jcarpent: you may choose between the and a

doc/b-examples/i-inverse-kinematics.md
```diff
@@ -1,7 +1,84 @@
...
+\until final
 
 ## C++
-\include i-inverse-kinematics.cpp
+The equivalent C++ implemetation is given below
+
+\includelineno i-inverse-kinematics.cpp
+
+### Explanation of the code
+The code follows exactly the same steps as Python.
```
jcarpent: as in Python.

doc/b-examples/i-inverse-kinematics.md
```diff
@@ -1,7 +1,84 @@
...
+We can identify two major differences. The first one concerns the Jacobian computation.
+In C++, you need to pre-allocate its space, set it to zero, and pass it as an input
```
jcarpent: space should be replaced by memory

doc/b-examples/i-inverse-kinematics.md
gabrielebndn: Ok

doc/b-examples/i-inverse-kinematics.md
gabrielebndn: Ok

doc/b-examples/i-inverse-kinematics.md
gabrielebndn: All right

doc/b-examples/i-inverse-kinematics.md
gabrielebndn: Oh! Good catch

doc/b-examples/i-inverse-kinematics.md
gabrielebndn: Ok

doc/b-examples/i-inverse-kinematics.md
gabrielebndn: All right

gabrielebndn: I wrote "memory space"

gabrielebndn:
Force-pushed the changes

jcarpent:
Sounds great now. @gabrielebndn Thanks for your fast modifications.

examples/i-inverse-kinematics.py
```diff
@@ -1,27 +1,31 @@
...
+else:
+    print "\nWarning: the iterative algorithm has not reached convergence to the desired precision"
 
-# Computing error for final q
-pinocchio.forwardKinematics(model,data,q)
-x   = data.oMi[JOINT_ID].translation
-err = x-xdes
-print '\nfinal error: ', err.T
+print '\nresult: ', q.flatten().tolist()
+print '\nfinal error: ', (x-xdes).T
```
nim65s: please use the print function instead of the statement, is this statement is not valid in Python 3.

To do so: add `from __future__ import print_function` before anything else in the file, and then use it with for example with `print('\nfinal error: %s' % (x-xdes).T)`

gabrielebndn: I also prefer the function syntax. In fact, I had considered using it, but then I did not do it because I did not want to add things which are not strictly necessary to the example. But I can change it, in fact it is much nicer. We just have to agree we will consistently use it throughout all of our examples.

One thing I have never tried/understood: what happens if you execute a file containing `from __future__ import print_function` in Python 3? I guess this statement is simply ignored?

PR #652, opened 2019-01-28
Load neutralconfiguration from srdf for joint.nq()>1

proyan:
Currently when loading neutralConfiguration from srdf, root_joint values are not loaded.
This PR should change that.

### DIFF
--- a/src/parsers/srdf.hxx
+++ b/src/parsers/srdf.hxx
@@ -256,8 +256,8 @@ namespace pinocchio
                   typename Model::ConfigVectorType joint_config(joint.nq());
                   const std::string joint_val = joint_tag.second.get<std::string>("<xmlattr>.value");
                   std::istringstream config_string(joint_val);
-                  const std::vector<double> config_vec((std::istream_iterator<double>(config_string)), std::istream_iterator<double>());
-                  joint_config = ModelTpl<Scalar,Options,JointCollectionTpl>::ConfigVectorType::Map(config_vec.data(), config_vec.size());
+                  std::vector<double> config_vec((std::istream_iterator<double>(config_string)), std::istream_iterator<double>());
+                  joint_config = Eigen::Map<typename Model::ConfigVectorType>(config_vec.data(), config_vec.size());
                   model.neutralConfiguration.segment(joint.idx_q(),joint.nq()) = joint_config;
                   if (verbose)
                   {
--- a/src/parsers/srdf.hxx
+++ b/src/parsers/srdf.hxx
@@ -209,7 +209,7 @@ namespace pinocchio
     {
       typedef ModelTpl<Scalar,Options,JointCollectionTpl> Model;
       typedef typename Model::JointModel JointModel;
-      
+
       // Check extension
       const std::string extension = filename.substr(filename.find_last_of('.')+1);
       if (extension != "srdf")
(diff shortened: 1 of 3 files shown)


src/parsers/srdf.hxx
```diff
@@ -247,19 +247,25 @@ namespace pinocchio
...
-                  model.neutralConfiguration(joint.idx_q()) = joint_config; // joint with 1 dof
-                  // model.neutralConfiguration.segment(joint.idx_q(),joint.nq()) = joint_config; // joint with more than 1 dof
+                  typename Model::ConfigVectorType joint_config(joint.nq());
+                  if(joint.nq() == 1)  joint_config[0] = (Scalar)joint_tag.second.get<double>("<xmlattr>.value");
+                  else
+                  {
+                    const std::string joint_val = joint_tag.second.get<std::string>("<xmlattr>.value");
+                    std::istringstream config_string(joint_val);
+                    const std::vector<double> config_vec((std::istream_iterator<double>(config_string)), std::istream_iterator<double>());
+                    joint_config = ModelTpl<Scalar,Options,JointCollectionTpl>::ConfigVectorType::Map(config_vec.data(), config_vec.size());
```
jcarpent: Use Model here instead of ModelTpl<Scalar,Options,JointCollectionTpl>

src/parsers/srdf.hxx
```diff
@@ -247,19 +247,25 @@ namespace pinocchio
               if (joint_tag.first == "joint")
               {
                 std::string joint_name = joint_tag.second.get<std::string>("<xmlattr>.name");
-                const Scalar joint_config = (Scalar)joint_tag.second.get<double>("<xmlattr>.value");
                 typename Model::JointIndex joint_id = model.getJointId(joint_name);
-
+
```
jcarpent: Can you remove that spaces?

src/parsers/srdf.hxx
```diff
@@ -242,30 +242,37 @@ namespace pinocchio
...
+                  typename Model::ConfigVectorType joint_config(joint.nq());
+                  if(joint.nq() == 1)  joint_config[0] = (Scalar)joint_tag.second.get<double>("<xmlattr>.value");
+                  else
+                  {
+                    const std::string joint_val = joint_tag.second.get<std::string>("<xmlattr>.value");
+                    std::istringstream config_string(joint_val);
+                    const std::vector<double> config_vec((std::istream_iterator<double>(config_string)), std::istream_iterator<double>());
+                    joint_config = ModelTpl<Scalar,Options,JointCollectionTpl>::ConfigVectorType::Map(config_vec.data(), config_vec.size());
+                  }
+                  model.neutralConfiguration.segment(joint.idx_q(),joint.nq()) = joint_config; // joint with more than 1 dof
```
jcarpent: You can also remove the ending comment which is now outdated!

jcarpent:
@proyan Thanks for this contribution. I just have some minor remarks that you can handle rapidly.

src/parsers/srdf.hxx
proyan: I tried using typename Model::ConfigVectorType.
gcc cannot does not understand the Map attribute when I do that, but it finds the attribute when I use the full name.

src/parsers/srdf.hxx
proyan: Done with force push.

src/parsers/srdf.hxx
jcarpent: because you miss the template keyword when doing so. You can see with @jmirabel for that point if he is at LAAS today to explain you this additional info.

src/parsers/srdf.hxx
proyan: Done with force push.

jcarpent:
By the way, you can also simplify your code by just handling the generic case. We do not care about the single DOF context in fact.

proyan:
I've taken care of all comments by the recent commit.

PR #653, opened 2019-01-30
Update cmake submodule and add ability to use externally passed in Eigen

wxmerkt:
This PR contains two minimal changes that allow us to fully build on Ubuntu 18.04 inside a catkin workspace (and fully contained with dependency resolution and packaging for the ROS ecosystem):

1. Update of the CMake submodule to get the fixed check for Python versions (otherwise fails on stock 18.04)
2. Search of Eigen3 in quiet mode using CMake logic - the original logic and pkg-config steps are still preserved. The reason for this is to be able to pass in the path of `eigenpy` directly in the cmake configuration step which would stop the inclusion of the eigenpy dependency eigen3 which is not otherwise resolved.


bindings/python/CMakeLists.txt
```diff
@@ -36,6 +36,10 @@ SET(PYWRAP ${PROJECT_NAME}_pywrap)
 SET(PYWRAP ${PYWRAP} PARENT_SCOPE)
 
 # --- DEPENDENCIES --- #
+find_package(Eigen3 QUIET)
```
jcarpent: Eigen3 is already included by the main CMakeLists.txt.
These lines are then not necessary.

jcarpent: >     2\. Search of Eigen3 in quiet mode using CMake logic - the original logic and pkg-config steps are still preserved. The reason for this is to be able to pass in the path of `eigenpy` directly in the cmake configuration step which would stop the inclusion of the eigenpy dependency eigen3 which is not otherwise resolved.

@wxmerkt I certainly do not understand well why you have to perform this step. Can you rephrase it?

jcarpent: Thanks for this clarification.
I just have a remaining question, if `PKG_CONFIG_PATH`variable is not available in Catkin, how do you manage the dependencies?

wxmerkt: We specify them in the `package.xml` and thus they will get installed automatically during the `rosdep install` (urdfdom, boost-python). For eigenpy, we pass it in directly. The rest is available automatically. We run Travis on Kinetic and Melodic and it passes well.

jcarpent:
The inclusion of Pinocchio inside ROS is very appreciated, and I want to thank you for this important contribution. For sure, it will help people to use Pinocchio for their own research as the installation will be simplified.
@wxmerkt Again, I'm greatly thankful for your contribution.

jcarpent:
I will merge after Travis checking.

wxmerkt:
Thank you very much, Justin! I will then begin to make the release requests for eigenpy (v1.5.0) and pinocchio (v2.0.0/latest develop from today) to Kinetic and Melodic if that is okay with you.

jcarpent:
Yes, sure. Your contribution is welcome!

PR #655, opened 2019-01-31
Support for catkinized hpp-fcl

wxmerkt:
This PR adds support for both the existing/robotpkg hpp-fcl as well as the [catkinized hpp-fcl](https://github.com/ipab-slmc/hpp-fcl_catkin). The default behaviour does not change.


jcarpent:
@wxmerkt Thanks for this contribution

PR #661, opened 2019-02-05
Correct macro naming

gabrielebndn:
Some Pinocchio internal macros did not adhere to the naming standard `PINOCCHIO_"MACRO_NAME"` (see #570). I corrected the ones I was able to spot. In particular:

- A few Eigen-related macros, contained in [src/eigen-macros.hpp](https://github.com/stack-of-tasks/pinocchio/blob/devel/src/eigen-macros.hpp), were beginning with `EIGEN_`. This is a bad practice. I added the `PINOCCHIO_` prefix to them. I am talking about
  - EIGEN_PLAIN_TYPE
  - EIGEN_PLAIN_COLUMN_MAJOR_TYPE
  - EIGEN_PLAIN_ROW_MAJOR_TYPE
  - EIGEN_REF_CONSTTYPE (renamed PINOCCHIO_EIGEN_REF_CONST_TYPE)
  - EIGEN_REF_TYPE
  - EIGEN_DOT_PRODUCT_RETURN_TYPE
  - EIGEN_CONST_CAST

- A few macros were prefixed with `SE3_`. I changed this prefix to `PINOCCHIO_` for consistency, except for two or three of them which I determined to be related to the SE3 class, where I added `PINOCCHIO_` instead of using it as a replacement.

- I removed file [src/assert.hpp](https://github.com/stack-of-tasks/pinocchio/blob/devel/src/assert.hpp). This file contained a single macro, called `SE3_STATIC_ASSERT`, which was never employed inside Pinocchio, and which has been superseded by `PINOCCHIO_STATIC_ASSERT` in [src/macros.hpp](https://github.com/stack-of-tasks/pinocchio/blob/devel/src/macros.hpp). The file *was* included by a few other headers, but said macro did not appear inside them, so I just removed the `#include` statement.


jmirabel:
Nice job !

jcarpent:
Thanks for your contribution!

PR #662, opened 2019-02-07
Introduce support for MeshCat viewer in RobotWrapper.

matthieuvigne:
RobotWrapper now supports transparently the use of MeshCat as a 3D viewer, alongside of gepetto-gui.

Added example script to show how to use the MeshCat viewer.
Fix the gepetto-gui example script.

Linked to https://github.com/stack-of-tasks/pinocchio/issues/657

### DIFF
--- a/bindings/python/pinocchio/robot_wrapper.py
+++ b/bindings/python/pinocchio/robot_wrapper.py
@@ -287,7 +287,7 @@ class RobotWrapper(object):
             visualizer: the meshcat.Visualizer instance to use.
             robot_name: name to give to the robot in the viewer
             robot_color: optional, color to give to the robot. This overwrites the color present in the urdf.
-                         Format is a list of three RGB floats (between 0 and 1)
+                         Format is a list of four RGBA floats (between 0 and 1)
         """
         import meshcat.geometry
         # Set viewer to use to gepetto-gui.
@@ -302,11 +302,11 @@ class RobotWrapper(object):
                 raise IOError("Visual mesh file not found for link {}.".format(visual.name))
             # Get file type from filename extension.
             _, file_extension = os.path.splitext(visual.meshPath)
-            if file_extension == ".dae" or file_extension == ".DAE":
+            if file_extension.lower() == ".dae":
                 obj = meshcat.geometry.DaeMeshGeometry.from_file(visual.meshPath)
-            elif file_extension == ".obj" or file_extension == ".OBJ":
+            elif file_extension.lower() == ".obj":
                 obj = meshcat.geometry.ObjMeshGeometry.from_file(visual.meshPath)
-            elif file_extension == ".stl" or file_extension == ".STL":
+            elif file_extension.lower() == ".stl":
                 obj = meshcat.geometry.StlMeshGeometry.from_file(visual.meshPath)
             else:
                 raise ImportError("Unknown mesh file format: {}.".format(visual.meshPath))
@@ -317,6 +317,10 @@ class RobotWrapper(object):
             else:
                 meshColor = robot_color
             material.color = int(meshColor[0] * 255) * 256**2 + int(meshColor[1] * 255) * 256 + int(meshColor[2] * 255)
+            # Add transparency, if needed.
+            if float(meshColor[3]) != 1.0:
+                material.transparent = True
+                material.opacity = float(meshColor[3])
             self.meshcat_viewer[viewer_name].set_object(obj, material)
 
     # Create the scene displaying the robot meshes in gepetto-viewer
--- a/bindings/python/pinocchio/robot_wrapper.py
+++ b/bindings/python/pinocchio/robot_wrapper.py
@@ -47,7 +47,7 @@ class RobotWrapper(object):
 
 
     def __init__(self, model = pin.Model(), collision_model = None, visual_model = None, verbose=False):
-        
+
         self.model = model
         self.data = self.model.createData()
 
@@ -72,6 +72,9 @@ class RobotWrapper(object):
 
         self.v0 = utils.zero(self.nv)
         self.q0 = pin.neutral(self.model)
+
+        # Two viewers are supported: meshcat or gepetto-gui. By default, gepetto-gui is to be used.
+        self.used_viewer = 'gepetto-gui'
 
     @property
     def nq(self):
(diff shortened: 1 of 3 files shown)


matthieuvigne:
This PR is not entirely complete, as currently it lacks DAE/STL support. For this reason, the example script I provided cannot work on the current romeo model - to test it, you need to place the following folder in the ```models``` subfolder. I think it's best not to introduce this folder in pinocchio - once we have DAE support, we should be able to simply change the path to load the romeo model.

[katana.zip](https://github.com/stack-of-tasks/pinocchio/files/2842925/katana.zip)

Also, so far there is no documentation to indicate
 - that MeshCat is supported
 - that, to enable MeshCat support, one needs to install MeshCat.

I feel it's the same for gepetto-gui : in RobotWrapper there is a ````import gepetto.corbaserver``` but it's nowhere listed as a dependency. Am I mistaken ? Do you see a good place to put such a documentation (it would make sense to talk about both viewers in the same document somewhere).

The current code is quite minimal, if there are extra features you would like to see (appart from STL/DAE support) feel free to ask :)

jcarpent:
@matthieuvigne Thanks for this quick job.

olivier-stasse:
I strongly support the remark of @nmansard regarding #663
We should create a new package and remove any reference to a GUI inside pinocchio.

olivier-stasse:
Dear @matthieuvigne thank you very much for this very nice effort. But in order to avoid any unnecessary dependency to pinocchio (we are using it also to do control) are you ok that we propose a new package with your PR with all the necessary infra-structure. I am also fine for a ROS package to avoid robotpkg if you do not like it.

aelkhour:
@olivier-stasse the PR is related to a python viewer client, whose dependency can be made optional and does not affect the C++ library. I'm personally against putting this in a separate package:
- It's less user-friendly for installation (not everyone uses robotpkg build tool or provided binaries)
- It's a hassle to maintain by experience.

Regarding #663, discussion is about splitting the wrapper into multiple classes, but not necessarily multiple packages (which IMHO is a bad idea, I would like to avoid reliving any PhD nightmares related to hpp ot sot dependency trees ;-) ).

olivier-stasse:
@aelkhour They are several way to simplify the compilation procedure of multiple packages. You can use Makefile such it is with hpp or rosinstall for catkin package. Putting everything in one single package is for me a way to brittle even more the source code of the package. Every generation of PhD comes with his own prefered GUI (and most of the time with very good reason). Why should we put this into pinocchio. This package is supposed to be about computation of mechanics. Just by adding geometry we now have an optional dependency to osg-dae. Following your policy, you can endup with project such as Choronoid (which is a very nice project btw) where even for specialist it is very difficult to compile.

From a company perspective if you want to use pinocchio in your robot, I doubt that you want to have meshcat inside the system. My experience on three robots in that on each of them we had to make specific adaptation for the code. Lowering the dependency on each package is one way to make the overall process easier, and simplify the CI.

olivier-stasse:
@aelkhour After discussing with Nicolas, I realized that the point is to have some code that will not be taken into account in the packaging. In that case, I am fine with the approach. But it should be clear that this code *will not* be provided in the binary packages.

aelkhour:
@olivier-stasse Perfect, fine by me!

matthieuvigne:
Currently pinnochio already integrates ```gepetto-gui``` as an unpackaged viewer dependency. I think we need a consistent handling of both ```gepetto-gui``` and ```meshcat```, and possibly other viewers, as discussed in https://github.com/stack-of-tasks/pinocchio/issues/663 - the idea being that pinocchio provides the interface to several viewers, but not the viewers themselves. This can easily take the form of a python package, which we can bundle inside pinocchio or make a new project (something like ```pinocchio-viewer-interface```). Personally I'm also in favor of keeping it in pinocchio, so as to limit the number of small projects.

matthieuvigne:
@wxmerkt Do you think you could handle adding DAE/STL support inside MeshCat? I think it makes more sense to add that before merging this PR in pinocchio. See https://github.com/rdeits/MeshCat.jl/issues/91

wxmerkt:
@matthieuvigne I'll have a look again later today/this week and ping once done.

wxmerkt:
@matthieuvigne I have now added DAE and STL support to MeshCat, pending PR feedback and merge:

- https://github.com/rdeits/meshcat/pull/48
- https://github.com/rdeits/meshcat-python/pull/36

matthieuvigne:
Great work @wxmerkt !

I updated the wrapper code to match https://github.com/rdeits/meshcat-python/pull/36. I tested STL support on katana, and that worked great. When I tried to load the DAE files from Romeo, they appeared with a huge scale factor in Meshcat. However, the problem seems to lie in the files themselves: I loaded the DAE files in Meshlab, and there as well I had a 100x scale factor: looks like a unit error.

I'm not familiar with the DAE file format, but since both parsers (three.js and Meshlab) find the same scale factor, I tend to believe that the files currently inside pinocchio are wrong. Have they been tested recently?

Btw, I'm not able to load them into gepetto-gui, I don't really know why. After instally OpenSceneGraph-Collada-3.4.1-6, I get

```
Load failed in COLLADA DOM
Load failed in COLLADA DOM conversion
Warning: Error in reading to "/home/matthieu/dev/src/pinocchio/models/romeo/romeo_description/meshes/V1/visual/PhalaxEnd.dae".
```

matthieuvigne:
Indeed, I just tested with other DAE files (conversion from katana stl meshes) and they display fine. Something is definitely wrong, or at least unconventional, with the Romeo meshes.

bindings/python/scripts/robot_wrapper.py
```diff
@@ -262,26 +267,63 @@ def initDisplay(self, windowName="python-pinocchio", sceneName="world", loadMode
...
+        self.viewerRootNodeName = robot_name
+
+        # Load robot meshes in MeshCat
+        for visual in self.visual_model.geometryObjects:
+            viewer_name = self.viewerRootNodeName + visual.name
+            if visual.meshPath == "":
+                raise IOError("Visual mesh file not found for link {}.".format(visual.name))
+            # Get file type from filename extension.
+            _, file_extension = os.path.splitext(visual.meshPath)
+            if file_extension == ".dae" or file_extension == ".DAE":
```
wxmerkt: Nit: You could use `file_extension.lower() == ".ext"` here

matthieuvigne:
As of today Meshcat support DAE/STL files (thanks to @wxmerkt) The code is this PR is thus fully functional, I think it can be merged as is. If needed we can discuss python viewer packaging in another issue and make a dedicated PR.

jcarpent:
Sounds good to me too. I will merge it now!
Thanks @matthieuvigne for this very nice contribution.

PR #664, opened 2019-02-11
Algorithm to compute the derivatives of the COM velocity

nmansard:
I put the algo in a new center-of-mass-derivative.hpp/xx file, with new unittest file, following the policy used for other *-derivatives.hpp/xx files.
I put the binding in expose-kinematic-derivatives because creating a new file at this point seemed overkilled, but I can modify that choice if need be.
The algorithm is computed in the local frame, which makes it compact and elegant. I validated it for the simple humanoid in the unittest, but also for talos arm and legs in python.

### DIFF
--- a/src/algorithm/center-of-mass-derivatives.hxx
+++ b/src/algorithm/center-of-mass-derivatives.hxx
@@ -1,0 +1,82 @@
+//
+// Copyright (c) 2019 CNRS
+//
+
+#ifndef __pinocchio_center_of_mass_derivatives_hxx__
+#define __pinocchio_center_of_mass_derivatives_hxx__
+
+#include "pinocchio/multibody/visitor.hpp"
+#include "pinocchio/algorithm/check.hpp"
+
+namespace pinocchio
+{
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl,
+           typename Matrix3xOut>
+  struct CoMVelocityDerivativesForwardStep
+    : public fusion::JointVisitorBase< CoMVelocityDerivativesForwardStep<Scalar,Options,JointCollectionTpl,Matrix3xOut> >
+  {
+    /* Assumes that both computeForwardKinematicsDerivatives and centerOfMass (or
+     * equivalent methods like computeAllTerms) have been computed beforehand. */
+
+    typedef ModelTpl<Scalar,Options,JointCollectionTpl> Model;
+    typedef DataTpl<Scalar,Options,JointCollectionTpl> Data;
+    
+    typedef boost::fusion::vector<const Model &,
+                                  Data &,
+                                  Matrix3xOut &
+                                  > ArgsType;
+    
+    template<typename JointModel>
+    static void algo(const JointModelBase<JointModel> & jmodel,
+                     JointDataBase<typename JointModel::JointDataDerived> & jdata,
+                     const Model & model,
+                     Data & data,
+                     const Eigen::MatrixBase<Matrix3xOut> & vcom_partial_dq_CONST)
+    {
+      typedef typename Model::JointIndex JointIndex;
+      typedef typename Data::Motion Motion;
+      
+      const JointIndex & i = jmodel.id();
+
+      Matrix3xOut & dvcom_dq = PINOCCHIO_EIGEN_CONST_CAST(Matrix3xOut,vcom_partial_dq_CONST);
+      typename SizeDepType<JointModel::NV>::template ColsReturn<typename Data::Matrix3x>::Type
+        dvcom_dqi = jmodel.jointCols(dvcom_dq);
+
+      Motion vpc = (model.parents[i]>0) ? (data.v[i]-(Motion)jdata.v()) : Motion::Zero();
+      vpc.linear() -= data.vcom[i]; // vpc = v_{parent+c} = [ v_parent+vc; w_parent ]
+      const Eigen::Matrix<Scalar,6,JointModel::NV,Options> & vxS = vpc.cross(jdata.S());
+      
+      dvcom_dqi = data.mass[i]*data.oMi[i].rotation()
+        *( vxS.template topRows<3>() - cross( data.com[i], vxS.template bottomRows<3>()) );
+    }
+  };
+  
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl,
+           typename Matrix3xOut>
+  inline void getCenterOfMassVelocityDerivatives(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                                                 DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                                                 const Eigen::MatrixBase<Matrix3xOut> & vcom_partial_dq_CONST)
+  {
+    EIGEN_STATIC_ASSERT_SAME_MATRIX_SIZE(Matrix3xOut,Data::Matrix3x);
+    
+    assert(vcom_partial_dq.cols() ==  model.nv);
+    assert(model.check(data) && "data is not consistent with model.");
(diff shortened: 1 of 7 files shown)


src/algorithm/center-of-mass-derivatives.hxx
```diff
@@ -0,0 +1,82 @@
...
+                                  Data &,
+                                  Matrix3xOut &
+                                  > ArgsType;
+    
+    template<typename JointModel>
+    static void algo(const JointModelBase<JointModel> & jmodel,
+                     JointDataBase<typename JointModel::JointDataDerived> & jdata,
+                     const Model & model,
+                     Data & data,
+                     const Eigen::MatrixBase<Matrix3xOut> & vcom_partial_dq_CONST)
```
jcarpent: `vcom_partial_dq_CONST` is not really elegant. I would keep `vcom_partial_dq`

src/algorithm/center-of-mass-derivatives.hxx
```diff
@@ -0,0 +1,82 @@
...
+      typedef typename Model::JointIndex JointIndex;
+      typedef typename Data::Motion Motion;
+      
+      const JointIndex & i = jmodel.id();
+
+      Matrix3xOut & dvcom_dq = PINOCCHIO_EIGEN_CONST_CAST(Matrix3xOut,vcom_partial_dq_CONST);
+      typename SizeDepType<JointModel::NV>::template ColsReturn<typename Data::Matrix3x>::Type
+        dvcom_dqi = jmodel.jointCols(dvcom_dq);
+
+      Motion vpc = (model.parents[i]>0) ? (data.v[i]-(Motion)jdata.v()) : Motion::Zero();
```
jcarpent: You can put this result in data.v[i] if you want.

src/algorithm/center-of-mass-derivatives.hxx
```diff
@@ -0,0 +1,82 @@
...
+      vpc.linear() -= data.vcom[i]; // vpc = v_{parent+c} = [ v_parent+vc; w_parent ]
+      const Eigen::Matrix<Scalar,6,JointModel::NV,Options> & vxS = vpc.cross(jdata.S());
+      
+      dvcom_dqi = data.mass[i]*data.oMi[i].rotation()
+        *( vxS.template topRows<3>() - cross( data.com[i], vxS.template bottomRows<3>()) );
```
jcarpent: you can use here Motion::LINEAR and Motion::ANGULAR to be generic

unittest/center-of-mass-derivatives.cpp
```diff
@@ -0,0 +1,71 @@
...
+  Data data(model), data_ref(model);
+  
+  model.lowerPositionLimit.head<3>().fill(-1.);
+  model.upperPositionLimit.head<3>().fill(1.);
+  VectorXd q = randomConfiguration(model);
+  VectorXd vq(VectorXd::Random(model.nv));
+  VectorXd aq(VectorXd::Random(model.nv));
+
+  q.fill(1);
+  q.head<7>() << 0,0,0,1,0,0,0;
```
jcarpent: Why?

unittest/center-of-mass-derivatives.cpp
```diff
@@ -0,0 +1,71 @@
...
+  for(int k = 0; k < model.nv; ++k)
+  {
+    dq[k] = alpha;
+    centerOfMass(model,data,integrate(model,q,dq),vq);
+    dvcom_dqn.col(k) = (data.vcom[0]-vcom0)/alpha;
+    dq[k] = 0;
+  }
+
+  // Check that algo result and finite-diff approx are similar.
+  //std::cout << dvcom_dqn << std::endl<< std::endl<< std::endl;
```
jcarpent: You can remove these two lines if they are not more needed.

unittest/center-of-mass-derivatives.cpp
```diff
@@ -0,0 +1,71 @@
...
+    
+  // Compute dvcom_dq using the algorithm
+  Data::Matrix3x dvcom_dq = Data::Matrix3x::Zero(3,model.nv);
+  computeForwardKinematicsDerivatives(model,data,q,vq,aq);
+  centerOfMass(model,data,q,vq);
+  getCenterOfMassVelocityDerivatives(model,data,dvcom_dq);
+
+  // Approximate dvcom_dq by finite diff.
+  Eigen::Vector3d vcom0 = data.vcom[0];
+  const double alpha = 1e-6;
```
jcarpent: 1e-8 is preferable here. To follow the other tests.

jcarpent:
By the way, you can have access to these derivatives through the centroidal dynamics derivatives that already exist and bind.

src/algorithm/center-of-mass-derivatives.hxx
nmansard: I don't agree. In other algo, you you XXX for the const ref and XXX_ for the const-cast ref. I think this convention is better. 
Never mind. I changed it.

unittest/center-of-mass-derivatives.cpp
nmansard: no reasons. Thx.

src/algorithm/center-of-mass-derivatives.hxx
nmansard: thanks for the trick.

unittest/center-of-mass-derivatives.cpp
nmansard: It is 1e-6 in the test_kinematics_derivatives_velocity this test is built on ...
Changed

unittest/center-of-mass-derivatives.cpp
nmansard: Sure? These two lines are the first two you would write again in case this test does not pass.
Changed

nmansard:
About computing COM first: the current algo seems quite minimal using vcom. And it is reasonable to expect that the user would have computed the vcom quantity before computing its derivative.

About centroidal algo: Agreed, but I wanted COM only. BTW, the centroidal algo seems an order of magnitude more complex. I did not check timings yet.

nmansard:
Additional question: I have a unittest in Python. What should I do with it?

nmansard:
Pushed. Thanks for the review.

unittest/center-of-mass-derivatives.cpp
jcarpent: Then, this is a mistake from my side. I will change it as soon as possible.

gabrielebndn:
@nmansard, we have a place for Python unit tests in [unittest/python](https://github.com/stack-of-tasks/pinocchio/tree/devel/unittest/python).
These tests are working but far from complete, and tend to be very simple, their main purpose being checking that the binding mechanism itself is working properly.

If you want to add your unit test, I would put it there, it will certainly not do any harm. All unit tests in that folder are implemented by employing the `unittest` framework, so I would just make sure your own test is structured in the same way.

Also, all unit tests can be run together using the script [unittest/python/tests.py](https://github.com/stack-of-tasks/pinocchio/blob/devel/unittest/python/tests.py). I am not sure this is super useful (there are many ways to tell Linux to run all files in a folder), but if you want this feature, you can just add the proper import in that file

jcarpent:
@nmansard Seems good to me. Maybe, if you have time to add the Python test it would be nice.

jcarpent:
Anyway, @nmansard thank you for your contribution.

nmansard:
I added the python tests. You can merge.

BTW: I really don't like the unittest in Python :(

jcarpent:
Ok. Sounds good. I'm just waiting for Travis.

PR #665, opened 2019-02-12
get the inverse of MJtJ matrix using the matrix decompositions performed in forwardDynamics or impulsedynamics

proyan:
Currently the matrix decompositions done in forwardDynamics are wasted when computing the inverse of the MJtJ matrix. This PR adds a function that returns the analytical inverse of MJtJ (via schur complement) using the computations already done in forwardDynamics. This is also true for impulsedynamics.

I add the unittest and python bindings for this function as well.

### DIFF
--- a/bindings/python/algorithm/expose-dynamics.cpp
+++ b/bindings/python/algorithm/expose-dynamics.cpp
@@ -27,6 +27,16 @@ namespace pinocchio
 
     BOOST_PYTHON_FUNCTION_OVERLOADS(impulseDynamics_overloads, impulseDynamics_proxy, 5, 7)
 
+    static const Eigen::MatrixXd getKKTContactDynamicMatrixInverse_proxy(const Model & model,
+                                                                         Data & data,
+                                                                         const Eigen::MatrixXd & J)
+    {
+      Eigen::MatrixXd MJtJ_inv(model.nv+J.rows(), model.nv+J.rows());
+      getKKTContactDynamicMatrixInverse(model, data, J, MJtJ_inv);
+      return MJtJ_inv;
+    }
+    
+
     void exposeDynamics()
     {
       using namespace Eigen;
@@ -56,6 +66,10 @@ namespace pinocchio
                        "Update kinematics (if true, it updates only the joint space inertia matrix)"),
               "Solve the impact dynamics problem with contacts, put the result in Data::dq_after and return it. The contact impulses are stored in data.impulse_c"
               )[bp::return_value_policy<bp::return_by_value>()]);
+      bp::def("getKKTContactDynamicMatrixInverse",getKKTContactDynamicMatrixInverse_proxy,
+              bp::args("Model","Data",
+                       "Contact Jacobian J(size nb_constraint * Model::nv)"),
+              "Computes the inverse of the constraint matrix [[M JT], [J 0]]. forward/impulseDynamics must be called first. The jacobian should be the same that was provided to forward/impulseDynamics.");
     }
     
   } // namespace python
--- a/src/algorithm/dynamics.hpp
+++ b/src/algorithm/dynamics.hpp
@@ -103,46 +103,47 @@ namespace pinocchio
     a.noalias() = J.transpose() * lambda_c;
     cholesky::solve(model, data, a);
     a += data.torque_residual;
+
     
     return a;
   }
 
   ///
-  /// \brief Computes the inverse of the constraint matrix [[M JT], [J 0]].
-  /// The matrix is defined when we call forwardDynamics. This method makes use of
-  /// the matrix decompositions performed during the forwardDynamics and returns the inverse.
-  /// The jacobian should be the same that was provided to forwardDynamics.
-  /// Thus you should call forward Dynamics first.
+  /// \brief Computes the inverse of the KKT matrix for dynamics with contact constraints, [[M JT], [J 0]].
+  /// The matrix is defined when we call forwardDynamics/impulsedynamics. This method makes use of
+  /// the matrix decompositions performed during the forwardDynamics/impulasedynamics and returns the inverse.
+  /// The jacobian should be the same that was provided to forwardDynamics/impulasedynamics.
+  /// Thus you should call forward Dynamics/impulsedynamics first.
   /// \param[in] model The model structure of the rigid body system.
   /// \param[in] data The data structure of the rigid body system.
   /// \param[out] MJtJ_inv inverse of the MJtJ matrix.
   ///
   template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl,
-           typename ConstraintMatrixType>
-  inline void getMJtJInverse(const ModelTpl<Scalar,Options,JointCollectionTpl> &
-                             model,
-                             const DataTpl<Scalar,Options,JointCollectionTpl> & data,
-                             const Eigen::MatrixBase<ConstraintMatrixType> & J,
-                             typename DataTpl<Scalar,Options,JointCollectionTpl>::MatrixXs& MJtJ_inv)
+           typename ConstraintMatrixType, typename KKTMatrixType>
+  inline void getKKTContactDynamicMatrixInverse(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                                                const DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                                                const Eigen::MatrixBase<ConstraintMatrixType> & J,
+                                                const Eigen::MatrixBase<KKTMatrixType> & MJtJ_inv)
   {
     typedef DataTpl<Scalar,Options,JointCollectionTpl> Data;
     assert(MJtJ_inv.cols() == data.JMinvJt.cols() + model.nv);
     assert(MJtJ_inv.rows() == data.JMinvJt.rows() + model.nv);
-    typename Data::MatrixXs::Index nc = data.JMinvJt.cols();
+    const typename Data::MatrixXs::Index& nc = data.JMinvJt.cols();
 
-    Eigen::Block<typename Data::MatrixXs> topLeft = MJtJ_inv.topLeftCorner(model.nv, model.nv);
-    Eigen::Block<typename Data::MatrixXs> topRight = MJtJ_inv.topRightCorner(model.nv, nc);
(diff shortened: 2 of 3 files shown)


bindings/python/algorithm/expose-dynamics.cpp
```diff
@@ -27,6 +27,16 @@ namespace pinocchio
 
     BOOST_PYTHON_FUNCTION_OVERLOADS(impulseDynamics_overloads, impulseDynamics_proxy, 5, 7)
 
+    static const Eigen::MatrixXd getMJtJInverse_proxy(const Model & model,
+                                                Data & data,
```
jcarpent: Can you align here?

src/algorithm/dynamics.hpp
```diff
@@ -106,7 +106,45 @@ namespace pinocchio
...
+  /// the matrix decompositions performed during the forwardDynamics and returns the inverse.
+  /// The jacobian should be the same that was provided to forwardDynamics.
+  /// Thus you should call forward Dynamics first.
+  /// \param[in] model The model structure of the rigid body system.
+  /// \param[in] data The data structure of the rigid body system.
+  /// \param[out] MJtJ_inv inverse of the MJtJ matrix.
+  ///
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl,
+           typename ConstraintMatrixType>
+  inline void getMJtJInverse(const ModelTpl<Scalar,Options,JointCollectionTpl> &
```
jcarpent: I think this is the wrong naming. It is JtMJ and not MJtJ

src/algorithm/dynamics.hpp
```diff
@@ -106,7 +106,45 @@ namespace pinocchio
     
     return a;
   }
-  
+
+  ///
+  /// \brief Computes the inverse of the constraint matrix [[M JT], [J 0]].
```
jcarpent: You can call this matrix the KKT matrix of the contact dynamics.

src/algorithm/dynamics.hpp
```diff
@@ -106,7 +106,45 @@ namespace pinocchio
...
+    Eigen::Block<typename Data::MatrixXs> bottomRight = MJtJ_inv.bottomRightCorner(nc, nc); 
+
+    bottomRight.setIdentity(); topLeft.setIdentity();
+    data.llt_JMinvJt.solveInPlace(bottomRight);    cholesky::solve(model, data, topLeft);
+
+    bottomLeft.noalias() = J*topLeft;
+    topRight.noalias() = bottomLeft.transpose() * bottomRight;
+    topLeft.noalias() -= topRight*bottomLeft;
+    bottomLeft = topRight.transpose();
+    bottomRight *=-1;
```
jcarpent: You can avoid this last operation to save computation times.

unittest/dynamics.cpp
```diff
@@ -79,7 +79,16 @@ BOOST_AUTO_TEST_CASE ( test_FD )
   
   Eigen::VectorXd dynamics_residual (data.M * data.ddq + data.nle - tau - J.transpose()*data.lambda_c);
   BOOST_CHECK(dynamics_residual.norm() <= 1e-12);
-  
+
```
jcarpent: Can you a dedicated test for this new feature inside this file?

src/algorithm/dynamics.hpp
proyan: [M Jt]
[J 0]
Reading from left to right, top to bottom.

jcarpent: Then, I would call it the KKTContactDynamicMatrix

src/algorithm/dynamics.hpp
proyan: I have the choice between
1) saving A; do computations using A; replace A with -A
2) saving -A; do computations using A by calling (-)(-A) everytime.

I chose the first one. Do you think that the second would be better?

unittest/dynamics.cpp
proyan: I would be reusing 99% of the same code as in forwarddynamics. Since this is only a getter function (no matter how many computations are needed), I think this should remain with the function which does the actual computations.

src/algorithm/dynamics.hpp
jcarpent: Yes, Eigen will do the job for you in this precise context!

bindings/python/algorithm/expose-dynamics.cpp
proyan: Done

src/algorithm/dynamics.hpp
proyan: Done.

unittest/dynamics.cpp
jcarpent: I understand your point, but I really prefer to have a different test for different functions, otherwise, it will become a mess.
In addition, if the content of forwardDynamics is changing (that will happen next week), then you won't be able to really say that the two codes are identical.

src/algorithm/dynamics.hpp
proyan: I would still need to do multiply by -1, since I am creating the A matrix by solveInPlace(-Identity)
I could do a *-1 on the diagonal only after setting identity. this would be faster.

jcarpent: You can set it to -Identity. Eigen will the optimization for you.

unittest/dynamics.cpp
proyan: Done.

src/algorithm/dynamics.hpp
proyan: getKKTContactDynamicMatrixInverse seems too long. are you sure of this?

jcarpent: But much more explicit than the current solution.
Maybe something like getContactDynamicMatrixInverse is a bit shorter and as explicit as the previous solution.

proyan: Done.

src/algorithm/dynamics.hpp
```diff
@@ -103,10 +103,48 @@ namespace pinocchio
...
+  /// \param[in] data The data structure of the rigid body system.
+  /// \param[out] MJtJ_inv inverse of the MJtJ matrix.
+  ///
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl,
+           typename ConstraintMatrixType>
+  inline void getKKTContactDynamicMatrixInverse(const ModelTpl<Scalar,Options,JointCollectionTpl> &
+                             model,
+                             const DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                             const Eigen::MatrixBase<ConstraintMatrixType> & J,
+                             typename DataTpl<Scalar,Options,JointCollectionTpl>::MatrixXs& MJtJ_inv)
```
jcarpent: This is my final remark, I think to me as generic as possible, the resulting matrix (which stores the result) should be something like EIgen::MatrixBase<ReturnMatrixType>

unittest/dynamics.cpp
```diff
@@ -79,6 +79,69 @@ BOOST_AUTO_TEST_CASE ( test_FD )
...
+  data.M.triangularView<Eigen::StrictlyLower>() = data.M.transpose().triangularView<Eigen::StrictlyLower>();
+
+  Eigen::MatrixXd MJtJ(model.nv+12, model.nv+12);
+  MJtJ << data.M, J.transpose(),
+    J, Eigen::MatrixXd::Zero(12, 12);
+
+  Eigen::MatrixXd MJtJ_inv(model.nv+12, model.nv+12);
+  getKKTContactDynamicMatrixInverse(model, data, J, MJtJ_inv);
+
+  BOOST_CHECK((MJtJ.inverse()-MJtJ_inv).norm() <=1e-12);
```
jcarpent: Can you use here isApprox?

unittest/dynamics.cpp
```diff
@@ -79,6 +79,69 @@ BOOST_AUTO_TEST_CASE ( test_FD )
...
+  const double r_coeff = 1.;
+  VectorXd v_before = VectorXd::Ones(model.nv);
+  pinocchio::impulseDynamics(model, data, q, v_before, J, r_coeff, true);
+  data.M.triangularView<Eigen::StrictlyLower>() = data.M.transpose().triangularView<Eigen::StrictlyLower>();
+  MJtJ << data.M, J.transpose(),
+    J, Eigen::MatrixXd::Zero(12, 12);
+
+  getKKTContactDynamicMatrixInverse(model, data, J, MJtJ_inv);
+
+  BOOST_CHECK((MJtJ.inverse()-MJtJ_inv).norm() <=1e-12);
```
jcarpent: Same apply here.

unittest/dynamics.cpp
proyan: Done.

src/algorithm/dynamics.hpp
proyan: Done.

unittest/dynamics.cpp
proyan: Done.

src/algorithm/dynamics.hpp
```diff
@@ -103,10 +103,47 @@ namespace pinocchio
...
+  /// \param[in] model The model structure of the rigid body system.
+  /// \param[in] data The data structure of the rigid body system.
+  /// \param[out] MJtJ_inv inverse of the MJtJ matrix.
+  ///
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl,
+           typename ConstraintMatrixType, typename KKTMatrixType>
+  inline void getKKTContactDynamicMatrixInverse(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                                                const DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                                                const Eigen::MatrixBase<ConstraintMatrixType> & J,
+                                                Eigen::MatrixBase<KKTMatrixType> & MJtJ_inv)
```
jcarpent: To make it work, you have to do a const_cast. See other similar functions in algo directory.

proyan: Done.

jcarpent:
Seems fine now! Thanks for your contribution!

Issue #656, opened 2019-02-01
Definition of neutral configuration


proyan:
We currently have two definition for a neutral configuration for a model.
1) model::neutralConfiguration, which can be set by srdf, and is being used for loading the half-sitting posture of a robot.

2) pinocchio::neutral(model) in joint-configurations.hpp, which returns the equivalent of zero for the joint manifold.

In my opinion, we should change one of these names. Both of these could be used for either of the purposes.

I would recommend changing neutralConfiguration to something else.

jcarpent:
We already discussed this point with @jmirabel and we decided to modify the neuralConfiguration.
I would be more to remove it completely and let the SRDF parser returns the value as a ConfigVector type.

proyan:
I would be okay with this change.
However, we still need to give a name to what used to be neutralConfiguration for loading with srdf.
If you are not saving it in model anymore, I would recommend defaultConfiguration as the replacement.

jcarpent:
I would prefer `referenceConfigurationVector`but this one should not be in Model class.

proyan:
I like that name.

I'll put neutralConfiguration and the srdf function to deprecated, and create a PR with the new names.

jcarpent:
OK. Sounds good to me!

nmansard:
If not added to model, where do you want to put it?

jcarpent:
I think this information should not be inside model because you may have various referenceConfigurationVector. But we may provide just a function to load it from a SRDF file.

jmirabel:
For me, referenceConfigurationVector means there are several configurations. I don't think Vector adds anything. I would name it referenceConfiguration(s) depending on the number of reference configuration.

If there are several configurations, I think I would store them into a map in order to assign a name to each.

Having a function to extract it from a SRDF is not convenient. It forces developers to store somewhere either the filename or the configurations. In most places where you want this configuration, you don't have the filename.

proyan:
I think srdf files are meant to be customizable by the developers. So saving this map `name->configuration` in an srdf file, and making this file available in the <robot>-description should not be a  problem.
As developers, we are storing the urdf filenames as well when we load a robot. i don't see why srdf should be different.

jcarpent:
@proyan Can you take in charge the requested modifications?

proyan:
I am not sure if we all agree on the same modifications. Are you okay if I implement the suggestion of @gabrielebndn?

jcarpent:
[...]

Issue #375, opened 2017-02-15
spurious output


florent-lamiraux:
Would it be possible not to pollute the standard output when using pinocchio ?
If necessary https://github.com/humanoid-path-planner/hpp-util is a light weight package that provides logging into files.

olivier-stasse:
:+1:

jmirabel:
:+1:

jcarpent:
Discussing with @jmirabel, I got your issue. We can indeed use LOG system. I would go for boost-log to avoid any further dependency.

But it seems that you obtain some outputs from verbosity mode when reading URDF model. This behavior is expected, isn't it?

So, we can also stream to LOG variables.

jmirabel:
IMHO, `verbose` should not be an arguments of functions, like in prototype of `buildModel`. Verbosity should be controlled project-wise (through macros or free-functions) and not function-wise.

jcarpent:
I think it really depends on your application and what you want to see. Sometines, one just needs local printed info for a specific function, like when one loads an URDF file.

jmirabel:
The functionality you talk about can be achieved in many other ways.

- Function-wise control pollutes function prototype and do not add extra functionality.
- The output of `buildModel` is very long and has to be written to a file for inspection. It wouldn't be less readable with other stuff in the file.
- Project-wise control can be made very easy to parse. This is easier than modifying some c++ code to enable verbosity.

jcarpent:
Ok, I'm open to propostions or pointees: what do you suggest to do?
The main idea is to avoid big files with thousands of lines.

jcarpent:
In addition, we want the log data be available from Python.

jmirabel:
> In addition, we want the log data to be available from Python.

Boost Python makes `std::cout` available in Python ? If you work with stream, you can redirect the stream wherever you want.

> The main idea is to avoid big files with thousands of lines.

In the first tutorial of HPP, Pinocchio prints 3308 lines to the standard output... I don't think you can avoid files with thousands of lines...

jmirabel:
My suggestion is either internal macros (you can probably get some inspiration from hpp-util) or external library like boost-log or glog.

jcarpent:
@florent-lamiraux Would you have time to solve this issue?

jcarpent:
With no activity for a while, I will close this issue.

PR #666, opened 2019-02-12
Create std map referenceConfigurations in model, and load from srdf

proyan:


proyan:
closes #656

src/multibody/fwd.hpp
```diff
@@ -8,6 +8,7 @@
 #include "pinocchio/fwd.hpp"
 
 # include <cstddef> // std::size_t
+# include <map>
```
jcarpent: Can you add this in model.hpp

src/parsers/srdf.hpp
```diff
@@ -94,6 +95,14 @@ namespace pinocchio
                                     const std::string & filename,
                                     const bool verbose = false) throw (std::invalid_argument)
     { return getNeutralConfiguration(model,filename,verbose); }
+
+
```
jcarpent: Can you add documentation here?

src/parsers/srdf.hxx
```diff
@@ -200,12 +200,35 @@ namespace pinocchio
...
       return false; // warning : uninitialized vector is returned
     }
-    
+
     template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
     typename ModelTpl<Scalar,Options,JointCollectionTpl>::ConfigVectorType
     getNeutralConfiguration(ModelTpl<Scalar,Options,JointCollectionTpl> & model,
                             const std::string & filename,
                             const bool verbose) throw (std::invalid_argument)
+    {
```
jcarpent: This function is now deprecated. It would be better to not modify its content.

src/parsers/srdf.hxx
```diff
@@ -200,12 +200,35 @@ namespace pinocchio
...
+      else
+      {
+        return it->second;
+      }   
+    }
+    
+    template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
+    void
+    loadReferenceConfigurations(ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                               const std::string & filename,
```
jcarpent: Can you correct the alignment here?

bindings/python/utils/std-aligned-map.hpp
```diff
@@ -0,0 +1,41 @@
...
+    
+    ///
+    /// \brief Expose an std::map from a type given as template argument. Uses the Eigen::aligned_allocater in the value.
+    ///
+    /// \tparam KeyType, ValType Type to expose as std::map<KeyType, ValType, Eigen::aligned_allocator<KeyType, ValType>>.
+    ///
+    /// \sa StdAlignedMapPythonVisitor
+    ///
+    template<class KyeType, class ValType, bool NoProxy = true>
+    struct StdAlignedMapPythonVisitor
```
jcarpent: As it is clearly similar to the not aligned version, you should put bot( in the same .hpp file

bindings/python/multibody/model.hpp
```diff
@@ -87,6 +89,11 @@ namespace pinocchio
                       make_getter(&Model::neutralConfiguration, bp::return_value_policy<bp::return_by_value>()),
                       make_setter(&Model::neutralConfiguration, bp::return_value_policy<bp::return_by_value>()),
                       "Joint's neutral configurations.")
+
+        .add_property("referenceConfigurations",
```
jcarpent: You should not return by value here!

bindings/python/multibody/model.hpp
```diff
@@ -7,6 +7,7 @@
 #define __pinocchio_python_model_hpp__
 
 #include <boost/python/suite/indexing/vector_indexing_suite.hpp>
+#include <boost/python/suite/indexing/map_indexing_suite.hpp>
```
jcarpent: Not needed here.

src/parsers/srdf.hxx
```diff
@@ -238,47 +261,48 @@ namespace pinocchio
...
-                else
+              // Search in model the joint and its config id
+              if (joint_id != model.joints.size()) // != model.njoints
+              {
+                const JointModel & joint = model.joints[joint_id];
+                typename Model::ConfigVectorType joint_config(joint.nq());
+                const std::string joint_val = joint_tag.second.get<std::string>("<xmlattr>.value");
+                std::istringstream config_string(joint_val);
+                std::vector<double> config_vec((std::istream_iterator<double>(config_string)), std::istream_iterator<double>());
+                joint_config = Eigen::Map<typename Model::ConfigVectorType>(config_vec.data(), config_vec.size());
```
jcarpent: This should be a map on Eigen::VectorXd

florent-lamiraux:
Would it be possible to replace "assert" by exceptions when reading files ?

src/multibody/fwd.hpp
proyan: Done.

src/parsers/srdf.hxx
proyan: Done.

bindings/python/multibody/model.hpp
proyan: Done.

bindings/python/multibody/model.hpp
proyan: Done.

src/parsers/srdf.hxx
proyan: Done.

bindings/python/utils/std-aligned-map.hpp
proyan: I am following the same standard as std-aligned-vector and std-vector.
Would you still prefer that I change this?

src/parsers/srdf.hxx
proyan: Since the function is still valid, the change is meant to keep its current behaviour unmodified in order to not affect any packages downstream.

I am not adding any changing the behaviour of the function here. The function remains the same for the users.

src/parsers/srdf.hpp
proyan: Done.

jcarpent:
As you are dealing with ConfigVector which is a dynamically sized vector, you don't need to have aligned map allocator.
I would suggest removing this useless feature as it does not help at all here, make the code less readable.

bindings/python/utils/std-aligned-map.hpp
jcarpent: I've replied to this issue bellow

src/parsers/srdf.hxx
jcarpent: Yes, but not the content. This also means that if you've introduced a bug here, then the bug will appear to the user too.

proyan: Would you prefer that I duplicate the code?

bindings/python/utils/std-aligned-map.hpp
proyan: I just created the python visitor for the sake of completion of the code.
I removed the visitors and the aligned-allocator. the binding is now directly using the bp::map_indexing class

src/parsers/srdf.hxx
jcarpent: I would prefer to not modify the code of getNeutralConfiguration, some having some duplicate codes. It is just a safer procedure when dealing with deprecated functions.

proyan:
@florent-lamiraux Currently there is an assert at the end, if the configvector loading does not work properly. There is no assert during the file loading.
@jcarpent can perhaps answer better regarding the policy regarding exceptions and asserts?

src/parsers/srdf.hxx
proyan: Done.

src/parsers/srdf.hpp
```diff
@@ -94,6 +95,22 @@ namespace pinocchio
...
+    ///
+    /// \brief Get the reference configurations of a given model associated to a SRDF file.
+    ///        It throws if the SRDF file is incorrect. The reference configurations are
+    ///        saved in a map indexed by the configuration name
```
jcarpent: Can you add missing points in the documentation?

proyan: What do you mean by missing points?

jcarpent: To close the sentences of your phrases. There are two missing points.
I know, this is a bit fullish, but it will avoid additional work later.

proyan: Done.

jcarpent:
Finally, we are done. @proyan Thanks for your contribution.

src/parsers/srdf.hxx
```diff
@@ -280,6 +280,87 @@ namespace pinocchio
...
+                joint_config = Eigen::Map<Eigen::VectorXd>(config_vec.data(), config_vec.size());
+                ref_config.segment(joint.idx_q(),joint.nq()) = joint_config;
+                if (verbose)
+                {
+                  std::cout << "(" << joint_name << " , " << joint_config.transpose() << ")" << std::endl;
+                }
+              }
+              else
+              {
+                if (verbose) std::cout << "The Joint " << joint_name << " was not found in model" << std::endl;
```
jmirabel: I think the handling of error messages is not very convenient. Either you enable logs and you get a whole bunch of useless message, with maybe an error in between, or you don't enable logs and you don't get any error messages.

I think would be smarter to put this in std::cerr, even when verbose if false.

This remark is related to https://github.com/stack-of-tasks/pinocchio/issues/375

src/parsers/srdf.hxx
```diff
@@ -280,6 +280,87 @@ namespace pinocchio
...
+      read_xml(srdf_stream, pt);
+      
+      // Iterate over all tags directly children of robot
+      BOOST_FOREACH(const ptree::value_type & v, pt.get_child("robot"))
+      {
+        // if we encounter a tag group_state
+        if (v.first == "group_state")
+        {
+          const std::string name = v.second.get<std::string>("<xmlattr>.name");
+          if ( !model.referenceConfigurations.insert(std::make_pair(name, Model::ConfigVectorType::Zero(model.nq))).second)
```
jmirabel: The configuration shouldn't be initialized with 0. It should be initialized using LieGroup::neutral.

jmirabel: Or rather `neutral (model, q)`. I will put this change on top of fcl-release.

src/parsers/srdf.hxx
proyan: Thanks @jmirabel
Error handling and logging is a much bigger issue, I mainly followed the existing practice when loading the srdf. Just editing this one error might not be sufficient for what is eventually desired.
I'll let the gurus take care of this :)

PR #673, opened 2019-02-13
Tests & Python3

nim65s:
Following https://github.com/stack-of-tasks/pinocchio/pull/645#issuecomment-462762306 I tried to see if we could make a release of pinocchio using hpp-fcl 1.0.0. 

Manually running tests on 14.04, 16.04 & 18.04 with python 2 and on 18.04 with python 3 raised a few issues. This PR address those (and a bit more).

I also found that the rnea test pass on 14.04 & 16.04, but not 18.04 (the python version doesn't matter), so I'll open a separate issue about that.

I think that this PR is easier to review commit after commit. If needed, I can also split it in more pieces.

I have Dockerfiles for 14.04, 16.04 & 18.04 with python 2 and on 18.04 with python 3 if someone want to run the same tests as I did, but it would require extra work to make them available outside of the LAAS, as some parts of the buildfarm that generate their base image contain private data. This extra work will be done after this release.


examples/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+  #overview-urdf
+  )
+
+FOREACH(EXAMPLE ${${PROJECT_NAME}_EXAMPLES})
+  IF(BUILD_PYTHON_INTERFACE)
+    ADD_PYTHON_UNIT_TEST("py-example-${EXAMPLE}" "examples/${EXAMPLE}.py" "bindings/python")
+  ENDIF(BUILD_PYTHON_INTERFACE)
+  ADD_UNIT_TEST("example-${EXAMPLE}" ${EXAMPLE})
+  PKG_CONFIG_USE_DEPENDENCY("example-${EXAMPLE}" eigen3)
+  PKG_CONFIG_USE_DEPENDENCY("example-${EXAMPLE}" hpp-fcl)
```
jcarpent: You should add a flag to check if HPP_FCL has been found

unittest/CMakeLists.txt
```diff
@@ -79,6 +79,12 @@ IF(BUILD_PYTHON_INTERFACE)
 
   TARGET_LINK_LIBRARIES(python_parser ${PYWRAP})
   TARGET_LINK_LIBRARIES(python_parser ${PYTHON_LIBRARIES})
+  SET(PYTHONPATH "${CMAKE_BINARY_DIR}/bindings/python")
+  IF(DEFINED ENV{PYTHONPATH})
```
jcarpent: You are not obliged to add this IF condition here

bindings/python/parsers/python/model.cpp
```diff
@@ -68,11 +69,38 @@ namespace pinocchio
...
       }
 
-      // close interpreter
+      // close the interpreter
+      // cf. https://github.com/numpy/numpy/issues/8097
+#if PY_MAJOR_VERSION < 3
       Py_Finalize();
+#else
+
+      PyObject * poMainModule = PyImport_AddModule("__main__");
```
jcarpent: Maybe you should add a ref for those lines to help people in the future for a diagnosis of this code.

nim65s:
I added too much tests, now travis complains about the logs being too long and aborts after the 10 000th line.

bindings/python/parsers/python/model.cpp
nim65s: It is on the new line 73: 
`       // cf. https://github.com/numpy/numpy/issues/8097`

nim65s:
The path of the scripts is for eb7d8422b223faa57cbe503073aba8f236e6c311 and explained in the commit message as follow:
> compiling all python scripts into CMAKE_BUILD_DIR/bindings/python/pinocchio, where the .so already is, and add this directory to the PYTHONPATH for this test.

In other words, my goal is to have `CMAKE_BUILD_DIR/bindings/python/pinocchio` with both all the `.pyc` files and the `libpinocchio_pywrap.so`. Those files must be located inside the same directory to have a valid python module.

And the `PYTHON_BUILD` from the cmake submodule requires that the source module and the built module have the same name.

jcarpent:
Ok. But if possible, it would be nice to not install the *.py files but only their compiled versions.

bindings/python/parsers/python/model.cpp
jcarpent: Ok. I did not see it. Thanks!

nim65s:
I don't understand why we would want to install .pyc files and not their corresponding .py.

nim65s:
Do you have references for this standard way to proceed ? All the linux distributions I use provide .py files in their packages.

I agree that there is no reason to have both, but as you can get the compiled version from the source version, and not get the source version if you only have the compiled version… Why would you provide only the compiled ?

jcarpent:
Ok. If you say so that all distributions provide twice information it is fine for me.
But, I do not think that the make uninstall is removing both.

nim65s:
I don't think that the make install actually installs the .pyc for now. And btw, if you provide only .pyc and not .py:
```
In [1]: from pinocchio import RobotWrapper

In [2]: RobotWrapper??
Init signature:
RobotWrapper(self, model=Nb joints = 1 (nq=0,nv=0)
  Joint universe: parent=0
, collision_model=None, visual_model=None, verbose=False)
Docstring:      <no docstring>
File:           /local/users/gsaurel/stack-of-tasks/pinocchio/build/bindings/python/pinocchio/robot_wrapper.pyc
Type:           type
```

gabrielebndn:
Nice job!

unittest/python/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+  bindings_force
+  bindings_frame
+  bindings_geometry_object
+  bindings_geometry_object_urdf
+  bindings_inertia
+  bindings_joint_composite
+  bindings_motion
+  bindings_SE3
+  #explog
+  model
```
gabrielebndn: Why are you not testing bindings_com, bindings_dynamics and explog?
Also, there is a script called tests which runs all Python tests one after the other. Do we still need it?

examples/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+  i-inverse-kinematics
+  overview-simple
+  #overview-urdf
+  )
+
+FOREACH(EXAMPLE ${${PROJECT_NAME}_EXAMPLES})
+  IF(BUILD_PYTHON_INTERFACE)
+    ADD_PYTHON_UNIT_TEST("py-example-${EXAMPLE}" "examples/${EXAMPLE}.py" "bindings/python")
+  ENDIF(BUILD_PYTHON_INTERFACE)
+  ADD_UNIT_TEST("example-${EXAMPLE}" ${EXAMPLE})
```
gabrielebndn: So you have decided to treat the examples as unit tests? How about treating them separately, with a separate compilation option BUILD_EXAMPLES?

examples/i-inverse-kinematics.cpp
```diff
@@ -10,37 +10,35 @@ int main(int argc, char ** argv)
...
+    q = pinocchio::integrate(model,q,v*DT);
+    if(!(i%10)) std::cout << "error = " << err.transpose() << std::endl;
 
-  if(i==IT_MAX)
-    std::cout << "\nWarning: the iterative algorithm has not reached convergence to the desired precision" << std::endl;
-
-  std::cout << "\nresult: " << q.transpose() << std::endl;
-  std::cout << "\nfinal error: " << err.transpose() << std::endl;
- 
+    if(i==IT_MAX)
```
gabrielebndn: Thank you for fixing this example :)
This test will never work because it is now inside the for loop, therefore i will never be equal to IT_MAX inside the loop.
Aside from the obvious error, why did you change this example?
Also, any change to this file should be reflected in the corresponding Python example, and in the corresponding doc page [b-examples/i-inverse-kinematics.md](https://github.com/stack-of-tasks/pinocchio/blob/devel/doc/b-examples/i-inverse-kinematics.md), which reference the file line by line and is now (I think) broken

unittest/python/CMakeLists.txt
nim65s: As I said in the commit message:
> 3 are not passing, they are commented for now

But obviously we have to fix them :)

nim65s: About the `tests.py` script, I don't think we need it. It is easier to run those tests with CTests, as it is already integrated in the test suite, and giving each test individually to the system helps to run them in parallel, and to easily point which one is failing

gabrielebndn: I hadn't seen the commit message. But that's weird, I was sure they were working. I am going to check this, but are they not passing at all, or just for a specific system, like 18.04 ?

jcarpent: @gabrielebndn Can you take in charge the fixing of the Python tests?

examples/CMakeLists.txt
nim65s: It was a request from @nmansard . I will add the `BUILD_EXAMPLES` option.

examples/i-inverse-kinematics.cpp
nim65s: `i==IT_MAX` can't be true inside the loop, that's right. But before, it was outside of the loop, and the test did not even compile, as `i` was defined inside the loop…

nim65s: The other parts I changed were changed for the very same reason: this file did not compile. A `;` was missing, and the call to `pinocchio::jointJacobian` was wrong, and the `err` which also defined in the loop was also accessed from outside of the loop.

bindings/python/CMakeLists.txt
```diff
@@ -162,7 +162,7 @@ SET(PYTHON_FILES
 FOREACH(python ${PYTHON_FILES})
   PYTHON_BUILD(${PROJECT_NAME} ${python})
   INSTALL(FILES
-    "${${PROJECT_NAME}_SOURCE_DIR}/bindings/python/scripts/${python}"
+    "${${PROJECT_NAME}_SOURCE_DIR}/bindings/python/pinocchio/${python}"
```
nim65s: Thanks :)

jcarpent:
Maybe, for this first round, you can deactivate the tests which are not passing. We will fix them later (next week). @nim65s What do you think?

unittest/python/CMakeLists.txt
nim65s: It is not about a specific system. The missing `;` was clearly making this file invalid.

examples/i-inverse-kinematics.cpp
gabrielebndn: Damn, I don't know how it happened. Do as you please

unittest/python/CMakeLists.txt
gabrielebndn: So, I tested these example both on my Mac (OSX El Capitan, 10.11) and on Ubuntu 16.04, current devel branch.
`explog` is passing on both systems.
`bindings_com` and `bindings_dynamics` seem to return correctly on Ubuntu 16.04, but they are triggering assertion errors on my Mac, where assertions are enabled (haven't tested enabling assertions on Ubuntu, I will do it soon).

nim65s: > It is not about a specific system. The missing `;` was clearly making this file invalid.

I was not in the right conversation thread, sorry, I misunderstood. I can try to test them again, I don't remember.

gabrielebndn: I confirm `bindings_com` and `bindings_dynamics` raise an assertion, while `explog` segfaults in Debug mode. I think we can go on with this PR. I will open a separate issue on this

examples/CMakeLists.txt
nmansard: +1 
I like the idea of having an automatic validation of the examples.

unittest/CMakeLists.txt
nim65s: Do you suggest to remove `IF(DEFINED ENV{PYTHONPATH})` and `ENDIF(DEFINED ENV{PYTHONPATH})` but keep `SET(PYTHONPATH "${PYTHONPATH}:$ENV{PYTHONPATH}")`, or also remove this one ?

for reference: https://github.com/jrl-umi3218/jrl-cmakemodules/pull/137#discussion_r191808174

jcarpent: Just keep as you mentioned, it is sufficient. Or I may be wrong?

nim65s: Well… In fact, this `IF` was meant to prevent having a trailing `:` at the end of the `PYTHONPATH`, as it would add the current working directory in `sys.path`. 
But firstly I am not even sure if this can raise issues, and secondly this `IF` is useless when `ENV{PYTHONPATH}` is defined but empty.

So let's do what you suggest, and if someone find a real case issue with that, we well have to use `IF(ENV{PYTHONPATH})`, and put that also in the cmake submodule.

nim65s:
I had to rebase it for #675 , and I addressed your comments.

CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ SET(PROJECT_DESCRIPTION "Rigid multi body dynamics algorithms")
 SET(PROJECT_URL "http://github.com/stack-of-tasks/pinocchio")
 SET(PROJECT_CUSTOM_HEADER_EXTENSION "hpp")
 OPTION(INSTALL_DOCUMENTATION "Generate and install the documentation" ON)
+OPTION(BUILD_EXAMPLES "Build and test the examples" ON)
```
gabrielebndn: I would set it OFF by default

examples/CMakeLists.txt
nim65s: done in 924953ee0a71aed189a5c158de52a2a375f0d8d3

unittest/CMakeLists.txt
nim65s: done in 3c677514bdb578a69e211e59c55c6c86b68fb85f

examples/CMakeLists.txt
nim65s: done in b925bce8bd54482a130893ecfdadcf7b8c009dd2

examples/i-inverse-kinematics.cpp
nim65s: fixed in 53b7fdba6be2689a9c96f142c3da1b105b91a31d

CMakeLists.txt
nim65s: Why ?

examples/i-inverse-kinematics.cpp
```diff
@@ -20,7 +20,7 @@ int main(int argc, char ** argv)
   pinocchio::Data::Matrix6x J(6,model.nv); J.setZero();
   unsigned int svdOptions = Eigen::ComputeThinU | Eigen::ComputeThinV;
 
-  for (int i=0 ; i<IT_MAX ; ++i)
+  for (int i=0 ; i<=IT_MAX ; ++i)
```
jcarpent: Why this change? And what does your commit message mean?

CMakeLists.txt
gabrielebndn: These are examples. They do not add any functionality, so I see no point in building them by default. I think the best approach is that whoever wants them should explicitly ask for them, otherwise nothing. Pinocchio already takes a lot of time to compile, let's not add by default an unnecessary other burden that needs to be switched off if you just want Pinocchio to be working and you want to keep compilation times below the span of a human life

examples/i-inverse-kinematics.cpp
gabrielebndn: It is to address [this comment](https://github.com/stack-of-tasks/pinocchio/pull/673/files#r256545913).
I do not think it is the best solution, but I guess we can think about it in another moment.

gabrielebndn:
@nim65s , I have just fixed issue #677 with PR #678.
Maybe, once this is merged, you can rebase your changes and uncomment the Python tests?

jcarpent:
@nim65s I have merged the PR #678. You can proceed when you want.

CMakeLists.txt
nim65s: > I see no point in building them by default.
These are tests. The point in building them is to check that the examples are still up to date.

>  think the best approach is that whoever wants them should explicitly ask for them, otherwise nothing
Then are you also asking to turn `BUILD_UNIT_TESTS` to `OFF` by default ?

> Pinocchio already takes a lot of time to compile
```
$ cmake .. ; time make -j8 ; time make test
0:32.21elapsed 621%CPU
0:05.32elapsed 319%CPU

$ cmake -DBUILD_EXAMPLES=OFF .. ; time make -j8 ; time make test
0:30.50elapsed 619%CPU
0:05.48elapsed 299%CPU

$ cmake -DBUILD_UNIT_TESTS=OFF ..; time make -j8
0:16.67elapsed 107%CPU
```

Once again, are you asking to turn `BUILD_UNIT_TESTS` to `OFF` by default ?.

If you let `BUILD_EXAMPLES` to `ON` but put `BUILD_UNIT_TESTS` to `OFF`, they will not be compiled. 
Therefore, I think we should remove this `BUILD_EXAMPLES` option, it really adds nothing, and if someone really want to turn this compilation off, they might as well just disable all tests.

jcarpent: I do agree.

nim65s: if you want me to remove ccache, I can run those benchmarks again, but my point is that no, for now, we have like ~50 loc for 2 .cpp examples with only one main function. The compilation time of that is neglegible.

examples/i-inverse-kinematics.cpp
nim65s: Yes, I took the quick & dirty path, sorry. This fixed an issue, but not with the best style. I think finding the best style can wait and be addressed outside of this PR.
@jcarpent please tell me if you want any change, or if we can resolve this conversation.

CMakeLists.txt
nim65s: @jcarpent thanks, fixed in 7b69be675d8cf42b1e1d1f9f456b959607ff620e

examples/i-inverse-kinematics.cpp
jcarpent: It seems fine for now. Thanks to you.

CMakeLists.txt
jcarpent: Sounds good to me.

nim65s:
I can't reproduce travis' errors…

jcarpent:
The problem seems to come from EigenPy. Is the version on Travis the latest one?

nim65s:
So if eigenpy's version was an issue, we should probably update 
https://github.com/stack-of-tasks/pinocchio/blob/master/bindings/python/CMakeLists.txt#L39

jcarpent:
Yes sure. I've just done it.

nim65s:
Let's rebase for the conflict in bindings/python/scripts/deprecated.py from #680

jcarpent:
It passes! Very nice job Guilhem!

PR #675, opened 2019-02-13
Fix center of mass derivatives debug compilation error

gabrielebndn:
With assertions enabled, center-of-mass-derivatives gave a compilation error due to a misnamed variable (which was not seen in Release mode because the assertion is simply skipped).
I fixed it. I grabbed the occasion to change `import pinocchio as pio` -> `import pinocchio as pin` in the corresponding unit test.


nmansard:
Thanks for that, sorry to have introduced a bug in debug mode. Indeed I 
did not valide the debug compilation, and it is not tested by travis (as 
we discussed it this morning with @nim65s).

gabrielebndn:
Makes sense. Actually I have no idea why I got this problem. I am supposed to be compiling in Debug mode, yet I still see the assertions. It's as if NDEBUG were not defined. Either there is something wrong in my configuration, or this is a very unexpected behavior from my compiler

Issue #671, opened 2019-02-13
Dealing with armature in crba, computeAllTerms and other dynamics algos


proyan:
Currently we can load the rotor parameters inside Pinocchio, but they are not used anywhere in the algorithms.
As a result, forwardDynamics, and ImpulseDynamics are useless if you need the armature, since both internally computes the mass during computeAllTerms.

I can take care of adding the armature values to crba and computeAllTerms by either:
1) Modify the current algos and add a (by default armature is zero) value to the mass diagonals()
2) create a separate algo.

Would (1) be okay with you?

jcarpent:
The problem of the armature is for multi-dof joints.
Currently, I do not see how to handle them correctly in this meaning.

jcarpent:
The correct way to deal with armature information would be by implementing #440 features, but no one is available to implement it.

proyan:
#440 would require plenty of work. It has been open for more than a year, and I don't see that it will be done anytime soon.
However, just adding the available aramature to the mass matrix would provide an acceptable solution for the time being. This would be beneficial for anyone doing control.

proyan:
Plus, we already save the information in model class.

jcarpent:
Indeed, we save it as it is now because we have not implemented a convenient and generic solution, as suggested in #440.

jcarpent:
To conclude, I think you can proceed as 1 but this is not the most elegant and generic solution.
By the way, you should assert that the armature is indeed a positive vector.

proyan:
I agree it isn't generic. The solution will just be a placeholder untill #440 is solved.
I'll work on it.

proyan:
Extending Armature to ABA and RNEA:

I understood how to add the armature to the aba (and by extension, abaderivatives). It requires me to:
1) call calc_aba in AbaForwardPass2
2) define a constraintAxisSelector in the constraintBase class (akin to the jointConfigSelector). It takes as input a 6D vector, and returns the segment where jdata.S() is active.

Proposals:
1) (1) seems more intrusive than simply adding a diagonal term (as in the case of crba and computeAllTerms). I propose to add a function abaWithArmature, which uses (1) as the ForwardPass2.
2) I will simply add the armature terms to RNEA, instead of creating a new function.It is again only one extra addition operation. Unfortunately, this means that we have two aba (aba and abaWithArmature), but one RNEA.

Would you be okay with that?

nmansard:
[...]

PR #676, opened 2019-02-14
Adding armature to computeallterms

proyan:
1) Remove rotorInertia and rotorGearRatio. Replace with armature.
2) add armature to compute all terms
3) test.

Currently it only adds the armature to the mass diagonal. In case you decide to add the coriolis and centrifugal effects, please reflect it in the computeallterms as well.

For now, this deals partially with what was discussed in #671

### DIFF
--- a/src/algorithm/compute-all-terms.hpp
+++ b/src/algorithm/compute-all-terms.hpp
@@ -233,6 +233,10 @@ namespace pinocchio
       Pass2::run(model.joints[i],data.joints[i],
                  typename Pass2::ArgsType(model,data));
     }
+
+    //Add the armature value. By default, this is a zero vector.
+    assert((model.armature.array() >= 0.0).all());
+    data.M.diagonal() += model.armature;
     
     // CoM
     data.com[0] /= data.mass[0];
--- a/unittest/compute-all-terms.cpp
+++ b/unittest/compute-all-terms.cpp
@@ -8,6 +8,7 @@
 #include "pinocchio/multibody/model.hpp"
 #include "pinocchio/multibody/data.hpp"
 #include "pinocchio/algorithm/crba.hpp"
+#include "pinocchio/algorithm/joint-configuration.hpp"
 #include "pinocchio/algorithm/rnea.hpp"
 #include "pinocchio/algorithm/jacobian.hpp"
 #include "pinocchio/algorithm/compute-all-terms.hpp"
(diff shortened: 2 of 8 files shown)


unittest/compute-all-terms.cpp
```diff
@@ -162,4 +162,61 @@ BOOST_AUTO_TEST_CASE ( test_against_algo )
...
+
+  pinocchio::Model model; buildModels::humanoidRandom(model);
+  pinocchio::Data data(model); data.M.fill (0.);
+  pinocchio::Data data_other(model); data_other.M.fill (0.);
+
+  VectorXd q (VectorXd::Random(model.nq));
+  VectorXd v (VectorXd::Random(model.nv));
+
+  // ----Check that Original computeAllTerms still works-
+  q.setZero ();
```
jcarpent: Why q et v set to Zero?

unittest/compute-all-terms.cpp
```diff
@@ -162,4 +162,61 @@ BOOST_AUTO_TEST_CASE ( test_against_algo )
...
+  potentialEnergy(model, data_other, q, true);
+
+  BOOST_CHECK (data.nle.isApprox(data_other.nle, 1e-12));
+  BOOST_CHECK (Eigen::MatrixXd(data.M.triangularView<Eigen::Upper>())
+              .isApprox(Eigen::MatrixXd(data_other.M.triangularView<Eigen::Upper>()), 1e-12));
+  BOOST_CHECK (data.J.isApprox(data_other.J, 1e-12));
+  BOOST_CHECK (data.Jcom.isApprox(data_other.Jcom, 1e-12));
+
+  // ----Check that computeAllTerms with armature works---
+  VectorXd armature(VectorXd::Random(model.nv));
```
jcarpent: armature are positive reals

src/algorithm/compute-all-terms.hpp
```diff
@@ -233,6 +233,9 @@ namespace pinocchio
       Pass2::run(model.joints[i],data.joints[i],
                  typename Pass2::ArgsType(model,data));
     }
+
+    //Add the armature value. By default, this is a zero vector.
```
jcarpent: You should add an assert here, or in the checking of the algo, to check if the armature is indeed positive.

proyan:
By the way if you want to account for the effect on the nle, we should put back the rotorInertia and gearRatio inside Model class.

unittest/compute-all-terms.cpp
proyan: This was a mistake. I will correct it.

jcarpent:
> By the way if you want to account for the effect on the nle, we should put back the rotorInertia and gearRatio inside Model class.

I think the armature is sufficient (but maybe I'm wrong), as it corresponds to the handling of the rotor inertia as observed from the joint axis.

unittest/compute-all-terms.cpp
jcarpent: Ok. Perfect.

nmansard:
Thanks for the contribution. I answer first because we discussed that orally. @jcarpent please react after me. 
I like the change in the parsing and storage of armature elements. 
I am afraid of the issue raised by @gabrielebndn in #676: the dynamic algorithms should all be consistent, i.e. we should also propagate the armature to e.g. mass-inverse, kinematic energy, etc.

I suggest that some post-processing algorithms are added in a dedicated algorithm/armature.hpp file, that would minimally modify the existing algorithm and offer all classical functionalities with armature. Two set of algos can be separated:
- for rnea, crba, computeallterms, this is pure post-processing: run the classical algo, add the armature to the result, returns.
- for aba and massinv algo: as you raised it, you need to re-write a dedicated 3-rd pass of the algorithm, and can reuse the pass 1 and 2.

nmansard:
For this PR, I suggest you only keep the modification of parser and model. This is a consensual one, it should be accepted quickly. In a second PR, please introduce a armature.hpp file with the post-processing of computeallterms. You will finally insert the ABA and Minv new algos in a later PR.

unittest/compute-all-terms.cpp
proyan: Done.

unittest/compute-all-terms.cpp
proyan: Done.

jcarpent:
>     * for rnea, crba, computeallterms, this is pure post-processing: run the classical algo, add the armature to the result, returns.

For RNEA, I think we should take into account centrifugal and Coriolis effects too.

src/algorithm/compute-all-terms.hpp
proyan: Done.

jcarpent:
I think we should try to avoid to put all the versions of armature algorithms in the same file. It will not help the maintenance and the readability, mostly if all the algorithms have to be double.

In terms of naming, I would suggest to add the namespace armature and keep the same naming for the algorithms.

@nmansard What do you think?

jcarpent:
I don't know what to do with this PR. 
I can merge it because it is problematic as the changes are currently transparent for the end users.
Or, we can start to address the issue raised by @nmansard by providing separates signatures when taking into account the armature information.

@proyan @nmansard I think both of you are the main concerned by this question.

proyan:
If we are okay with duplicating and adding separate wrappers for the algorithms, I think you should wait before merging this PR, and I can create a CATWithArmature in armature.hpp. Otherwise, you should merge.
However, I don't know what the resolution of #671 is, so I can't suggest the way forward.

(I personally would prefer the separate signatures)

jcarpent:
In my case, I don't like the withArmature suffix. While the namespace seems more elegant, and easier to use by simply writing `using armature::algo` for instance.
In addition, it allows us to put directly the code inside the dedicated place, instead of adding a single file with a lot of signatures, not relevant if you want to use only `crba` for instance.

proyan:
That is okay with me. I'll close this and make a separate PR which puts CAT it in a separate namespace.

jcarpent:
Ok. Sounds great. Thanks a lot!

PR #678, opened 2019-02-14
Fix python unit tests, fix #677

### LINKED ISSUE #667: ssh-path in .gitmodules limits easy access for certain users.
Hi all,
when cloning the source-code of the current master, certain users may encounter the following problem: currently, there is a SSH-address hard-coded in ".gitmodules", namely
`
url = git://github.com/jrl-umi3218/jrl-cmakemodules.git
`
From within some institutions and universities, it is impossible to access external sources via SSH. It would be great if you could change the URL to
`url = https://github.com/jrl-umi3218/jrl-cmakemodules`, which avoids unnecessary overhead when cloning pinocchio.

Keep up the great work! :)

### LINKED ISSUE #677: Python unit tests failing in non-Release mode
Python unit tests [bindings_com.py](https://github.com/stack-of-tasks/pinocchio/blob/devel/unittest/python/bindings_com.py), [bindings_dynamics.py](https://github.com/stack-of-tasks/pinocchio/blob/devel/unittest/python/bindings_dynamics.py) and [explog.py](https://github.com/stack-of-tasks/pinocchio/blob/devel/unittest/python/explog.py) seem to pass in Release mode, but they fail badly in other compilation modes.

In particular:  when assertions are enabled (not only in Debug mode, tested on Mac OS 10.11), `bindings_com.py` and `bindings_dynamics.py` raise an assertion.
In Debug mode, additionally, `explog.py` segfaults (tested on Ubuntu 16.04).

Although the tests nominally work in Release mode, these error might reflect some subtler problems which were somehow not appearing in the unit tests, therefore it is worth to investigate this.

I will take a look

Related to #673

gabrielebndn:
Fixes #677. Related to #673. The error on `explog.py` was actually due to the binding of `SE3.isIdentity()` which was crashing. I do not understand why there was this problem and why only in Debug mode, but I fixed it.

The error on `bindings_com.py` and `bindings_dynamics.py` was just a trivial error in the unit test itself (wrong vector size). I corrected it. Which makes me think, wouldn't it be better to enable assertions the Python library? If yes, is there any way to treat a failing assertion as an exception instead of having Python crash completely when one fails?

### DIFF
--- a/bindings/python/spatial/se3.hpp
+++ b/bindings/python/spatial/se3.hpp
@@ -25,6 +25,8 @@ namespace pinocchio
   {
     namespace bp = boost::python;
 
+    BOOST_PYTHON_MEMBER_FUNCTION_OVERLOADS(isIdentity_overload,SE3::isIdentity,0,1)
+
     template<typename SE3>
     struct SE3PythonVisitor
       : public boost::python::def_visitor< SE3PythonVisitor<SE3> >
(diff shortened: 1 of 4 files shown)


jcarpent:
@gabrielebndn Cool. Nice fix of the overloading.
I'm waiting for Travis and will merge afterward.

jcarpent:
@gabrielebndn Thanks for your contribution.

nim65s:
> Which makes me think, wouldn't it be better to enable assertions the Python library?

During the tests, yes. I'll add that to #673 .

PR #679, opened 2019-02-15
Reimplement joint-configuration algorithms, fix #658

### LINKED ISSUE #658: Joint-configuration algorithms are performing memory allocation
A lot of emphasis is put on the idea that Pinocchio algorithms *never* perform any memory allocation, relying on storage inside `Data` or in the user passing the required memory,  as in the case of `getJointJacobian(model, data, jointId, rf, J)`, where the result is written in matrix `J`.

However, pretty much all algorithms contained in [joint-configuration.hpp](https://github.com/stack-of-tasks/pinocchio/blob/devel/src/algorithm/joint-configuration.hpp) do not follow this pattern. Instead, the result vector is allocated and returned.

Should we adopt a pattern which mimics the Jacobian case? Do you think it is worth it? In this case, should we deprecate the current signatures, or leave them as a more friendly option? This last question arises from the fact that for instance in most cases `randomConfiguration` and `neutral` are just used as one-shot calls with the specific purpose of initializing a vector with a bogus value at the beginning of a program, and forcing the user to pre-allocate their memory would just make their use clunkier with no real benefit.

gabrielebndn:
Fix #658.
Now all joint-configuration algorithms have two versions: one with return value as argument, and one which allocates memory.

Notice that most algorithms will probably give wrong results if the output is the same object as one of the inputs. This is true in particular for `integrate`. I might implement `integrateInPlace` in the future.

Additionally, I re-implemented `distance` so that it does not employ the auxiliary vector provided by `squaredDistance`.
Also, I am providing the square of the distance with method `squaredDistanceSum`, because `squaredDistance` was already taken. By the way, I think `squaredDistance` is pretty useless now that it is not employed by `distance`, should it be eliminated, and its name recycled for what I called `squaredDistanceSum`?

### DIFF
--- a/src/algorithm/joint-configuration.hpp
+++ b/src/algorithm/joint-configuration.hpp
@@ -15,11 +15,11 @@ namespace pinocchio
   /// \{
 
   /**
-   * @brief      Integrate a configuration for the specified model for a tangent vector during one unit time
+   * @brief      Integrate a configuration vector for the specified model for a tangent vector during one unit time
    *
    * @param[in]  model   Model that must be integrated
    * @param[in]  q       Initial configuration (size model.nq)
-   * @param[in]  v       Velocity (size model.nv)
+   * @param[in]  v       Joint velocity (size model.nv)
    * @param[out] qout    The integrated configuration (size model.nq)
    */
   template<typename LieGroup_t, typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType, typename TangentVectorType, typename ReturnType>
@@ -30,11 +30,11 @@ namespace pinocchio
             const Eigen::MatrixBase<ReturnType> & qout);
 
   /**
-   * @brief      Integrate a configuration for the specified model for a tangent vector during one unit time
+   * @brief      Integrate a configuration vector for the specified model for a tangent vector during one unit time
    *
    * @param[in]  model   Model that must be integrated
    * @param[in]  q       Initial configuration (size model.nq)
-   * @param[in]  v       Velocity (size model.nv)
+   * @param[in]  v       Joint velocity (size model.nv)
    * @param[out] qout    The integrated configuration (size model.nq)
    */
   template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType, typename TangentVectorType, typename ReturnType>
@@ -48,7 +48,7 @@ namespace pinocchio
   }
 
   /**
-   * @brief      Interpolate the model between two configurations
+   * @brief      Interpolate two configurations for a given model
    *
    * @param[in]  model   Model to be interpolated
    * @param[in]  q0      Initial configuration vector (size model.nq)
@@ -65,7 +65,7 @@ namespace pinocchio
               const Eigen::MatrixBase<ReturnType> & qout);
 
   /**
-   * @brief      Interpolate the model between two configurations
+   * @brief      Interpolate two configurations for a given model
    *
    * @param[in]  model   Model to be interpolated
    * @param[in]  q0      Initial configuration vector (size model.nq)
@@ -87,34 +87,34 @@ namespace pinocchio
   /**
    * @brief      Compute the tangent vector that must be integrated during one unit time to go from q0 to q1
    *
-   * @param[in]  model   Model to be differenced
+   * @param[in]  model   Model of the system
    * @param[in]  q0      Initial configuration (size model.nq)
    * @param[in]  q1      Wished configuration (size model.nq)
-   * @param[out] dqout   The corresponding velocity (size model.nv)
+   * @param[out] dvout   The corresponding velocity (size model.nv)
    */
   template<typename LieGroup_t, typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorIn1, typename ConfigVectorIn2, typename ReturnType>
   void
   difference(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
              const Eigen::MatrixBase<ConfigVectorIn1> & q0,
              const Eigen::MatrixBase<ConfigVectorIn2> & q1,
-             const Eigen::MatrixBase<ReturnType> & dqout);
+             const Eigen::MatrixBase<ReturnType> & dvout);
 
   /**
    * @brief      Compute the tangent vector that must be integrated during one unit time to go from q0 to q1
    *
-   * @param[in]  model   Model to be differenced
+   * @param[in]  model   Model of the system
    * @param[in]  q0      Initial configuration (size model.nq)
    * @param[in]  q1      Wished configuration (size model.nq)
-   * @param[out] dqout   The corresponding velocity (size model.nv)
+   * @param[out] dvout   The corresponding velocity (size model.nv)
    */
   template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorIn1, typename ConfigVectorIn2, typename ReturnType>
   void
   difference(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
              const Eigen::MatrixBase<ConfigVectorIn1> & q0,
              const Eigen::MatrixBase<ConfigVectorIn2> & q1,
-             const Eigen::MatrixBase<ReturnType> & dqout)
+             const Eigen::MatrixBase<ReturnType> & dvout)
   {
-    difference<LieGroupMap,Scalar,Options,JointCollectionTpl,ConfigVectorIn1,ConfigVectorIn2,ReturnType>(model,q0.derived(),q1.derived(),dqout.derived());
+    difference<LieGroupMap,Scalar,Options,JointCollectionTpl,ConfigVectorIn1,ConfigVectorIn2,ReturnType>(model,q0.derived(),q1.derived(),dvout.derived());
   }
 
   /**
@@ -219,7 +219,7 @@ namespace pinocchio
    *
    * @param[in]  model   Model that must be integrated
    * @param[in]  q       Initial configuration (size model.nq)
-   * @param[in]  v       Velocity (size model.nv)
+   * @param[in]  v       Joint velocity (size model.nv)
    * @param[out] J       Jacobian of the Integrate operation, either with respect to q or v (size model.nv x model.nv).
    * @param[in]  arg     Argument (either q or v) with respect to which the
    *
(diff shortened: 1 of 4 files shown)


jmirabel:
`squaredDistance` is required for people who wants to tune the metric on the config space i.e. use different weights on each joint. However, this might be achievable with the difference operator.

In HPP, I think we do not use directly these algorithms because we need to put weights on rotational parts of SE(n).

gabrielebndn:
@jmirabel I see. Now that I think about it, summing the rotational difference to the positional one does not make a lot of sense

jmirabel:
By the way, why don't you overload the function for squaredDistance ?

gabrielebndn:
> By the way, why don't you overload the function for squaredDistance ?

Because it's impossible. The method returning the vector and the one returning `double` would have the same signatures, only the return type would change, which is not allowed in. A more coherent naming would be `squaredDistance` for the double and `squaredDistances` or `squareDistanceVector` for the vector, but that would break current code employing `squaredDistance`.

jcarpent:
> @jmirabel I see. One think I realised while doing this, is that `distance` is summing the rotational difference to the positional one does not make a lot of sense

This indeed why we have allowed the overloading of this function through the LieGroupeMap. But for a robotic arm, this is fine.

src/multibody/liegroup/liegroup-algo.hxx
```diff
@@ -276,6 +276,38 @@ namespace pinocchio
...
+    template<typename ConfigVectorIn1, typename ConfigVectorIn2>
+    static void run(const JointModelBase<JointModel> & jmodel,
+                    const Eigen::MatrixBase<ConfigVectorIn1> & q0,
+                    const Eigen::MatrixBase<ConfigVectorIn2> & q1,
+                    typename ConfigVectorIn1::Scalar & squaredDistance)
+    {
+      typedef typename Visitor::LieGroupMap LieGroupMap;
+      typename LieGroupMap::template operation<JointModel>::type lgo;
+      squaredDistance += lgo.squaredDistance(jmodel.jointConfigSelector(q0.derived()),
+                                                 jmodel.jointConfigSelector(q1.derived()));
```
jcarpent: Can you correctly align the code here?

gabrielebndn:
Ok

src/algorithm/joint-configuration.hxx
```diff
@@ -9,226 +9,173 @@
...
 
-  template<typename LieGroup_t, typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorIn1, typename ConfigVectorIn2>
-  inline typename PINOCCHIO_EIGEN_PLAIN_TYPE(ConfigVectorIn1)
+  template<typename LieGroup_t, typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorIn1, typename ConfigVectorIn2, typename ReturnType>
+  void
   difference(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
              const Eigen::MatrixBase<ConfigVectorIn1> & q0,
-             const Eigen::MatrixBase<ConfigVectorIn2> & q1)
+             const Eigen::MatrixBase<ConfigVectorIn2> & q1,
+             const Eigen::MatrixBase<ReturnType> & dqout)
```
jcarpent: This is not a dqout but rather a dv

src/algorithm/joint-configuration.hpp
```diff
@@ -41,7 +230,228 @@ namespace pinocchio
...
+                      const Eigen::MatrixBase<ConfigVectorIn2> & q2,
+                      const Scalar & prec);
+
+  /**
+   * @brief         Return true if the given configurations are equivalents
+   * @warning       Two configurations can be equivalent but not equally coefficient wise (e.g for quaternions)
+   *
+   * @param[in]     model     Model
+   * @param[in]     q1        The first configuraiton to compare
+   * @param[in]     q2        The Second configuraiton to compare
```
jcarpent: second and not Second.

src/algorithm/joint-configuration.hpp
```diff
@@ -41,7 +230,228 @@ namespace pinocchio
...
+  isSameConfiguration(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                      const Eigen::MatrixBase<ConfigVectorIn1> & q1,
+                      const Eigen::MatrixBase<ConfigVectorIn2> & q2,
+                      const Scalar & prec);
+
+  /**
+   * @brief         Return true if the given configurations are equivalents
+   * @warning       Two configurations can be equivalent but not equally coefficient wise (e.g for quaternions)
+   *
+   * @param[in]     model     Model
```
jcarpent: Model of the system

src/algorithm/joint-configuration.hpp
```diff
@@ -41,7 +230,228 @@ namespace pinocchio
...
+  inline Scalar
+  distance(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+           const Eigen::MatrixBase<ConfigVectorIn1> & q0,
+           const Eigen::MatrixBase<ConfigVectorIn2> & q1)
+  {
+    return distance<LieGroupMap,Scalar,Options,JointCollectionTpl,ConfigVectorIn1,ConfigVectorIn2>(model, q0.derived(), q1.derived());
+  }
+
+  /**
+   * @brief         Normalize a configuration
```
jcarpent: Normalize a configuration vector

src/algorithm/joint-configuration.hpp
```diff
@@ -6,25 +6,214 @@
...
+  void
+  difference(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+             const Eigen::MatrixBase<ConfigVectorIn1> & q0,
+             const Eigen::MatrixBase<ConfigVectorIn2> & q1,
+             const Eigen::MatrixBase<ReturnType> & dqout);
+
+  /**
+   * @brief      Compute the tangent vector that must be integrated during one unit time to go from q0 to q1
+   *
+   * @param[in]  model   Model to be differenced
```
jcarpent: `Model to be differenced` is not clear

src/algorithm/joint-configuration.hpp
```diff
@@ -6,25 +6,214 @@
...
-            const Eigen::MatrixBase<TangentVectorType> & v);
-  
+            const Eigen::MatrixBase<TangentVectorType> & v,
+            const Eigen::MatrixBase<ReturnType> & qout)
+  {
+    integrate<LieGroupMap,Scalar,Options,JointCollectionTpl,ConfigVectorType,TangentVectorType,ReturnType>(model, q.derived(), v.derived(), qout.derived());
+  }
+
+  /**
+   * @brief      Interpolate the model between two configurations
```
jcarpent: The model is not interpolated

src/algorithm/joint-configuration.hpp
```diff
@@ -6,25 +6,214 @@
...
+   */
+  template<typename LieGroup_t, typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType, typename TangentVectorType, typename ReturnType>
+  void
+  integrate(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+            const Eigen::MatrixBase<ConfigVectorType> & q,
+            const Eigen::MatrixBase<TangentVectorType> & v,
+            const Eigen::MatrixBase<ReturnType> & qout);
+
+  /**
+   * @brief      Integrate a configuration for the specified model for a tangent vector during one unit time
```
jcarpent: a configuration vector seems better

src/algorithm/joint-configuration.hpp
```diff
@@ -41,7 +230,228 @@ namespace pinocchio
                   const Eigen::MatrixBase<TangentVectorType> & v,
                   const Eigen::MatrixBase<JacobianMatrixType> & J,
                   const ArgumentPosition arg);
-  
+
+  /**
+   * @brief      Computes the Jacobian of a small variation of the configuration vector or the tangent vector into the tangent space at identity.
+   *
+   * @param[in]  model   Model that must be integrated
+   * @param[in]  q       Initial configuration (size model.nq)
```
jcarpent: Joint configuration

src/algorithm/joint-configuration.hpp
```diff
@@ -41,7 +230,228 @@ namespace pinocchio
...
+   * @param[in]  v       Velocity (size model.nv)
```
jcarpent: Joint velocity

src/algorithm/joint-configuration.hpp
```diff
@@ -251,6 +251,38 @@ namespace pinocchio
...
+   * @brief      Overall squared distance between two configuration vectors
+   *
+   * @param[in]  model      Model we want to compute the distance
+   * @param[in]  q0         Configuration 0 (size model.nq)
+   * @param[in]  q1         Configuration 1 (size model.nq)
+   *
+   * @return     The squared distance between the two configurations
+   */
+  template<typename LieGroup_t, typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorIn1, typename ConfigVectorIn2>
+  Scalar squaredDistanceSum(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
```
jcarpent: I don't this version is really worthy

jcarpent: @gabrielebndn Do you have ideas of when this function might be useful?
Maybe @jmirabel has some hints.

gabrielebndn:
Well, it's just that to compute distance you need to compute the sum of
squared distances and finally take the squared root... One might want to
avoid the square root, so I gave this option, much in the same way we
sometimes use squaredNorm() instead of norm().

Il giorno ven 15 feb 2019 alle 10:13 Justin Carpentier <
notifications@github.com> ha scritto:

jcarpent:
Yes sure. I'm wondering if someone will use it or not. As it is done now, keep it like.
My only wish is to not add two much signatures to avoid the code becoming a mess.

gabrielebndn:
I've implemented the requested changes. I will wait for the resolution of  #681 and rebase on that

jcarpent:
Thanks for these modifications.
I will proceed to the merge before, as I think it will be easier and the PR are independent.

jcarpent:
@gabrielebndn Is it fine for you?

gabrielebndn:
@jcarpent as I had already rebased on #673, I checked everything is fie and I pushed it. Anyhow I am not getting the compilation error referenced in #681

Issue #683, opened 2019-02-16
Why do we have the JointData::F variable?


proyan:
This is just a small question. Why do we have the F variable inside some JointData classes? I tried to look for the use, but couldn't find one. Perhaps I haven't been thorough enough.

Is it useful?

jcarpent:
We don't need it at all. It has been implemented at the origin of the project.
It should be removed in a future release.

proyan:
Ok. Thanks. I'll close this.

PR #684, opened 2019-02-16
Create python objects for JointData classes

proyan:
This PR primarily creates the python bindings for the joint data classes.

In order to achive that, there are the following changes inside the JointData structs:
1) adding a shortname/classname visitor and function
2) adding a default constructor for JointDataComposite
3) create an ostream operator for making the classes printable

The rest of the PR deals with the python bindings of the JointData variants.
I don't expose JointData::F, since I didn't really understand why it is there. I'll add it if #683 is resolved that way.

### DIFF
--- a/bindings/python/multibody/joint/joints-datas.hpp
+++ b/bindings/python/multibody/joint/joints-datas.hpp
@@ -5,8 +5,7 @@
 #ifndef __pinocchio_python_joints_datas_hpp__
 #define __pinocchio_python_joints_datas_hpp__
 
-#include <eigenpy/exception.hpp>
-#include <eigenpy/eigenpy.hpp>
+#include<boost/python.hpp>
 
 namespace pinocchio
 {
--- a/bindings/python/multibody/joint/joints-datas.hpp
+++ b/bindings/python/multibody/joint/joints-datas.hpp
@@ -7,8 +7,6 @@
 
 #include <eigenpy/exception.hpp>
 #include <eigenpy/eigenpy.hpp>
-//#include "pinocchio/multibody/joint/joint-collection.hpp"
-//#include "pinocchio/multibody/joint/joint-composite.hpp"
 
 namespace pinocchio
 {
@@ -24,6 +22,41 @@ namespace pinocchio
       return cl;
     }
 
+    // specialization for JointDataRevoluteUnaligned
+    template<>
+    inline bp::class_<JointDataRevoluteUnaligned>& expose_joint_data<JointDataRevoluteUnaligned> (bp::class_<JointDataRevoluteUnaligned> & cl)
+    {
+      return cl
+        .def(bp::init<Eigen::Vector3d> (bp::args("axis"), "Init JointDataRevoluteUnaligned from an axis with x-y-z components"))
+        ;
+    }
+
+    // specialization for JointDataPrismaticUnaligned
+    template<>
+    inline bp::class_<JointDataPrismaticUnaligned>& expose_joint_data<JointDataPrismaticUnaligned> (bp::class_<JointDataPrismaticUnaligned> & cl)
+    {
+      return cl
+        .def(bp::init<Eigen::Vector3d> (bp::args("axis"), "Init JointDataPrismaticUnaligned from an axis with x-y-z components"))
+               ;
+    }
+
+    template<>
+    inline bp::class_<JointDataPlanar>& expose_joint_data<JointDataPlanar> (bp::class_<JointDataPlanar> & cl)
+    {
+      return cl
+        .add_property("StU",&JointDataPlanar::StU)
+        ;
+    }
+
+
+    template<>
+    inline bp::class_<JointDataSphericalZYX>& expose_joint_data<JointDataSphericalZYX> (bp::class_<JointDataSphericalZYX> & cl)
+    {
+      return cl
+        .add_property("StU",&JointDataSphericalZYX::StU)
+        ;
+    }
+    
     template<>
     inline bp::class_<JointDataComposite>& expose_joint_data<JointDataComposite> (bp::class_<JointDataComposite> & cl)
     {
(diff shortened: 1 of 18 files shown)


bindings/python/multibody/joint/joints-datas.hpp
```diff
@@ -0,0 +1,79 @@
+//
+// Copyright (c) 2019 CNRS
+//
+
+#ifndef __pinocchio_python_joints_datas_hpp__
+#define __pinocchio_python_joints_datas_hpp__
+
+#include <eigenpy/exception.hpp>
+#include <eigenpy/eigenpy.hpp>
+//#include "pinocchio/multibody/joint/joint-collection.hpp"
```
jcarpent: You can remove these lines.

proyan:
I just need the classname and shortname to make the definition of bp classes automatic. This is the same strategy that was used for the jointmodel classes as well.

jcarpent:
Ok. Sounds good.
My only concerns are then with the commented include.

proyan:
Pickling only creates a copy of the elements in the __dict__. Thus, if the jointdata classes are not present inside the class members, it would not be pickled. The copy of pinocchio data that will be created after pickling and unpickling will be useless for any future computations then, you will have to find a way to copy the members which were pickled into any new class that you create.

bindings/python/multibody/joint/joints-datas.hpp
proyan: Done.

bindings/python/multibody/joint/joints-datas.hpp
```diff
@@ -0,0 +1,77 @@
+//
+// Copyright (c) 2019 CNRS
+//
+
+#ifndef __pinocchio_python_joints_datas_hpp__
+#define __pinocchio_python_joints_datas_hpp__
+
+#include <eigenpy/exception.hpp>
```
jcarpent: They are not needed in fact. Only boos/python.hpp is requested here

bindings/python/multibody/joint/joints-datas.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+namespace pinocchio
+{
+  namespace python
+  {
+    namespace bp = boost::python;
+
+
+    // generic expose_joint_data : do nothing special
+    template <class T>
```
jcarpent: I would prefer to have struct specialization here, as the rest of the bindings.
Indeed they are more flexible and we allows to expose other Scalar types in the very near future.

bindings/python/multibody/joint/joints-variant.hpp
```diff
@@ -31,7 +33,23 @@ namespace pinocchio
       }
     };
 
-    struct exposer
```
jcarpent: Please use the naming convention used in the rest of the projet.

src/multibody/joint/joint-composite.hpp
```diff
@@ -73,6 +73,16 @@ namespace pinocchio
     typedef container::aligned_vector<JointDataVariant> JointDataVector;
     
     // JointDataComposite()  {} // can become necessary if we want a vector of JointDataComposite ?
+
+    JointDataCompositeTpl()
+    : joints()
+    , iMlast(0), pjMi(0)
+    , S(0)
+    , M(),v(),c()
+    , U(0,0), Dinv(0,0), UDinv(0,0)
```
jcarpent: The dimensions are not all equal to 0 here.

jcarpent:
By the way, why do you need to expose all these methods exactly?
Do you know that you accces though them the JointDataGeneric?

bindings/python/multibody/joint/joints-datas.hpp
proyan: I don't understand what you mean. There is the struct exposer which calls the expose_joint_data. (and a similar one which calls the expose_joint_model. Would you prefer to move the expose_joint_* within the structs?

proyan:
We have to create a fully accessible state for each joint (such that the class can be recreated in python from all its members). This is needed in order to not destroy the classes after pickling. That is why I am exposing all the members of a JointData class

bindings/python/multibody/joint/joints-datas.hpp
jcarpent: You have done function specialization, but if we want to go towards genericity and being able to also expose code with automatic differentiation, we will need to use struct specialization which is less limited than function specialization.
I will take in charge this change latter inside another PR.

jcarpent:
@proyan Can you address the small remaining remarks that I've made?
Thanks a lot in advance.

bindings/python/multibody/joint/joints-datas.hpp
proyan: Done.

bindings/python/multibody/joint/joints-variant.hpp
proyan: I am using DataExposer and ModelExposer now.

bindings/python/multibody/joint/joints-datas.hpp
proyan: Thanks

src/multibody/joint/joint-composite.hpp
proyan: What would you prefer? I am using nv=0 as the default, since it is the same as the default JointModelComposite constructor.

jcarpent: For me, UDinv and U are of dimension (6,0)

proyan: Ah yes. ofcourse. I'll fix this.

proyan: Done.

jcarpent:
Ok. Sounds great now. I'm going to merge it now.

Issue #681, opened 2019-02-15
examples/overview-simple.cpp does not compile


proyan:
After the recent commits, examples/overview-simple.cpp is not compiling anymore.
I get the error: fatal error: cppad/cppad.hpp: No such file or directory

I believe cppad needs to be linked.

jcarpent:
@proyan Thanks for raising this issue.
I found the bug, I will fix it asap.

jcarpent:
Fixed by #688.

PR #688, opened 2019-02-20
Mostly fix some warnings and adjust code to work on APPLE

jcarpent:


src/multibody/model.hpp
```diff
@@ -220,7 +224,11 @@ namespace pinocchio
         res.frames[k] = frames[k].template cast<NewScalar>();
       }
       
+      /// TODO: remove this pragma when neutralConfiguration will be removed
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
       return res;
+#pragma GCC diagnostic pop
```
gabrielebndn: Why here? Is `return res;` causing warnings?

jcarpent: Maybe this one is not useful. You're certainly right.

jcarpent:
Related to #687 and fixes #681.

PR #691, opened 2019-02-20
Deprecate neutralConfiguration in Python

gabrielebndn:
`neutralConfiguration` is now deprecated in Python too. To achieve this, I exposed it as `_neutralConfiguration` and then I created a `neutralConfiguration` property raising the warning.


jcarpent:
Seems good to me. I will merge it.

PR #693, opened 2019-02-21
Define contact-dynamics and minor cleanings

jcarpent:


jcarpent:
I will merge. This is only a minor refactoring.

Issue #694, opened 2019-02-21
Eigen minimum version


jmirabel:
Currently, the minimum version is 3.0.4 and it is not correct. I know that 3.2.4 does not work and 3.2.92 does. Is it fine to update the minimum version to 3.2.92 (ubuntu 16.04) ?

nim65s:
I think Pinocchio is still working on 14.04 with eigen 3.2.0

jmirabel:
@ksyy had this error with 3.2.4:
```
/usr/localDev/HPP/install/include/pinocchio/multibody/joint/joint-revolute.hpp:157:63:
error: could not convert ‘Eigen::DenseBase<Derived>::Zero(Eigen::DenseBase<Derived>::Index)
  [with Derived = Eigen::CwiseNullaryOp<Eigen::internal::scalar_constant_op<double>, Eigen::Matrix<double, 3, 1> >;
           Eigen::DenseBase<Derived>::ConstantReturnType = Eigen::CwiseNullaryOp<Eigen::internal::scalar_constant_op<double>, Eigen::CwiseNullaryOp<Eigen::internal::scalar_constant_op<double>, Eigen::Matrix<double, 3, 1> > >;
            typename Eigen::internal::traits<T>::Scalar = double;
            Eigen::DenseBase<Derived>::Index = long int]()’
from ‘const ConstantReturnType
  {aka const Eigen::CwiseNullaryOp<Eigen::internal::scalar_constant_op<double>, Eigen::CwiseNullaryOp<Eigen::internal::scalar_constant_op<double>, Eigen::Matrix<double, 3, 1> > >}’
to ‘pinocchio::TransformRevoluteTpl<double, 0, 0>::LinearType
  {aka Eigen::CwiseNullaryOp<Eigen::internal::scalar_constant_op<double>, Eigen::Matrix<double, 3, 1> >}’
     LinearType translation() const { return LinearType::Zero(3); };
```

nim65s:
https://travis-ci.org/stack-of-tasks/pinocchio/builds/496407364 : job .1 is trusty, and job .2 is xenial,

jmirabel:
Then, likely, there is one function which is not unit-tested.

jcarpent:
The minimal version is 3.0.5.

jmirabel:
@jcarpent Then, there is a bug at `multibody/joint/joint-revolute.hpp:157`

jmirabel:
[...]

PR #695, opened 2019-02-21
joints: fix return type of translation() or rotation() methods

jcarpent:
Related to #694.

### DIFF
--- a/src/multibody/joint/joint-revolute.hpp
+++ b/src/multibody/joint/joint-revolute.hpp
@@ -154,7 +154,7 @@ namespace pinocchio
     void setValues(const OtherScalar & sin, const OtherScalar & cos)
     { m_sin = sin; m_cos = cos; }
 
-    LinearType translation() const { return LinearType::Zero(3); };
+    LinearType translation() const { return LinearType::PlainObject::Zero(3); };
     AngularType rotation() const {
       AngularType m(AngularType::Identity(3));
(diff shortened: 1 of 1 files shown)


jcarpent:
@jmirabel This patch should fix your issue. Please let me know if it works.

jcarpent:
To make this fix consistent with time, it would be necessary to provide some unit tests to check that compiles well. I don't have time to do it currently, but at least if this patch is working on your setup, this will be fine.

@jmirabel Can you provide a case where you are calling this method (translation()), because externally, this should not happen, except if you call a joint-generic.

jmirabel:
I don't have time to fully validate it neither.

I did a small test. I don't think this change will fix the issue as the error is not on the constness but on the type itself.
```cpp
#include <Eigen/Core>

using namespace Eigen;

typedef typename Vector3d::ConstantReturnType LinearType;

// This does not work
// LinearType foo () { return LinearType::Zero (3); }
// This works
LinearType foo () { return LinearType::PlainObject::Zero (3); }

int main ()
{
  Vector3d v = foo();
}

```

jcarpent:
Ok. I will set this case then.

jcarpent:
I will merge it now. @jmirabel Thanks for this report, hoping it will be helpful but I'm not sure as it seems to be a bug from Eigen side.

PR #696, opened 2019-02-21
Update to new hpp-fcl namespace wrapping

olivier-roussel:
Namespaces in hpp-fcl are now encapsulated inside hpp namespace (hpp-fcl commit 788afa0d )

### DIFF
--- a/bindings/python/multibody/fcl/collision-geometry.hpp
+++ b/bindings/python/multibody/fcl/collision-geometry.hpp
@@ -23,7 +23,7 @@ namespace pinocchio
       
       struct CollisionGeometryPythonVisitor : public bp::def_visitor<CollisionGeometryPythonVisitor>
       {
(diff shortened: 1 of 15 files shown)


jcarpent:
Good to see you here.
I think this PR is worth to combine with `topic/fcl-release` branch. Can you rebase on it or see with @jmirabel?

nim65s:
I think that there is no issue with the hpp-fcl namespace, and that #645 was enough, especially with the `namespace fcl = hpp::fcl;`in src/multibody/fcl.hpp.

nim65s:
But yes, @olivier-roussel you should use the `topic/fcl-release` branch if you want to use hpp-fcl v1.0+

olivier-roussel:
My bad, trying to catch your workflow. Thanks, I guess we can close the PR.

PR #698, opened 2019-02-22
Align Python Jacobian to C++, fix #545

### LINKED ISSUE #545: [jointJacobian] different behavior between C++ and python
Python method `jointJacobian` DOES NOT call its corresponding C++ function. Instead, it internally calls `computeJointJacobians` followed by `getJointJacobian`. This causes three differences in the behavior:
- first, the unimportant, obvious ones: at the end of the python method, the overall Jacobian is computed, while at the end of the C++ method it is not. Also, the python method defeats the purpose of being more efficient, but who cares about efficiency in python
- second: the python method allows to specify a Reference frame (LOCAL/WORLD), while the C++ method does not and only offers the local version.
- third: `computeJointJacobians` internally does the equivalent of `forwardKinematics`, which C++ `jointJacobian` does not. This is particularly bad, because the doxygen C++ comment to `computeJointJacobians(model, data, q)` explicitly states:
> This function computes also the forwardKinematics of the model.

while the one for `jointJacobian` says:
> This function is equivalent to call first computeJacobians(model,data,q) and then call getJointJacobian\<LOCAL\>(model,data,jointId,J).

but this is not true. If you think this is not relevant, well, it was causing problems in one example Nicolas prepared in PR #541.
Should this be fixed? In which way? Or, should we leave things as they are in order not to break everybody's code, but explicitly document the difference in behavior?

gabrielebndn:
Mostly four things:
- Add a benchmark called `timings-jacobian` to compare between `getJointJacobian`, `computeJointJacobians` and related algorithms.
- Use the corresponding C++ function for the Python bindings of `jointJacobian`. Now, the signature and side-effects of `jointJacobian` in Python are completely aligned to C++. The old signature is now deprecated. This fixes #545
- Implemented missing `frameJacobian` function in C++, and use the newly-created function for the Python bindings of `frameJacobian`. The old signature is now deprecated. This also fixes #545
- Change the C++ return policy `getFrameVelocity` and `getFrameAcceleration`. Now, the value is returned instead of being written in a reference

### DIFF
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -75,3 +75,7 @@ ADD_BENCH(timings-eigen)
 IF(URDFDOM_FOUND AND HPP_FCL_FOUND)
   ADD_BENCH(timings-geometry TRUE)
 ENDIF(URDFDOM_FOUND AND HPP_FCL_FOUND)
+
+# timings-jacobian
+#
+ADD_BENCH(timings-jacobian TRUE)
--- a/src/algorithm/frames.hpp
+++ b/src/algorithm/frames.hpp
@@ -67,15 +67,78 @@ namespace pinocchio
    * @param[in]  model       The kinematic model
    * @param[in]  data        Data associated to model
    * @param[in]  frame_id    Id of the operational Frame
+   *
+   * @return     The spatial velocity of the Frame expressed in the coordinates Frame.
+   *
+   * @warning    Fist or second order forwardKinematics should have been called first
+   */
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
+  inline MotionTpl<Scalar, Options>
+  getFrameVelocity(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                   const DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                   const typename ModelTpl<Scalar,Options,JointCollectionTpl>::FrameIndex frame_id);
+
+  /**
+   * @brief      Returns the spatial velocity of the frame expressed in the LOCAL frame coordinate system.
+   *             You must first call pinocchio::forwardKinematics to update placement and velocity values in data structure.
+   *
+   * @param[in]  model       The kinematic model
+   * @param[in]  data        Data associated to model
+   * @param[in]  frame_id    Id of the operational Frame
    * @param[out] frame_v     The spatial velocity of the Frame expressed in the coordinates Frame.
+   *
+   * @deprecated This function is now deprecated. Use the return-value version instead (since: 19 feb 2019)
    *
    * @warning    Fist or second order forwardKinematics should have been called first
    */
   template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename MotionLike>
-  void getFrameVelocity(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
-                        const DataTpl<Scalar,Options,JointCollectionTpl> & data,
-                        const typename ModelTpl<Scalar,Options,JointCollectionTpl>::FrameIndex frame_id,
-                        const MotionDense<MotionLike> & frame_v);
+  PINOCCHIO_DEPRECATED
+  inline void getFrameVelocity(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                               const DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                               const typename ModelTpl<Scalar,Options,JointCollectionTpl>::FrameIndex frame_id,
+                               const MotionDense<MotionLike> & frame_v)
+  {
+    frame_v.derived() = getFrameVelocity(model, data, frame_id);
+  }
+
+  /**
+   * @brief      Returns the spatial acceleration of the frame expressed in the LOCAL frame coordinate system.
+   *             You must first call pinocchio::forwardKinematics to update placement values in data structure.
+   *
+   * @param[in]  model       The kinematic model
+   * @param[in]  data        Data associated to model
+   * @param[in]  frame_id    Id of the operational Frame
+   *
+   * @return The spatial acceleration of the Frame expressed in the coordinates Frame.
+   *
+   * @warning    Second order forwardKinematics should have been called first
+   */
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
+  inline MotionTpl<Scalar, Options>
+  getFrameAcceleration(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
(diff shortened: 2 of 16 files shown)


src/algorithm/frames.hpp
```diff
@@ -69,16 +87,61 @@ namespace pinocchio
...
-                        const DataTpl<Scalar,Options,JointCollectionTpl> & data,
-                        const typename ModelTpl<Scalar,Options,JointCollectionTpl>::FrameIndex frame_id,
-                        const MotionDense<MotionLike> & frame_v);
+  PINOCCHIO_DEPRECATED
+  inline void getFrameVelocity(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                               const DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                               const typename ModelTpl<Scalar,Options,JointCollectionTpl>::FrameIndex frame_id,
+                               const MotionDense<MotionLike> & frame_v)
+  {
+    const_cast<MotionLike &>(frame_v.derived()) = getFrameVelocity(model, data, frame_id);
```
jcarpent: You don’t have to do a const_cast here

src/algorithm/frames.hpp
```diff
@@ -69,16 +87,61 @@ namespace pinocchio
...
+   *
+   * @warning    Second order forwardKinematics should have been called first
+   */
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename MotionLike>
+  PINOCCHIO_DEPRECATED
+  inline void getFrameAcceleration(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                                   const DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                                   const typename ModelTpl<Scalar,Options,JointCollectionTpl>::FrameIndex frame_id,
+                                   const MotionDense<MotionLike> & frame_a)
+  { const_cast<MotionLike &>(frame_a.derived()) = getFrameAcceleration(model, data, frame_id); }
```
jcarpent: The same apply here

benchmark/CMakeLists.txt
```diff
@@ -75,3 +75,7 @@ ADD_BENCH(timings-eigen)
 IF(URDFDOM_FOUND AND HPP_FCL_FOUND)
   ADD_BENCH(timings-geometry TRUE)
 ENDIF(URDFDOM_FOUND AND HPP_FCL_FOUND)
+
+# timings-jacobian
+#
+ADD_BENCH(timings-jacobian TRUE)
```
jcarpent: Jacobiand are already bench in the main timings file.
Maybe it is not needed to over populate this directory

src/algorithm/frames.hpp
```diff
@@ -69,16 +87,61 @@ namespace pinocchio
...
+   * @param[in]  data        Data associated to model
+   * @param[in]  frame_id    Id of the operational Frame
+   *
+   * @return The spatial acceleration of the Frame expressed in the coordinates Frame.
+   *
+   * @warning    Second order forwardKinematics should have been called first
+   */
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
+  inline MotionTpl<Scalar, Options>
+  getFrameAcceleration(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
```
jcarpent: Why did you decide to change this return policy? 
You should keep both I think.

src/algorithm/frames.hpp
gabrielebndn: The previous implementation had `const_cast`... You mean I should remove `const` the `frame_v` argument? It does seem unnecessary... Or, leave the signature like this and just remove the `const_cast`, would that work?

benchmark/CMakeLists.txt
gabrielebndn: That benchmark only has `computeJointJacobians(model,data,q)` along with many unrelated algorithms. The purpose of this benchmark is specifically comparing the speed of the different ways of computing the Jacobian (mostly, `computeJointJacobians` + `getJointJacobian` vs `jointJacobian`).
It may have meaningful results. For instance, on my machine, it seems `forwardKinematics` + `jointJacobian` is already a bit slower than `computeJointJacobians` (which also computes `forwardKinematics` as a side effect) + `getJointJacobian`. In other words, if, other than the Jacobian, you also need the result of `forwardKinematics`, then it is always better to use `computeJointJacobians`.
I may remove it if you want, but I think it is useful

src/algorithm/frames.hpp
jcarpent: Yes, it should work without the const_cast.

benchmark/CMakeLists.txt
jcarpent: Yes, you can not compare `forwardKinematics` + `jointJacobian` vs `computeJointJacobians` because this is not the same purpose. In general, we do not need at all `jointJacobian`.
But these results are already known.

jcarpent:
By the way, when you commit, can you add before the commit message which topic you are then considering (like algo, spatial, ...). It is much more readable and this is also the main way of committing on Pinocchio.

src/algorithm/frames.hpp
gabrielebndn: Isn't that completely redundant? The return-value-as-argument for this method was my original implementation, but I think it was misguided. In truth, instantiation of a Motion object does not in general dynamically allocate memory, as it only holds a fixed-size 6-element matrix vector. Also, with return-value optimization, I think `getFrameVelocity(model,data,v)` should generate more or less the same code as `v = getFrameVelocity(model,data)`. I see no reason for keeping this signature.

gabrielebndn:
> By the way, when you commit, can you add before the commit message which topic you are then considering (like algo, spatial, ...). It is much more readable and this is also the main way of committing on Pinocchio.

You are totally right, I did it for the first 3 commits but then I forgot for the rest. I can easily edit the commit messages

jcarpent:
@gabrielebndn It will be fine for now concerning the commit messages. I just need you to remove the explicit const_cast.

gabrielebndn:
@jcarpent I have really no means to work on it right now (traveling). I will be home this late afternoon, I'll try to fix it then

jcarpent:
@gabrielebndn No problem. It was just to summarize the remaining tasks to perform before merging.

jcarpent:
@gabrielebndn It would be nice if we can add this feature to the new upcoming release of Pinocchio. Will you have time to deal with it today?

gabrielebndn:
I am working on it. I am not on my own computer, so I need to install pretty much everything

gabrielebndn:
I have removed the explicit `const_cast` and I have fixed the commit messages. Also, if you want to remove the Jacobian benchmark, you can just discard the last commit

jcarpent:
@gabrielebndn Thank you very much for this contribution.

Issue #720, opened 2019-03-06
Error about omniORB.CORBA.BAD_PARAM?


littleggghost:
Hi, there
I test the tutorial in [4.1) Direct dynamics](http://projects.laas.fr/gepetto/doc/stack-of-tasks/pinocchio/master/md_doc_d-practical-exercises_4-dyn.html)
and got an error:
```
Traceback (most recent call last):
  File "exercise_4_1_move.py", line 15, in <module>
    robot.display(robot.q0)
  File "/home/xxx/pinocchio/doc/d-practical-exercises/src/robot_hand.py", line 316, in display
    visual.place(self.viewer, self.data.oMi[visual.jointParent])
  File "/home/xxx/pinocchio/doc/d-practical-exercises/src/robot_hand.py", line 36, in place
    display.place(self.name, oMbody, False)
  File "/home/xxx/pinocchio/doc/d-practical-exercises/src/display.py", line 53, in place
    se3ToXYZQUAT(M))
  File "/opt/openrobots/lib/python2.7/site-packages/gepetto/corbaserver/graphical_interface_idl.py", line 393, in applyConfiguration
    return _omnipy.invoke(self, "applyConfiguration", _0_gepetto.corbaserver.GraphicalInterface._d_applyConfiguration, args)
omniORB.CORBA.BAD_PARAM: CORBA.BAD_PARAM(omniORB.BAD_PARAM_WrongPythonType, CORBA.COMPLETED_NO)
```
This error **not** happens with v2.0.0, and occurs with v2.1.0.

jmirabel:
Related to https://github.com/stack-of-tasks/pinocchio/pull/702.

Try to replace `se3ToXYZQUAT(M)` by `se3.se3ToXYZQUATtuple(M)`.

littleggghost:
Thanks. I think some docs should be updated.

jcarpent:
@jmirabel Thanks for this quick reply.
@littleggghost Thanks for raising this issue.
I'm fixing it on the devel branch.

PR #721, opened 2019-03-06
Fix doc issues related to new gepetto-viewer API modifications

jcarpent:
Related to #720

### DIFF
--- a/bindings/python/pinocchio/utils.py
+++ b/bindings/python/pinocchio/utils.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2015 CNRS
+# Copyright (c) 2015-2019 CNRS INRIA
 #
 
 from __future__ import print_function
@@ -28,9 +28,13 @@ def skew(p):
     return np.matrix([[0, -z, y], [z, 0, -x], [-y, x, 0]], np.double)
 
 
-se3ToXYZQUAT = pin.se3ToXYZQUAT
+@deprecated('Now useless. You can directly have access to this function from the main scope of Pinocchio')
+def se3ToXYZQUAT(M):
+    return pin.se3ToXYZQUATtuple(M)
 
-XYZQUATToSe3 = pin.XYZQUATToSe3
+@deprecated('Now useless. You can directly have access to this function from the main scope of Pinocchio')
+def XYZQUATToSe3(vec):
+    return pin.XYZQUATToSe3(vec)
 
 
 @deprecated('Now useless.')
--- a/doc/d-practical-exercises/src/display.py
+++ b/doc/d-practical-exercises/src/display.py
@@ -50,6 +50,6 @@ class Display():
         only at the end of the list.
         '''
         self.viewer.gui.applyConfiguration(objName,
-                                           se3ToXYZQUAT(M))
+                                           se3.se3ToXYZQUATtuple(M))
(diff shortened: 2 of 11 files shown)


doc/d-practical-exercises/src/display.py
```diff
@@ -50,6 +50,6 @@ def place(self,objName,M,refresh=True):
         only at the end of the list.
         '''
         self.viewer.gui.applyConfiguration(objName,
-                                           se3ToXYZQUAT(M))
+                                           pin.pin.oXYZQUATtuple(M))
```
gabrielebndn: I think this is a mistake

doc/d-practical-exercises/src/mobilerobot.py
```diff
@@ -1,6 +1,6 @@
-import pinocchio as se3
+import pinocchio as pin
 from pinocchio.robot_wrapper import RobotWrapper
-from pinocchio.utils import eye, se3ToXYZQUAT
+from pinocchio.utils import eye, pin.oXYZQUATtuple
```
gabrielebndn: Also here

doc/d-practical-exercises/src/mobilerobot.py
```diff
@@ -58,13 +58,13 @@ def initDisplay(self, loadModel):
     def display(self, q):
         RobotWrapper.display(self, q)
         M1 = self.data.oMi[1]
-        self.viewer.gui.applyConfiguration('world/mobilebasis', se3ToXYZQUAT(M1))
-        self.viewer.gui.applyConfiguration('world/mobilewheel1', se3ToXYZQUAT(M1))
-        self.viewer.gui.applyConfiguration('world/mobilewheel2', se3ToXYZQUAT(M1))
+        self.viewer.gui.applyConfiguration('world/mobilebasis', pin.oXYZQUATtuple(M1))
```
gabrielebndn: And here (the same error repeats below)

bindings/python/pinocchio/utils.py
```diff
@@ -27,12 +27,6 @@ def skew(p):
     x, y, z = p
     return np.matrix([[0, -z, y], [z, 0, -x], [-y, x, 0]], np.double)
 
-
-se3ToXYZQUAT = pin.se3ToXYZQUAT
```
jmirabel: I added them in #720 for backward compatibility. Shouldn't we keep them ?

By the way,  `se3ToXYZQUAT = pin.se3ToXYZQUAT` should have been  `se3ToXYZQUAT = pin.se3ToXYZQUATtuple`

gabrielebndn: I think they should be kept as deprecated

jcarpent: They are already exposed in the main scope, but we can indeed retrieve the previous behavior

jcarpent: with deprecation.

jmirabel: > They are already exposed in the main scope

Yes, but in many examples/tutorials, I see
```python
from pinocchio.utils import *
import pinocchio as pin
```
In this case, they become unavailable.

A more concise declaration would be
```python
from pinocchio import se3ToXYZQUATtuple as se3ToXYZQUAT, XYZQUATToSe3
```
but then, they wouldn't be deprecated.

PR #722, opened 2019-03-06
[SRDF]  Fix parsing of reference config + Add ability to read a SRDF stream. 

jmirabel:


unittest/srdf.cpp
```diff
@@ -60,6 +60,38 @@ BOOST_AUTO_TEST_CASE(readReferenceConfig)
...
+    "<group_state name='reference' group='all'>"
+    "     <joint name='joint'  value='0.0' />"
+    "</group_state>"
+    "</robot>";
+  
+  Model model;
+  pinocchio::urdf::buildModelFromXML(urdf, model);
+
+  std::istringstream iss (srdf);
+  pinocchio::srdf::loadReferenceConfigurations(model,iss,false);
```
jmirabel: This was failing before because unbounded revolute joints where not handled properly.

jmirabel:
> A simple thing you can add is the Python binding of this new signature

Can you easily convert a Python object into a std::istream object ? If not, then it is not possible without having a different name.

jcarpent:
See https://github.com/stack-of-tasks/pinocchio/blob/db8fe0add7d10ea31ff13aa677e30fcc3b569e42/bindings/python/parsers/parsers.hpp#L75 for further details.

jmirabel:
The functions have different names in C++. I wanted to avoid that.

PR #724, opened 2019-03-07
[Model] Total mass

gabrielebndn:
Quite often, I find myself in need of just getting the total mass of the robot.
However, there is no simple one-shot algorithm to get it in Pinocchio.
The standard today is to check `data.mass[0]`. However, the only way of having the correct value is to first call an unrelated method which fills it as a side-effect, such as `centerOfMass`, possibly with meaningless arguments. This is annoying, and somewhat error-prone.

With this PR, I add an instance method `Model.getTotalMass()`.
I implemented it as an instance method inside the `Model` class because it is very simple, I did not know in which existing header to put it, and it seemed like an overkill to create a new one just for this.

I am returning the value, instead of putting it in `data.mass[0]`. This is partly due to the choice of developing this algorithm as an instance method, and partly because, as I am not filling the other values, it seemed inconsistent to just fill the zeroth element.

### DIFF
--- a/bindings/python/algorithm/expose-com.cpp
+++ b/bindings/python/algorithm/expose-com.cpp
@@ -47,6 +47,16 @@ namespace pinocchio
     {
       using namespace Eigen;
       
+      bp::def("computeTotalMass",
+              (double (*)(const Model &))&computeTotalMass<double,0,JointCollectionDefaultTpl>,
+              bp::args("Model"),
+              "Compute the total mass of the model and return it.");
+
+      bp::def("computeTotalMass",
+              (double (*)(const Model &, Data &))&computeTotalMass<double,0,JointCollectionDefaultTpl>,
+              bp::args("Model", "Data"),
+              "Compute the total mass of the model, put it in data.mass[0] and return it.");
+
       bp::def("centerOfMass",com_0_proxy,
               bp::args("Model","Data",
                        "Joint configuration q (size Model::nq)"),
--- a/src/algorithm/center-of-mass.hpp
+++ b/src/algorithm/center-of-mass.hpp
@@ -11,6 +11,26 @@
 namespace pinocchio
 {
   
+  /// \brief Compute the total mass of the model and return it.
+  ///
+  /// \param[in] model The model structure of the rigid body system.
+  ///
+  /// \return Total mass of the model.
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
+  inline Scalar computeTotalMass(const ModelTpl<Scalar,Options,JointCollectionTpl> & model);
+
+  /// \brief Compute the total mass of the model, put it in data.mass[0] and return it.
+  ///
+  /// \param[in] model The model structure of the rigid body system.
+  /// \param[in] data The data structure of the rigid body system.
+  ///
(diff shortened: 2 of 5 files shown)


jcarpent:
Maybe @nmansard, can you give our feedback on this PR?

gabrielebndn:
I am fine with renaming it to `computeMass`. I would prefer `computeTotalMass` so that people do not think it is the mass matrix of that it has anything to do with `data.mass`.
As for making it an independent algorithm, I agree it would be preferable, but where to put it? `joint-configuration.hpp`? `center-of-mass.hpp`? It really seems inconvenient to create a separate header. I put it as an instance method because it is really so small and simple.

jcarpent:
I would go for centerOfMass.h** with the name you suggest. You can also but the result in data.mass[0].

gabrielebndn:
Fine with `center-of-mass.hpp`.
I propose overloading the algorithm: `computeTotalMass(model)` and `computeTotalMass(model,data)`, the latter filling `data.mass[0]`.
I'll wait a bit for @nmansard feedback

cmastalli:
@gabrielebndn so far I know `data.mass[0]` represents the value for the body 0 (often floating-base body). So I don't understand why you want to fills values there from `computeTotalMass`.

Additionally, Mass values (as CoMs) are intrinsic properties to a model. If you agree with that, then I right place to store this constant values is inside model class.

gabrielebndn:
@cmastalli no, `mass[i]` is the mass of the subtree supported by joint `i`, and `data.mass[0]` corresponds to the total mass of the model. The mass of the link attached to joint `i` can be found as `model.inertias[i].mass()`

gabrielebndn:
Also, I don't think computation results should be put inside `Model`.

cmastalli:
Indeed it's debatable if we should put total mass in `Model `. From one point of view, a total mass shouldn't be seen as a computed value, it's always the constant!. And just it's technically needed to computed due to URDF doesn't express it explicitly.

gabrielebndn:
@cmastalli the output of `neutral` is constant, but we don't have a `Model` variable for it (indeed, we have just deprecated `neutralConfiguration`, which was ill-defined because it was also used for values read from SRDF).
One way to do it would be to initialize the model with `0` and update it each time a body is added, but it would be harder to maintain, and it would not be robust in case somebody manually changes the values inside `model.inertias`. I *really* much prefer a separate, neat algorithm.
I do not feel the need of storing the result somewhere, but if it need be, I think it should be `data.mass[0]`. We should not put the results of an algorithm inside the `Model` class, those go inside `Data`. `data.mass[0]` is where it has always been, and I see no point in adding an instance variable to the `Data` class.

nmansard:
I agree that this should be a new algorithm, not a modification of model, and that the results should be stored in data.mass[0] and not in model. I would also vote for putting it with the COM algo.

nmansard:
@jcarpent could you add the derivatives of this algorithm as well?

jcarpent:
@nmansard Derivatives of computeTotalMass? It is independent of any configuration vector.

gabrielebndn:
I implemented the requested changes. We now have `computeTotalMass(model)` and `computeTotalMass(model,data)`, the latter filling `data.mass[0]` and returning its value.

PR #732, opened 2019-03-08
unittest/srdf: fix vector size

jcarpent:
This was causing bugs in DEBUG mode

### DIFF
--- a/unittest/srdf.cpp
+++ b/unittest/srdf.cpp
@@ -91,7 +91,37 @@ BOOST_AUTO_TEST_CASE(readReferenceConfig_stream)
   Eigen::VectorXd qexpected(1); qexpected << 0;
   BOOST_CHECK(q.size() == model.nq);
   BOOST_CHECK(q.isApprox(qexpected));
+}
+
+BOOST_AUTO_TEST_CASE(test_continuous_joint)
+{
+  const string urdf =
+  "<robot name='test'>"
+  "<link name='base_link'/>"
+  "<link name='child_link'/>"
+  "<joint type='continuous' name='joint'>"
+  "  <parent link='base_link'/>"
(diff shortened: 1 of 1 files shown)


unittest/srdf.cpp
```diff
@@ -88,9 +88,9 @@ BOOST_AUTO_TEST_CASE(readReferenceConfig_stream)
   pinocchio::srdf::loadReferenceConfigurationsFromXML(model,iss,false);
 
   Eigen::VectorXd q = model.referenceConfigurations["reference"];
-  Eigen::VectorXd qexpected (2); qexpected << 1,0;
+  Eigen::VectorXd qexpected(1); qexpected << 1;
```
jmirabel: The bug is somewhere else then. The size of q should be 2...

jcarpent: You put a revolute model, not a continuous one.

jmirabel: You are right. I though a revolute joint without bounds was a continuous joint but it is a mistake.

I meant "continuous" as it will also check the loadReferenceConfigurations handles continuous joints properly.

BTW, I am surprised this test succeeded before. qexpected should have been 0, not 1...

jmirabel: > BTW, I am surprised this test succeeded before. qexpected should have been 0, not 1...

understood:
```
BOOST_CHECK(!q.isApprox(qexpected));
```

jcarpent: Yes, I changed the value by forced-pushed.

jcarpent: I think to test the continuous case, we should add an additional test.

jcarpent: I've just done it.

Issue #700, opened 2019-02-25
Serialization of Model in Pinocchio


jcarpent:
I would like to raise the question of serialization of Model in Pinocchio.
A lot of users are asking for this feature, which seems to be common for many applications.

I suggest relying on Boost.Serialization to achieve this objective, which will require to also serialize Spatial classes.
Concerning Data, I do not think worthy to serialize such a class, because data contained in it are just temporary for computations.

What do you think?

jmirabel:
I implemented serialization of BVH objects in hpp-fcl, using Boost. It wouldn't be hard to extend it to shapes. It is very easy to serialize Eigen Matrices so I guess spatial classes won't be hard either.

I haven't been very far in comparing with other serialization libraries. It was really amazed by the easiness of use of Boost.Serialization.

olivier-stasse:
I would also be for boost::serialization, mostly because it would allow to have a clean separation with specific format. Maintaining serialization for various formats is known to be a pain.

jcarpent:
OK. Thanks for your feedback.
I will then rely on Boost.Serialization, and make explicit serialization for binaries, XML and text files.

proyan:
I have seen before that boost::serialization is not forward compatible. This was especially a problem when I serialized an object in 16.04 (boost 1.58) and deserialized in 14.04 (boost version 1.54). I can forsee that this would be a problem again when alternating between 16.04 and 18.04 (boost version 1.65). And I don't think the solution is in the pipeline. See for reference https://github.com/boostorg/serialization/issues/88

Please keep this is mind in case an issue is raised in the future because of this.

jcarpent:
@proyan Thanks for this useful feedback. But I guess you were using binaries and not text or XML?

proyan:
I was using text serialization.

jcarpent:
Solved by #734

PR #734, opened 2019-03-10
Add serialization of Model and related quantities

jcarpent:
Related to #700


jmirabel:
I suggest to think now of support for class versioning. For now, there is not much to do.
Version 0 is implied when no version is specified.
```
BOOST_CLASS_VERSION(class, 0)
```

For reference, see section *Class versioning* here: https://www.boost.org/doc/libs/1_69_0/libs/serialization/doc/tutorial.html

For instance, when Model::neutralConfiguration will be deleted, we can replaced corresponding lines by
```
if (version <= 0) {
  VectorXd neutralConfiguration;
  ar & make_nvp("neutralConfiguration",neutralConfiguration);
}
...
BOOST_CLASS_VERSION(Model, 1)
```

jcarpent:
@jmirabel Currently, with this PR, we cannot do more to support versioning. Isn't it?

jmirabel:
There is nothing to do with the code. Maybe a note for developers to make it easier someone else to update the code. For instance, the //TODO in Model serialization function could be changed to include the above code snippet.

PR #736, opened 2019-03-18
[com] computeSubtreeMasses + CoM bindings

gabrielebndn:
Related to #724.
This PR adds an algorithm called `computeSubtreeMasses` which fills the whole `data.mass` vector.
An alternative implementation would be providing an optional boolean to `computeTotalMass(model,data)`, but I find the current interface is neater and clearer.

Also, I've exposed all missing overloads of `centerOfMass`.

### DIFF
--- a/unittest/com.cpp
+++ b/unittest/com.cpp
@@ -110,6 +110,29 @@ BOOST_AUTO_TEST_CASE ( test_mass )
   BOOST_CHECK_CLOSE(data2.mass[0], mass, 1e-12);
 }
 
+BOOST_AUTO_TEST_CASE ( test_subtree_masses )
+{
+  using namespace Eigen;
+  using namespace pinocchio;
+
+  pinocchio::Model model;
+  pinocchio::buildModels::humanoidRandom(model);
+
+  pinocchio::Data data1(model);
+
+  computeSubtreeMasses(model,data1);
+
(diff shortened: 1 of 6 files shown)


unittest/com.cpp
```diff
@@ -110,6 +110,29 @@ BOOST_AUTO_TEST_CASE ( test_mass )
...
+  computeSubtreeMasses(model,data1);
+
+  pinocchio::Data data2(model);
+  VectorXd q = VectorXd::Ones(model.nq);
+  q.middleRows<4> (3).normalize();
+  centerOfMass(model,data2,q);
+
+  for(size_t i=0; i<(size_t)(model.njoints);++i)
+  {
+    BOOST_CHECK_CLOSE(data1.mass[i], data2.mass[i], 1e-12);
```
jcarpent: They should be equal here.

gabrielebndn: True. However, as a general rule, I always prefer to check for approximate equal, for two reasons
1) Not breaking my head tracking which values are supposed to be exactly equal and which ones can be just approximately equal
2) Having more robust tests, in case some implementation details  are changed
The fact that the internal calculation is exactly the same should be irrelevant to the user. Users should not rely on fractional values being identical up to the least relevant bit.

PR #739, opened 2019-03-20
Revert "[CI] work with hpp-fcl v1, which is not yet in the binary repository"

nim65s:
This reverts commit eb73311af732c6efbbbeec4eade59f37cdc03ad9, because hpp-fcl v1 is in the binary respository now.

### DIFF
--- a/.ci-deps
+++ b/.ci-deps
@@ -1,1 +1,0 @@
-path/hpp-fcl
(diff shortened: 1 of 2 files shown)


nim65s:
https://gepgitlab.laas.fr/gsaurel/pinocchio/pipelines/3690/builds :)

PR #743, opened 2019-03-22
[serialization] [python] Fix compilation bug in Python on (old) Mac OS and FreeBSD

gabrielebndn:
I had a hard time solving this.

I'm on Mac OSX 10.11 (El Capitan).
I found this fix was necessary for the serialization bindings to compile.

It originates from a known [bug](https://bugs.python.org/issue10910) in file pyport.h. My solution was inspired from [here](https://gitlab.kitware.com/vtk/vtk/commit/a4fa4480994822c957d624762c55f29a41634f21#42ca2d64cee3c4b8f3adde83c1959ecd974599cc).

However, my Mac version is pretty old and the bug is marked as "resolved". So I think it should not be a problem on newer versions.

@jcarpent, I know you often work on Mac. Which version are you on? Do you need the fix?

If the fix is needed on the latest Mac versions, then it should definitely be added.

If it is not needed on the latest Mac versions, it may still be added as a way of providing support for old versions, up to you to decide.

### DIFF
--- a/src/serialization/archive.hpp
+++ b/src/serialization/archive.hpp
@@ -18,6 +18,17 @@
 #include <boost/archive/binary_iarchive.hpp>
 #include <boost/archive/binary_oarchive.hpp>
 
+/* undo some macro defs in pyport.h */
+#if defined(_PY_PORT_CTYPE_UTF8_ISSUE)
+#undef isalnum
+#undef isalpha
+#undef islower
+#undef isspace
+#undef isupper
+#undef tolower
+#undef toupper
+#endif
+
 // Handle NAN inside TXT or XML archives
 #include <boost/math/special_functions/nonfinite_num_facets.hpp>
 


src/serialization/archive.hpp
```diff
@@ -18,6 +18,17 @@
 #include <boost/archive/binary_iarchive.hpp>
 #include <boost/archive/binary_oarchive.hpp>
 
+/* undo some macro defs in pyport.h */
```
jcarpent: This fix should not appear here if it is only related to Python compilation issues.
It may be dangerous for projects which depend on Pinocchio to under some many macros like that.

gabrielebndn: It is related to Python, but this is the only place where I found it was working. I could not just put it in the bindings. I can take a closer look and see if I can manage to change this.

jcarpent: You should be able to put them in the right place. You should also be able to active this fix for certain Python versions, not all.

gabrielebndn: I will check. By the way, do you have the problem, or not?

jcarpent: Not at all. It is passes on all the Ubuntu versions of GitLab + Travis + MacOS Mojave

jcarpent: Any news?

gabrielebndn: Sorry, haven't had time to work on it. Are you still interested in this fix?

jcarpent: Sure. If it allows using Pinocchio on a large spectrum of OS.

jcarpent: I fixed this issue from my side, I will provide a fix soon.
It is a matter of headers inclusion.

jcarpent:
Solved by #750

Issue #742, opened 2019-03-21
Fusion of Models


jmirabel:
I would like to be able to append a Model to a specific joint of another Model. I cannot think of any easy way of doing it.

Using Pinocchio, I would have to write:
- an algo like buildModel. That's the easy part.
- an algo that reorder the joints so that all the children of a joint are contiguous. This looks a bit more tricky as they are many fields to be updated, both in Model and GeomModel. it's going to be a nightmare to maintain...

I could also create a new Model and append joints to it in the right order. It seems a bit easier.

Any hints or comments on the above ?

jcarpent:
>     * an algo that reorder the joints so that all the children of a joint are contiguous. This looks a bit more tricky as they are many fields to be updated, both in Model and GeomModel. it's going to be a nightmare to maintain...

You should follow this strategy if you want Pinocchio to be efficient and it is mandatory in the context of CRBA algorithms for instance.
I would say that is better to create a new model from two existing ones. It is just a call to appendJoint method, which is pretty sufficient.

jmirabel:
For my application, it doesn't have to be efficient.

I will have a look at the second option.

jcarpent:
I would prefer to implement directly the final solution, which means having continuous ordering for directly parent joints. It will avoid additional work later and full compatibility with all the existing algo.

jmirabel:
> I would prefer to implement directly the final solution

what do you mean ?

jcarpent:
I mean, the solution of merging two models in a new model, having all the index of a given subtree represented by a continuous collection (again, this is mandatory for most dynamics algo).

jmirabel:
In which file should this be ? My first thought is `algorithm/model.hpp`.

jmirabel:
It could also go into `algorithm/copy.hpp`.

jmirabel:
It could also be a constructor of Model class. It may be a little awkward though.

gabrielebndn:
Related to #470.
A new file called `algorithm/model.hpp` could contain all algorithms related to directly manipulating the model, such as the proposed fusion of models, the extraction of subtrees, and possibly joint reordering (see #379).
My feeling though is that all these features would be best implemented as instance methods of the `Model` class.

jcarpent:
I would do it in `multibody/model.hxx`

jcarpent:
solved by #752.

PR #751, opened 2019-03-27
Inertia of universe if initialized with Zero.

jmirabel:
A long time ago, I added the following lines in the constructor of Model:
```cpp
      // Inertia of universe has no sense.
      inertias[0].mass() = std::numeric_limits<Scalar>::quiet_NaN();
      inertias[0].lever().fill (std::numeric_limits<Scalar>::quiet_NaN());
      inertias[0].inertia().fill (std::numeric_limits<Scalar>::quiet_NaN());
```

While implementing #742, *I found a sense to the inertia of the universe* :)

When append modelB to joint a of modelA, the inertia of the universe of modelB ought to be added to the inertia of joint a. Sadly, the information was lost because it was initialized with NaN values.

### DIFF
--- a/src/multibody/model.hpp
+++ b/src/multibody/model.hpp
@@ -149,7 +149,7 @@ namespace pinocchio
     , njoints(1)
     , nbodies(1)
     , nframes(0)
-    , inertias(1)
+    , inertias(1, Inertia::Zero())
     , jointPlacements(1, SE3::Identity())
     , joints(1)
     , parents(1, 0)
@@ -162,10 +162,6 @@ namespace pinocchio
       // in the list of joints ? See comment in definition of
       // Model::addJointFrame and Model::addBodyFrame
       addFrame(Frame("universe", 0, 0, SE3::Identity(), FIXED_JOINT));
-      // Inertia of universe has no sense.
-      inertias[0].mass() = std::numeric_limits<Scalar>::quiet_NaN();
-      inertias[0].lever().fill (std::numeric_limits<Scalar>::quiet_NaN());
-      inertias[0].inertia().fill (std::numeric_limits<Scalar>::quiet_NaN());
     }
     ~ModelTpl() {} // std::cout << "Destroy model" << std::endl; }
     


jmirabel:
By the way, at the moment, `Inertia::Zero() + Inertia::Zero()` gives a NaN Inertia. Is this behavior desired ?

jcarpent:
As far as I understand, you should never read this information when merging two models. Am I correct?

jmirabel:
If modelB is a manipulator arm, then there are geometries which are not moving. In Pinocchio, they are thus attached to the universe joint and their inertia is lost.
Appending modelB to any other model will then produce a wrong inertia.

I agree that, in most cases, it is not used and I can implement a workaround for this in hpp-pinocchio. So it is not blocking for me.

jcarpent:
> By the way, at the moment, `Inertia::Zero() + Inertia::Zero()` gives a NaN Inertia. Is this behavior desired ?

The sum of two Inertia with Zero entries has no meaning because there is no notion of barycenter.

jcarpent:
> If modelB is a manipulator arm, then there are geometries which are not moving. In Pinocchio, they are thus attached to the universe joint and their inertia is lost.
> Appending modelB to any other model will then produce a wrong inertia.
> 
> I agree that, in most cases, it is not used and I can implement a workaround for this in hpp-pinocchio. So it is not blocking for me.

This is fine for me. The idea of having these NaN values was for debugging purposes only.

jmirabel:
> The sum of two Inertia with Zero entries has no meaning because there is no notion of barycenter.

It isn't hard to find a meaning to the barycenter of (0,0,0), with weight 0, and (0,0,0), with weight 0...

jcarpent:
Yes, you're right. But you then need to check that two CoM are equal, which is not warranty.
Anyway, I think a special case to check whether the two inertia are Zero will slow down the computations a lot. 
We can just add an assert to be sure that the two Inertias are not equal to Zero.

jmirabel:
>  We can just add an assert to be sure that the two Inertias are not equal to Zero.

I am fine with keeping it as it is now or adding an assert.

Issue #470, opened 2018-06-06
Extracting subtree for independent computation


olivier-stasse:
Dear all,
I need to test and port old fashion control algorithm. They need to perform computation on subtree of a humanoid robot. Typically it would consist in making inverse kinematic computations of the left and right legs seperately.
I found subtrees in the structure of pinocchio but not a proper way to extract path of joints as it is done in other libraries.
Thanks for your help.

cmastalli:
I'm working in similar topic, i.e. computation of kinematic quantities (FK and IK) at the subtree level and other functions. Those functions are implemented in a library on top of Pinocchio, with the possibility of having Python bindings too. However, I'm currently testing the code (i.e. creating unittest).

The reason to implement them in an another library is due to the fact that these functions need another level of semantic description needed in motion control or planning. Just to give you a quick taste of this, imagine that you want to compute the FK of only feet branches.

This might interested topic of discussion for other people. @olivier-stasse doesn't hesitate to contact me if you're interested.

jcarpent:
@olivier-stasse There is now the notion of subtrees in Pinocchio, allowing to extract all the joints ids of this subtree. I plan to add the `forwardKinematics` of subtree in the very next future. Are you still interested by this feature?

Neotriple:
Just wanted to chime in to say that I would definitely love this feature. Pretty similar to what @cmastalli was referring to, having a subtree of the feet would be a really nice feature.

rshum19:
Hi @jcarpent ,  

I have a mobile manipulator model (robot base + robot arm). I'm trying to extract the Lagrangian dynamics for only the robot arm. Is there a way of extracting  just the robot arm as a subtree and using function like crba() and computeCoriolisMatrix()? 

thanks,
Roberto

jcarpent:
If you want to isolate the robot arm as a model, you have the function: `buildReducedModel` which does the work and then allows to reuse all the algorithms as usual.
Currently, the features are not yet available, as we are mostly lacking manpower.

rshum19:
`buildReducedModel` seem to be what I'm looking for. However, I didn't quite understand :

> Currently, the features are not yet available, as we are mostly lacking manpower.

Does this mean that the `buildReducedModel` function has been defined but not yet implemented or that the feature described to perform `forwardKinematics` in subtrees is not yet available?

jcarpent:
`buildReducedModel` works perfectly well. What I meant is that working with subtrees is not yet possible

[...]

Issue #379, opened 2017-02-28
Add option to reorder joint after parsing URDF


thomasfla:
To be compatible with other dynamic libraries or robot drivers, it would be good to be able to choose joint order after parsing URDF file, taking as input a vector of joint names.

Even if to be efficient Pinocchio needs to place parents before their child, there is still freedom in ordering the sub-chains of the tree. If the reordering asked by user doesn't respect this parent-child constraint, Pinocchio could report an error.

This need isn't a priority as memory map is possible, but so far it has been express by T. KO, G. BUONDONNO and myself.

nmansard:
I think it is a good idea. An algorithm could be implemented to re-order 
Model (no need to reorder Data, that would be re-created from the new 
model after reordering). It would also be necessary to re-order 
GeomModel, although it can be implemented in a second part of the work. 
I do not expect any blocking difficulty in implementing it, although it 
is a tedious work to have all the indexes correct afterward. I could 
supervise the implementation. Who could lead the task?

jmirabel:
For the information, the only reason why the joints are reordered is that `urdfdom` stores them in a map, which is an ordered set. They are sorted by joint name. IMHO, this is a mistake in `urdfdom`...

For the record, the problem is in `Model::initTree`, which should not loop on a map but on a vector. The fix is very easy here, but it will be a pain to get the change accepted by the package maintainers...

jcarpent:
@jmirabel The request of the feature is not only due to URDF choice of traversing tree, but depends on the target application. For instance, with the HRP-2, the joint order is not the same in the VRML file than in the URDF file. Then, you may want that in Model, the joint ordering be the same than in old application to avoid unnecessary mapping.

cmastalli:
Joint desired order could be described through a SDF.

tlemarec:
Hello, any news on this issue ?

jcarpent:
Sorry, but we are terribly lacking in manpower. Your contribution @tlemarec will be really appreciated for implementing this feature!

tlemarec:
I'll see what we can do (@davidfauthoux), our understanding of the inner workings of pinocchio is still limited

jcarpent:
Normally, this should not be so hard, knowing we already support the build of a reduced model from a given model.
For efficiency reasons, you just need to have the correct ordering, meaning that the index of a parent is always below that the ones of its children + for a given subtree, the indexes should be successive.

york-fu:
[...]

PR #752, opened 2019-03-27
[C++] Add algo appendModel to merge two Models.

jmirabel:
Solves #742 

It also:
- corrects the inertia of humanoid model,
- make Model printing function easier to read.


jmirabel:
It shouldn't be hard to tweak the code of this PR to solve #470.

src/algorithm/model.hxx
```diff
@@ -0,0 +1,211 @@
...
+
+#ifndef __pinocchio_algorithm_model_hxx__
+#define __pinocchio_algorithm_model_hxx__
+
+namespace pinocchio
+{
+  namespace details {
+    template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
+    void appendUniverseToModel (
+        const ModelTpl<Scalar,Options,JointCollectionTpl> & modelAB,
```
jcarpent: Can you align with the bracket here?

src/algorithm/model.hxx
```diff
@@ -0,0 +1,211 @@
...
+        const GeometryModel& geomModelAB,
+        FrameIndex parentFrame,
+        SE3 pfMAB, /* parent frame shift */
```
jcarpent: Missing const reference here

src/algorithm/model.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#ifndef __pinocchio_algorithm_model_hpp__
+#define __pinocchio_algorithm_model_hpp__
+
+#include "pinocchio/multibody/model.hpp"
+#include "pinocchio/multibody/geometry.hpp"
+
+namespace pinocchio
+{
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
```
jcarpent: Can you add a documentation of the algo here?

src/algorithm/model.hxx
```diff
@@ -0,0 +1,211 @@
...
+      }
+    };
+  }
+
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
+  void
+  appendModel(const ModelTpl<Scalar,Options,JointCollectionTpl> & modelA,
+              const ModelTpl<Scalar,Options,JointCollectionTpl> & modelB,
+              const GeometryModel& geomModelA,
+              const GeometryModel& geomModelB,
```
gabrielebndn: This algorithm appends models and geometry models at the same time. Not everybody who is using models also uses geometry models. Can you separate these?

src/algorithm/model.hxx
```diff
@@ -0,0 +1,211 @@
...
+              FrameIndex frameInModelA,
+              SE3 aMb,
```
gabrielebndn: I would put this as optional argument, defaulting to identity

src/algorithm/model.hxx
jmirabel: I am not gonna split this into two algorithms. It's more work and I never received any request in that direction.

If you do not have geometry models, you can pass empty models. If requested later, we can add a prototype without the two GeometryModel.

src/algorithm/model.hxx
jmirabel: I was about to do that but I changed my mind. This function is not going to be used a lot and I prefer to have the outputs as last arguments.

gabrielebndn: Yes, it sound logical to have the outputs at the end

src/algorithm/model.hxx
```diff
@@ -0,0 +1,210 @@
...
+              const GeometryModel& geomModelB,
+              FrameIndex frameInModelA,
+              const SE3Tpl<Scalar, Options>& aMb,
+              ModelTpl<Scalar,Options,JointCollectionTpl>& model,
+              GeometryModel& geomModel)
+  {
+    typedef details::AppendJointOfModelAlgoTpl<Scalar, Options, JointCollectionTpl> AppendJointOfModelAlgo;
+    typedef typename AppendJointOfModelAlgo::ArgsType ArgsType;
+    const Frame& frame = modelA.frames[frameInModelA];
+    static const SE3 id = SE3::Identity();
```
jcarpent: It is not the correct SE3 type here.

src/algorithm/model.hxx
```diff
@@ -0,0 +1,210 @@
...
+    for (JointIndex jid = 1; jid <= frame.parent; ++jid) {
+      ArgsType args (modelA, geomModelA, 0, id, model, geomModel);
+      AppendJointOfModelAlgo::run (modelA.joints[jid], args);
+    }
+
+    // Copy modelB joints
+    details::appendUniverseToModel (modelB, geomModelB,
+        model.getFrameId (frame.name, frame.type), aMb, model, geomModel);
+    for (JointIndex jid = 1; jid < modelB.joints.size(); ++jid) {
+      SE3 pMi = (jid == 1 ? frame.placement * aMb : id);
```
jcarpent: The same comment applies here.

src/algorithm/model.hxx
jmirabel: I mark this as resolved. @gabrielebndn unresolve if you disagree.

src/algorithm/model.hxx
jmirabel: Good catch.

src/algorithm/model.hxx
jmirabel: What about this one ?

src/algorithm/model.hxx
```diff
@@ -0,0 +1,212 @@
...
+              FrameIndex frameInModelA,
+              const SE3Tpl<Scalar, Options>& aMb,
+              ModelTpl<Scalar,Options,JointCollectionTpl>& model,
+              GeometryModel& geomModel)
+  {
+    typedef details::AppendJointOfModelAlgoTpl<Scalar, Options, JointCollectionTpl> AppendJointOfModelAlgo;
+    typedef typename AppendJointOfModelAlgo::ArgsType ArgsType;
+    typedef SE3Tpl<Scalar, Options> SE3;
+
+    const Frame& frame = modelA.frames[frameInModelA];
```
gabrielebndn: `Frame` should be typedefed to the good class too.

src/algorithm/model.hxx
```diff
@@ -0,0 +1,212 @@
...
+  namespace details {
+    template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
+    void appendUniverseToModel (const ModelTpl<Scalar,Options,JointCollectionTpl> & modelAB,
+                                const GeometryModel& geomModelAB,
+                                FrameIndex parentFrame,
+                                const SE3Tpl<Scalar, Options>& pfMAB,
+                                ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                                GeometryModel& geomModel)
+    {
+      const Frame& pframe = model.frames[parentFrame];
```
gabrielebndn: Here too, `Frame`

jmirabel:
It could also be extended to solve #379

src/algorithm/model.hxx
jcarpent: Yes, I agree with @gabrielebndn. It would be worthy to split it. But I do not know if you are willing to provide this additional effort?

jmirabel: I was about to propose to add a folder like `unsupported` or `wip` for stuff that has not yet reached the desired level of maturity.

I added an API without the geometry models.

jcarpent: OK. But maybe it should be put in the corpus of the doc, and maybe also as a warning during the compilation of the function, for instance, to inform the user about the status of the functionality.

jmirabel: In the documentation, it is easy to put it in a separate doc, like Eigen does.
I don't think a compilation warning is a good idea. Adding `#include <pinocchio/wip/algorithm/model.hpp>`, eventually in the namespace `pinocchio::wip::...` is explicit enough.

jcarpent:
Thanks @jmirabel for this new feature. It will be required to update it later, but for the time being, your implementation is already sufficient.

src/algorithm/model.hxx
gabrielebndn: Nope, agreed :)

src/algorithm/model.hxx
jmirabel: Another simpler option is to add a doxygen macro (called alias in Doxygen syntax) like "this_feature_is_not_mature_yet" which would either:
- put the function into a special page,
- or add a clear warning message.

jmirabel:
Thanks for having been flexible.

src/algorithm/model.hxx
gabrielebndn: If you ever do this, please do not call it wip, I can already see lots of confusion with robotpkg/wip ("it's in the wip" ^^). "unsupported" is clear enough and Eigen-like

PR #760, opened 2019-04-08
[python][bindings][centroidalDynamicsDerivatives] add bindings for dhg

proyan:


proyan:
I've seen that most of the derivative values are not stored in pinocchio data. So I've kept the same format here, the output of the binding is directly returned to python, and not saved in pinocchio data.

However, I think saving the derivative values (just like other algorithm outputs) in data would be a better option.
So in case there is motivation to store the other derivative algorithm outputs inside data as well, I would update this PR to store the centroidaldynamicsderivatives inside data.

bindings/python/algorithm/expose-centroidal-derivatives.cpp
```diff
@@ -0,0 +1,65 @@
...
+      
+      bp::def("computeCentroidalDynamicsDerivatives",
+              computeCentroidalDynamicsDerivatives_proxy,
+              bp::args("Model","Data",
+                       "q: configuration vector (size model.nq)",
+                       "v: velocity vector (size model.nv)",
+                       "a: acceleration vector (size model.nv)"),
+              "Computes the analytical derivatives of the centroidal dynamics\n"
+              "with respect to the joint configuration vector, velocity and acceleration.");
+      bp::def("getDynamicsDerivatives",
```
jcarpent: This is not the right naming here. Can you correct it?

bindings/python/algorithm/expose-centroidal-derivatives.cpp
```diff
@@ -0,0 +1,65 @@
+//
+// Copyright (c) 2019 CNRS
+//
+
+#include "pinocchio/bindings/python/algorithm/algorithms.hpp"
+#include "pinocchio/algorithm/centroidal-derivatives.hpp"
+#include <boost/python/tuple.hpp>
```
jcarpent: Can you add a blank line here?

bindings/python/algorithm/expose-centroidal-derivatives.cpp
proyan: This was by mistake. done.

bindings/python/algorithm/expose-centroidal-derivatives.cpp
proyan: Done.

Issue #759, opened 2019-04-08
Meaning of joint-configuration::dIntegrate


matthieuvigne:
I'm a bit confused about the meaning of the ```dIntegrate_dq``` and ```dIntegrate_dv``` functions. I read https://github.com/stack-of-tasks/pinocchio/issues/542, but I still feel there is a dimension error: as you write it, if

```
M_{next} = f(M, v)
```

then f takes values in SE3 (i.e. dim nq), and so does ```\dot{f}``` - however ```dIntegrate``` takes values in se3 (i.e. dim nv). From what I understand, ```dIntegrate_dq``` represents the partial derivative of f over q, and thus should have dimension nq x nq, not nv x nv.

To give some perspective, I'm trying to implement a second-order joint configuration integrate function, taking into account the acceleration. Indeed, the current integrate function performs first-order Euler integration:
<a href="https://www.codecogs.com/eqnedit.php?latex=q(t&space;&plus;&space;dt)&space;=&space;q(t)&space;&plus;&space;dt&space;\dot{q}(t)&space;=&space;integrate(q,&space;v&space;*&space;dt)" target="_blank"><img src="https://latex.codecogs.com/gif.latex?q(t&space;&plus;&space;dt)&space;=&space;q(t)&space;&plus;&space;dt&space;\dot{q}(t)&space;=&space;integrate(q,&space;v&space;*&space;dt)" title="q(t + dt) = q(t) + dt \dot{q}(t) = integrate(q, v * dt)" /></a>

I want to implement the second-order integration

<a href="https://www.codecogs.com/eqnedit.php?latex=q(t&space;&plus;&space;dt)&space;=&space;q(t)&space;&plus;&space;dt&space;\dot{q}(t)&space;&plus;&space;\frac{dt^2}{2}&space;\ddot{q}(t)" target="_blank"><img src="https://latex.codecogs.com/gif.latex?q(t&space;&plus;&space;dt)&space;=&space;q(t)&space;&plus;&space;dt&space;\dot{q}(t)&space;&plus;&space;\frac{dt^2}{2}&space;\ddot{q}(t)" title="q(t + dt) = q(t) + dt \dot{q}(t) + \frac{dt^2}{2} \ddot{q}(t)" /></a>

If ```\dot{q}(t) = g(q, v)```  then this writes

[...]

PR #761, opened 2019-04-10
Add documentation for dIntegrate function.

matthieuvigne:
Fix https://github.com/stack-of-tasks/pinocchio/issues/759

Add some mathematical insight to the documentation of ```integrate```, ```difference``` and ```dIntegrate```.

### DIFF
--- a/src/algorithm/joint-configuration.hpp
+++ b/src/algorithm/joint-configuration.hpp
@@ -17,6 +17,9 @@ namespace pinocchio
   /**
    * @brief      Integrate a configuration vector for the specified model for a tangent vector during one unit time
    *
+   * @details This function corresponds to the exponential map of the joint configuration Lie Group.
+   *          Its output can be interpreted as the "sum" from the Lie algebra to the joint configuration space \f$ q \oplus v \f$.
+   *
    * @param[in]  model   Model that must be integrated
    * @param[in]  q       Initial configuration (size model.nq)
    * @param[in]  v       Joint velocity (size model.nv)
@@ -31,6 +34,9 @@ namespace pinocchio
 
   /**
    * @brief      Integrate a configuration vector for the specified model for a tangent vector during one unit time
+   *
+   * @details This function corresponds to the exponential map of the joint configuration Lie Group.
+   *          Its output can be interpreted as the "sum" from the Lie algebra to the joint configuration space \f$ q \oplus v \f$.
    *
    * @param[in]  model   Model that must be integrated
    * @param[in]  q       Initial configuration (size model.nq)
@@ -87,6 +93,9 @@ namespace pinocchio
   /**
    * @brief      Compute the tangent vector that must be integrated during one unit time to go from q0 to q1
    *
+   * @details This function corresponds to the log map of the joint configuration Lie Group.
+   *          Its output can be interpreted as a difference from the joint configuration space to the Lie algebra \f$ q_1 \ominus q_0 \f$.
+   *
    * @param[in]  model   Model of the system
    * @param[in]  q0      Initial configuration (size model.nq)
    * @param[in]  q1      Wished configuration (size model.nq)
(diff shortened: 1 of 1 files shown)


src/algorithm/joint-configuration.hpp
```diff
@@ -17,6 +17,9 @@ namespace pinocchio
   /**
    * @brief      Integrate a configuration vector for the specified model for a tangent vector during one unit time
    *
+   * @details This function corresponds to the exponential map of the joint configuration Lie Group.
+   *          Its output can be interpreted as a sum in the Lie algebra \f$ q \oplus v \f$.
```
jcarpent: I would say here: `as the sum from the Lie algebra to the joint configuration space`

src/algorithm/joint-configuration.hpp
```diff
@@ -217,6 +229,13 @@ namespace pinocchio
   /**
    * @brief      Computes the Jacobian of a small variation of the configuration vector or the tangent vector into the tangent space at identity.
    *
+   * @details This jacobian has to be interpreted in terms of Lie group, not vector space: as such,
+   *          it is expressed in the tangent space only, not the configuration space.
+   *          Calling \f$ f(q, v) \f$ the integrate function, these jacobians satisfy the following relationships in the
+   *          tangent space:
+   *           - Jacobian relative to q: \f$ f(q \oplus \delta q, v) \ominus f(q, v) = J_q + o(\delta q)\f$.
+   *           - Jacobian relative to v: \f$ f(q, v + \delta v) \ominus f(q, v) = J_v + o(\delta v)\f$.
```
gabrielebndn: You are missing the multiplication of the Jacobians by  `\delta q`  and `\delta v` here

src/algorithm/joint-configuration.hpp
```diff
@@ -234,6 +253,13 @@ namespace pinocchio
   /**
    * @brief      Computes the Jacobian of a small variation of the configuration vector or the tangent vector into the tangent space at identity.
    *
+   * @details This jacobian has to be interpreted in terms of Lie group, not vector space: as such,
+   *          it is expressed in the tangent space only, not the configuration space.
+   *          Calling \f$ f(q, v) \f$ the integrate function, these jacobians satisfy the following relationships in the
+   *          tangent space:
+   *           - Jacobian relative to q: \f$ f(q \oplus \delta q, v) \ominus f(q, v) = J_q(q, v) + o(\delta q)\f$.
+   *           - Jacobian relative to v: \f$ f(q, v + \delta v) \ominus f(q, v) = J_v(q, v) + o(\delta v)\f$.
+   *
```
gabrielebndn: Same here

jmirabel:
Generally speaking, this doc is a good thing. I think there is a need for a page which summarizes the notations. There are a few tricky points which ought to be highlighted:
1. q0 \oplus (q1 \ominus q0) can be different from q1 (in cases which are however easy to understand).
2. q0 \oplus (v0 + v1) is generally different from (q0 \oplus v0) \oplus v1.
3. You *should* never check equality using q0 == q1 (neither (q0-q1).norm() < thr). This can work if you now what you are doing. If you don't know, better use (q0 \ominus q1).norm() < thr.
4. **You should prefer \oplus and \ominus  operations over + and - ones on configuration vectors**.
5. Moreover, when computing derivatives, you should always think of these operators as functions with their own non-trivial derivatives. It will help you avoid *hard-to-find* wrong derivative (and they are a lot easier to write using the Lie Algebra)

I am sure the list is not exhaustive.

jcarpent:
@matthieuvigne can you address the comments of @gabrielebndn. I will then open a new issue concerning the comments of @jmirabel 
Thanks in advance.

matthieuvigne:
@jmirabel I definitely agree with you on this. It's easy to confuse the joint configuration for a vector space (in particular because in practice, it is implemented as a vector) when you work with only 1DoF joints, and then everything stats to fall apart when you look at more complex kinematics. Viewing everything in terms of Lie group is a lot more powerful, and really well handled by pinocchio, but it takes some practice and it's easy to take the wrong habits.

PR #766, opened 2019-04-12
Minor changes in repository

proyan:
1) Expose data.h
2) Expose computeCentroidalDynamics
3) small simplification in computation of v in computecentroidaldynamics.

### DIFF
--- a/bindings/python/algorithm/expose-centroidal.cpp
+++ b/bindings/python/algorithm/expose-centroidal.cpp
@@ -13,6 +13,23 @@ namespace pinocchio
     void exposeCentroidal()
     {
       using namespace Eigen;
+
+      bp::def("computeCentroidalDynamics",
+              &computeCentroidalDynamics<double,0,JointCollectionDefaultTpl,VectorXd,VectorXd>,
+              bp::args("Model","Data",
+                       "Joint configuration q (size Model::nq)",
+                       "Joint velocity v (size Model::nv)"),
+              "Computes the Centroidal dynamics, a.k.a. the total momenta of the system expressed around the center of mass.",
+              bp::return_value_policy<bp::return_by_value>());
+      
+      bp::def("computeCentroidalDynamics",
+              &computeCentroidalDynamics<double,0,JointCollectionDefaultTpl,VectorXd,VectorXd,VectorXd>,
+              bp::args("Model","Data",
+                       "Joint configuration q (size Model::nq)",
+                       "Joint velocity v (size Model::nv)",
+                       "Joint acceleration a (size Model::nv)"),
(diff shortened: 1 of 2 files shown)


src/algorithm/centroidal.hxx
```diff
@@ -53,14 +53,12 @@ namespace pinocchio
 
       data.mass[i] = Y.mass();
       data.com[i] = Y.mass() * Y.lever();
-      
+
+      data.v[i] = jdata.v();
```
jcarpent: The change are not optimal. The Original version is optimal with respect to the Joint API. Can you remove this commit?

bindings/python/algorithm/expose-centroidal.cpp
```diff
@@ -13,6 +13,23 @@ namespace pinocchio
...
     {
       using namespace Eigen;
+
+      bp::def("computeCentroidalDynamics",
+              &computeCentroidalDynamics<double,0,JointCollectionDefaultTpl,VectorXd,VectorXd>,
+              bp::args("Model","Data",
+                       "Joint configuration q (size Model::nq)",
+                       "Joint velocity v (size Model::nv)"),
+              "Computes the Centroidal dynamics, a.k.a. the total momenta of the system expressed around the center of mass.",
+              bp::return_value_policy<bp::return_by_value>());
```
jcarpent: There is no need of return by value here.

bindings/python/algorithm/expose-centroidal.cpp
```diff
@@ -13,6 +13,23 @@ namespace pinocchio
...
+      
+      bp::def("computeCentroidalDynamics",
+              &computeCentroidalDynamics<double,0,JointCollectionDefaultTpl,VectorXd,VectorXd,VectorXd>,
+              bp::args("Model","Data",
+                       "Joint configuration q (size Model::nq)",
+                       "Joint velocity v (size Model::nv)",
+                       "Joint acceleration a (size Model::nv)"),
+              "Computes the Centroidal dynamics and its time derivatives, a.k.a. the total momenta of the system and its time derivative expressed around the center of mass.",
+              bp::return_value_policy<bp::return_by_value>());
```
jcarpent: The same applies here.

src/algorithm/centroidal.hxx
proyan: I tried to keep the same loop as in forwardkinematics.
nevermind. changed.

bindings/python/algorithm/expose-centroidal.cpp
proyan: Every function that I see is using return by value as the policy.
what would you prefer here?

jcarpent: It is because they are returning a matrix. But maybe it is fine like that avoiding miss understanding by the user.

PR #767, opened 2019-04-12
add computation of dhg/dq inside computecentroidaldynamicsderivatives

proyan:
I extend the forceSet::motionAction to include the case when multiple column vectors are being applied to a single force vector.

Except that, the implementation is easy to understand.

### DIFF
--- a/src/algorithm/centroidal-derivatives.hxx
+++ b/src/algorithm/centroidal-derivatives.hxx
@@ -318,6 +318,8 @@ namespace pinocchio
       typename Data::Matrix6x & Ftmp = data.Fcrb[0];
 
       ColsBlock J_cols = jmodel.jointCols(data.J);
+      ColsBlock dVdq_cols = jmodel.jointCols(data.dVdq);
+      ColsBlock dHdq_cols = jmodel.jointCols(data.dHdq);
       ColsBlock dFdq_cols = jmodel.jointCols(data.dFdq);
       ColsBlock Ftmp_cols = jmodel.jointCols(Ftmp);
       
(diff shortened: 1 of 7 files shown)


src/algorithm/centroidal-derivatives.hxx
```diff
@@ -332,14 +342,18 @@ namespace pinocchio
...
       }
+
+      forceSet::motionActions(J_cols, data.oh[i], dHdq_cols);
+      motionSet::inertiaAction<ADDTO>(data.oYcrb[i], dVdq_cols, dHdq_cols);
     }
   }; // struct GetCentroidalDynDerivativesBackwardStep
   
   template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl,
-           typename Matrix6xLike1, typename Matrix6xLike2, typename Matrix6xLike3>
+           typename Matrix6xLike0,typename Matrix6xLike1, typename Matrix6xLike2, typename Matrix6xLike3>
```
jcarpent: Missing space here.

src/spatial/act-on-set.hpp
```diff
@@ -64,6 +64,22 @@ namespace pinocchio
...
                              Eigen::MatrixBase<MatRet> const & jF);
+
+    ///
+    /// \brief Action of a set of motions on a force. Motion is
+    ///        represented by a 6xN matrix whose each
+    ///        column represent a spatial motion.
+    ///
+    template<int Op, typename Mat, typename ForceDerived, typename MatRet>
+    static void motionActions(const Eigen::MatrixBase<Mat> & iV,
+                             const ForceDense<ForceDerived> & f,
```
jcarpent: Missing indentation

src/algorithm/centroidal-derivatives.hxx
proyan: Done.

src/spatial/act-on-set.hpp
proyan: Done.

proyan:
It seems that I missed the motionSet::act, it is the same implementation as forceSet::motionAction.
(Which is a bit confusing, since all implementation of motion and force operations is done in the Force class, or in the forceSet namespace. motionSet::act is an outlier here.)

I've removed the commit which added forceset::motionActions.

Issue #764, opened 2019-04-11
Strange behavior of geom.removeCollisonPair


nmansard:
I don't understand what happens here:

p=gmodel.collisionPairs[1]  
print(p)    # display    collision pair (4,0)
gmodel.removeCollisionPair(gmodel.collisionPairs[0])
print(p)   # display collision pair (6,0)
print( gmodel.collisionPairs[0]) # display collision pair (4,0)
print(gmodel.collisionPairs[1]) # display collision pair (6,0)

Is p a fake index on the collision-pair array?

jmirabel:
Did you check that pair (4,0) is not duplicated in the beginning ?

nmansard:
It is not (the first is (0,3)). Even if it was, I don't see how this would explain the behavior.

gabrielebndn:
The problem is in the way the bindings of `collisionPairs` are implemented. When you do
```
p=gmodel.collisionPairs[1]
```
`p` is a *pointer* to the location of `gmodel.collisionPairs[1]`.

Then, you do
```
gmodel.removeCollisionPair(gmodel.collisionPairs[0])
```
Here, you are removing element zero from `gmodel.collisionPairs`. `collisionPairs` is a `std::vector`. Internally, it stores its data in an array. To execute this operation, it shifts all elements of the array back one position.

```
print(p) # display collision pair (6,0)
```
Here, `p` is still pointing at element 1 of the array. This position is now filled by what used to be element 2, before `removeCollisionPair` was called. In this case, `(6,0)`

```
print( gmodel.collisionPairs[0]) # display collision pair (4,0)
```
Element zero now holds what used to be element 1, before `removeCollisionPair` was called, i.e. `(4,0)`.

```
print(gmodel.collisionPairs[1]) # display collision pair (6,0)
```
Element 1 now holds what used to be element 2, before `removeCollisionPair` was called, i.e. `(6,0)`.

It gets worse: after a certain number of removals, the `std::vector` could decide to completely reallocate the array in order to free some memory. In this case, any previously-taken pointer would be pointing at garbage.

To make this clearer: what you are doing in Python is equivalent to the following C++ code
```
CollisionPair & p = gmodel.collisionPairs[1] // notice the reference
cout << p << endl; // display collision pair (4,0)
gmodel.removeCollisionPair(gmodel.collisionPairs[0])
cout << p << endl; // display collision pair (6,0)
[...]

PR #769, opened 2019-04-12
[bindings] copyable collision pair

gabrielebndn:
Allows to copy a `CollisionPair` object in Python by doing `pair.copy()`. Related to #764


gabrielebndn:
@jcarpent I'll see what I can do.
I think it is a bit tricky because, to things right, we want
```
p = v[i]
```
to avoid the problem in #764, but
```
v[i] = p
```
should still be possible and work as expected.
I think the final solution will be given by #555.

Anyhow, since a `CollisionPair` object is not immutable, an explicit copy method may still be useful, so I think the change is worth merging as it is for now.

PR #770, opened 2019-04-13
Fix bug in computation of dVdq and dAdq

proyan:
There is a small bug in the computation of dVdq and dAdq.

To explain using dVdq as the example, the computation of dVdq for the ith joint is given by ov[parent].cross(Ji), where Ji is the jacobian of the ith joint.

The current implementation of dVdq is done using dJ, which is ov[i].cross(Ji).

Normally, in most cases, the two are equal. The difference between the two is given by an S x S term, which is zero for all one DoF joint. And when we have a free flyer as the first joint, ov[parent] is zero, so we don't see the effect of S x S term again. This is why there are no problems with the unittest.

You can see the problem when you add a spherical joint ( or any joint in which SxS is not zero, such as freeflyer) at a position other than root joint. I added the spherical joint in centroidal-derivatives unittest, and the test does not pass anymore.

This PR takes care of this.

Surprisingly, the implementation in aba-derivatives is correct. The problem is only in the centroidal-derivatives and rnea-derivatives. So I have kept the same implementation for all three.

### DIFF
--- a/src/algorithm/centroidal-derivatives.hxx
+++ b/src/algorithm/centroidal-derivatives.hxx
@@ -85,17 +85,16 @@ namespace pinocchio
       J_cols = data.oMi[i].act(jdata.S());
       motionSet::motionAction(ov,J_cols,dJ_cols);
       motionSet::motionAction(data.oa[parent],J_cols,dAdq_cols);
-      
+      dAdv_cols = dJ_cols;      
       if(parent > 0)
       {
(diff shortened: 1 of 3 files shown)


jcarpent:
Yes. I intended made this modifications to simplify the computations. I did only that for the RNEA derivatives that I then propagate by copy past to the centroidal ones. That explains why the implementation of Aba derivatives are correct.

Issue #776, opened 2019-04-17
Questionable hard-coded joint limit value in urdf-parsers for "continous" joints.


markusgft:
https://github.com/stack-of-tasks/pinocchio/blob/43e43293033b7f0b239869241c6cf780917164cf/src/parsers/urdf/model.hxx#L312

Dear all,
why is the joint limit value set to 1.01 ?

jcarpent:
There is no good reason for that. It just to have something greater than one. Because continuous revolute joints do not have bounds.

cmastalli:
I would expect to have the greatest value that it's possible, e.g.:

```c++
const ConfigVector_t::Scalar u = numeric_limits<double>::max()
```

jcarpent:
@cmastalli This is indeed one solution.

jmirabel:
If I remember well, I think I am the one who set it to 1.01. The motivations were:
- to help tracking errors. Since then, we added a isNormalized, which is the proper way of tracking an error.
- to easily (i.e. without visitor) know whether a joint is bounded or not. If you put the max double, then you lose the information that this value is bounded.

The bounds are on the configuration vector parameters, and not on the rotation angle. The ideal bound ought to be 1, and not max double. But then, there will be numerical issues... I admit 1.01 is too big.

gabrielebndn:
I don't think max double makes sense. A continuous joint is never expected to exceed `. It should be as close as possible to 1, but not exactly 1, just slightly greater.

Also, putting max double will likely create problems to `randomConfiguration(model)`, which uses the limits

gabrielebndn:
> Also, putting max double will likely create problems to randomConfiguration(model), which uses the limits

Actually, I've just read the code computing a random configuration for an unbounded joint. It seems the bounds are ignored. You can find it in [src/multibody/liegroup/special-orthogonal.hpp](https://github.com/stack-of-tasks/pinocchio/blob/devel/src/multibody/liegroup/special-orthogonal.hpp). Method `randomConfiguration_impl` does not employ the bounds. Is this right? I know it does not make a lot of sense to put bounds on "unbounded" joints, but are we OK just ignoring them?

gabrielebndn:
Also, I've noticed an incoherence: when *parsing* an URDF model, the default limit is 1.01. But when you use `addJoint`, *without* specifying the limits, then the limits are indeed max double.

(also, `addJoint` with limits is not exposed in Python, I'll fix this)

cmastalli:
[...]

PR #779, opened 2019-04-18
Fix first order normalization of unit complex.

jmirabel:
#776 made me look at this and realize the error.


jmirabel:
Cos and sin omega are on the unit circle with a fairly good precision. But the error on sa and sa may not, and this error can be cumulative...

jmirabel:
So, yes, I agree it is for numerical stability.

PR #780, opened 2019-04-18
algo/center-of-mass-derivatives: improve efficiency

jcarpent:
Number of operations are reduced and it also avoid memory allocations


jcarpent:
An additional improvement would be to use directly the quantities expressed in the world frame, avoiding unnecessary computations.

PR #781, opened 2019-04-19
Add support for LOCAL_CARTESIAN_ORIENTED in jacobian computation

### LINKED ISSUE #773: Compute jacobians in local world oriented frame
Hi folks,

at this point the jacobians can be computed in the world or local frame. I propose to add another reference frame that is the local_cartesian_oriented frame. This is useful when dealing with point contacts.

The code to compute this boils down to:

```
        jac = se3.getFrameJacobian(self.model, self.data, index, se3.ReferenceFrame.LOCAL)
        world_R_joint = se3.SE3(self.data.oMf[index].rotation, zero(3))
        return world_R_joint.action.dot(jac)
```

As this is a convenient method to have, I would like to add it to the pinocchio se3/RobotWrapper Jacobian's C++ API but would like to get feedback of adding a PR with this feature has changes to getting accepted.

Best,
Julian

jviereck:
This is to fix #773. 

This is NOT done yet but there are a few points I thought it's better to get feedback on earlier then later.

I haven't tested the code yet. The code compute the frame orientation I've copied from some other internal code we have, but I don't know if the jointJacobian code is working as expected.

A few things:
- Does the name `LOCAL_CARTESIAN_ORIENTED` make sense?
- I have to use a temporary `J_tmp` to avoid allocations (this was a problem we ran into with tsid). Though I am not 100% sure yet if using a local `J_tmp` and resizing works around the memory allocation problem. Does someone has an idea how to write the SE3 rotation of the jacobian without doing the `toActionMatrix() * J` matrix product - e.g. in a form where no new resulting matrix must be allocated and the results can be directly stored in J.

### DIFF
--- a/src/algorithm/frames.hxx
+++ b/src/algorithm/frames.hxx
@@ -125,11 +125,7 @@ namespace pinocchio
     if(rf == WORLD)
     {
       getJointJacobian(model,data,joint_id,WORLD,PINOCCHIO_EIGEN_CONST_CAST(Matrix6xLike,J));
-      return;
-    }
-    
-    if(rf == LOCAL)
-    {
+    } else if(rf == LOCAL || rf == LOCAL_CARTESIAN_ORIENTED) {
       Matrix6xLike & J_ = PINOCCHIO_EIGEN_CONST_CAST(Matrix6xLike,J);
       const typename Data::SE3 & oMframe = data.oMf[frame_id];
       const int colRef = nv(model.joints[joint_id])+idx_v(model.joints[joint_id])-1;
@@ -138,7 +134,13 @@ namespace pinocchio
       {
         J_.col(j) = oMframe.actInv(Motion(data.J.col(j))).toVector(); // TODO: use MotionRef
       }
-      return;
+
+      if (rf == LOCAL_CARTESIAN_ORIENTED) {
+        Matrix6xLike J_tmp;
+        J_tmp.resize(6, model.nv);
+        J_tmp = SE3(data.oMf[frame_id].rotation(), Eigen::Vector3d::Zero()).toActionMatrix() * J;
+        J_ = J_tmp;
+      }
     }
   }
   
--- a/src/algorithm/frames.hxx
+++ b/src/algorithm/frames.hxx
@@ -126,6 +126,7 @@ namespace pinocchio
     if(rf == WORLD)
     {
       getJointJacobian(model,data,joint_id,WORLD,PINOCCHIO_EIGEN_CONST_CAST(Matrix6xLike,J));
+      return;
     } 
     else if(rf == LOCAL || rf == LOCAL_WORLD_ALIGNED) 
     {
--- a/src/algorithm/frames.hxx
+++ b/src/algorithm/frames.hxx
@@ -122,22 +122,28 @@ namespace pinocchio
     
     const Frame & frame = model.frames[frame_id];
     const JointIndex & joint_id = frame.parent;
+    
     if(rf == WORLD)
     {
       getJointJacobian(model,data,joint_id,WORLD,PINOCCHIO_EIGEN_CONST_CAST(Matrix6xLike,J));
-    } else if(rf == LOCAL || rf == LOCAL_WORLD_ALIGNED) {
+    } 
+    else if(rf == LOCAL || rf == LOCAL_WORLD_ALIGNED) 
+    {
       Matrix6xLike & J_ = PINOCCHIO_EIGEN_CONST_CAST(Matrix6xLike,J);
       const typename Data::SE3 & oMframe = data.oMf[frame_id];
       const int colRef = nv(model.joints[joint_id])+idx_v(model.joints[joint_id])-1;
       
       for(Eigen::DenseIndex j=colRef;j>=0;j=data.parents_fromRow[(size_t) j])
       {
-        if (rf == LOCAL) {
-        J_.col(j) = oMframe.actInv(Motion(data.J.col(j))).toVector(); // TODO: use MotionRef
-        } else {
+        if (rf == LOCAL) 
+        {
+          J_.col(j) = oMframe.actInv(Motion(data.J.col(j))).toVector(); // TODO: use MotionRef
+        } 
+        else 
+        {
           J_.col(j) = data.J.col(j);
           J_.col(j).template segment<3>(Motion::LINEAR) -= oMframe.translation().cross(J_.col(j).template segment<3>(Motion::ANGULAR));
-      }
+        }
       }
     }
   }
--- a/src/algorithm/frames.hxx
+++ b/src/algorithm/frames.hxx
@@ -125,21 +125,19 @@ namespace pinocchio
     if(rf == WORLD)
     {
       getJointJacobian(model,data,joint_id,WORLD,PINOCCHIO_EIGEN_CONST_CAST(Matrix6xLike,J));
-    } else if(rf == LOCAL || rf == LOCAL_CARTESIAN_ORIENTED) {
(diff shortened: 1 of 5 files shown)


gabrielebndn:
I believe the local J_tmp does not solve the allocation problem. One way of doing it would be declaring it as static, I think, but it would be really non-standard for Pinocchio. Normally, we put temporary variables inside Data.

I am pretty sure there is a way to avoid that variable, but one would need to study the algorithm better

jmirabel:
Don't make it static. It would break thread-safety.

You can allocate a vector of fixed size 6 and do the matrix product manually, column-wise on the Jacobian.

jmirabel:
I didn't check Eigen's code, but that might be done in Eigen::Matrix::applyOnTheLeft (or applyOnTheRight, I never know to who the left and right applies...)

gabrielebndn:
Yes, I would do as @jmirabel says

gabrielebndn:
@jviereck, thanks a lot for your contribution, really appreciate it.

To answer your question about temporaries: @jmirabel's answer is certainly correct, but I think the best way to implement what you want is to think about it as a *global* frame which is *translated*. This will save you computations and will spare you the need of any temporary variable.

You can just copy-paste the code used for `LOCAL`. Only, intead of
```
J_.col(j) = oMframe.actInv(Motion(data.J.col(j))).toVector();
```
you will have
```
J_.col(j).segment<3>(Motion::LINEAR) -= oMframe.translation().cross(J_.col(j).segment<3>(Motion::ANGULAR);
```
equivalent to
```
J_.col(j) = SE3(Eigen::MatrixXd::Identity(), oMframe.translation()).actInv(Motion(data.J.col(j))).toVector();
```
but without the intermediate `SE3` object and without all redundant multiplications.

I think `LOCAL_CARTESIAN_ORIENTED` is quite long, I would favor a more compact name, such as simply `CARTESIAN`, but I am neutral about it.

src/algorithm/frames.hxx
```diff
@@ -138,7 +134,13 @@ namespace pinocchio
       {
         J_.col(j) = oMframe.actInv(Motion(data.J.col(j))).toVector(); // TODO: use MotionRef
       }
-      return;
+
+      if (rf == LOCAL_CARTESIAN_ORIENTED) {
+        Matrix6xLike J_tmp;
+        J_tmp.resize(6, model.nv);
+        J_tmp = SE3(data.oMf[frame_id].rotation(), Eigen::Vector3d::Zero()).toActionMatrix() * J;
```
gabrielebndn: `Eigen::Vector3d` should not be employed directly. Pinocchio is templated by scalar type. The same holds true for `SE3`. Appropriate typedefs parametrized bu `Scalar` are to be preferred.

src/algorithm/jacobian.hxx
```diff
@@ -141,6 +141,13 @@ namespace pinocchio
         J_.col(j) = oMjoint.actInv(mref).toVector();
       }
     }
+
+    if (rf == LOCAL_CARTESIAN_ORIENTED) {
+        Matrix6Like J_tmp;
+        J_tmp.resize(6, model.nv);
+        J_tmp = SE3(oMjoint.rotation(), Eigen::Vector3d::Zero()).toActionMatrix() * J;
```
gabrielebndn: Same here

jcarpent:
@jviereck `LOCAL_CARTESIAN_ORIENTED` seems not as obvious as naming.
Maybe `LOCAL_ALIGN_WITH_WORLD` or `LOCAL_WORLD_ALIGNED` may be better. Any other ideas are welcome.

gabrielebndn:
> Maybe `LOCAL_ALIGN_WITH_WORLD` or `LOCAL_WORLD_ALIGNED` may be better. Any other ideas are welcome.

+1 for `LOCAL_WORLD_ALIGNED`

src/algorithm/frames.hxx
jviereck: I was not thinking about the parametrized by `Scalar` thing. Thanks for pointing this out.

jviereck:
hi folks, 

thanks for all the feedback and input.

> LOCAL_WORLD_ALIGNED

+1 for this one.

> J_.col(j).segment<3>(Motion::LINEAR) -= oMframe.translation().cross(J_.col(j).segment<3>(Motion::ANGULAR));

Ohh, I haven't thought about computing the Jacobian in this way - that looks like the much better way to go and I am going to adjust the code.

> In addition, I would suggest making unitary test to be sure that what you've implemented is correct.

Agreed. I am traveling over the weekend and plan to work on this during the ride.

jcarpent:
@jviereck Seems good to me.

jviereck:
Hi folks, I've implemented your feedback and added a simple unit test. However, the unit test is not passing. I am not sure yet what is going on and going to take  a closer look later.

jviereck:
Hi folks, the commit just pushed fixes the unit test. The issue was that I queried the frame before calling `forwardFrames` etc in the unittest. After switching the order, this unit test works now.

For the joint jacobian, what's the frame orientation the LOCAL joint jacobians are expressed in? Are the LOCAL joint jacobians already expressed in the world oriented frame?

jcarpent:
> For the joint jacobian, what's the frame orientation the LOCAL joint jacobians are expressed in? Are the LOCAL joint jacobians already expressed in the world oriented frame?

The LOCAL means that the Jacobian is expressed in the frame of the joint directly. While data.J corresponds to the collection of the joint spatial spaces expressed at the origin of the world, aligned with the wold axes.

jcarpent:
@jviereck Thanks a lot for your contribution.

src/algorithm/frames.hxx
```diff
@@ -126,6 +126,7 @@ namespace pinocchio
     if(rf == WORLD)
     {
       getJointJacobian(model,data,joint_id,WORLD,PINOCCHIO_EIGEN_CONST_CAST(Matrix6xLike,J));
+      return;
```
jviereck: Is there a reason to have the return here? If that's part of the pinocchio coding style convention to have `returns` early in a function, then there should also be a `return` for the `else if(rf == LOCAL || rf == LOCAL_WORLD_ALIGNED) ` branch.

jcarpent: You removed it in one of your commits. I've just restored it.
Concerning the convention, it is just helpful for the compiler, and to avoid checking other flags. Similar to the break instruction in a switch.

jviereck:
hi @jcarpent, thanks for merging this PR! However, I am not sure this PR is done yet. There are no unit tests for `getJointJacobian(..., LOCAL_WORLD_ALIGNED)`.

> The LOCAL means that the Jacobian is expressed in the frame of the joint directly. While data.J corresponds to the collection of the joint spatial spaces expressed at the origin of the world, aligned with the wold axes.

Thanks for this clearification!

I tried to write a similar unit test to the `getFrameJacobian` one. However, when I compared the result between the result from `LOCAL_WORLD_ALIGNED` and manually rotating the `WORLD` jacobian (as in the `getFrameJacobian` unit test), the resulting Jacobians did not match.

Any idea why this could be the case?

jcarpent:
This is my mistake. I was thinking you were done.
Feel free to open a new PR and add the other unit tests.


> I tried to write a similar unit test to the `getFrameJacobian` one. However, when I compared the result between the result from `LOCAL_WORLD_ALIGNED` and manually rotating the `WORLD` jacobian (as in the `getFrameJacobian` unit test), the resulting Jacobians did not match.
> 
> Any idea why this could be the case?

The WORLD jacobian is expressed at the origin of the WORLD. You then need a shift.
You need to rotate the jacobian computing in the LOCAL frame to retrieve the expected results.
Is it clear enough?

PR #782, opened 2019-04-19
[bindings] Make GeometryModel.collisionPairs return-by-value, fix #764

### LINKED ISSUE #764: Strange behavior of geom.removeCollisonPair

gabrielebndn:
The return policy of `GeometryModel.collisionPairs` is now return-by-value. In other words, `gmodel.collisionPairs[i]` returns a copy of the element at `i`, instead of a reference. However, it is still possible to assign the vector elements by doing `gmodel.collisionPairs[i] = c`. This fixes #764.

In fact, boost-python allows to choose very easily between the two policies - just setting a boolean parameter!
I found that boost-python vector wrappers handle removals and insertions quite well. However, the problem with `gmodel.collisionPairs` is that it was directly manipulated by C++ code, so the wrapper could not see it. That is why I only made this change to vectors of `CollisionPair` elements -- the others are untouched.

As a bonus, I made the naming of `std::vector` bindings uniform.
Before, as noticed in #555 and #626, some were named beginning with `StdVec` while others are with `StdVect`. Now they are all called `StdVec`. I retained the old `StdVect` as deprecated for back-compatibility.

### DIFF
--- a/bindings/python/algorithm/expose-aba-derivatives.cpp
+++ b/bindings/python/algorithm/expose-aba-derivatives.cpp
@@ -51,7 +51,7 @@ namespace pinocchio
               "put the result in data.ddq_dq, data.ddq_dv and data.Minv\n"
               "which correspond to the partial derivatives of the acceleration output with respect to the joint configuration,\n"
               "velocity and torque vectors.\n"
-              "The forces are of type StdVect_Force.");
+              "The forces are of type StdVec_Force.");
     }
   } // namespace python
 } // namespace pinocchio
--- a/bindings/python/multibody/data.hpp
+++ b/bindings/python/multibody/data.hpp
@@ -31,22 +31,22 @@ namespace pinocchio
 
     public:
 
-#define ADD_DATA_PROPERTY(TYPE,NAME,DOC)				 \
-      def_readwrite(#NAME,						 \
-      &Data::NAME,		 \
+#define ADD_DATA_PROPERTY(TYPE,NAME,DOC)    \
+      def_readwrite(#NAME,                  \
+      &Data::NAME,                          \
       DOC)
       
-#define ADD_DATA_PROPERTY_READONLY(TYPE,NAME,DOC)				 \
-      def_readonly(#NAME,						 \
-      &Data::NAME,		 \
+#define ADD_DATA_PROPERTY_READONLY(TYPE,NAME,DOC) \
+      def_readonly(#NAME,                         \
+      &Data::NAME,                                \
       DOC)
       
-#define ADD_DATA_PROPERTY_READONLY_BYVALUE(TYPE,NAME,DOC)				 \
-      add_property(#NAME,						 \
-      make_getter(&Data::NAME,bp::return_value_policy<bp::return_by_value>()), \
-      DOC)
-      
-	 
+#define ADD_DATA_PROPERTY_READONLY_BYVALUE(TYPE,NAME,DOC)                       \
+      add_property(#NAME,                                                       \
+      make_getter(&Data::NAME,bp::return_value_policy<bp::return_by_value>()),  \
+      DOC) 
+
+
       /* --- Exposing C++ API to python through the handler ----------------- */
       template<class PyClass>
       void visit(PyClass& cl) const 
--- a/bindings/python/multibody/expose-frame.cpp
+++ b/bindings/python/multibody/expose-frame.cpp
@@ -14,7 +14,7 @@ namespace pinocchio
     void exposeFrame()
     {
       FramePythonVisitor::expose();
-      StdAlignedVectorPythonVisitor<Frame>::expose("StdVect_Frame");
+      StdAlignedVectorPythonVisitor<Frame>::expose("StdVec_Frame");
     }
     
   } // namespace python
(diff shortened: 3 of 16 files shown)


gabrielebndn:
> I would to not return by value from a vector, this is the spirit of Python with the list and we followed this policy since the beginning of Pinocchio.

Uh? I thought this is what you were suggesting in #769.

I do not see any other way solve the problem in #764. There's at least four methods in Pinocchio which directly remove elements from `gmodel.collisionPairs`:
- https://github.com/stack-of-tasks/pinocchio/blob/dbcfbcb283eea486e1f5279c5e40cecdb6e87fad/src/multibody/geometry.hpp#L154 
- https://github.com/stack-of-tasks/pinocchio/blob/dbcfbcb283eea486e1f5279c5e40cecdb6e87fad/src/multibody/geometry.hpp#L158
- https://github.com/stack-of-tasks/pinocchio/blob/dbcfbcb283eea486e1f5279c5e40cecdb6e87fad/src/parsers/srdf.hpp#L29
- https://github.com/stack-of-tasks/pinocchio/blob/dbcfbcb283eea486e1f5279c5e40cecdb6e87fad/src/parsers/srdf.hpp#L55

plus their deprecated aliases `removeCollisionPairsFromSrdf` and `removeCollisionPairsFromSrdfString`. These methods remove elements from the `std::vector` itself, so any sort of wrapper we can possibly make will not be able to see it. It seems to me that the only way to make these methods work in Python with return-by-reference would be to reimplement them in Python, which is unfeasible from the maintenance point of view.

> We cannot have various policies for the various list/vectors that are contained in the Pinocchio structures.

I agree it might be confusing. This is why I added a note to the Python doc. I do not see it as a terrible thing, after all boost-python offers many possibilities exactly because each case might need to be treated differently. Maybe it would be clearer if we called the exposed `std::vector<CollisionPair>` in a different way, such as `StdCopyVec_CollisionPair`, so to make it clear it is a separate class and should be treated differently?

> In the meantime, can you avoid PR with structural modifications and the one with just spelling adjustments? It will make the review easier for me.

I did not think the first and second commit deserved a PR of their own, but I'll keep it in mind for the future. Meanwhile, what do you want me to do, should I split this PR? If you want I can do it, but each individual commit is self-contained and can be easily analyzed on its own. If you are not convinced by the choice of returning by value, I can make a separate PR with all commits but the third one.

<br/>

SIDE NOTE 1: is `std::vector` a good choice for `gmodel.collisionPairs`? Wouldn't `std::list` be better? Just asking, I imagine that the access to `gmodel.collisionPairs` is preferably sequential rather than random, but I might be wrong. Using `std::list` *might* solve the problem of #764 while keeping return-by-reference in Python, but we shouldn't be influenced by this, only the efficiency of the container really counts.

SIDE NOTE 2: even if we expose `std::vector` as a Python list, as discussed in #555, I don't think it will  solve the problem in  #764.

jcarpent:
In the meantime, when thinking about the problem encountered in #764, the behavior is expected.
When deleting an element from a vector, its reference is also deleted: this behavior is expected. 
So, I would suggest keeping as it is, and to simply keep your commit related to the indentation and change of naming.

gabrielebndn:
> In the meantime, when thinking about the problem encountered in #764, the behavior is expected.

I don't agree. It is easily explained in terms of references, that's for sure, but it is not intuitive at all and definitely not Python-like, much less Python-like than return-by-reference in IMHO.

Anyhow, I'll open a separate PR for the naming, the indentation, and the equality-inequality operators for collision pairs.

jcarpent:
The expected behavior is in terms of the current implementation. The ideal would be to rely on shared ptr for this issue, which is a reasonable solution. The problem is then that the memory is no more aligned, as in the case of std::vector. 
@gabrielebndn If you open a new PR, I would suggest closing this one, because the solution is not appropriate with respect to #764.

gabrielebndn:
> The expected behavior is in terms of the current implementation.

I really don't agree. A Python user does not *expect* the pointed object to change just because the container has rearranged its elements. The problem is `std::vector` really changes the values contained in that memory location, so anything pointing there will see the values changed. In the end, one should take care to always make a copy if it is foreseeable that the vector is changed behind the curtains. And, as one can never be sure, one should *always* make a copy in order to avoid potential bugs.

> The ideal would be to rely on shared ptr for this issue, which is a reasonable solution.

I think the shared ptr would probably solve the problem of undefined behavior (and potential segfaults) in case the reference starts pointing *out* of the array and/or `std::vector` decides to reallocate the memory, but still, the resulting behavior would be quite erratic and unpredictable for such cases.

> The problem is then that the memory is no more aligned, as in the case of std::vector.

I don't think we have alignment issues with `gmodel.collisionPairs`: it does not contain Eigen objects. Indeed, it is a regular `std::vector`, not an aligned vector.

> @gabrielebndn If you open a new PR, I would suggest closing this one,

If you are really convinced of dropping this change, I'll close this PR as soon as the new one is ready.

> because the solution is not appropriate with respect to #764.

So, if I understand what you mean, the appropriate answer to #764 is just saying this behavior is expected, suggest using #769, and closing it?

gabrielebndn:
Closing this PR. I opened #783 containing all changes it is wished to retain (namely, all commits except for the third one).

PR #788, opened 2019-04-29
remove deprecated throw specification

christian-rauch:
This PR removes the throw specifications that are deprecated since C++11:
`warning: dynamic exception specifications are deprecated in C++11 [-Wdeprecated]`.


christian-rauch:
Yes. Most, if not all, C++ deprecation warnings become errors at some point. It makes sense to resolve them as soon as they appear.

Issue #663, opened 2019-02-08
Separating the functionnalities of RobotWrapper


nmansard:
This is related to #657 

RobotWrapper implements 3 very different sets of functionalities:
- it loads the 3 models (kinematics, visuals, collisions) and create the corresponding data.
- it implements nice object-oriented wrappers around pinocchio main algorithms, useful for newcommers.
- it implements the bindings with the viewer, by creating the viewer client, loading the geoms, computing their placement and forwarding the placements to the viewer.

I believe we should separate those 3 sets. 
- by creating a loadModelsFromUrdf function that returns the 3 models (no need to put that in an object)
- by keeping the nice wrapping of the algorithms in a RobotWrapper object that would be initialize by an instance of Pinocchio::Model.
- by introducing a new GepettoViewerClient that would take care of the bindings with the viewer. This would be implemented as a new class, whose constructor would take a pinocchio::Model and a pinocchio::GeometryModel.

2 positive consequences of that would be:
- robot wrapper would not be systematically used, and only used for the algorithms. In particular, in TSID and Crocoddyl, the RobotWrapper is just their for functionnality 1 and 3, not for 2.
- when introducing new display like MeshCat in #662, we would not need to make RobotWrapper more complex but simply to introduce a new viewer_meshcat.py file with the corresponding clien class in it.

jcarpent:
I would change also the syntax:  not using `loadModelsFromUrdf` but rather `loadModelsFromURDF`

jcarpent:
Concerning `RobotWrapper`, is already compatible with the creation from a Pinocchio `Model`.

jcarpent:
>     * robot wrapper would not be systematically used, and only used for the algorithms. In particular, in TSID and Crocoddyl, the RobotWrapper is just their for functionnality 1 and 3, not for 2.

@nmansard Can you say in your discussion what are 1, 2 and 3?

nim65s:
Conventions in python are `RobotWrapper` for class names, and `load_model_from_urdf` for functions / methods: https://www.python.org/dev/peps/pep-0008/#class-names / https://www.python.org/dev/peps/pep-0008/#id45

jcarpent:
@nim65s But we need to stick to the C++ conventions I think, because we are only doing Python bindings and we want our code to interoperate between both languages.

nim65s:
I disagree. I don't see why we could not respect c++ standards in c++ and python standards in python. 

[...]

PR #791, opened 2019-05-01
[python] Define buildModelsFromUrdf and createDatas function

gabrielebndn:
This PR partially addresses #663 by providing convenience Python methods for creating models and geometry models in a single call, and analogously for the data objects.

This PR is autonomous, so if it is agreed upon it might be safely accepted. I thought it might be better to split the changes in two separate PRs so that this part may be more easily reviewed.

In summary:
- expose `buildGeomFromUrdf` accepting a single string for the package_directory, so we don't always have to put it inside a list
- define a function called `buildModelsFromUrdf` which parses an URDF and returns a Pinocchio Model plus one or two GeometryModel(s), depending on the inputs (see docstring for more details)
- define a Python method called `createData()` for `GeometryModel`, in analogy to `Model`
- define a function called `createDatas(*models)`, taking any number of `Model` or `GeometryModel` objects and returning corresponding data objects (see docstring for more details)
- use `buildModelsFromUrdf` and `createDatas` inside `RobotWrapper`

Thus, the recommended Python workflow is now the following:
```
model, collision_model, visual_model = pin.buildModelsFromUrdf(model_path, hint_list, root_joint)
data, collision_data, visual_data = pin.createDatas(model,collision_model,visual_model)
```
saving a lot of typing in cases when `RobotWrapper` is not used (two lines instead of six).

A few notes:
- As an exception to the general rule of implementing things in C++ and exposing them in Python, I implemented the new methods purely in Python. They are in [shortcuts.py](https://github.com/gabrielebndn/pinocchio/blob/c7e39ca8da3971e31914b6f247f006ead79c4664/bindings/python/pinocchio/shortcuts.py). This for two reasons:
  - These are basically shortcuts thought for rapid prototyping and for reducing the dependency on `RobotWrapper`. As such, they are not really needed in C++
  - I figured it would be much easier to write and to maintain: implementing it in C++ would have meant creating at least 2 or 3 C++ overloads and and exposing 7 or 8 signatures depending whether the package directory is a string, a list, a vector, or not provided, and wether the mesh loader is provided or not. This way it is much more compact.
- I greatly simplified the implementation of `buildModelsFromUrdf` with respect to the original code in `RobotWrapper`. This because it seemed to me that `RobotWrapper` was making a lot of redundant safety checks, such as `if "buildGeomFromUrdf" not in dir(pin)` (true if and only if the URDF module is not provided) or wether the elements in `package_dirs` are indeed all strings (let boost-python handle that). If you want to check, you may put extra care in reviewing commit 
8ae5cf4 (and possibly c7e39ca)

Hope this helps!

### DIFF
--- a/bindings/python/multibody/geometry-model.hpp
+++ b/bindings/python/multibody/geometry-model.hpp
@@ -43,6 +43,7 @@ namespace pinocchio
                                                             "Add a GeometryObject to a GeometryModel and set its parent joint by reading its value in model")
         .def("getGeometryId",&GeometryModel::getGeometryId)
         .def("existGeometryName",&GeometryModel::existGeometryName)
+        .def("createData",&GeometryModelPythonVisitor::createData)
 #ifdef PINOCCHIO_WITH_HPP_FCL
         .add_property("collisionPairs",
                       &GeometryModel::collisionPairs,
@@ -71,7 +72,7 @@ namespace pinocchio
         ;
       }
       
-      
+      static GeometryData createData(const GeometryModel & geomModel) { return GeometryData(geomModel); }
 
       /* --- Expose --------------------------------------------------------- */
       static void expose()
--- a/bindings/python/parsers/parsers.hpp
+++ b/bindings/python/parsers/parsers.hpp
@@ -95,9 +95,8 @@ namespace pinocchio
                         const GeometryType type
                         )
       {
-        std::vector<std::string> hints;
         GeometryModel geometry_model;
-        pinocchio::urdf::buildGeom(model,filename,type,geometry_model,hints);
+        pinocchio::urdf::buildGeom(model,filename,type,geometry_model);
         
         return geometry_model;
       }
@@ -124,6 +123,19 @@ namespace pinocchio
       {
         std::vector<std::string> package_dirs_ = extractList<std::string>(package_dirs);
         return buildGeomFromUrdf(model,filename,package_dirs_,type);
+      }
+
+      static GeometryModel
+      buildGeomFromUrdf(const Model & model,
+                        const std::string & filename,
+                        const std::string & package_dir,
+                        const GeometryType type
+                        )
+      {
+        GeometryModel geometry_model;
+        pinocchio::urdf::buildGeom(model,filename,type,geometry_model,package_dir);
+
+        return geometry_model;
       }
 
 #ifdef PINOCCHIO_WITH_HPP_FCL
@@ -166,7 +178,21 @@ namespace pinocchio
         std::vector<std::string> package_dirs_ = extractList<std::string>(package_dirs);
         return buildGeomFromUrdf(model,filename,package_dirs_,type,meshLoader);
       }
-      
+
+      static GeometryModel
+      buildGeomFromUrdf(const Model & model,
+                        const std::string & filename,
+                        const std::string & package_dir,
+                        const GeometryType type,
+                        const fcl::MeshLoaderPtr& meshLoader
+                        )
+      {
+        GeometryModel geometry_model;
+        pinocchio::urdf::buildGeom(model,filename,type,geometry_model,package_dir,meshLoader);
+
+        return geometry_model;
+      }
+
       BOOST_PYTHON_FUNCTION_OVERLOADS(removeCollisionPairs_overload,
                                       srdf::removeCollisionPairs,
                                       3,4)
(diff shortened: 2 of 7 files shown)


jcarpent:
@gabrielebndn I also want to thank you a lot for this nice and useful PR.

gabrielebndn:
> I think a function should not change the number of returned arguments according to the input arguments. This might lead to bugs easily.
> Can you change `buildModelsFromUrdf` accordingly?

I had thought about it, but I think it is nice to keep it this way. It is more flexible.
What if only one of the two geometry models is specified in the package? Even if both are present, if only one is needed, why wasting memory and time loading a useless model?

The alternatives would be:
- not providing the case for a single model and let the user just employ the standard separate algorithms (I don't like this solution)
- creating a separate method, with a different name, taking a `GeometryType` parameter and returning only the Model and one GeometryModel of the specified type (uselessly heavy, I think)
- keeping the `geometry_type` argument, but returning `None` for the non-specified model (leading to uglier user code and not removing the problem of bugs).

The current usage is sufficiently clear and well-documented to ensure that users employ this method correctly, I think (especially now that I have slightly updated the docstring).

jcarpent:
The problem of having different return type according to the type of arguments is really error-prone. I would like to avoid such a bad situation.

I would suggest implementing two functions: buildModel and buildGeomModel for simplicity.
And maybe, but not needed, a function which calls both.

jcarpent:
@gabrielebndn What do you think?

gabrielebndn:
@jcarpent as I said, I think it would just be heavier to have two different functions, conceptually they would be quite similar, but the name would be different, it would be more confusing than not, I think.

I think the main source of problems would be knowing which one is the collision model and which one is the visual model, but this would not be eliminated by splitting the two functions, it would be just the same, wether you are using the two-outputs function or the three-outputs one.

Unless someone is doing weird things such as keeping the whole output as a tuple and assuming stuff about it (which they shouldn't), errors in the number of output values would be immediately clear as a runtime exception when unpacking too many or too few arguments.

If we do split the two cases, I would turn the existing unified function into an identical auxiliary function, only hidden (`_` at the beginning) while the user functions would call that one. This way we would avoid code duplication.

Also, if we split the two cases, we should find clear, unambiguous names. `buildModel` and `buildGeomModel` are not good, because they are too easily mixed with existing methods. Maybe something like `buildAllModelsFromUrdf` for the three-outputs case? What about the two-outputs case? `buildModelAndGeometryModelFromUrdf`? Long and boring to type. `buildTwoModelsFromUrdf`? Sounds funny.

jcarpent:
`twoModels` is not meaningful I think. 
I think we should keep transform the tupple return type into a dict , where each field corresponds to: `model`, `collision_model`, `visual_model` and set to None the respected fields that we do not care.
In this way, the number of return arguments remains the same.
In the end, I just want to avoid multiple different namings that may be explained in the doc, but are not reveled in the naming of the function.

gabrielebndn:
No, a dict would make it longer to retrieve the values and defeat the purpose of coinciseness. I'd rather have two functions, plus a map is definitely an overkill I think.
buildSpecifiedModels + buildAllModels?

jcarpent:
> No, a dict would make it longer to retrieve the values and defeat the purpose of coinciseness. I'd rather have two functions, plus a map is definitely an overkill I think.
> buildSpecifiedModels + buildAllModels?

You're in Python, and you're calling it only one. A dict seems a reasonable solution.
Another solution (but not a Python standard) would be to return a struct (empty class) by filling the different models as field of this class.

gabrielebndn:
> Another solution (but not a Python standard) would be to return a struct (empty class) by filling the different models as field of this class.

Then everybody would be passing around that struct, which we want to avoid.

I don't see the point of a map for something which we know is going to be either two or three known objects, and I think resulting user code would be less nice.

As I said, my preference is to leave things as they are, but I am open to splitting in two functions. I don't see what is wrong in having two different functions doing two different things

jcarpent:
The naming of the functions is a bit weird: from the naming of the function, the user should be able to guess what the function is doing.
I would suggest keep the buildModelsFromURDF and then return a single type:

- either a tuple where each element corresponds to model, collision_model, visual_model with respect to this order
- either a dict

And I do not think that the option geometry_type is necessary or the current behiavior is the best I woudl say. If geometry_type is None, it would mean that the user does not care of Geometries.

gabrielebndn:
> either a tuple where each element corresponds to model, collision_model, visual_model with respect to this order

Then, not giving the option of retrieving one geometry model only? Or leaving None?

> If geometry_type is None, it would mean that the user does not care of Geometries.

None is widely used to represent a default, but I Agee it might be confusing. One alternative is replacing geometry_type by a list called geometry_types, which is either [] or [COLLISION,VISUAL] by default. Then, the returned number of arguments would still be variable, but according to the list. I actually like this behavior! It is more intuitive, and less error-prone I think

jcarpent:
Yes, I think not giving the option is fair. And if the object is not used, it will be destroyed. I like your proposal, and I will vote for it definitely.

gabrielebndn:
> I like your proposal, and I will vote for it definitely.

You mean the list or not giving the option? If the list, which one do you think is a better default, [] or [COLLISION, VISUAL]? In the first case, the default would be equivalent to buildModel, quite useless, and it would be more verbose if you want three arguments, but it would be more intuitive and more robust to human error, probably. Either one is ok for me

Either way, I'll probably think about it tomorrow morning

jcarpent:
I would go for the default list with full options.
Thanks a lot for including this change in your PR.

gabrielebndn:
Pushed changes. Thanks to you for your feedback!

jcarpent:
@gabrielebndn Thanks a lot!

PR #795, opened 2019-05-07
Make visualization functionalities independent from RobotWrapper, closes #663

### LINKED ISSUE #663: Separating the functionnalities of RobotWrapper

gabrielebndn:
This PR allows to use visualization routines independently from `RobotWrapper`.
Along with #791, this closes #663.

In short, this is what I did:
- create a submodule `pinocchio.display`
- create a class `pinocchio.display.AbstractDisplay`, which is meant to be the base class for all present and future displays. It defines a common interface and implements a few common functionalities
- create classes `pinocchio.display.GepettoDisplay` and `pinocchio.display.AbstractDisplay`, which subclass `AbstractDisplay` and implement the necessary functionalities for GepettoViewer and meshcat, respectively
- put a display instance inside `RobotWrapper` and delegate all previous `RobotWrapper` viewing methods to this object. The display is settable through `setDisplay`. If `initDisplay` is called without the display being set, it defaults to `GepettoDisplay`, without any warning. `initMeshcatDisplay` is deprecated.
- also, I added an algorithm called `scaleGeometryModel(gmodel,scale)` which allows to easily set the same scale (either a 3D-vector or a scalar) to each `GeometryObject` of a given `GeometryModel` (C++ with Python bindings).

To test `GepettoDisplay`: [examples/python/gepetto-viewer.py](https://github.com/gabrielebndn/pinocchio/blob/85b0c4a61d578fd3e1ec1ec80cb623fbab4d3ae7/examples/python/gepetto-viewer.py)

To test `MeshcatDisplay`: [examples/python/meshcat-viewer.py](https://github.com/gabrielebndn/pinocchio/blob/85b0c4a61d578fd3e1ec1ec80cb623fbab4d3ae7/examples/python/meshcat-viewer.py)

To test them together: [examples/python/load-urdf.py](https://github.com/gabrielebndn/pinocchio/blob/85b0c4a61d578fd3e1ec1ec80cb623fbab4d3ae7/examples/python/load-urdf.py)

To test them within `RobotWrapper`: [examples/python/robot-wrapper-viewer.py](https://github.com/gabrielebndn/pinocchio/blob/85b0c4a61d578fd3e1ec1ec80cb623fbab4d3ae7/examples/python/robot-wrapper-viewer.py)

I decided to put the displays in a separate submodule because it seemed more orderly and because
> Namespaces are one honking great idea -- let's do more of those!

The changes in `RobotWrapper` should be perfectly backward-compatible. No code should be broken because of this PR. The only partial exception is `initMeshcatDisplay`, which has been deprecated because it should be handled differently, but still works.

A few notes about `MeshcatDisplay`:
- I installed Meshcat using `pip install --user meshcat`. However, I could not start the server with `meshcat-server`. What do I have to do in order to have this command?
- Compared to the previous implementation, the new one is improved.
  - objects are added to the viewer respecting the tree structure
  - scaling is handled. With this modification and with the addition of `scaleGeometryModel`, [examples/python/meshcat-viewer.py](https://github.com/gabrielebndn/pinocchio/blob/85b0c4a61d578fd3e1ec1ec80cb623fbab4d3ae7/examples/python/meshcat-viewer.py) is showing the Romeo model with the correct scaling
- Compared to the previous implementation, the API is a bit different. What used to be `initMeshcatDisplay` is now split between `initDisplay` and `loadDisplayModel`. This for keeping things separate and for uniformity with `GepettoDisplay`
- I chose not to automatically perform `viewer.open()` because I figured that's not how you guys are working, maybe it can be a boolean option of `initDisplay`, defaulting to false?
- Right now, `MeshcatDisplay` only loads the visual objects. It might be interesting to load the collision objects too, in order to display them if desired. This is what `GepettoDisplay` is doing. Adding this functionality would make the two viewers more similar to each other. It can certainly be done very easily. The main reason I did not do it is because I did not know how to toggle visibility of objects through Python. I'll leave it for others to implement
- Right now, `MeshcatDisplay` is not able to handle textures. This is the reason why Romeo does not show up unless the color is manually set. However, it seems Meshcat itself is able to handle textures, so I think it should be implemented. This would remove the need of specifying the color for Romeo. Again, I'll leave it for others.

@wxmerkt and @matthieuvigne, I would very much appreciate your feedback on this.

### DIFF
--- a/bindings/python/pinocchio/display/__init__.py
+++ b/bindings/python/pinocchio/display/__init__.py
@@ -1,4 +1,0 @@
-
-from .abstract_display import AbstractDisplay
-from .gepetto_display import GepettoDisplay
-from .meshcat_display import MeshcatDisplay
--- a/bindings/python/pinocchio/display/__init__.py
+++ b/bindings/python/pinocchio/display/__init__.py
@@ -1,3 +1,4 @@
 
 from abstract_display import AbstractDisplay
 from gepetto_display import GepettoDisplay
+from meshcat_display import MeshcatDisplay
--- a/bindings/python/pinocchio/display/__init__.py
+++ b/bindings/python/pinocchio/display/__init__.py
@@ -1,0 +1,2 @@
+
+from abstract_display import AbstractDisplay
--- a/bindings/python/pinocchio/display/__init__.py
+++ b/bindings/python/pinocchio/display/__init__.py
@@ -1,2 +1,3 @@
 
 from abstract_display import AbstractDisplay
+from gepetto_display import GepettoDisplay
--- a/bindings/python/pinocchio/display/__init__.py
+++ b/bindings/python/pinocchio/display/__init__.py
@@ -1,4 +1,4 @@
 
-from abstract_display import AbstractDisplay
-from gepetto_display import GepettoDisplay
-from meshcat_display import MeshcatDisplay
+from .abstract_display import AbstractDisplay
+from .gepetto_display import GepettoDisplay
+from .meshcat_display import MeshcatDisplay
--- a/bindings/python/pinocchio/display/abstract_display.py
+++ b/bindings/python/pinocchio/display/abstract_display.py
@@ -1,0 +1,70 @@
+from .. import libpinocchio_pywrap as pin
+from ..shortcuts import buildModelsFromUrdf, createDatas
+
+import time
+
+class AbstractDisplay(object):
+    """Pinocchio displays are employed to easily display a model at a given configuration.
+    AbstractDisplay is not meant to be directly employed, but only to provide a uniform interface and a few common methods.
+    New displays should extend this class and override its methods as neeeded.
+    """
+
+    @classmethod
+    def BuildFromURDF(cls,filename, package_dirs=None, root_joint=None, verbose=False, meshLoader=None):
+        """Construct a display from the given URDF file"""
+
+        display = cls()
+        display.initFromURDF(filename, package_dirs, root_joint, verbose, meshLoader)
+        return display
+
+    def initFromURDF(self,filename, package_dirs=None, root_joint=None, verbose=False, meshLoader=None):
+        """Initialize a display from the given URDF file"""
+
+        model, collision_model, visual_model = buildModelsFromUrdf(filename, package_dirs, root_joint, verbose, meshLoader)
+        cls = type(self)
+        cls.__init__(self,model=model,collision_model=collision_model,visual_model=visual_model)
+
+    def __init__(self, model = pin.Model(), collision_model = None, visual_model = None):
+        """Construct a display from the given model, collision model, and visual model"""
+
+        self.model = model
+        self.collision_model = collision_model
+        self.visual_model = visual_model
+
+        self.data, self.collision_data, self.visual_data = createDatas(model,collision_model,visual_model)
+
+    def getViewerNodeName(self, geometry_object, geometry_type):
+        """Return the name of the geometry object inside the viewer"""
+        pass 
+
+    def initDisplay(self, *args, **kwargs):
+        """Init display by loading the gui and creating a window."""
+        pass
+
+    def loadDisplayModel(self, *args, **kwargs):
+        """Create the scene displaying the robot meshes in gepetto-viewer"""
+        pass
+
+    def display(self, q):
+        """Display the robot at configuration q in the viewer by placing all the bodies."""
+        pass
+
+    def displayCollisions(self,visibility):
+        """Set whether to diplay collision objects or not"""
+        pass
+ 
+    def displayVisuals(self,visibility):
+        """Set whether to diplay visual objects or not"""
+        pass
+
+    def play(self, q_trajectory, dt):
+        """Play a trajectory with given time step"""
+        for k in range(q_trajectory.shape[1]):
+            t0 = time.time()
+            self.display(q_trajectory[:, k])
+            t1 = time.time()
+            elapsed_time = t1 - t0
+            if elapsed_time < dt:
+                time.sleep(dt - elapsed_time)
+
+__all__ = ['AbstractDisplay']
--- a/bindings/python/pinocchio/display/abstract_display.py
+++ b/bindings/python/pinocchio/display/abstract_display.py
@@ -9,27 +9,18 @@ class AbstractDisplay(object):
     New displays should extend this class and override its methods as neeeded.
     """
 
-    @classmethod
-    def BuildFromURDF(cls,filename, package_dirs=None, root_joint=None, verbose=False, meshLoader=None):
-        """Construct a display from the given URDF file"""
+    def __init__(self, model = pin.Model(), collision_model = None, visual_model = None, copy_models=False):
+        """Construct a display from the given model, collision model, and visual model.
+        If copy_models is True, the models are copied. Otherwise, they are simply kept as a reference."""
 
-        display = cls()
-        display.initFromURDF(filename, package_dirs, root_joint, verbose, meshLoader)
-        return display
-
-    def initFromURDF(self,filename, package_dirs=None, root_joint=None, verbose=False, meshLoader=None):
-        """Initialize a display from the given URDF file"""
-
-        model, collision_model, visual_model = buildModelsFromUrdf(filename, package_dirs, root_joint, verbose, meshLoader)
-        cls = type(self)
-        cls.__init__(self,model=model,collision_model=collision_model,visual_model=visual_model)
-
-    def __init__(self, model = pin.Model(), collision_model = None, visual_model = None):
-        """Construct a display from the given model, collision model, and visual model"""
-
-        self.model = model
-        self.collision_model = collision_model
-        self.visual_model = visual_model
+        if copy_models:
+            self.model = model.copy()
+            self.collision_model = collision_model.copy()
+            self.visual_model = visual_model.copy()
+        else:
+            self.model = model
+            self.collision_model = collision_model
+            self.visual_model = visual_model
 
         self.data, self.collision_data, self.visual_data = createDatas(model,collision_model,visual_model)
 
@@ -50,11 +41,11 @@ class AbstractDisplay(object):
         pass
 
     def displayCollisions(self,visibility):
-        """Set whether to diplay collision objects or not"""
+        """Set whether to display collision objects or not"""
         pass
  
     def displayVisuals(self,visibility):
-        """Set whether to diplay visual objects or not"""
+        """Set whether to display visual objects or not"""
         pass
 
     def play(self, q_trajectory, dt):
--- a/examples/python/robot-wrapper-viewer.py
+++ b/examples/python/robot-wrapper-viewer.py
@@ -1,19 +1,5 @@
 ##
-## Copyright (c) 2018 CNRS
-##
-## This file is part of Pinocchio
-## Pinocchio is free software: you can redistribute it
-## and/or modify it under the terms of the GNU Lesser General Public
-## License as published by the Free Software Foundation, either version
-## 3 of the License, or (at your option) any later version.
-##
-## Pinocchio is distributed in the hope that it will be
-## useful, but WITHOUT ANY WARRANTY; without even the implied warranty
-## of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
-## General Lesser Public License for more details. You should have
-## received a copy of the GNU Lesser General Public License along with
-## Pinocchio If not, see
-## <http:##www.gnu.org/licenses/>.
+## Copyright (c) 2018-2019 CNRS
 ##
 
 ##
@@ -24,20 +10,18 @@
 import pinocchio as pin
 from pinocchio.robot_wrapper import RobotWrapper
 from pinocchio.display import *
+import os
 
 DISPLAY = None
 # DISPLAY = GepettoDisplay
 # DISPLAY = MeshcatDisplay
 
-## Load Romeo with RomeoWrapper
-import os
-current_path = os.getcwd()
-
-# The model of Romeo is contained in the path PINOCCHIO_GIT_REPOSITORY/models/romeo
-model_path = current_path + "/" + "../../models/romeo"
+# Load the URDF model with RobotWrapper
+# Conversion with str seems to be necessary when executing this file with ipython
+current_path =  str(os.path.dirname(os.path.abspath(__file__)))
+model_path = str(os.path.abspath(os.path.join(current_path, '../../models/romeo')))
 mesh_dir = model_path
-urdf_filename = "romeo_small.urdf"
-urdf_model_path = model_path + "/romeo_description/urdf/" + urdf_filename
+urdf_model_path = str(os.path.abspath(os.path.join(model_path, 'romeo_description/urdf/romeo_small.urdf')))
 
 robot = RobotWrapper.BuildFromURDF(urdf_model_path, mesh_dir, pin.JointModelFreeFlyer())
 
@@ -54,12 +38,12 @@ com2 = pin.centerOfMass(model,data,q0)
 
 ## load model into gepetto-gui
 if DISPLAY:
-    robot.setDisplay(DISPLAY())
+    robot.setDisplay(DISPLAY(),copy_models = (DISPLAY == MeshcatDisplay))
     robot.initDisplay()
     if DISPLAY == GepettoDisplay:
         robot.loadDisplayModel("pinocchio")
     elif DISPLAY == MeshcatDisplay:
-        pin.setGeometryMeshScales(robot.visual_model,0.01)
+        pin.setGeometryMeshScales(robot.disp.visual_model,0.01)
         robot.viewer.open()
         robot.loadDisplayModel("pinocchio",color=[0.,0.,0.,1.])
     else:
--- a/examples/python/robot-wrapper-viewer.py
+++ b/examples/python/robot-wrapper-viewer.py
@@ -39,11 +39,11 @@ com2 = pin.centerOfMass(model,data,q0)
 ## load model into gepetto-gui
 if VISUALIZER:
     robot.setVisualizer(VISUALIZER())
-    robot.initDisplay()
+    robot.initViewer()
     if VISUALIZER == MeshcatVisualizer:
-        robot.loadDisplayModel("pinocchio", color=[0., 0., 0., 1.])
+        robot.loadViewerModel("pinocchio", color=[0., 0., 0., 1.])
     else:
-        robot.loadDisplayModel("pinocchio")
+        robot.loadViewerModel("pinocchio")
     robot.display(q0)
 
 raw_input("Press enter to exit.")
--- a/examples/python/robot-wrapper-viewer.py
+++ b/examples/python/robot-wrapper-viewer.py
@@ -9,12 +9,12 @@
 
 import pinocchio as pin
 from pinocchio.robot_wrapper import RobotWrapper
-from pinocchio.display import *
+from pinocchio.visualize import *
 import os
 
-DISPLAY = None
-# DISPLAY = GepettoDisplay
-# DISPLAY = MeshcatDisplay
+VISUALIZER = None
+# VISUALIZER = GepettoVisualizer
+# VISUALIZER = MeshcatVisualizer
 
 # Load the URDF model with RobotWrapper
 # Conversion with str seems to be necessary when executing this file with ipython
@@ -37,10 +37,10 @@ com = robot.com(q0)
 com2 = pin.centerOfMass(model,data,q0)
 
 ## load model into gepetto-gui
-if DISPLAY:
-    robot.setDisplay(DISPLAY())
+if VISUALIZER:
+    robot.setVisualizer(VISUALIZER())
     robot.initDisplay()
-    if DISPLAY == MeshcatDisplay:
+    if VISUALIZER == MeshcatVisualizer:
         robot.loadDisplayModel("pinocchio", color=[0., 0., 0., 1.])
     else:
         robot.loadDisplayModel("pinocchio")
--- a/examples/python/robot-wrapper-viewer.py
+++ b/examples/python/robot-wrapper-viewer.py
@@ -38,16 +38,12 @@ com2 = pin.centerOfMass(model,data,q0)
 
 ## load model into gepetto-gui
 if DISPLAY:
-    robot.setDisplay(DISPLAY(),copy_models = (DISPLAY == MeshcatDisplay))
+    robot.setDisplay(DISPLAY())
     robot.initDisplay()
-    if DISPLAY == GepettoDisplay:
-        robot.loadDisplayModel("pinocchio")
-    elif DISPLAY == MeshcatDisplay:
-        pin.setGeometryMeshScales(robot.disp.visual_model,0.01)
-        robot.viewer.open()
-        robot.loadDisplayModel("pinocchio",color=[0.,0.,0.,1.])
+    if DISPLAY == MeshcatDisplay:
+        robot.loadDisplayModel("pinocchio", color=[0., 0., 0., 1.])
     else:
-        raise Exception("Unknown display")
+        robot.loadDisplayModel("pinocchio")
     robot.display(q0)
 
 raw_input("Press enter to exit.")
--- a/examples/python/robot-wrapper-viewer.py
+++ b/examples/python/robot-wrapper-viewer.py
@@ -59,7 +59,7 @@ if DISPLAY:
     if DISPLAY == GepettoDisplay:
         robot.loadDisplayModel("pinocchio")
     elif DISPLAY == MeshcatDisplay:
-        pin.scaleGeometryModel(robot.visual_model,0.01)
+        pin.setGeometryMeshScales(robot.visual_model,0.01)
         robot.viewer.open()
         robot.loadDisplayModel("pinocchio",color=[0.,0.,0.,1.])
     else:
--- a/examples/python/robot-wrapper-viewer.py
+++ b/examples/python/robot-wrapper-viewer.py
@@ -22,6 +22,7 @@
 ##
 
 import pinocchio as pin
+from pinocchio.robot_wrapper import RobotWrapper
 from pinocchio.display import *
 
 DISPLAY = None
@@ -38,27 +39,31 @@ mesh_dir = model_path
 urdf_filename = "romeo_small.urdf"
 urdf_model_path = model_path + "/romeo_description/urdf/" + urdf_filename
 
-model, collision_model, visual_model = pin.buildModelsFromUrdf(urdf_model_path, mesh_dir, pin.JointModelFreeFlyer())
+robot = RobotWrapper.BuildFromURDF(urdf_model_path, mesh_dir, pin.JointModelFreeFlyer())
 
-# In this example, we do not explicitely need collision data and visual data
-data = model.createData()
+# alias
+model = robot.model
+data = robot.data
 
 # do whatever, e.g. compute the center of mass position expressed in the world frame
-q0 = pin.neutral(model)
-com = pin.centerOfMass(model,data,q0)
+q0 = robot.q0
+com = robot.com(q0)
+
+# This last command is similar to:
+com2 = pin.centerOfMass(model,data,q0)
 
 ## load model into gepetto-gui
 if DISPLAY:
-    display = DISPLAY(model, collision_model, visual_model)
-    display.initDisplay()
+    robot.setDisplay(DISPLAY())
+    robot.initDisplay()
     if DISPLAY == GepettoDisplay:
-        display.loadDisplayModel("pinocchio")
+        robot.loadDisplayModel("pinocchio")
     elif DISPLAY == MeshcatDisplay:
-        pin.scaleGeometryModel(visual_model,0.01)
-        display.viewer.open()
-        display.loadDisplayModel("pinocchio",color=[0.,0.,0.,1.])
+        pin.scaleGeometryModel(robot.visual_model,0.01)
+        robot.viewer.open()
+        robot.loadDisplayModel("pinocchio",color=[0.,0.,0.,1.])
     else:
         raise Exception("Unknown display")
-    display.display(q0)
+    robot.display(q0)
 
 raw_input("Press enter to exit.")
--- a/src/algorithm/geometry.hpp
+++ b/src/algorithm/geometry.hpp
@@ -50,6 +50,33 @@ namespace pinocchio
                                        const GeometryModel & geomModel,
                                        GeometryData & geomData);
 
+
+  ///
+  /// \brief     Apply a scaling to each GeometryObject inside the GeometryModel
+  ///
+  /// param[in]  geomModel The geometry model containing the collision objects.
+  /// param[in]  meshScale The scale to be applied to each GeometryObject
+  ///
+  template<typename Vector3Like>
+  inline void scaleGeometryModel(GeometryModel & geomModel, const Eigen::MatrixBase<Vector3Like> & meshScale)
+  {
+    EIGEN_STATIC_ASSERT_VECTOR_SPECIFIC_SIZE(Vector3Like,3);
+    for(GeomIndex index=0; index<geomModel.ngeoms; index++)
+      geomModel.geometryObjects[index].meshScale = meshScale;
+  }
+
+
+  ///
+  /// \brief     Apply a scaling to each GeometryObject inside the GeometryModel, equally in all directions
+  ///
+  /// param[in]  geomModel The geometry model containing the collision objects.
+  /// param[in]  meshScale The scale, to be applied to each GeometryObject, equally in all directions
+  ///
+  inline void scaleGeometryModel(GeometryModel & geomModel, const double meshScale)
+  {
+    scaleGeometryModel(geomModel, Eigen::Vector3d::Constant(meshScale));
+  }
+
 #ifdef PINOCCHIO_WITH_HPP_FCL
 
   ///
--- a/src/algorithm/geometry.hpp
+++ b/src/algorithm/geometry.hpp
@@ -50,31 +50,29 @@ namespace pinocchio
                                        const GeometryModel & geomModel,
                                        GeometryData & geomData);
 
-
   ///
-  /// \brief     Apply a scaling to each GeometryObject inside the GeometryModel
+  /// \brief     Set a mesh scaling vector to each GeometryObject contained in the the GeometryModel.
   ///
   /// param[in]  geomModel The geometry model containing the collision objects.
   /// param[in]  meshScale The scale to be applied to each GeometryObject
   ///
   template<typename Vector3Like>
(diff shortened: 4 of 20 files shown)


jmirabel:
> also, I added an algorithm called scaleGeometryModel(gmodel,scale) which
allows to easily set the same scale (either a 3D-vector or a scalar) to
each GeometryObject of a given GeometryModel (C++ with Python bindings).

What do you mean (viewer, hpp-fcl, both) ? Why do you have to do that ?

gabrielebndn:
@jmirabel, ultimately, the method amounts to a loop over GeometryModel.geometryObjects

```
for(GeomIndex index=0; index<geomModel.ngeoms; index++)
    geomModel.geometryObjects[index].meshScale = meshScale;
```

It allows to scale a whole model.
I needed to scale the Romeo model in order to display it properly in Meshcat. I saw the code based on GepettoViewer was using this property. As I figured this need might be relatively common, I just thought to make a simple algorithm to address it.

bindings/python/pinocchio/display/__init__.py
```diff
@@ -0,0 +1,4 @@
+
+from abstract_display import AbstractDisplay
```
nim65s: I think you should use`from .abstract_display import AbstractDisplay` for python 3, and the same on the next lines.

gabrielebndn: You are totally right. I'll fix this.

jmirabel:
It sounds weird to me that you have to change the scale for displaying the
model. Either there is a bug in the creation of the geometry model or the
meshes in the viewer and in Pinocchio won't have the same scale...

Shouldn't the scale be set once and for all by the code that creates the
geometries ? I thought this was the case...

wxmerkt:
@gabrielebndn Thank you for your hard work on creating an abstract interface to visualization tools. Indeed, while the MeshCat API/viewer supports scale (cf. https://github.com/rdeits/meshcat/blob/4fc7279bfd3557d437ebe6210f5e2d70c54b7611/src/index.js#L207-L209), the Python wrapper currently does not have it implemented yet (cf. https://github.com/rdeits/meshcat-python/issues/40).

gabrielebndn:
> It sounds weird to me that you have to change the scale for displaying the model. Either there is a bug in the creation of the geometry model or the meshes in the viewer and in Pinocchio won't have the same scale... Shouldn't the scale be set once and for all by the code that creates the geometries ? I thought this was the case...

@jmirabel I do not fully understand where the problem comes from as I do not know the internal mechanisms of GepettoViewer or Meshcat. There is a 100x scale factor.
It seems a problem of measurement units. It's as if the Romeo meshes were in centimeters, whereas Meshcat expects them in meters (just a hypothesis). Cf. https://github.com/stack-of-tasks/pinocchio/pull/662#issuecomment-463814384.
It seems a problem with the Romeo model itself. But then, I do not know why GepettoViewer displays it correctly. What measurement units does GepettoViewer expect? What measurement units does Meshcat expect?

What do you mean with "_Shouldn't the scale be set once and for all by the code that creates the geometries?_" You mean, by `MeshcatDisplay`, when the meshes are loaded inside the viewer? This I couldn't do because as @wxmerkt says there are no bindings for this, that's why I resorted to the workaround of using a transformation when displaying

bindings/python/pinocchio/display/abstract_display.py
```diff
@@ -0,0 +1,70 @@
...
+import time
+
+class AbstractDisplay(object):
+    """Pinocchio displays are employed to easily display a model at a given configuration.
+    AbstractDisplay is not meant to be directly employed, but only to provide a uniform interface and a few common methods.
+    New displays should extend this class and override its methods as neeeded.
+    """
+
+    @classmethod
+    def BuildFromURDF(cls,filename, package_dirs=None, root_joint=None, verbose=False, meshLoader=None):
```
jcarpent: This method is really specific to Gepetto viewer and should be set as a common method to all the viewers.
In addition, this option is in Gepetto viewer for historical reason, but Pinocchio allows to bypass it easily.

bindings/python/pinocchio/display/abstract_display.py
```diff
@@ -0,0 +1,70 @@
...
+        """Construct a display from the given URDF file"""
+
+        display = cls()
+        display.initFromURDF(filename, package_dirs, root_joint, verbose, meshLoader)
+        return display
+
+    def initFromURDF(self,filename, package_dirs=None, root_joint=None, verbose=False, meshLoader=None):
```
jcarpent: The same comment applies here.

bindings/python/pinocchio/display/abstract_display.py
```diff
@@ -0,0 +1,70 @@
...
+        """Initialize a display from the given URDF file"""
+
+        model, collision_model, visual_model = buildModelsFromUrdf(filename, package_dirs, root_joint, verbose, meshLoader)
+        cls = type(self)
+        cls.__init__(self,model=model,collision_model=collision_model,visual_model=visual_model)
+
+    def __init__(self, model = pin.Model(), collision_model = None, visual_model = None):
+        """Construct a display from the given model, collision model, and visual model"""
+
+        self.model = model
```
jcarpent: Should we copy the model here? I think so.

gabrielebndn:
> @gabrielebndn Thank you for your hard work on creating an abstract interface to visualization tools. Indeed, while the MeshCat API/viewer supports scale (cf. https://github.com/rdeits/meshcat/blob/4fc7279bfd3557d437ebe6210f5e2d70c54b7611/src/index.js#L207-L209), the Python wrapper currently does not have it implemented yet (cf. [rdeits/meshcat-python#40](https://github.com/rdeits/meshcat-python/issues/40)).

@wxmerkt the way I am supporting scaling is by a little hack, i.e. by manually applying a scaling transformation at display time. I know about https://github.com/rdeits/meshcat-python/issues/40, that's exactly where I got this idea (see https://github.com/rdeits/meshcat-python/issues/40#issuecomment-474430463).

So, we might say that although Python Meshcat does not *directly* support scaling, it does with a little hack, and thats' how `MeshcatDisplay` is supporting it.

I agree the best would be to fix https://github.com/rdeits/meshcat-python/issues/40 and use the proper bindings, but previously the scale was not managed *at all*, and as for now this is valid code. Once https://github.com/rdeits/meshcat-python/issues/40 is fixed, it will be possible to change the implementation to something more fitting.

bindings/python/pinocchio/display/abstract_display.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def displayCollisions(self,visibility):
+        """Set whether to diplay collision objects or not"""
+        pass
+ 
+    def displayVisuals(self,visibility):
+        """Set whether to diplay visual objects or not"""
+        pass
+
+    def play(self, q_trajectory, dt):
```
jcarpent: I have implemented this play function a while ago, but it is not safe and it is not well synchronized. Can we make it deprecated?

bindings/python/pinocchio/display/abstract_display.py
```diff
@@ -0,0 +1,70 @@
+from .. import libpinocchio_pywrap as pin
+from ..shortcuts import buildModelsFromUrdf, createDatas
+
+import time
+
+class AbstractDisplay(object):
```
jcarpent: Maybe AbstractViewer is better here as naming.

examples/python/robot-wrapper-viewer.py
```diff
@@ -0,0 +1,69 @@
+##
+## Copyright (c) 2018 CNRS
+##
+## This file is part of Pinocchio
```
jcarpent: This signature is deprecated and has to be removed.

src/algorithm/geometry.hpp
```diff
@@ -50,6 +50,33 @@ namespace pinocchio
                                        const GeometryModel & geomModel,
                                        GeometryData & geomData);
 
+
+  ///
+  /// \brief     Apply a scaling to each GeometryObject inside the GeometryModel
```
jcarpent: `a scaling vector to each GeometryObject contained in the the GeometryModel`

bindings/python/pinocchio/display/abstract_display.py
gabrielebndn: Why is it specific to Gepetto viewer? All viewer will need the model, the visual model, and the collision model. You mean it should *not* be set as a common method? Why? It's a convenient way to build the viewer.

bindings/python/pinocchio/display/abstract_display.py
gabrielebndn: I don't see why. This is not what we were doing in `RobotWrapper` anyway. The model object is basically immutable, for the use we make of it. This way we can share some information. I thought this was one of the reasons for the whole idea of splitting model and data in the first place. Notice the data objects are not shared, but they are created with the viewer.

bindings/python/pinocchio/display/abstract_display.py
jcarpent: Because Gepetto Viewer comes with an internal URDF parser, which allows the direct loading of the geometries. But this is not the case of all the viewers we target, and a simple way to a load a model inside a viewer is simply by looking to the meshes path stored inside the Geometries. This is the main way to achieve it, without requiring urdfdom and others to be connected to your viewer.

jcarpent: In addition, a viewer does not that there is a kinematic tree behind the scene. Its main functionnality is to display meshes.

bindings/python/pinocchio/display/abstract_display.py
gabrielebndn: Yeah, I don't like it a lot, it does not handle time very well at all. It is practical though, and it can always be overridden by the more specific viewers, if need be. Rather than deprecating it, I would rather leave it here with some sort of caveats in the doc, maybe keeping in mind it should be improved (though I do not see how).

bindings/python/pinocchio/display/abstract_display.py
jcarpent: Yes, but RobotWrapper is not well done. As you are working on the splitting, we should do it completely. 
The splitting between model and data goes beyond this scope, it is very helpful for performances issues.
Here, I would like to think twice before merging how far the memory used for the viewer has to be shared with the rest of the application. One thing to add as the last argument should be a copy option with a default value to False I think to really allow a strict splitting between all the functionalities.

examples/python/robot-wrapper-viewer.py
gabrielebndn: Uh, You mean, I should simply remove all the copyright-related comments on top of the file? I actually, I don't know how this one ended up here. I will remove it.

src/algorithm/geometry.hpp
gabrielebndn: Ok

examples/python/robot-wrapper-viewer.py
jcarpent: You can keep the copyright, but you just need to remove the license.

bindings/python/pinocchio/display/abstract_display.py
gabrielebndn: So, the reason why I chose the word "Display", and not "Viewer", is that these are not really the viewers, but just wrappers around them. A "Display" instance usually contains a variable called `viewer`, which represents the actual viewer doing the job of displaying the stuff. So I think calling the Pinocchio classes `<Something>Viewer` might be confusing. The word "Display" comes from a tiny class called `Display` which sometimes pops up in lots of Gepetto code, and which is basically a primitive version of `GepettoDisplay` (thus, it is now made redundant).
I don't care a lot about this, but I would rather spare myself the work of changing all the namings.

bindings/python/pinocchio/display/abstract_display.py
gabrielebndn: In Python, it is absolutely normal to share objects through the same pointers.
Meh, I do not think the option would add a lot. If one *really* wants the display to use a copy of the models, they could simply pass copies in input, like this
```
display = GepettoDisplay(model.copy(),collision_model.copy(),visual_model.copy())
```

Actually, I see geometry models are *not* copyable in Python. I will make them copyable.

examples/python/robot-wrapper-viewer.py
gabrielebndn: Ok

bindings/python/pinocchio/display/abstract_display.py
jcarpent: I do not think adding this option is hard and may save some times and energy beforehand. 
I think this is a feature requested to really make both independent, without having to use the signature you've just suggested.

bindings/python/pinocchio/display/abstract_display.py
jcarpent: I don't like the way of having two namings for the same functionality. In this case, the naming ViewerWrapper is better I would say.

bindings/python/pinocchio/display/abstract_display.py
gabrielebndn: Either there is a misunderstanding or I just do not follow you.

This method is just to instantiate the Pinocchio `Model` and the two `GeometryModel` objects.
If they are stored in a URDF, you need to read them, no matter the viewer.

The loading of the meshes inside the viewer is then handled by `loadDisplayModel`. This method is specialized for each viewer.

> a viewer does not know that there is a kinematic tree behind the scene. Its main purpose is to display meshes.

The viewer does not, but the Display does. Method `display(q)` calls `forwardKinematics` and `updateGeometryPlacements` in order to compute the placement of each mesh for the given joint configuration `q`, and then passes the resulting placements to the viewer. You need knowledge about the models for that.

Otherwise, the `GeometryData` objects should be updated externally. It would then make sense to share them with external code. But I thought it was preferred to keep them separate.

jcarpent: Ok. I got your point.
In this case, I do not this worthy to add these two methods at the viewer level.

The ideal and normal usage case is:
1. load or create your Pinocchio model
2. display it calling the default constructor of the Viewer taking as arguments the various models

I would be afraid that adding the URDF methods will lead to bad habits at the user level, with some questions about the sharing of the model for instance. And from a conceptual point of view again, a generic Viewer should not be aware of the URDF convention and so on. It should only take care of the meshes or the basic geometries, nothing more. We already discussed all these points a while ago with @jmirabel, @nmansard, @olivier-stasse, @florent-lamiraux to name a few.

bindings/python/pinocchio/display/abstract_display.py
gabrielebndn: > I don't like the way of having two namings for the same functionality.

Why two namings? Classes are all consistently called `<Something>Display`. `GepettoDisplay` uses Gepetto Viewer, and `MeshcatDisplay` uses Meshcat Viewer. It seems perfectly consistent to me.

If you want ViewerWrapper, then I guess we should call the three classes `AbstractViewerWrapper`, `GepettoViewerWrapper` and `MeshcatViewerWrapper`? If you prefer... I think it's just a more complicated name and it will be a pain to change.

bindings/python/pinocchio/display/abstract_display.py
gabrielebndn: I think this is somehow unusual for Python. However, if you want it, it will be easy to add. Still requires to expose `.copy()` to `GeometryModel` though

bindings/python/pinocchio/display/abstract_display.py
jcarpent: Display seems not adapted, and we should focus on a single naming to name the same thing.
`MeshcatWrapper` is short and sufficient. `ViewerWrapper` for the parent class (we can forget the Abstract which is not needed). And GepettoViewer is the naming of the visualization tool, so we should go for `GepettoViewerWrapper`.

bindings/python/pinocchio/display/abstract_display.py
jcarpent: This is for instance how NumPy works too.

bindings/python/pinocchio/display/abstract_display.py
gabrielebndn: > I would be afraid that adding the URDF methods will lead to bad habits at the user level, with some questions about the sharing of the model for instance.

Uhm, I see what you mean. This encourages loading the models through the viewer and then keeping using the models from inside the viewer. Then, I may remove this functionalities so that usage of the dedicated tools (`buildModelsFromUrfs` and `createDatas`) is enforced.

jcarpent: Yes, this is exactly what I mean.

matthieuvigne:
@gabrielebndn Nice work implementing all this !

>I installed Meshcat using pip install --user meshcat. However, I could not start the server with meshcat-server. What do I have to do in order to have this command?

You should have ```meshcat-server``` available after installing the python package. You might have an issue with your path variables, could you check where ```pip``` is installing the package.

> @jmirabel I do not fully understand where the problem comes from as I do not know the internal mechanisms of GepettoViewer or Meshcat. There is a 100x scale factor.
It seems a problem of measurement units. It's as if the Romeo meshes were in centimeters, whereas Meshcat expects them in meters (just a hypothesis).

From what I observed when doing  #662 (comment), Meshcat wasn't the only tool having issues with the Romeo meshes: when I loaded them with the Meshlab viewer I had the same scalling issue. Both tools however worked fine with other meshes. I don't know how units are handled inside DAE files, but I would guess that the Romeo meshes current in Pinocchio are in cm and don't have unit declared - some viewer then default to meters, but not Gepetto.

jcarpent:
> I don't know how units are handled inside DAE files, but I would guess that the Romeo meshes current in Pinocchio are in cm and don't have unit declared - some viewer then default to meters, but not Gepetto.

@matthieuvigne There are some scaling fields in the DAE file that are correctly handled by OSG, but apparently not by the other viewers.

jcarpent:
@jmirabel @gabrielebndn 
A final point concerns the scaling of the meshes. I think having a function to automatically rescaled the meshes is good, but as mentioned by @jmirabel, this has to be handled for all the stages (visual + collision for FCL too). @gabrielebndn I have not got deep into the code, but do you know if it the case with your current implementation?

matthieuvigne:
>  There are some scaling fields in the DAE file that are correctly handled by OSG, but apparently not by the other viewers.

Sure, that explains what I saw, thanks :)

gabrielebndn:
> @jmirabel @gabrielebndn
> A final point concerns the scaling of the meshes. I think having a function to automatically rescaled the meshes is good, but as mentioned by @jmirabel, this has to be handled for all the stages (visual + collision for FCL too). @gabrielebndn I have not got deep into the code, but do you know if it the case with your current implementation?

@jcarpent 
- `GepettoDisplay` only includes code which was taken straight out of `RobotWrapper`. In this code, for each `GeometryObject`, property `meshScale` is read at the time of loading and set inside the viewer. Whether these values are equal for the collision model and for the visual model, this is a matter of the provided models.
- `MeshcatDisplay` only really loads visual models, because this is what I found (if anybody wants to implement plotting of the collision models, welcome, I think everything is in place to do that). Since I could not set the scale at the time of loading, what I am doing is for each `GeometryObject` I am reading property `meshScale` at the time of plotting and I am using this property to compute a scaling transformation.

Both solutions rely on reading property `meshScale` for each `GeometryObject` and acting accordingly. The `meshScale` is a property of the model, so if this is wrong, or not consistent between visual and collision, this is a problem of the model(s), but the plotting code is still correct in and by itself.

Method `scaleGeometryModel` is only a quick way to set `meshScale` property for each `GeometryObject` inside a `GeometryModel`. It is invoked by the user, should it be necessary. I found it *was* necessary for plotting Romeo inside Meshcat. Therefore, I am using it in the examples employing Meshcat, i.e. [examples/python/meshcat-viewer.py](https://github.com/gabrielebndn/pinocchio/blob/85b0c4a61d578fd3e1ec1ec80cb623fbab4d3ae7/examples/python/meshcat-viewer.py), [examples/python/load-urdf.py](https://github.com/gabrielebndn/pinocchio/blob/85b0c4a61d578fd3e1ec1ec80cb623fbab4d3ae7/examples/python/load-urdf.py) and [examples/python/robot-wrapper-viewer.py](https://github.com/gabrielebndn/pinocchio/blob/85b0c4a61d578fd3e1ec1ec80cb623fbab4d3ae7/examples/python/robot-wrapper-viewer.py).

I am thinking this might be a case where it is interesting to have the viewer hold a copy of the visual model instead of the visual model itself.

gabrielebndn:
@jcarpent just to be clear: it is incorrect to say that 

> @jmirabel @gabrielebndn
> A final point concerns the scaling of the meshes. I think having a function to automatically rescaled the meshes is good

@jcarpent I will ad, just to be 100% clear: technically, it is incorrect to say that `scaleGeometryModel` *scales* the meshes. In no way the mesh itself is ever dealt with. All this method does is setting the field `meshScale` for each `GeometryObject`. Ultimately, making a method for it was a great necessity, I might as well have let users do it themselves with a simple loop in Python, but I thought providing a simple convenience tool for what I expect to be a common problem might be nice.

As I have the impression there is some confusion around what `scaleGeometryModel` does, I am thinking to rename `setGeometryMeshScales` or something like that. What do you think?

bindings/python/pinocchio/display/abstract_display.py
gabrielebndn: Finally, I don't like calling it `Wrapper`. I'd rather stay with the current naming. I think it's clear enough, and I do not see an issue of duplicate naming: a Pinocchio display uses a viewer, there's a clear separation.

gabrielebndn:
> You should have `meshcat-server` available after installing the python package. You might have an issue with your path variables, could you check where `pip` is installing the package.

@matthieuvigne thanks, I thought it would be something like that, I found it in `~/.local/bin/meshcat-server`. I find it a bit awkward that the command is there, but I guess it's inevitable when you install with `pip --user`

jcarpent:
Another point is if you should be able to plot collisions in both the viewers. It is indeed a nice feature and it is only few lines of code with respect to what you have done so far.
Can you add it too?

jcarpent:
Another suggestion is also to change or simplify the API. For instance by replacing initDisplayModel by init. @gabrielebndn Is it fine?

gabrielebndn:
> Another point is if you should be able to plot collisions in both the viewers. It is indeed a nice feature and it is only few lines of code with respect to what you have done so far.
> Can you add it too?

As I said, I was planning to leave it for others to implement, but I might do it as well. Only, I do not know how to toggle visibility in Meshcat. @wxmerkt, @matthieuvigne, can you help me on that?

gabrielebndn:
> Another suggestion is also to change or simplify the API. For instance by replacing initDisplayModel by init. @gabrielebndn Is it fine?

The API is fine this way, I think. I want to keep the names as close as possible to the ones that already exist in `RobotWrapper`, to make the transition smoother. `init` is confusing I think, as it sounds too mush like the constructor

wxmerkt:
I am afraid I don't have cycles to add these features at the present time, sorry. Perhaps let's add it as a comment/TODO in the code and have it be added when it becomes used/required?

gabrielebndn:
> Perhaps let's add it as a comment/TODO in the code and have it be added when it becomes used/required?

@wxmerkt that's partially what I did. I am going to leave more comments/notes.
Also, I think I will probably put the visual model in node called `/visual/`

To be clear: I can easily load both collision objects and visual objects, the problem is that collision objects will always show up and I do not know how to hide them.

src/algorithm/geometry.hpp
gabrielebndn: I changed the comment to make more clear what the method is actually doing. Also, I changed the name to `setGeometryMeshes`

bindings/python/pinocchio/display/abstract_display.py
gabrielebndn: I added an argument called `copy_models`. I did the same for `setDisplay` in `RobotWrapper`

gabrielebndn:
So, I pushed some changes. To summarize the most important changes:
- Fix the problem with the implicit relative import
- Made `GeometryModel` and `GeometryData` copyable in Python
- added a `copy_models` option to the constructor, and so to `setDisplay` in `RobotWrapper`
- removed `BuildFromUrdf` and `initFromUrdf`
- changed `scaleGeometryModel` to `setGeometryMeshScales`
- added an `open` optional argument to `initDisplay` in `MeshcatDisplay`

jmirabel:
> > I don't know how units are handled inside DAE files, but I would guess that the Romeo meshes current in Pinocchio are in cm and don't have unit declared - some viewer then default to meters, but not Gepetto.
> 
> @matthieuvigne There are some scaling fields in the DAE file that are correctly handled by OSG, but apparently not by the other viewers.

That does not explain the issue. On my computer, all the DAE files of romeo contains the following line:
```
<unit meter="1" name="meter"/>
```

jmirabel:
> @jmirabel @gabrielebndn
> A final point concerns the scaling of the meshes. I think having a function to automatically rescaled the meshes is good, but as mentioned by @jmirabel, this has to be handled for all the stages (visual + collision for FCL too). @gabrielebndn I have not got deep into the code, but do you know if it the case with your current implementation?

It sounds to me like a dirty fix of an obscure problem. If the URDF file tells you to load a set of DAE files, with a (potentially different) scaling factor for each of them, you should either stick to that or correct the URDF file.

Moreover, rescaling the meshes in the back of the user is very unexpected for this user.

Finally, changing the scale attribute of the geometry does not rescale the FCL object itself. You end-up with a couple (mesh filename, scale) not consistent with the FCL object in memory. And the FCL object in memory is correct, this I am sure.

My final point on this is:
**Don't make Pinocchio dirty because there is a bug in a third-party viewer**. I would rather explore the bug in the third-party lib.

jmirabel:
I found the bug: the DAE files also contain the following lines:
```
<scale sid="scale">0.01000 0.01000 0.01000</scale>
```
I guess the DAE loader of MeshCat ought to be fixed in order to take this field into account.

jcarpent:
> > > I don't know how units are handled inside DAE files, but I would guess that the Romeo meshes current in Pinocchio are in cm and don't have unit declared - some viewer then default to meters, but not Gepetto.
> > 
> > 
> > @matthieuvigne There are some scaling fields in the DAE file that are correctly handled by OSG, but apparently not by the other viewers.
> 
> That does not explain the issue. On my computer, all the DAE files of romeo contains the following line:
> 
> ```
> <unit meter="1" name="meter"/>
> ```

There is a scale field too with value 0.01 value

gabrielebndn:
> Moreover, rescaling the meshes in the back of the user is very unexpected for this user.
> 
> Finally, changing the scale attribute of the geometry does not rescale the FCL object itself. You end-up with a couple (mesh filename, scale) not consistent with the FCL object in memory. And the FCL object in memory is correct, this I am sure.
> 
> My final point on this is:
> **Don't make Pinocchio dirty because there is a bug in a third-party viewer**. I would rather explore the bug in the third-party lib.

Dear @jmirabel, I totally agree with you that Pinocchio should stay clean. That's what I have done. Maybe what I did is not sufficiently clear.

1) As I said above, yes, changing the scale attribute does not rescale the FCL object itself. That's why I changed the name from `scaleGeometryModel` to `setGeometryMeshScale`: for it to be clear.

2) the provided algorithm, `setGeometryMeshScale`, is a perfectly legal and reasonable algorithm. I can see people wanting to set the `meshScale` property to the same value for each `GeometryObject` in a `GeometryModel`. So I think it's useful.

3) The ideal way of managing `meshScale` in the display would be to set it once and for all, as it was said. This is what `GepettoDisplay` is doing. I found this in `RobotWrapper`. It seemed to me entirely logical.
Otherwise, have I misunderstood something? What is the `meshScale` property supposed to exist for, if not for this?
Now, to replicate the same thing with Meshcat, I found no bindings. As a workaround, I decided to use property `meshScale` in order to compute a scaling transformation *inside the plotter*. If it is not clear enough, this is what I am doing:
```
self.visual_data.oMg[self.visual_model.getGeometryId(visual.name)]
S = np.diag(np.concatenate((visual.meshScale,np.array([[1.0]]))).flat)
T = np.array(M.homogeneous).dot(S)
self.viewer[self.getViewerNodeName(visual,pin.GeometryType.VISUAL)].set_transform(T)
```
3. (continued): now, this is a workaround, but it is only a workaround to the problem of Meshcat *not being able to set the scale*. However, this is perfectly valid code, which does what it is supposed to do. The final behavior is exactly equivalent to what I would have done if Meshcat provided a way to set the scale in Python

As a final point
**No user code I provided contains any "dirty" stuff with respect to the scaling problem**. The hack of manually changing the scale is only inside the example files:
[examples/python/meshcat-viewer.py](https://github.com/gabrielebndn/pinocchio/blob/85b0c4a61d578fd3e1ec1ec80cb623fbab4d3ae7/examples/python/meshcat-viewer.py)
[examples/python/load-urdf.py](https://github.com/gabrielebndn/pinocchio/blob/85b0c4a61d578fd3e1ec1ec80cb623fbab4d3ae7/examples/python/load-urdf.py)
[examples/python/robot-wrapper-viewer.py](https://github.com/gabrielebndn/pinocchio/blob/85b0c4a61d578fd3e1ec1ec80cb623fbab4d3ae7/examples/python/robot-wrapper-viewer.py)
which directly employ Meshcat. This is *inevitable* if you want to properly display Romeo in the current version of Meshcat. I may remove this if you want (after all, I only did it to convince myself everything was properly showing), on which case Romeo model will not display properly, but who cares, that's what I found in the first place anyways. But it will only affect the examples, not the "real" Pinocchio code, because the "real" Pinocchio code is fine.

jmirabel:
> As I said above, yes, changing the scale attribute does not rescale the FCL object itself. That's why I changed the name from scaleGeometryModel to setGeometryMeshScale: for it to be clear.

I wouldn't guess from the name that this does not actually rescale the FCL object. For me, it is logical that the scale attribute is the scale of the FCL object. If you change one, the other ought to be changed.

> The ideal way of managing meshScale in the display would be to set it once and for all, as it was said. This is what GepettoDisplay is doing. I found this in RobotWrapper. It seemed to me entirely logical.

Yes, I think it's the only thing you should do: passing the scale from Pinocchio to the viewer.

Honestly, I find the lines below dirty:
```
# Scale the model
# This is necessary for Romeo, as the provided model has wrong scaling
pin.scaleGeometryModel(display.visual_model,0.01)
```
for reasons like:
- nobody will do that in the first shot,
- you should do that if you use MeshCat, and not GepettoViewer, so you cannot switch from one viewer to the other,
- it is something you must do for some models, and not for others,
- it changes the scale of meshes, but not primitive shapes,
- it makes the geometry model inconsistent (the scale does not correspond to the FCL geometry anymore),

By the way, the bug is in MeshCat, not in the Romeo models.

And if you don't want to dig into MeshCat, I believe an optional argument to MeshcatDisplay.BuildFromURDF makes more sense. This argument would trigger a rescaling when sending the scale to Meshcat and wouldn't touch the geometry model.

gabrielebndn:
Dear @jmirabel

> For me, it is logical that the scale attribute is the scale of the FCL object. If you change one, the other ought to be changed.

I ask you again: why having the meshScale property at all then? This is just a loop changing that property. Users might even do it themselves in Python, if they wanted to, and nothing would stop them from that

> Honestly, I find the lines below dirty:

Again, this is in the *examples*, just to make it work. If it really gives you creeps, I will simply remove the line and the example will not show up properly until the problem fixes itself when Meshcat is updated, I am fine with that

> nobody will do that in the first shot

You don't. You see there is a problem with the scaling, you solve it. That's how it works. That's what I did in the first place

> you should do that if you use MeshCat, and not GepettoViewer

If Meshcat handles data differently from GepettoViewer, it's inevitable that you use them differebtly

> so you cannot switch from one viewer to the other

I tried to make things as uniform as I could. However, it's not always possible

> it is something you must do for some models, and not for others,

Some models need special treatment, others don't

> it changes the scale of meshes, but not primitive shapes

Really? I had not thought about that. Maybe you can explain this to me in private later

> it makes the geometry model inconsistent (the scale does not correspond to the FCL geometry anymore)

I updated the examples. Now, I am passing copies, to avoid any problem

> By the way, the bug is in MeshCat, not in the Romeo models.

Agreed. Check the new comment in the updated code 

> I believe an optional argument to MeshcatDisplay.BuildFromURDF makes more sense. This argument would trigger a rescaling when sending the scale to Meshcat and wouldn't touch the geometry model

First of all, I removed BuildFromUrdf. Second, *that's* what I would call making Pinocchio dirty: adding a questionable addition to the user API to handle a buggy corner case in a specific viewer. I think we should not do that

jcarpent:
I do agree with @jmirabel. The problem should be solved at the Meshcat level and not at the level of Pinocchio.

gabrielebndn:
> I do agree with @jmirabel. The problem should be solved at the Meshcat level and not at the level of Pinocchio.

I really think you are missing the point. I am *not* solving the problem at the level of Pinocchio.
I am solving the problem *in the examples* the way a *user* would *if* he wanted to correctly display Romeo model.

If you think I should not care about it, I will just remove those lines from the examples. Then the examples will display Romeo incorrectly, which I guess is fine.

Is that what you want?

gabrielebndn:
OK. As requested I pushed some API changes
- the sub-module is now called `pinocchio.visualize`
- the three classes are called `BaseVisualizer`, `GepettoVisualizer` and `MeshcatVisualizer`
- In `RobotWrapper`, `setDisplay` has been changed to `setVisualizer`
- Methods `initDisplay` has been changed to `initViewer` and `loadDisplayModel` has been changed to `loadViewerModel`. In `RobotWrapper`, the old methods still exist, but they are deprecated
- the internal `RobotWrapper` variable holding the visualizer is now called `viz` instead of `disp`

Also:
- Added the possibility to pass viewer instance in `GepettoVisualizer`, similar to `MeshcatVisualizer`
- Removed calls to `setGeometryMeshScales` in the examples

Hope this solves all issues

bindings/python/pinocchio/display/abstract_display.py
gabrielebndn: See changes

jcarpent:
@gabrielebndn Thanks a lot for these modifications!

Issue #809, opened 2019-06-06
Using CasADi automatic differentiation is not possible


mkatliar:
[The paper](https://hal.laas.fr/hal-01866228v2/document) says:
```
In addition to analytical derivatives, Pinocchio supports
automatic differentiation. This is made possible through the
full scalar templatization of the whole C++ code and the use
of any external library that does automatic differentiation:
ADOL-C [23], CasADi [24], CppAD [25] and others.
```

I tried to use `casadi::SX` type as a scalar type for Pinocchio algorithms, but the code does not compile. Here is a code snippet:
```c++
pinocchio::ModelTpl<casadi::SX> model;
pinocchio::DataTpl<casadi::SX> data(model);

Eigen::Matrix<casadi::SX, N_JOINTS, 1> q;
Eigen::Matrix<casadi::SX, N_JOINTS, 1> v;
Eigen::Matrix<casadi::SX, N_JOINTS, 1> tau;

aba(model, data, q, v, tau);
```
Here is the relevant compiler output: [make.log](https://github.com/stack-of-tasks/pinocchio/files/3261876/make.log)

jcarpent:
Nice to try with Casadi.
Before starting with Pinocchio, I would suggest to test with Eigen first and to make sure that everything works well. 
Do you have a unitary example of Eigen + Casadi?

From what I've seen from your log, the problem comes from there.

mkatliar:
I see at least two sources of the problem:
1. The non-specialized implementation of `SINCOSAlgo` calls `std::sin()` and `std::cos()`:
    ```c++
    template<typename Scalar>
    struct SINCOSAlgo
    {
      static void run(const Scalar & a, Scalar * sa, Scalar * ca) 
      {   
        (*sa) = std::sin(a); (*ca) = std::cos(a);
      }   
    };
    ```
    It would be more appropriate to just do `sin(a)`, `cos(a)` and let [argument-dependent lookup](https://en.cppreference.com/w/cpp/language/adl) do the job. For `casadi::SX` argument type, the calls would resolve to `casadi::sin()` and `casadi::cos()`. Notice that if we rely on ADL, the following code fragment becomes unnecessary:
    ```c++
    #ifdef PINOCCHIO_WITH_CPPAD_SUPPORT
  
    /// Implementation for CppAD scalar types.
    template<typename Scalar>
    struct SINCOSAlgo< CppAD::AD<Scalar> >
    {
[...]

PR #813, opened 2019-06-06
add support of CASADI

jcarpent:
@mkatliar @jmirabel I've just opened a new PR for checking the integration of Pinocchio inside CASADI. This is still work in progress but I will try to add the different tests that @mkatliar is adding.

Do not hesitate to add your contributions.

Related to #809


jcarpent:
Minor info: I have found some bugs. I will try to push them by tomorrow.

src/math/casadi.hpp
```diff
@@ -0,0 +1,73 @@
...
+    // Specialization of Eigen::internal::cast_impl for CppAD input types
+    template<typename Scalar>
+    struct cast_impl<casadi::SX,Scalar>
+    {
+#if EIGEN_VERSION_AT_LEAST(3,2,90)
+      EIGEN_DEVICE_FUNC
+#endif
+      static inline Scalar run(const casadi::SX & x)
+      {
+        return (Scalar)x;
```
jmirabel: Eigen's implementation of cast_impl is:
```cpp
    return static_cast<NewType>(x);
```
Why do you have to C-style cast rather that C++-style cast operator ?

jcarpent: I'm just calling the cast of Casadi Matrix type directly.

jmirabel:
Another good reason is that it would give a good example of how one can use another auto-diff module.

jcarpent:
> As a remark, I think support of external libraries like casADi, AutoDiff, CppAD should not be added this way. I would put all the template specialization code into separate files.
> 
> There are good reasons for this:
> 
>     * putting the specialization into the main files prevents user to provide their own specialization, e.g. for an application specific case or to fix a bugged specialization.
> 
>     * it should be provided not as a complete but rather as a start. User may need to complete it. As far as I know (correct me if I am wrong), the main developers of Pinocchio haven't use auto-diff a lot. #809 is one example.

I do agree with all these points, but I do not have any great solution for that.
Can you suggest something better?

Maybe we can think to something like the unsupported directory of Eigen. 
The only thing to keep in mind is that for some reasons, we need to include CppAD headers before Eigen's ones.

jcarpent:
I suggest to first make everything in such a way Pinocchio works fine with Casadi, which requires some works. And to propose a clean way of integrating all AutoDiff frameworks later. It will be easier for me.

jcarpent:
By the way, at least CppAD should be an optional dependency because it is the framework I use to do code generation.

jmirabel:
> The only thing to keep in mind is that for some reasons, we need to include CppAD headers before Eigen's ones.

I believe this is an easy task which can be the responsibility of the user. I don't think this is the responsibility of the user.

> Can you suggest something better?

Add a file casadi.hpp anywhere in Pinocchio, with the following properties:
- contains all the template specializations.
- does not include casadi
- add a line which won't compile if casadi is not included, with a comment explaining that the user must include casadi himself, before including any Pinocchio/Eigen headers.

Eventually, it could be a folder with several files.
Would this work ?

src/math/casadi.hpp
jmirabel: I don't see any difference between the Eigen's implementation of `Eigen::internal::cast_impl<casadi::SX, Scalar>` and yours. Don't they do exactly the same ?

jmirabel: Eigen's implementation of `Eigen::internal::cast_impl<casadi::SX, Scalar>` boils down to
```cpp
static inline Scalar run(const casadi::SX & x)
{
    return static_cast<Scalar>(x);
}
```

jcarpent:
Yes, I guess. I will try to do it after fixing all the unit test for the integration of CASADI.

jmirabel:
Thank you for your work !

mkatliar:
A stupid question, how can I add my contributions to this pull request (instead of creating another one)?

jcarpent:
> A stupid question, how can I add my contributions to this pull request (instead of creating another one)?

You're now a collaborator of the project. You should be able to push on the current PR. @mkatliar Thanks for your contributions.

mkatliar:
FYI, I am currently working on making `Eigen::LLT` work with `casadi::SX`.

mkatliar:
I never did *push to a pull request* before, so maybe I am doing something wrong here. I added git@github.com:jcarpent/pinocchio.git as a remote, then checked out `topic/casadi`, made some changes and committed. Then following [these instructions](https://help.github.com/en/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork) I do:

```
$ git push origin topic/casadi 
ERROR: Permission to stack-of-tasks/pinocchio.git denied to mkatliar.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

What is wrong?

jcarpent:
You did not accept the invitation to join the project that you should have received few hours ago.

src/math/casadi.hpp
```diff
@@ -6,6 +6,48 @@
 #define __pinocchio_math_casadi_hpp__
 
 #include <casadi/casadi.hpp>
+// #include <Eigen/Core>
+#include <Eigen/Dense>
+
+// Copy casadi matrix to Eigen matrix
+template <typename MT>
+inline void copy(casadi::SX const& src, Eigen::MatrixBase<MT>& dst)
```
jcarpent: @mkatliar These functions should appear in an external namespace

jcarpent: Like pinocchio::casadi

src/math/casadi.hpp
```diff
@@ -68,6 +112,246 @@ namespace Eigen
...
+      static bool inplace_decomposition(MatrixType& m)
+      { return llt_inplace<typename MatrixType::Scalar, Upper>::blocked(m)==-1; }
+    };
+  }
+
+
+  /// @brief LLT template specialization supporting casadi::SX
+  ///
+  template<int M, int N, int Options, int _UpLo>
+  class LLT<Matrix<casadi::SX, M, N, Options>, _UpLo>
```
jcarpent: I do not think that you should specialize all the Eigen::LLT for your type.
You should use partial template specialization of methods instead (see https://en.cppreference.com/w/cpp/language/partial_specialization)

unittest/casadi-basic.cpp
```diff
@@ -52,8 +49,8 @@ BOOST_AUTO_TEST_CASE(test_example)
   {
     for (Eigen::Index j = 0; j < A.cols(); ++j)
     {
-      A(i, j) = 10 * i + j;
-      B(i, j) = -10 * i - j;
+      A(i, j) = 10. * static_cast<double>(i) + static_cast<double>(j);
```
jcarpent: Did you have some compiler warnings here?

mkatliar:
On my system, I get the following output from cmake when trying to build the `topic/casadi` branch:

```
[cmake] eigen3 >= 3.0.5 is required.
[cmake] Checking for module 'eigen3 >= 3.0.5'
[cmake]   Found eigen3 , version 3.3.7
[cmake] Pkg-config module eigen3 v3.3.7 has been detected with success.
[cmake] urdfdom >= 0.2.0 is required.
[cmake] Checking for module 'urdfdom >= 0.2.0'
[cmake]   Found urdfdom , version 1.0.3
[cmake] Pkg-config module urdfdom v1.0.3 has been detected with success.
[cmake] cppad >= 20180000.0 is required.
[cmake] Checking for module 'cppad >= 20180000.0'
[cmake]   Found cppad , version 20190601
[cmake] Pkg-config module cppad v20190601 has been detected with success.
[cmake] casadi >= 3.4.5 is required.
[cmake] Checking for module 'casadi >= 3.4.5'
[cmake]   No package 'casadi' found
[cmake] CMake Error at /usr/share/cmake-3.13/Modules/FindPkgConfig.cmake:452 (message):
[cmake]   A required package was not found
[cmake] Call Stack (most recent call first):
[cmake]   /usr/share/cmake-3.13/Modules/FindPkgConfig.cmake:622 (_pkg_check_modules_internal)
[cmake]   cmake/pkg-config.cmake:298 (PKG_CHECK_MODULES)
[cmake]   cmake/pkg-config.cmake:510 (ADD_DEPENDENCY)
[cmake]   CMakeLists.txt:120 (ADD_REQUIRED_DEPENDENCY)
[cmake] 
[cmake] 
[cmake] Configuring incomplete, errors occurred!
[cmake] See also "/home/mkatliar/software/pinocchio/build/CMakeFiles/CMakeOutput.log".
[cms-driver] Error during CMake configure: [cmake-server] Configuration failed.
```

But it instead of
```cmake
ADD_REQUIRED_DEPENDENCY("casadi >= 3.4.5")
```
I do
```cmake
find_package(CasADi 3.4.5 REQUIRED)
set(CASADI_FOUND True)
```
then cmake does not complain.

I have built casadi from sources and installed in `/usr/local`.

unittest/casadi-basic.cpp
mkatliar: Yes

src/math/casadi.hpp
```diff
@@ -68,6 +112,246 @@ namespace Eigen
...
+        // Copy the triangular part of matrix that will be used for the decompositon to a self-adjoint m_matrix,
+        // taking into account the UpLo flag.
+        m_matrix = SelfAdjointView<InputType const, UpLo>(matrix);
+
+        // The Eigen version of LLT::compute() does not compile for matrices of casadi::SX,
+        // because operator>() for casadi::SX is not defined.
+        // Therefore, we use casadi::chol() to do the decomposition, and then copy the result to an Eigen matrix of casadi::SX.
+
+        // Copy m_matrix to a temporary casadi::SX matrix
+        casadi::SX cs_matrix;
```
jcarpent: Avoid memory allocation here. It should be better to make it as a class member.

unittest/casadi-basic.cpp
jcarpent: Which compiler are you using?

jcarpent:
@mkatliar I have a basic question, why do you need to specialize Eigen::LLT, is it because you have some warnings when compiling `pinocchio::aba?

src/math/casadi.hpp
mkatliar: What needs to be specialized is
```c++
template<typename InputType>
      LLT& compute(const EigenBase<InputType>& matrix)
```

I would need to do a partial specialization of a template member function of a template class. I am not sure if such thing is possible in C++. At least I tried and my C++ skills were not sufficient to do that.

src/math/casadi.hpp
mkatliar: Agree

jcarpent:
> On my system, I get the following output from cmake when trying to build the `topic/casadi` branch:
> 
> ```
> [cmake] eigen3 >= 3.0.5 is required.
> [cmake] Checking for module 'eigen3 >= 3.0.5'
> [cmake]   Found eigen3 , version 3.3.7
> [cmake] Pkg-config module eigen3 v3.3.7 has been detected with success.
> [cmake] urdfdom >= 0.2.0 is required.
> [cmake] Checking for module 'urdfdom >= 0.2.0'
> [cmake]   Found urdfdom , version 1.0.3
> [cmake] Pkg-config module urdfdom v1.0.3 has been detected with success.
> [cmake] cppad >= 20180000.0 is required.
> [cmake] Checking for module 'cppad >= 20180000.0'
> [cmake]   Found cppad , version 20190601
> [cmake] Pkg-config module cppad v20190601 has been detected with success.
> [cmake] casadi >= 3.4.5 is required.
> [cmake] Checking for module 'casadi >= 3.4.5'
> [cmake]   No package 'casadi' found
> [cmake] CMake Error at /usr/share/cmake-3.13/Modules/FindPkgConfig.cmake:452 (message):
> [cmake]   A required package was not found
> [cmake] Call Stack (most recent call first):
> [cmake]   /usr/share/cmake-3.13/Modules/FindPkgConfig.cmake:622 (_pkg_check_modules_internal)
> [cmake]   cmake/pkg-config.cmake:298 (PKG_CHECK_MODULES)
> [cmake]   cmake/pkg-config.cmake:510 (ADD_DEPENDENCY)
> [cmake]   CMakeLists.txt:120 (ADD_REQUIRED_DEPENDENCY)
> [cmake] 
> [cmake] 
> [cmake] Configuring incomplete, errors occurred!
> [cmake] See also "/home/mkatliar/software/pinocchio/build/CMakeFiles/CMakeOutput.log".
> [cms-driver] Error during CMake configure: [cmake-server] Configuration failed.
> ```
> 
> But it instead of
> 
> ```cmake
> ADD_REQUIRED_DEPENDENCY("casadi >= 3.4.5")
> ```
> 
> I do
> 
> ```cmake
> find_package(CasADi 3.4.5 REQUIRED)
> set(CASADI_FOUND True)
> ```
> 
> then cmake does not complain.
> 
> I have built casadi from sources and installed in `/usr/local`.
> 
> How can we fix this?

I have forked `casadi` to allow it to handle pkgconfig configuration. I have made a PR so far. 
You can see it here: https://github.com/casadi/casadi/pull/2446

mkatliar:
> @mkatliar I have a basic question, why do you need to specialize Eigen::LLT, is it because you have some warnings when compiling `pinocchio::aba?

Because I have errors when compiling `pinocchio::aba` with `casadi::SX` as a scalar type.

mkatliar:
> I have forked `casadi` to allow it to handle pkgconfig configuration. I have made a PR so far.
> You can see it here: [casadi/casadi#2446](https://github.com/casadi/casadi/pull/2446)

@jcarpent casadi installs its `/usr/local/lib/cmake/casadi/casadi-config.cmake` file, after which cmake can find casadi automatically. Do we really need to rely on the pkgconfig mechanism here?

jcarpent:
> > @mkatliar I have a basic question, why do you need to specialize Eigen::LLT, is it because you have some warnings when compiling `pinocchio::aba?
> 
> @jcarpent because I have errors when compiling `pinocchio::aba` with `casadi::SX` as a scalar type.

This is expected. Some computations involve the usage of Eigen::LLT. I suggest to specialize these computations according to the Scalar types: if the Scalar type is classic (float, double, long double), then we need to call the classic LLT, otherwise, we have to the inversion (without relying on LLT, and without any comparisons).

mkatliar:
> This is expected. Some computations involve the usage of Eigen::LLT. I suggest to specialize these computations according to the Scalar types: if the Scalar type is classic (float, double, long double), then we need to call the classic LLT, otherwise, we have to the inversion (without relying on LLT, and without any comparisons).

@jcarpent this makes sense to me. Could you then write the specialized code for there computations, since you are much more familiar with the algorithms? I could help you on the casadi side if you need.

jcarpent:
Yes, I will do the specialization of the algorithms.

jcarpent:
> > I have forked `casadi` to allow it to handle pkgconfig configuration. I have made a PR so far.
> > You can see it here: [casadi/casadi#2446](https://github.com/casadi/casadi/pull/2446)
> 
> @jcarpent casadi installs its `/usr/local/lib/cmake/casadi/casadi-config.cmake` file, after which cmake can find casadi automatically. Do we really need to rely on the pkgconfig mechanism here?

A lot of projects (among them Pinocchio) rely on pkg-config for dependency check. I think it would be worthy to add this feature to CASADI, even if, for the time being, we can use the CMake approach.

jcarpent:
@mkatliar I have done most of the job to get Casadi working with Pinocchio. 
The remaining thing is to make the clean splitting between autodiff frameworks and Pinocchio core functionalities.

@mkatliar Do you also need the analytical derivatives?

mkatliar:
> @mkatliar Do you also need the analytical derivatives?

@jcarpent you mean if I need the `abaDerivatives()` etc. to work with casadi types? At the moment no, because if `aba()` works with casadi types then I can obtain the analytical derivatives from casadi.

jcarpent:
> @jcarpent you mean if I need the `abaDerivatives()` etc. to work with casadi types? At the moment no, because if `aba()` works with casadi types then I can obtain the analytical derivatives from casadi.

Yes, I totally understand. But, it will be worthy to use the one provided by Pinocchio I guess because they are fast and they work in the tangent of the configuration manifold. I can try to include them too in the unit test.

unittest/casadi-basic.cpp
mkatliar: g++ (Ubuntu 8.3.0-6ubuntu1) 8.3.0

jcarpent:
@mkatliar I think I'm done with this PR. Everything works well on my computer. Can you try it from your side?

jcarpent:
Following the discussion in #814, I will merge it and make the rebase for #814.

PR #814, opened 2019-06-17
Implementation of Mimic joints

jcarpent:


jmirabel:
@jcarpent before reviewing this, I am waiting for #813 to be merged and this PR to be re-based on top of devel branch.

jcarpent:
> @jcarpent before reviewing this, I am waiting for #813 to be merged and this PR to be re-based on top of devel branch.

Done!

jcarpent:
`q.segment (jmodel.idx_q(), jmodel.nq())` will affect the reference joint configuration.

I think indeed we need to provide clear documentation. I can change the way it works by using an internal `jmodel.nq()` and setting it to 0, and `jmodel.idx_q()`returning -1. In this way, Eigen will raise an issue.

@jmirabel What do you think?

jcarpent:
>     * Is the result correct (i.e. scaling * [1,..] + offset) or incorrect (i.e. [1, ...] ) ?
>       `jmodel.configVector(q)` ?

It is returning the incorrect result. But, the definition of configVector is that it returns the slice in the configuration vector, nothing more.

jcarpent:
Currently, only forwardKinematics will work well with JointMimic. 
I have to change other algos (crba, jointJacobians, ...) to correctly behave.

jmirabel:
I think it is better to have nq and nv equal 0. I am not sure what value is best for idx_q and idx_v. Setting it to -1 has the advantage of raising assertion in Eigen. Setting it to the reference joint model index has the advantage that the following loop works:
```cpp
for (int i=1; i < model.njoints; ++i)
{
  jmodel.configVector(q) = ...;
}
```

jcarpent:
>     * with your implementation, should users be aware of the presence of mimic joints ?

Not at all if I'm correct. It is transparent at the API level.

jcarpent:
>     * Is there anything special to modify in user code to handle them ?

No normally. This is why I have tried to not change the API. But as I was mentionning, we need to change a lot of algos (which were not thought for JointMimic) to make them compliant.
For instance, the Jacobian of a mimic joint has to sum the Jacobian of the reference joints + the Jacobian of the mimic joint.

jmirabel:
The issue is not only the API. Most loops on joints, in user code, must be revised in order to correctly handle mimic joints. I guess many people assumed that nq and nv are always strictly positive.

Did you think of putting the mimic joints in a separate vector ?

It is a bit too early to decide but introducing mimic joint may force us to increase the major version...
A good measure of the impact on user code is the impact on unit-tests.

jcarpent:
> The issue is not only the API. Most loops on joints, in user code, must be revised in order to correctly handle mimic joints. I guess many people assumed that nq and nv are always strictly positive.

It won't change anything at the API level, just in the way the algorithms are written internally when using Constraint classes. 
I guess only a few users are currently using them up to now. But this is just a guess.

> Did you think of putting the mimic joints in a separate vector ?

This is not compatible with the rigid body dynamics algorithms

> It is a bit too early to decide but introducing mimic joint may force us to increase the major version...
A good measure of the impact on user code is the impact on unit-tests.

Yes, certainly.

jmirabel:
> > The issue is not only the API. Most loops on joints, in user code, must be revised in order to correctly handle mimic joints. I guess many people assumed that nq and nv are always strictly positive.
> 
> It won't change anything at the API level, just in the way the algorithms are written internally when using Constraint classes.
> I guess only a few users are currently using them up to now. But this is just a guess.

I agree we can consider constraint internally. I was thinking of more elementary issues. There can be several joints:
- with the same idx_q and idx_v,
- with nq and nv = 0.

I consider them as *soft* API changes. Introducing whose changes will break someone's code.

jcarpent:
@jmirabel I did the modifications.

jmirabel:
Thank you.

What is the strategy you have in mind regarding this ?
Merging this PR now and leaving the non-implemented features for another PR ?

jcarpent:
@jmirabel Yes, I will merge it and as it remains some work to be done, I will push it later on.
We already have the structure, and I will work on making all the algos compliant with this new type of joints.

PR #818, opened 2019-06-21
Introduce joint torque regressor

gabrielebndn:
Summary:
this PR introduces a joint torque regressor, such that

<a href="https://www.codecogs.com/eqnedit.php?latex=\tau&space;=&space;Y(q,\dot{q},\ddot{q})\pi" target="_blank"></a>

where

<a href="https://www.codecogs.com/eqnedit.php?latex=\pi&space;=&space;(\pi_1^T\&space;\dots\&space;\pi_n^T)^T" target="_blank"></a>

<a href="https://www.codecogs.com/eqnedit.php?latex=\pi_i&space;=&space;(m,&space;mc_x,&space;mc_y,&space;mc_z,&space;I_{xx},&space;I_{xy},&space;I_{yy},&space;I_{xz},&space;I_{yz},&space;I_{zz})^T" target="_blank"></a>

<a href="https://www.codecogs.com/eqnedit.php?latex=I&space;=&space;I_C&space;&plus;&space;mS^T(c)S(c)" target="_blank"></a>

and <a href="https://www.codecogs.com/eqnedit.php?latex=I_C" target="_blank"></a> has its origin at the barycenter.

Also, it introduces several methods which, other than being employed by the function, might be useful *per se*, and a couple of minor additions and modifications.

This may be useful for calibration, but also for the computation of partial derivatives w.r.t. the dynamic parameters.

In detail:
- introduce methods `inertia.toDynamicParameters()` and `Inertia.FromDynamicParameters(Vector10)` to convert between a Pinocchio `Inertia` and its corresponding <a href="https://www.codecogs.com/eqnedit.php?latex=\pi_i" target="_blank"></a>
- introduce algorithm `bodyRegressor`, computing the regressor for the dynamic parameters of a single rigid body, of which the velocity and acceleration are given
- introduce algorithm `jointBodyRegressor`, computing the regressor for the dynamic parameters of a rigid body attached to a given joint, assuming RNEA has been called
- introduce algorithm `frameBodyRegressor`, computing the regressor for the dynamic parameters of a rigid body attached to a given frame, assuming RNEA has been called
- introduce algorithm `computeJointTorqueRegressor`, computing the joint torque regressor described above, which is written in `data.jointTorqueRegressor`
- expose `data.staticRegressor` in Python
- move `computeStaticRegressor` to main `pinocchio` namespace. Deprecating the version in `regressor` namespace. I did this because the existence of this namespace has always seemed incoherent with the style of the rest of Pinocchio, which is having all algorithms in the main Pinocchio namespace

NOTE 1: notice that, as forward step of the algorithm, I am using `RneaForwardStep`. I did it to go quickly, and also to avoid the need of compiling another step. I now realize calling `RneaForwardStep` will modify `data.f`, leaving it in an inconsistent state. This should be fixed. Therefore, I will either create a custom forward step or find another way to fix the problem which does not require creating a new step type (I can think of at least two,both of which would slightly increase execution time).

NOTE 2: because the matrices are small enough, I used fixed-size matrices: `Matrix<Scalar,10, 1>` and `Matrix<Scalar,6,10>`. As for `Matrix<Scalar,10, 1>`, I typedefed it to `Vector10` inside `InertiaTpl`. `Matrix<Scalar,6,10>`, I do not know where to put a convenient typedef. Also, neither of them is exposed in eigenpy. Should we think about it, especially for `Vector10`?

### DIFF
--- a/src/algorithm/regressor.hpp
+++ b/src/algorithm/regressor.hpp
@@ -47,8 +47,8 @@ namespace pinocchio
   /// \return The regressor of the body.
   ///
   template<typename MotionVelocity, typename MotionAcceleration>
-  inline Eigen::Matrix<typename MotionVelocity::Scalar,6,10,MotionVelocity::Options>
-  bodyRegressor(const MotionBase<MotionVelocity> & v, const MotionBase<MotionAcceleration> & a);
+  inline Eigen::Matrix<typename MotionVelocity::Scalar,6,10,PINOCCHIO_EIGEN_PLAIN_TYPE(typename MotionVelocity::Vector3)::Options>
+  bodyRegressor(const MotionDense<MotionVelocity> & v, const MotionDense<MotionAcceleration> & a);
 
   ///
   /// \brief Computes the regressor for the dynamic parameters of a rigid body attached to a given joint.
--- a/src/algorithm/regressor.hpp
+++ b/src/algorithm/regressor.hpp
@@ -34,6 +34,21 @@ namespace pinocchio
                            DataTpl<Scalar,Options,JointCollectionTpl> & data,
                            const Eigen::MatrixBase<ConfigVectorType> & q);
   }
+
+  ///
+  /// \brief Computes the regressor for the dynamic parameters of a single rigid body.
+  ///
+  /// The result is such that
+  /// \f$ I a + v \times I v = bodyRegressor(v,a) * I.toDynamicParameters() \f$
+  ///
+  /// \param[in] v Velocity of the rigid body
+  /// \param[in] a Acceleration of the rigid body
+  ///
+  /// \return The regressor of the body.
+  ///
+  template<typename MotionVelocity, typename MotionAcceleration>
+  inline Eigen::Matrix<typename MotionVelocity::Scalar,6,10,MotionVelocity::Options>
+  bodyRegressor(const MotionBase<MotionVelocity> & v, const MotionBase<MotionAcceleration> & a);
   
 } // namespace pinocchio
 
--- a/src/algorithm/regressor.hpp
+++ b/src/algorithm/regressor.hpp
@@ -15,6 +15,11 @@ namespace pinocchio
   /// \brief Computes the static regressor that links the center of mass positions of all the links
   ///        to the center of mass of the complete model according to the current configuration of the robot.
   ///
+  /// The result is stored in `data.staticRegressor` and it corresponds to a matrix \f$ Y \f$ such that
+  /// \f$ c = Y(q,\dot{q},\ddot{q})\tilde{\pi} \f$
+  /// where \f$ \tilde{\pi} = (\tilde{\pi}_1^T\ \dots\ \tilde{\pi}_n^T)^T \f$ and
+  /// \f$ \tilde{\pi}_i = \text{model.inertias[i].toDynamicParameters().head<4>()} \f$
+  ///
   /// \tparam JointCollection Collection of Joint types.
   /// \tparam ConfigVectorType Type of the joint configuration vector.
   ///
@@ -79,7 +84,7 @@ namespace pinocchio
   /// This algorithm assumes RNEA has been run to compute the acceleration and gravitational effects.
   ///
   /// The result is such that
-  /// \f$ f = jointBodyRegressor(model,data,jointId) * I.toDynamicParameters() \f$
+  /// \f$ f = \text{jointBodyRegressor(model,data,jointId) * I.toDynamicParameters()} \f$
   /// where \f$ f \f$ is the net force acting on the body, including gravity
   ///
   /// \param[in] model The model structure of the rigid body system.
@@ -100,7 +105,7 @@ namespace pinocchio
   /// This algorithm assumes RNEA has been run to compute the acceleration and gravitational effects.
   ///
   /// The result is such that
-  /// \f$ f = frameBodyRegressor(model,data,frameId) * I.toDynamicParameters() \f$
+  /// \f$ f = \text{frameBodyRegressor(model,data,frameId) * I.toDynamicParameters()} \f$
   /// where \f$ f \f$ is the net force acting on the body, including gravity
   ///
   /// \param[in] model The model structure of the rigid body system.
@@ -115,31 +120,35 @@ namespace pinocchio
                      DataTpl<Scalar,Options,JointCollectionTpl> & data,
                      FrameIndex frameId);
 
-    ///
-    /// \brief Computes the joint torque regressor that links the joint torque
-    ///        to the dynamic parameters of each link according to the current the robot motion.
-    ///
-    /// \tparam JointCollection Collection of Joint types.
-    /// \tparam ConfigVectorType Type of the joint configuration vector.
-    /// \tparam TangentVectorType1 Type of the joint velocity vector.
-    /// \tparam TangentVectorType2 Type of the joint acceleration vector.
-    ///
-    /// \param[in] model The model structure of the rigid body system.
-    /// \param[in] data The data structure of the rigid body system.
-    /// \param[in] q The joint configuration vector (dim model.nq).
-    /// \param[in] v The joint velocity vector (dim model.nv).
-    /// \param[in] a The joint acceleration vector (dim model.nv).
-    ///
-    /// \return The joint torque regressor of the system.
-    ///
-    template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType, typename TangentVectorType1, typename TangentVectorType2>
-    inline typename DataTpl<Scalar,Options,JointCollectionTpl>::MatrixXs &
-    computeJointTorqueRegressor(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
-                                DataTpl<Scalar,Options,JointCollectionTpl> & data,
-                                const Eigen::MatrixBase<ConfigVectorType> & q,
-                                const Eigen::MatrixBase<TangentVectorType1> & v,
-                                const Eigen::MatrixBase<TangentVectorType2> & a);
-  
+  ///
+  /// \brief Computes the joint torque regressor that links the joint torque
+  ///        to the dynamic parameters of each link according to the current the robot motion.
+  ///
+  /// The result is stored in `data.jointTorqueRegressor` and it corresponds to a matrix \f$ Y \f$ such that
+  /// \f$ \tau = Y(q,\dot{q},\ddot{q})\pi \f$
+  /// where \f$ \pi = (\pi_1^T\ \dots\ \pi_n^T)^T \f$ and \f$ \pi_i = \text{model.inertias[i].toDynamicParameters()} \f$
+  ///
+  /// \tparam JointCollection Collection of Joint types.
+  /// \tparam ConfigVectorType Type of the joint configuration vector.
+  /// \tparam TangentVectorType1 Type of the joint velocity vector.
+  /// \tparam TangentVectorType2 Type of the joint acceleration vector.
+  ///
+  /// \param[in] model The model structure of the rigid body system.
+  /// \param[in] data The data structure of the rigid body system.
+  /// \param[in] q The joint configuration vector (dim model.nq).
+  /// \param[in] v The joint velocity vector (dim model.nv).
+  /// \param[in] a The joint acceleration vector (dim model.nv).
+  ///
+  /// \return The joint torque regressor of the system.
+  ///
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType, typename TangentVectorType1, typename TangentVectorType2>
+  inline typename DataTpl<Scalar,Options,JointCollectionTpl>::MatrixXs &
+  computeJointTorqueRegressor(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                              DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                              const Eigen::MatrixBase<ConfigVectorType> & q,
+                              const Eigen::MatrixBase<TangentVectorType1> & v,
+                              const Eigen::MatrixBase<TangentVectorType2> & a);
+
 } // namespace pinocchio
 
 /* --- Details -------------------------------------------------------------------- */
--- a/src/algorithm/regressor.hpp
+++ b/src/algorithm/regressor.hpp
@@ -91,6 +91,31 @@ namespace pinocchio
   frameBodyRegressor(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
                      DataTpl<Scalar,Options,JointCollectionTpl> & data,
                      FrameIndex frameId);
+
+    ///
+    /// \brief Computes the joint torque regressor that links the joint torque
+    ///        to the dynamic parameters of each link according to the current the robot motion.
+    ///
+    /// \tparam JointCollection Collection of Joint types.
+    /// \tparam ConfigVectorType Type of the joint configuration vector.
+    /// \tparam TangentVectorType1 Type of the joint velocity vector.
+    /// \tparam TangentVectorType2 Type of the joint acceleration vector.
+    ///
+    /// \param[in] model The model structure of the rigid body system.
+    /// \param[in] data The data structure of the rigid body system.
+    /// \param[in] q The joint configuration vector (dim model.nq).
+    /// \param[in] v The joint velocity vector (dim model.nv).
+    /// \param[in] a The joint acceleration vector (dim model.nv).
+    ///
+    /// \return The joint torque regressor of the system.
+    ///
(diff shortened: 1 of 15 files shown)


src/algorithm/regressor.hxx
```diff
@@ -7,51 +7,209 @@
...
+  {
+    typedef typename MotionVelocity::Scalar Scalar;
+    enum { Options = PINOCCHIO_EIGEN_PLAIN_TYPE(typename MotionVelocity::Vector3)::Options };
+
+    typedef Symmetric3Tpl<Scalar,Options> Symmetric3;
+    typedef typename Symmetric3::SkewSquare SkewSquare;
+    typedef Eigen::Matrix<Scalar,6,10,Options> ReturnType;
+    using ::pinocchio::details::bigL;
+
+    Eigen::Matrix<Scalar, 3, 1, Options> acc = a.linear() + v.angular().cross(v.linear());
```
jcarpent: Avoid allocation here. You can use the res to store your result.

src/algorithm/regressor.hxx
```diff
@@ -7,51 +7,209 @@
...
+    typedef RneaForwardStep<Scalar,Options,JointCollectionTpl,ConfigVectorType,TangentVectorType1,TangentVectorType2> Pass1;
+    typename Pass1::ArgsType arg1(model,data,q.derived(),v.derived(),a.derived());
+    for(JointIndex i=1; i<(JointIndex)model.njoints; ++i)
+    {
+      Pass1::run(model.joints[i],data.joints[i],
+                 arg1);
+    }
+
+    typedef JointTorqueRegressorBackwardStep<Scalar,Options,JointCollectionTpl> Pass2;
+    Eigen::Matrix<Scalar,6,10,Options> jointRegressor;
```
jcarpent: Avoid memory allocation. Store this matrix in data.

src/spatial/inertia.hpp
```diff
@@ -266,6 +268,42 @@ namespace pinocchio
...
+    * The parameters are given as
+    * \f$ v = [m, mc_x, mc_y, mc_z, I_{xx}, I_{xy}, I_{yy}, I_{xz}, I_{yz}, I_{zz}]^T \f$
+    * where \f$ I = I_C + mS^T(c)S(c) \f$ and \f$ I_C \f$ has its origin at the barycenter.
+    */
+    Vector10 toDynamicParameters() const
+    {
+      Vector10 v;
+
+      v[0] = mass();
+      v.template segment<3>(1) = mass() * lever();
```
jcarpent: Use noalias here

src/spatial/inertia.hpp
```diff
@@ -266,6 +268,42 @@ namespace pinocchio
       return M;
     }
 
+    /** Returns the representation of the matrix as a vector of dynamic parameters.
+    * The parameters are given as
+    * \f$ v = [m, mc_x, mc_y, mc_z, I_{xx}, I_{xy}, I_{yy}, I_{xz}, I_{yz}, I_{zz}]^T \f$
```
jcarpent: It would be useful to define S in the documentation.

src/algorithm/regressor.hpp
```diff
@@ -28,12 +46,99 @@ namespace pinocchio
...
+  /// \param[in] frameId The id of the frame.
+  ///
+  /// \return The regressor of the body.
+  ///
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
+  inline Eigen::Matrix<Scalar,6,10,Options>
+  frameBodyRegressor(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                     DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                     FrameIndex frameId);
+
```
jcarpent: This is not correctly indented.

src/algorithm/regressor.hpp
```diff
@@ -28,12 +46,99 @@ namespace pinocchio
...
+    ///
+    /// \brief Computes the joint torque regressor that links the joint torque
+    ///        to the dynamic parameters of each link according to the current the robot motion.
+    ///
```
jcarpent: Can you add a single line to explain the math here. Similarly of what you have done previously in the doc.

src/algorithm/regressor.hxx
gabrielebndn: My reasoning was that, since this is a fixed-size matrix, it was ok to construct it inside the method itself. Plus I am always reusing the same matrix, within the method. Do you think it is important? I thought if I could avoid putting an extra temporary instance variable inside data, it would be better

src/spatial/inertia.hpp
gabrielebndn: Wait. Really, noalias? That's a product between a scalar and a vector. Is noalias needed?

gabrielebndn:
@jcarpent thanks for your comments. Concerning the final two notes I left to the original post, do you have suggestions?

src/spatial/inertia.hpp
jcarpent: It will do allocation for it.

src/algorithm/regressor.hxx
jcarpent: It will do stack allocation here I guess, because of the size which is larger than a certain value and Eigen will force it.

src/spatial/inertia.hpp
gabrielebndn: I am surprised, I thought a scalar product would not do it. However, it costs nothing to add it.

src/algorithm/regressor.hxx
gabrielebndn: Oh. Ok. But then I should probably rewrite the methods `bodyRegressor`, `jointBodyRegressor` and `frameBodyRegressor` so that they have versions not doing allocation either, I think, otherwise we will have the same problem. `bodyRegressor` and `jointBodyRegressor`, in particular, are employed by `jointTorqueRegressor`.

Is it the still worth it to employ `Eigen::Matrix<Scalar,6,10,Options>`, rather than `Eigen::Matrix<Scalar,Dynamic,Dynamic,Options>`?

src/algorithm/regressor.hpp
gabrielebndn: Actually, it is the algorithm just after it which is not correctly indented, with four spaces instead of the usual two

src/algorithm/regressor.hxx
jcarpent: > Is it the still worth it to employ `Eigen::Matrix<Scalar,6,10,Options>`, rather than `Eigen::Matrix<Scalar,Dynamic,Dynamic,Options>`?

Yes, for static dimensions checking and unrolling of operations.

jcarpent: > Oh. Ok. But then I should probably rewrite the methods `bodyRegressor`, `jointBodyRegressor` and `frameBodyRegressor` so that they have versions not doing allocation either, I think, otherwise we will have the same problem. `bodyRegressor` and `jointBodyRegressor`, in particular, are employed by `jointTorqueRegressor`.

I do agree. Thanks a lot.

jcarpent:
> @jcarpent thanks for your comments. Concerning the final two notes I left to the original post, do you have suggestions?

I have answered in the post. The rest seems fine to me.

src/algorithm/regressor.hxx
gabrielebndn: Done. I added a typedef `Data::BodyRegressorType` and a field `data.bodyRegressor`, which is employed by `jointBodyRegressor` and `frameBodyRegressor` to store their results, and by `computeJointTorqueRegressor` as an auxiliary variable.
Let me know if you like the solution.
Wait before you merge, as I still have to fix my NOTE 1.

gabrielebndn:
> > @jcarpent thanks for your comments. Concerning the final two notes I left to the original post, do you have suggestions?
> 
> I have answered in the post. The rest seems fine to me.

I think you might have missed what I mean when I mentioned the "final notes". I meant this:

> NOTE 1: notice that, as forward step of the algorithm, I am using RneaForwardStep. I did it to go quickly, and also to avoid the need of compiling another step. I now realize calling RneaForwardStep will modify data.f, leaving it in an inconsistent state. This should be fixed. Therefore, I will either create a custom forward step or find another way to fix the problem which does not require creating a new step type (I can think of at least two,both of which would slightly increase execution time).

Do you agree it might be a concern? I think I will need to make a custom forward step for it, which is identical to RNEA except that it does not touch `data.f`

src/algorithm/regressor.hxx
jmirabel: I think a matrix with 60 doubles is far below the stack allocation limit. From Eigen source code:
```cpp
#ifndef EIGEN_STACK_ALLOCATION_LIMIT
// 131072 == 128 KB
#define EIGEN_STACK_ALLOCATION_LIMIT 131072
#endif
```

src/spatial/inertia.hpp
jmirabel: I think Eigen does not consider aliasing of the form `a.segment<1,2>() = a.segment<0,2>();`. So there is no aliasing the case mentioned above.

Moreover, noalias for vectors of very small size is counter-productive. It can be *more efficient* to use a temporary.

jcarpent:
> Do you agree it might be a concern? I think I will need to make a custom forward step for it, which is identical to RNEA except that it does not touch `data.f`

Yes, I think this is not a problem. I will prefer that to have minimum number of operations, and self contained algorithms

src/spatial/inertia.hpp
gabrielebndn: Yes, I do not see where the temporary allocation might arise.

src/algorithm/regressor.hxx
gabrielebndn: Well, if it is not necessary, I would rather revert the changes

src/spatial/inertia.hpp
jcarpent: > I think Eigen does not consider aliasing of the form `a.segment<1,2>() = a.segment<0,2>();`. So there is no aliasing the case mentioned above.
> 
> Moreover, noalias for vectors of very small size is counter-productive. It can be _more efficient_ to use a temporary.

By default, Eigen is considering aliasing for any product expression.

gabrielebndn:
> > Do you agree it might be a concern? I think I will need to make a custom forward step for it, which is identical to RNEA except that it does not touch `data.f`
> 
> Yes, I think this is not a problem. I will prefer that to have minimum number of operations, and self contained algorithms

Fixed

src/spatial/inertia.hpp
gabrielebndn: > > I think Eigen does not consider aliasing of the form `a.segment<1,2>() = a.segment<0,2>();`. So there is no aliasing the case mentioned above.
> > Moreover, noalias for vectors of very small size is counter-productive. It can be _more efficient_ to use a temporary.
> 
> By default, Eigen is considering aliasing for any product expression. Indeed, it has no clue to guest it at compile time.

Any *matrix* product. This is a *scalar* product. Eigen knows it and there is no reasoin to fear aliasing. From Eigen's doc:

[https://eigen.tuxfamily.org/dox/group__TopicAliasing.html](https://eigen.tuxfamily.org/dox/group__TopicAliasing.html)
> Aliasing is harmless with coefficient-wise computations; **this includes scalar multiplication** and matrix or array addition.
**When you multiply two matrices, Eigen assumes that aliasing occurs.** If you know that there is no aliasing, then you can use noalias().
**In all other situations, Eigen assumes that there is no aliasing issue** and thus gives the wrong result if aliasing does in fact occur. To prevent this, you have to use eval() or one of the xxxInPlace() functions.

(emphasis mine)

I will remove the noalias() as it is useless

src/algorithm/regressor.hxx
```diff
@@ -7,51 +7,258 @@
...
+    PINOCCHIO_ASSERT_MATRIX_SPECIFIC_SIZE(OutputType, params, 6, 10);
+
+    typedef typename MotionVelocity::Scalar Scalar;
+    enum { Options = PINOCCHIO_EIGEN_PLAIN_TYPE(typename MotionVelocity::Vector3)::Options };
+
+    typedef Symmetric3Tpl<Scalar,Options> Symmetric3;
+    typedef typename Symmetric3::SkewSquare SkewSquare;
+    using ::pinocchio::details::bigL;
+
+    Eigen::MatrixBase<OutputType> & res = const_cast< Eigen::MatrixBase<OutputType> & >(regressor);
```
jcarpent: you can use here PINOCCHIO_EIGEN_CONST_CAST macro, simpler to use

src/spatial/inertia.hpp
jmirabel: > By default, Eigen is considering aliasing for any product expression.

for any matrix-matrix product, not scalar-matrix product, which falls into unary operation in Eigen.

src/algorithm/regressor.hxx
```diff
@@ -7,51 +7,258 @@
...
+
+    res.template block<3,1>(MotionVelocity::LINEAR,0) = a.linear() + v.angular().cross(v.linear());
+    const Eigen::Matrix<Scalar, 3, 1, Options> & acc = res.template block<3,1>(MotionVelocity::LINEAR,0);
```
jcarpent: You are making a copy here for nothing.

src/algorithm/regressor.hxx
```diff
@@ -7,51 +7,258 @@
...
+    }
+    
+    return data.staticRegressor;
+  }
+
+  namespace details {
+    // auxiliary function for bodyRegressor: bigL(omega)*I.toDynamicParameters().tail<6>() = I.inertia() * omega
+    template<typename Vector3Like>
+    inline Eigen::Matrix<typename Vector3Like::Scalar,3,6,PINOCCHIO_EIGEN_PLAIN_TYPE(Vector3Like)::Options>
+    bigL(const Eigen::MatrixBase<Vector3Like> & omega)
```
jcarpent: to be efficient, bigL must take its return type as argument (no copy, no allocation)

src/spatial/inertia.hpp
jcarpent: Ok, sounds good to me then. I was not aware of it before.

gabrielebndn: I removed it

jcarpent:
@gabrielebndn Do you want and have time to address the last comments have made?
Otherwise, we can keep it as it is, but it won't be the best we can perform, which is not an issue in its own.

src/algorithm/regressor.hxx
gabrielebndn: What? No. I am using a reference. This is a change I introduce with last commits following one of your previous remarks. `acc` is used later

gabrielebndn:
> @gabrielebndn Do you want and have time to address the last comments have made?
> Otherwise, we can keep it as it is, but it won't be the best we can perform, which is not an issue in its own.

Sorry no time, I'll be away the whole weekend and I really have to run

src/algorithm/regressor.hxx
gabrielebndn: You are right about this. Maybe we can change this in a second moment

src/algorithm/regressor.hxx
gabrielebndn: Oh. Forgot about that. Again, I may think about it in the future

src/algorithm/regressor.hxx
jcarpent: As you want. But because of Variant, it will allocate the memory 60 doubles x number of joints types.

gabrielebndn: What about this? Should I revert the changes? In which case, I will ask you to leave the PR open

src/algorithm/regressor.hxx
jcarpent: Yes, but a block is not an Eigen::Matrix, so it will allocate the memory. You have to use a block expression to address it.

src/algorithm/regressor.hxx
gabrielebndn: > As you want. But because of Variant, it will allocate the memory 60 doubles x number of joints types.

Mmmh, I understand... Then, we may keep the changes

src/algorithm/regressor.hxx
jcarpent: I will do it on top of your PR.

src/algorithm/regressor.hxx
gabrielebndn: Mmh, I thought Eigen was smarter than that... This is a really quick fix, I'll do it now

jcarpent:
It is fine for me. I will wait travis to merge.

src/algorithm/regressor.hxx
gabrielebndn: Actually, no, no time. I'll fix it in the future

jcarpent: The problem is not coming from Eigen here but from C++ itself ;)

jcarpent:
@gabrielebndn Thanks a lot again for this very nice work!

PR #819, opened 2019-06-21
[doc] Documentation for SE3 with examples and nice figures for Joints.

tdouin:


.gitignore
```diff
@@ -3,3 +3,5 @@ Xcode*
 *~
 *.pyc
 coverage*
+.vscode*
+_build*
```
jcarpent: This line is specific to your settings. I would prefer to not put it here. @tdouin Do you agree?

tdouin: yes sure.

nim65s: @tdouin: for information, you can add eg.
```
[core]
    excludesfile = ~/.global_gitignore
```
In your `~/.gitconfig`, and put those `.vscode*` & `_build*` in your `~/.global_gitignore`, and never worry about those again :)

jcarpent:
@tdouin I'm making some modifications on your branch. These are just minor improvements in your excellent text to make it a bit clearer. I hope you won't mine.

.gitignore
tdouin: Ok i get it, thanks for the advice

jmirabel:
@tdouin nice work !

jcarpent:
> May I ask why? Some people like to call their build folders _build-something. Not me, but others do. Why removing this possibility?

We cannot handle all the cases, and at this stage, the common way to do it is to cover the classic cases with Visual Studio, etc. The rest is very dependent on the user setup.

jcarpent:
@jmirabel Do you want to add other changes or comments?

jmirabel:
> @jmirabel Do you want to add other changes or comments?

No

jcarpent:
Very nice work @tdouin. Thanks a lot!

Issue #790, opened 2019-04-30
Automatically setting PYTHON_DEB_LAYOUT


wxmerkt:
Hi,
While this issue technically belongs to https://github.com/jrl-umi3218/jrl-cmakemodules, the discussion upstream has been postponed several times and referred to user configuration (e.g., https://github.com/jrl-umi3218/jrl-cmakemodules/pull/47). 
One of the most common problems users have experienced with the JRL-UMI cmake module is that the Python paths are not set correctly and require manually passing an option (`-DPYTHON_USE_DEB_LAYOUT=1`, cf. [here](https://github.com/jrl-umi3218/jrl-cmakemodules/blob/d4feb8a5220b196a8375ca2ec5fa505272e82c33/python.cmake#L83-L84)) during configuration.

In a catkin environment, the system automatically provides `SETUPTOOLS_DEB_LAYOUT`. Would it be an acceptable option to add the following to eigenpy and pinocchio? [Or, better yet upstream to the cmake modules]:

```cmake
if(SETUPTOOLS_DEB_LAYOUT)
  SET(PYTHON_DEB_LAYOUT ON CACHE BOOL "Python directories for debian distributions")
endif(SETUPTOOLS_DEB_LAYOUT)
```

This would allow to build and package eigenpy/pinocchio etc. easier without the current wrappers (simply by adding a `package.xml`).

What do you think?

jmirabel:
A few questions regarding SETUPTOOLS_DEB_LAYOUT:
- Who provides it ? catkin ?
- Is its meaning documented somewhere ?
- Why don't they also provide PYTHON_DEB_LAYOUT along with SETUPTOOLS_DEB_LAYOUT automatically ?

jcarpent:
I am not so familiar with ROS, but regarding your comments, I would suggest putting it directly in the JRL Cmake module.

wxmerkt:
- Yes, catkin sets it if `find_package(catkin)` is called in a project CMake/or if it's a catkin_make/catkin-tools built workspace. It is set [here](https://github.com/ros/catkin/blob/928c2025206e1a3f0d9d504ef5b26efe1cc95d8b/cmake/python.cmake#L10-L14) and was introduced in https://github.com/ros/catkin/pull/114
[...]

PR #822, opened 2019-06-25
cmake: do not check for Catkin when it is not needed

jcarpent:
This is for compilation issues in the CI process.
@wxmerkt Is it fine for you? Could you make your ROS building code compliant this new way of checking for Catkin?


wxmerkt:
This likely is a regression for us and makes things more complicated. 

What is the compilation issue with a _quiet find_package_? Perhaps we can fix that issue instead.

Practically speaking, we could get rid off this catkin-type search entirely if Pinocchio uses the correct Python package path (i.e., whatever is standard on the distro and not forcing `site-packages` for all).

jcarpent:
> Practically speaking, we could get rid off this catkin-type search entirely if Pinocchio uses the correct Python package path (i.e., whatever is standard on the distro and not forcing `site-packages` for all).

Ok. We can go towards this solution. Could you provide me more details?

wxmerkt:
That would be ideal - we could totally go away with the catkin-wrappers then by using a simple `package.xml` and the jrl cmake modules.

The key here is to detect whether the system is debian-flavoured and set `PYTHON_DEB_LAYOUT` (exposed by jrl cmake modules in [2]) accordingly. [3] was attempting to automatically set it, but got blocked. It is now up to the individual user projects to add that logic (i.e., we'd need here, hpp-fcl, and eigenpy).

References:
1. https://github.com/jrl-umi3218/jrl-cmakemodules/issues/38, issue discussing incorrect Python-paths on debian-flavoured distributions.
2. https://github.com/jrl-umi3218/jrl-cmakemodules/pull/39, added Python-install location option
3. https://github.com/jrl-umi3218/jrl-cmakemodules/pull/47, would activate the correct flag automatically on a debian-based distribution - yet, blocked as it would affect too many downstream and suggested to handle on a per-project basis.
4. https://github.com/stack-of-tasks/pinocchio/issues/790, an ill-fated work-around suggestion to achieve the same thing.

jcarpent:
@wxmerkt Thanks for this quick reply.
I would go towards #790, but the problem is that if catkin is present on the current machine, it means that Pinocchio will be installed in `dist-packages` while maybe this not intended by the user.

Is there a way to detect that Catkin is trying to build Pinocchio and then set on the respective flags?

nim65s:
For the records:

```
File "/root/robotpkg/math/py-pinocchio/work.runner-7gqkM7SY-project-5-concurrent-35/pinocchio-2.1.2/catkin_generated/generate_cached_setup.py", line 20, in <module>
from catkin.environment_cache import generate_environment_script
ImportError: No module named catkin.environment_cache
CMake Error at /opt/ros/kinetic/share/catkin/cmake/safe_execute_process.cmake:11 (message):
execute_process(/usr/bin/python2.7
"/root/robotpkg/math/py-pinocchio/work.runner-7gqkM7SY-project-5-concurrent-35/pinocchio-2.1.2/catkin_generated/generate_cached_setup.py")
returned error code 1
Call Stack (most recent call first):
/opt/ros/kinetic/share/catkin/cmake/all.cmake:198 (safe_execute_process)
/opt/ros/kinetic/share/catkin/cmake/catkinConfig.cmake:20 (include)
CMakeLists.txt:93 (find_package)
```
In https://gepgitlab.laas.fr/stack-of-tasks/pinocchio/-/jobs/21970

I have no time to allow to this right now, sorry.

jcarpent:
> For the records:
> 
> ```
> Traceback (most recent call last):
> File "/root/robotpkg/math/py-pinocchio/work.runner-7gqkM7SY-project-5-concurrent-35/pinocchio-2.1.2/catkin_generated/generate_cached_setup.py", line 20, in <module>
> from catkin.environment_cache import generate_environment_script
> ImportError: No module named catkin.environment_cache
> CMake Error at /opt/ros/kinetic/share/catkin/cmake/safe_execute_process.cmake:11 (message):
> execute_process(/usr/bin/python2.7
> "/root/robotpkg/math/py-pinocchio/work.runner-7gqkM7SY-project-5-concurrent-35/pinocchio-2.1.2/catkin_generated/generate_cached_setup.py")
> returned error code 1
> Call Stack (most recent call first):
> /opt/ros/kinetic/share/catkin/cmake/all.cmake:198 (safe_execute_process)
> /opt/ros/kinetic/share/catkin/cmake/catkinConfig.cmake:20 (include)
> CMakeLists.txt:93 (find_package)
> ```
> 
> In https://gepgitlab.laas.fr/stack-of-tasks/pinocchio/-/jobs/21970
> 
> I have no time to allow to this right now, sorry.

@nim65s I think we need to deactivate the search of catkin under robotpkg. This is easy to do.

wxmerkt:
> I would go towards #790, but the problem is that if catkin is present on the current machine, it means that Pinocchio will be installed in dist-packages while maybe this not intended by the user.

Having thought more about it, I don't like the approach I initially proposed in #790 and the switch should be distro-dependent, not catkin-dependent. 
I'd like to remove all catkin-related modifications that are required workarounds from the CMakeLists. The correct way would be to follow the standard of the platform, and only do non-standard paths when absolutely necessary (i.e., if you choose to install always to `site-packages` in robotpkg, activate that flag only during robotpkg builds).

> Is there a way to detect that Catkin is trying to build Pinocchio and then set on the respective flags?

I tried and it got overcomplicated/wasn't possible. Ideally, resolving the Python package path situation would be best and would completely remove all catkin from the CMakeLists (which are a required workaround to deal with non-standard Python package paths).

My hunch to the error above is that Python35 does not have catkin installed - but CMake triggered it and forces `alias python=python3`. So the failures likely are only on Python3 builds.

jmirabel:
>  (i.e., if you choose to install always to site-packages in robotpkg, activate that flag only during robotpkg builds).

That sounds good to me. I like the rule: *if the user did not specify anything, use the target OS settings*. 

Sadly, the t-OS (target OS), different from the b-OS (build OS) cannot be specified with CMake, can it ?
Does CMake assume that the t-OS is the b-OS ?
I believe the best way to answer the question @wxmerkt raised, and that has been regularly raised, is to find a good solution (and implementation in the JRL cmake module) to the problem:
*We need a way of knowing what the target OS is.*

Here are some suggestions:
- assume the t-OS is the b-OS, except if a particular CMake variable set the t-OS to something else.
- do target specific choices only if a CMake variable is set (otherwise, do not do any specific, not even for b-OS).

Maybe the good solution is to rely on `CMAKE_SYSTEM_NAME`: https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/CrossCompiling
This requires us to write a `Platform/${CMAKE_SYSTEM_NAME}.cmake` for each platform we want to support. I believe **this is a hack** which may break things. Maybe a PROJECT_SYSTEM_NAME, with a file in the JRL cmake module, is better.


Any thoughts or suggestions ?

wxmerkt:
Hi @jcarpent, sorry I have been out of the loop, what is the outcome of your conversations/discussions to package according to platform-standard unless deactivated?

jcarpent:
@jcarpent I will provide a new PR for that. 
So, the idea is to make Pinocchio looking for the current OS policy as you mentioned with @jmirabel.
We will force the flag if needed when using robotpkg.
Is it fine for you?

jcarpent:
I would even suggest to use Python directly to set `PYTHON_PACKAGES_DIR`. What do you think?

wxmerkt:
Thank you very much for the change to cmake upstream, I made a comment there :-)

jcarpent:
With the update on the jrl-cmake-modules, @wxmerkt do you think we can now remove the references to Catkin?

wxmerkt:
Yes, I will make a series of pull requests to adjust. I want to first do a release to the ROS buildfarm of those so we can then adjust the dependent packages (but it will be a breaking changing of that version for some). E.g., eigenpy is currently (but will be removed because of the packaging issues) part of MoveIt and we also have dependent packages. 

Is it okay if I make the PR to remove it in a few days? And can we tag new releases with the updated cmake-module synced into hpp-fcl, eigenpy, and pinocchio?

jcarpent:
Ok. I can make the PR for Pinocchio if you want?

jcarpent:
with @lesteve, we will work soon on the packaging of Pinocchio inside `anaconda`. So we need to proceed as follows for eigenpy too.

wxmerkt:
If you can sync the cmake submodule and tag releases, I will test the new ROS packaging (without the catkin-cmake wrapper) and test its impact

wxmerkt:
Actually fine to remove catkin right now, we have released, tagged versions

jcarpent:
> If you can sync the cmake submodule and tag releases, I will test the new ROS packaging (without the catkin-cmake wrapper) and test its impact

Ok sure. I've made a first proposal in #826.

Issue #839, opened 2019-07-22
Compute CoM jacobian of subtrees


matthieuvigne:
Currently pinocchio computes the CoM of subtrees, but only the jacobian of the total model CoM. To implement tasks on a specific limb of a robot, I would like to have the jacobian of the subtree CoM as well. 

This could be done by adding a ```std::vector<Matrix3x> Jcom;``` to the Data object that would be filled during the JacobianCenterOfMassBackwardStep

jcarpent:
Adding `std::vector<Matrix3x> Jcom` would not be efficient, because I guess you only need one center of mass Jacobian for a specific subtree, not for all them.

I would suggest implementing a dedicated function called `computeCoMSubtreeJacobian` which will fill a matrix given as output argument, and doing the job you want. 

@matthieuvigne Are you willing to fulfill this task?

matthieuvigne:
I felt doing a vector was closer to what pinocchio usually does in having one algorithm compute everything as it goes and storing the results for latter use ; this is exactly what is being done for subtree CoMs for instance (in ```container::aligned_vector<Vector3> com```). So I thought it made sense to do the same thing for the Jacobian. But yes for my use case a function would be enough.

And yes I can implement this, just wanted to get you opinion before getting started.

jmirabel:
Jacobian for all subtrees consumes a lot of memory. Typically, you will flood your CPU cache. This can cause inefficiencies.

matthieuvigne:
> Typically, you will flood your CPU cache.

Good point, I hadn't thought of that. I'll go with a basic function then.

jcarpent:
> Jacobian for all subtrees consumes a lot of memory. Typically, you will flood your CPU cache. This can cause inefficiencies.

I do agree. This is also why we only store the Jacobians of all the joints in `data.J` and extract then the right Jacobians when needed.

jcarpent:
> And yes I can implement this

Sounds good. Thanks a lot!

jcarpent:
See #843 for a full and efficient solution.

jcarpent:
fixed by #843.

PR #841, opened 2019-07-23
Add subtree CoM jacobian

matthieuvigne:
Fix https://github.com/stack-of-tasks/pinocchio/issues/839

Add jacobianSubtreeCenterOfMass and associated python bindings.
This function computes the jacobian of the center of mass of
a subtree, as a barycenter of individual body jacobian.

### DIFF
--- a/src/algorithm/center-of-mass.hpp
+++ b/src/algorithm/center-of-mass.hpp
@@ -10,7 +10,7 @@
 
 namespace pinocchio
 {
-  
+
   /// \brief Compute the total mass of the model and return it.
   ///
   /// \param[in] model The model structure of the rigid body system.
@@ -62,7 +62,7 @@ namespace pinocchio
                DataTpl<Scalar,Options,JointCollectionTpl> & data,
                const Eigen::MatrixBase<ConfigVectorType> & q,
                const bool computeSubtreeComs = true);
-  
+
   ///
   /// \brief Computes the center of mass position and velocity of a given model according to a particular joint configuration and velocity.
   ///        The result is accessible through data.com[0], data.vcom[0] for the full body com position and velocity.
@@ -87,7 +87,7 @@ namespace pinocchio
                const Eigen::MatrixBase<ConfigVectorType> & q,
                const Eigen::MatrixBase<TangentVectorType> & v,
                const bool computeSubtreeComs = true);
-  
+
   ///
   /// \brief Computes the center of mass position, velocity and acceleration of a given model according to a particular joint configuration, velocity and acceleration.
   ///        The result is accessible through data.com[0], data.vcom[0], data.acom[0] for the full body com position, velocity and acceleation.
@@ -115,7 +115,7 @@ namespace pinocchio
                const Eigen::MatrixBase<TangentVectorType1> & v,
                const Eigen::MatrixBase<TangentVectorType2> & a,
                const bool computeSubtreeComs = true);
-  
+
   ///
   /// \brief Computes the center of mass position, velocity and acceleration of a given model according to the current kinematic values contained in data and the requested level.
   ///        The result is accessible through data.com[0], data.vcom[0] and data.acom[0] for the full body com position and velocity.
@@ -133,7 +133,7 @@ namespace pinocchio
                            DataTpl<Scalar,Options,JointCollectionTpl> & data,
                            const int LEVEL,
                            const bool computeSubtreeComs = true);
-  
+
   ///
   /// \brief Computes the center of mass position, velocity and acceleration of a given model according to the current kinematic values contained in data.
   ///        The result is accessible through data.com[0], data.vcom[0] and data.acom[0] for the full body com position and velocity.
@@ -150,7 +150,7 @@ namespace pinocchio
                            DataTpl<Scalar,Options,JointCollectionTpl> & data,
                            const bool computeSubtreeComs = true)
   { centerOfMass(model,data,2,computeSubtreeComs); }
-  
+
   ///
   /// \brief Computes both the jacobian and the the center of mass position of a given model according to a particular joint configuration.
   ///        The results are accessible through data.Jcom and data.com[0] and are both expressed in the world frame. In addition, the algorithm also computes the Jacobian of all the joints (\sa pinocchio::computeJointJacobians).
@@ -177,7 +177,7 @@ namespace pinocchio
                        const Eigen::MatrixBase<ConfigVectorType> & q,
                        const bool computeSubtreeComs,
                        const bool updateKinematics);
-  
+
   ///
   /// \brief Computes both the jacobian and the the center of mass position of a given model according to a particular joint configuration.
   ///        The results are accessible through data.Jcom and data.com[0] and are both expressed in the world frame. In addition, the algorithm also computes the Jacobian of all the joints (\sa pinocchio::computeJointJacobians).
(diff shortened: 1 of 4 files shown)


src/algorithm/center-of-mass.hpp
```diff
@@ -221,6 +221,46 @@ namespace pinocchio
...
+  /// \param[in] parentJointIdIn Index of the parent joint supporting the subtree.
+  /// \param[out] J A reference on the Jacobian matrix where the results will be stored in (dim 3 x model.nv). You must fill J with zero elements, e.g. J.fill(0.).
+  ///
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType>
+  inline void
+  jacobianSubtreeCenterOfMass(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                              DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                              const Eigen::MatrixBase<ConfigVectorType> & q,
+                              const JointIndex & parentJointIdIn,
+                              typename DataTpl<Scalar,Options,JointCollectionTpl>::Matrix3x & JacobianOut);
```
jcarpent: It would be better to use here: `const Eigen::MatrixBase<Matrix3xLike> & JacobianOut` to allow any Eigen type

src/algorithm/center-of-mass.hpp
```diff
@@ -221,6 +221,46 @@ namespace pinocchio
...
+  /// \param[in] data The data structure of the rigid body system.
+  /// \param[in] parentJointIdIn Index of the parent joint supporting the subtree.
+  /// \param[out] J A reference on the Jacobian matrix where the results will be stored in (dim 3 x model.nv). You must fill J with zero elements, e.g. J.fill(0.).
+  ///
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
+  inline void
+  jacobianSubtreeCenterOfMass(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                              DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                              const JointIndex & parentJointIdIn,
+                              typename DataTpl<Scalar,Options,JointCollectionTpl>::Matrix3x & JacobianOut);
```
jcarpent: Same applies here.

src/algorithm/center-of-mass.hpp
```diff
@@ -221,6 +221,46 @@ namespace pinocchio
...
+  ///
+  /// \brief Computes the jacobian of the center of mass of the given subtree according to the current value stored in data.
+  ///        It assumes that forwardKinematics has been called first.
+  ///
+  /// \tparam JointCollection Collection of Joint types.
+  /// \tparam ConfigVectorType Type of the joint configuration vector.
+  ///
+  /// \param[in] model The model structure of the rigid body system.
+  /// \param[in] data The data structure of the rigid body system.
+  /// \param[in] parentJointIdIn Index of the parent joint supporting the subtree.
```
jcarpent: I would suggest using a better naming: replacing `parentJointIdIn` with `root_subtree_id`

src/algorithm/center-of-mass.hxx
```diff
@@ -295,74 +296,119 @@ namespace pinocchio
...
+
+    if (parentJointIdIn == 0)
+    {
+      // Joint 0 is universe: this means that the user is asking for the entier model CoM jacobian.
+      JacobianOut = jacobianCenterOfMass(model, data);
+    }
+    else
+    {
+      // Iterate over all bodies of subtree.
+      for (JointIndex id : model.subtrees[parentJointIdIn])
```
jcarpent: This is a C++11 feature. Please use the classic way of setting a `for` loop

src/algorithm/center-of-mass.hxx
```diff
@@ -295,74 +296,119 @@ namespace pinocchio
...
+      {
+        // Get joint jacobian in world.
+        Matrix6x jointJacobian = Matrix6x::Zero(6, model.nv);
+        getJointJacobian(model, data, id, ReferenceFrame::LOCAL, jointJacobian);
```
jcarpent: This code is not efficient. You may proceed without requiring any call to `getJointJacobian`.
I think you might derive an efficient algo from the existing `jacobianCenterOfMass` with very few minor changes.

src/algorithm/center-of-mass.hpp
jcarpent: This is just a suggestion. Any other idea is welcome.

matthieuvigne:
@jcarpent Thanks for the review. I had a look at the way jacobianCenterOfMass was implemented (I guess you are talking about the `JacobianCenterOfMassBackwardStep` algorithm), but I felt that, since the jacobians of the joints are already computed, the code would be much more readable by reusing them. It's less efficient but I was ready to live with it. I'll probably need some help reimplementing it the other way around...

Since I was only adding one function, with associated unit test and python bindings, I felt this all belonged to one commit. Sorry about the trailing spaces, my IDE actually automatically removes them on file save - and I though it made more sense to leave it that way (in fact I would suggest you remove all trailing spaces from pinocchio and all configure your editors to work that way, it makes the code look cleaner). Do you want me to revert the files to their previous states ?

jcarpent:
> I'll probably need some help reimplementing it the other way around...

You made the structure, I may provide a proposition. Is it fine for you?

jcarpent:
> Since I was only adding one function, with associated unit test and python bindings, I felt this all belonged to one commit. Sorry about the trailing spaces, my IDE actually automatically removes them on file save - and I though it made more sense to leave it that way (in fact I would suggest you remove all trailing spaces from pinocchio and all configure your editors to work that way, it makes the code look cleaner). Do you want me to revert the files to their previous states ?

I do understand, and unfortunately, we haven't been serious on this point since the beginning of the project. I would suggest not wasting your time, but next time it would be nice if you can split your work into several pieces.

src/algorithm/center-of-mass.hpp
matthieuvigne: I though you were also using upperCamelCase for variable names when applicable ?

src/algorithm/center-of-mass.hpp
matthieuvigne: What do you mean by Matrix3xLike ? Isn't the code here templated by `Scalar` already ?

jcarpent: You need to have an input which can represent any Matrix3 (which can be a dense matrix, a simple block, etc.). This why in all our algorithms, we have always something like `Eigen::MatrixBase<MatrixType> & ` which says that we can take as input any Eigen Matrix like type.

I encourage you yo read this excellent post in the Eigen documentation: https://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html

jcarpent: You need to have an input which can represent any Matrix3 (which can be a dense matrix, a simple block, etc.). This why in all our algorithms, we have always something like `Eigen::MatrixBase<MatrixType> & ` which says that we can take as input any Eigen Matrix like type.

I encourage you to read this excellent post in the Eigen documentation: https://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html

src/algorithm/center-of-mass.hpp
jcarpent: My suggestion is much more related to the naming choice.

src/algorithm/center-of-mass.hxx
matthieuvigne: Do you think that you will have time to do this is the coming days ? If not I may try to have a look at it tonight.

jcarpent: Tomorrow if you don’t mind.

jcarpent:
I will close this PR as #843 offers a better solution.

PR #854, opened 2019-08-08
Python 2/3 compatibility of examples.

lesteve:
I was running the examples to check out the recent `pinocchio` `conda-forge` package and bumped into a few minor issues.

### DIFF
--- a/examples/python/gepetto-viewer.py
+++ b/examples/python/gepetto-viewer.py
@@ -8,9 +8,14 @@
 import pinocchio as pin
 import numpy as np
 import os
-from future.builtins import input
 
 from pinocchio.visualize import GepettoVisualizer
+
+try:
+    # Python 2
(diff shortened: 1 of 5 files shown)


examples/python/gepetto-viewer.py
```diff
@@ -8,10 +8,15 @@
 import pinocchio as pin
 import numpy as np
 import os
-from future.builtins import input
 
 from pinocchio.visualize import GepettoVisualizer
 
+try:
```
lesteve: This is a bit manual to do it like this in order to have `input` mean the same thing on Python 2 and Python 3 but you don't have the additional dependency on the `future` package.

jcarpent: Sounds good to me. @lesteve Thanks for the fixes.

PR #871, opened 2019-09-12
travis: add bionic

jcarpent:


jcarpent:
Seems we are reaching the limits :(

nim65s:
To workaround those limits, we can also decide that some tests should be run only in certain cases. 

For examples, on Gitlab, it is usual to have some tests only on merge requests, and others only when the devel branch is updated. This way, we can still run all tests on all merge requests, but not at once.

jcarpent:
I was thinking to switch to other services for CI, like Azureus or Action (by GitHub) to remove these annoying limitations.

nim65s:
That's an option. As long as those services can run docker containers, I can provide pre-built images.

jmirabel:
You should consider using `-DCMAKE_CXX_FLAGS_DEBUG="-O1"`. I don't think you will miss any bugs (NDEBUG is still not defined).

At least, I suggest to remove the `-g` which is useless on Travis.

jcarpent:
@nim65s It seems to be possible.

jcarpent:
Seems to be good now thanks to parallel jobs.
@nim65s and @jmirabel Thanks a lot for your feedback.

PR #873, opened 2019-09-13
Fix memory errors in unit tests

gabrielebndn:
A few unit test contained some memory errors, as they where using uninitialized values.
These errors would not make the tests fail, but they caused valgrind to complain.

I fixed the errors. Now the tests are cleaner: in one case the input matrices where not set to zero and in another one a test was performed which did not make sense (the third error being negligible).

I would like to point out that valgrind complains a lot about python-parser-test too... However, since there is Python in the middle, it is much more difficult to analyze.

### DIFF
--- a/unittest/aba-derivatives.cpp
+++ b/unittest/aba-derivatives.cpp
@@ -283,7 +283,6 @@ BOOST_AUTO_TEST_CASE(test_multiple_calls)
   
   BOOST_CHECK(data1.dFdq.isApprox(data2.dFdq));
   BOOST_CHECK(data1.dFdv.isApprox(data2.dFdv));
-  BOOST_CHECK(data1.dFda.isApprox(data2.dFda));
   
   BOOST_CHECK(data1.dtau_dq.isApprox(data2.dtau_dq));
   BOOST_CHECK(data1.dtau_dv.isApprox(data2.dtau_dv));
--- a/unittest/contact-dynamics.cpp
+++ b/unittest/contact-dynamics.cpp
@@ -292,8 +292,10 @@ BOOST_AUTO_TEST_CASE (timings_fd_llt)
   const std::string LF = "lleg6_joint";
   
   Data::Matrix6x J_RF (6, model.nv);
+  J_RF.setZero();
   getJointJacobian(model, data, model.getJointId(RF), LOCAL, J_RF);
   Data::Matrix6x J_LF (6, model.nv);
+  J_LF.setZero();
   getJointJacobian(model, data, model.getJointId(LF), LOCAL, J_LF);
   
   Eigen::MatrixXd J (12, model.nv);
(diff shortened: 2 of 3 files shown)


unittest/aba-derivatives.cpp
```diff
@@ -283,7 +283,6 @@ BOOST_AUTO_TEST_CASE(test_multiple_calls)
   
   BOOST_CHECK(data1.dFdq.isApprox(data2.dFdq));
   BOOST_CHECK(data1.dFdv.isApprox(data2.dFdv));
-  BOOST_CHECK(data1.dFda.isApprox(data2.dFda));
```
jcarpent: Why?

jcarpent:
> I would like to point out that valgrind complains a lot about python-parser-test too... However, since there is Python in the middle, it is much more difficult to analyze.

This is expected as EigenPy is doing allocation. What do you mean by complaining?

unittest/contact-dynamics.cpp
```diff
@@ -292,8 +292,10 @@ BOOST_AUTO_TEST_CASE (timings_fd_llt)
   const std::string LF = "lleg6_joint";
   
   Data::Matrix6x J_RF (6, model.nv);
+  J_RF.setZero();
   getJointJacobian(model, data, model.getJointId(RF), LOCAL, J_RF);
   Data::Matrix6x J_LF (6, model.nv);
+  J_LF.setZero();
```
jcarpent: Thanks for it!

unittest/aba-derivatives.cpp
jcarpent: You can a second argument to isApprox if there are some sensitivities with respect to rounding errors.

gabrielebndn: This value is not actually computed by computeABADerivatives. It was a mistake to put it here.

jcarpent: Yes, you're right.

jcarpent: A copy-paste error of mine ;)

gabrielebndn:
> > I would like to point out that valgrind complains a lot about python-parser-test too... However, since there is Python in the middle, it is much more difficult to analyze.
> 
> This is expected as EigenPy is doing allocation. What do you mean by complaining?

I get lots of "invalid read", "use of uninitialised value", and even a " Conditional jump or move depends on uninitialised value(s)". It may be that, as in the other cases, the error is simply in the unit test itself but I cannot be sure.

jcarpent:
> > > I would like to point out that valgrind complains a lot about python-parser-test too... However, since there is Python in the middle, it is much more difficult to analyze.
> > 
> > 
> > This is expected as EigenPy is doing allocation. What do you mean by complaining?
> 
> I get lots of "invalid read", "use of uninitialised value", and even a " Conditional jump or move depends on uninitialised value(s)". It may be that, as in the other cases, the error is simply in the unit test itself but I cannot be sure.

Weird indeed.

jcarpent:
@gabrielebndn Thanks a lot for the fixes.

Issue #875, opened 2019-09-16
Adding damping factor to impulse dynamics


cmastalli:
It would be great to have a damping factor in the impulse dynamics. I could quickly add it if @jcarpent is OK with this.

jcarpent:
In similar way of constrained Dynamics you mean?

cmastalli:
Yes, I do

jcarpent:
Sure. Please you can go ahead.

jcarpent:
Solved by #876

PR #876, opened 2019-09-17
Topic/constrained dynamics

jcarpent:
Related to #875.
I decided to handle this issue as it is more complex than expected at first guess.
@cmastalli Can you give it a try and tell if it convenient for you?

I would also need feedback from @gabrielebndn  if possible (for Python bindings notably).

### DIFF
--- a/bindings/python/algorithm/expose-contact-dynamics.cpp
+++ b/bindings/python/algorithm/expose-contact-dynamics.cpp
@@ -30,20 +30,20 @@ namespace pinocchio
                                                        const Eigen::VectorXd & q,
                                                        const Eigen::VectorXd & v_before,
                                                        const Eigen::MatrixXd & J,
-                                                       const double inv_damping,
-                                                       const double r_coeff)
+                                                       const double r_coeff,
+                                                       const double inv_damping)
     {
-      return impulseDynamics(model, data, q, v_before, J, inv_damping, r_coeff);
+      return impulseDynamics(model, data, q, v_before, J, r_coeff, inv_damping);
     }
     
     static const Eigen::VectorXd impulseDynamics_proxy_no_q(const Model & model,
                                                             Data & data,
                                                             const Eigen::VectorXd & v_before,
                                                             const Eigen::MatrixXd & J,
-                                                            const double inv_damping,
-                                                            const double r_coeff)
+                                                            const double r_coeff,
+                                                            const double inv_damping)
     {
-      return impulseDynamics(model, data, v_before, J, inv_damping, r_coeff);
+      return impulseDynamics(model, data, v_before, J, r_coeff, inv_damping);
     }
 
     static const Eigen::MatrixXd getKKTContactDynamicMatrixInverse_proxy(const Model & model,
@@ -80,8 +80,9 @@ namespace pinocchio
                        "Joint configuration q (size Model::nq)",
                        "Joint velocity before impact v_before (size Model::nv)",
                        "Contact Jacobian J (size nb_constraint * Model::nv)",
-                       "Damping factor when J is rank deficient.",
-                       "Coefficient of restitution r_coeff (0 = rigid impact; 1 = fully elastic impact)"),
+                       "Coefficient of restitution r_coeff (0 = rigid impact; 1 = fully elastic impact)",
+                       "Damping factor when J is rank deficient."
+                       ),
               "Solves the impact dynamics problem with contacts, put the result in Data::dq_after and return it. The contact impulses are stored in data.impulse_c"
               );
       
@@ -90,8 +91,8 @@ namespace pinocchio
               bp::args("Model","Data",
                        "Joint velocity before impact v_before (size Model::nv)",
                        "Contact Jacobian J (size nb_constraint * Model::nv)",
-                       "Damping factor when J is rank deficient.",
-                       "Coefficient of restitution r_coeff (0 = rigid impact; 1 = fully elastic impact)"),
+                       "Coefficient of restitution r_coeff (0 = rigid impact; 1 = fully elastic impact)",
+                       "Damping factor when J is rank deficient."),
               "Solves the impact dynamics problem with contacts, put the result in Data::dq_after and return it. The contact impulses are stored in data.impulse_c"
               );
       
--- a/bindings/python/algorithm/expose-contact-dynamics.cpp
+++ b/bindings/python/algorithm/expose-contact-dynamics.cpp
@@ -24,21 +24,6 @@ namespace pinocchio
     }
     
     BOOST_PYTHON_FUNCTION_OVERLOADS(forwardDynamics_overloads, forwardDynamics_proxy, 7, 9)
-
-//    // TODO: overloading impulseDynamics directly, as done for forwardDynamics, was apparently not working (it crashed for 5 arguments)
-//    // Therefore, it was necessary to resort to a proxy
-//    static const Eigen::VectorXd impulseDynamics_proxy(const Model & model,
-//                                                       Data & data,
-//                                                       const Eigen::VectorXd & q,
-//                                                       const Eigen::VectorXd & v_before,
-//                                                       const Eigen::MatrixXd & J,
-//                                                       const double r_coeff = 0.0,
-//                                                       const bool updateKinematics = true)
-//    {
-//      return impulseDynamics(model, data, q, v_before, J, r_coeff, updateKinematics);
-//    }
-//
-//    BOOST_PYTHON_FUNCTION_OVERLOADS(impulseDynamics_overloads, impulseDynamics_proxy, 5, 7)
     
     static const Eigen::VectorXd impulseDynamics_proxy(const Model & model,
                                                        Data & data,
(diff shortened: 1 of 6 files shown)


gabrielebndn:
I am checking. If I understand, you managed to do it, but not in a clean way as you would like? If I can find a fix, I will directly make a PR on your branch

cmastalli:
@jcarpent yes, it seems good for me. Maybe you want to add default values for damping factor and restitution coefficient (i.e. both equals zero).

Thanks for handling this

bindings/python/algorithm/expose-contact-dynamics.cpp
```diff
@@ -25,20 +25,41 @@ namespace pinocchio
     
     BOOST_PYTHON_FUNCTION_OVERLOADS(forwardDynamics_overloads, forwardDynamics_proxy, 7, 9)
 
-    // TODO: overloading impulseDynamics directly, as done for forwardDynamics, was apparently not working (it crashed for 5 arguments)
-    // Therefore, it was necessary to resort to a proxy
+//    // TODO: overloading impulseDynamics directly, as done for forwardDynamics, was apparently not working (it crashed for 5 arguments)
```
gabrielebndn: This is normal. You cannot use BOOST_PYTHON_FUNCTION_OVERLOADS here. BOOST_PYTHON_FUNCTION_OVERLOADS only works when you are overloading the SAME function with optional arguments. But here you have to completely different signatures.
Example:
```
foo(double x, double y=0., int z=0.)
```
--> You can use BOOST_PYTHON_FUNCTION_OVERLOADS to have `foo(x)`, `foo(x,y)` and `foo(x,y,z)` in Python
```
foo(double x, int y)
foo(int y, double x)
```
--> You cannot use BOOST_PYTHON_FUNCTION_OVERLOADS

src/algorithm/contact-dynamics.hpp
```diff
@@ -72,6 +72,70 @@ namespace pinocchio
...
+  ///
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType, typename TangentVectorType, typename ConstraintMatrixType>
+  inline const typename DataTpl<Scalar,Options,JointCollectionTpl>::TangentVectorType &
+  impulseDynamics(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                  DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                  const Eigen::MatrixBase<ConfigVectorType> & q,
+                  const Eigen::MatrixBase<TangentVectorType> & v_before,
+                  const Eigen::MatrixBase<ConstraintMatrixType> & J,
+                  const Scalar inv_damping,
+                  const Scalar r_coeff);
```
gabrielebndn: I think a better signature would be: `impulseDynamics(model, data, q, v_before, J, r_coeff=0., inv_damping=0.)` ; i.e., switch `r_coeff` and `inv_damping`, so that `inv_damping` could be easily left unspecified.

src/algorithm/contact-dynamics.hpp
```diff
@@ -72,6 +72,70 @@ namespace pinocchio
...
+  /// \return A reference to the generalized velocity after impact stored in data.dq_after. The Lagrange Multipliers linked to the contact impulsed are available throw data.impulse_c vector.
+  ///
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType, typename TangentVectorType, typename ConstraintMatrixType>
+  inline const typename DataTpl<Scalar,Options,JointCollectionTpl>::TangentVectorType &
+  impulseDynamics(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                  DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                  const Eigen::MatrixBase<TangentVectorType> & v_before,
+                  const Eigen::MatrixBase<ConstraintMatrixType> & J,
+                  const Scalar inv_damping,
+                  const Scalar r_coeff);
```
gabrielebndn: Same here: I think `impulseDynamics(model, data, v_before, J, r_coeff=0., inv_damping=0.) would be better

bindings/python/algorithm/expose-contact-dynamics.cpp
gabrielebndn: On the other hand, if you have
```
foo(double x, double y=0., int z=0.)
foo(int w, double a=0., int b=0.)
```
I am not sure whether you can use BOOST_PYTHON_OVERLOADS separately on the first one and on the second one. My intuition is yes

src/algorithm/contact-dynamics.hpp
gabrielebndn: Also, this way calls like `impulseDynamics(model, data, q, v_before, J)` or `impulseDynamics(model, data, q, v_before, J, r_coeff)` would still be valid and return exactly the same result

bindings/python/pinocchio/deprecated.py
```diff
@@ -195,3 +195,42 @@ def log6FromSE3(transform):
...
+@deprecated("This function has been renamed getJointJacobian and will be removed in future releases of Pinocchio. Please change for new getJointJacobian function.")
+def getJacobian(model,data,jointId,local):
+  if local:
+    return pin.getJointJacobian(model,data,jointId,pin.ReferenceFrame.LOCAL)
+  else:
+    return pin.getJointJacobian(model,data,jointId,pin.ReferenceFrame.WORLD)
+
+# This function is only deprecated when using a specific signature. Therefore, it needs special care
+# Marked as deprecated on 16 Sept 2019
+def impactDynamics(model, data, q = None, *args):
```
gabrielebndn: Uhm... I think there are some problems... This way, Python calls such as `impactDynamics(model,data,v_before,J[,...])` will not work...

src/algorithm/contact-dynamics.hpp
jcarpent: Yes, but 
`impulseDynamics(model, data, q, v_before, J, r_coeff=0., inv_damping=0.)`
has conflitcs with
`impulseDynamics(model, data, q, v_before, J, r_coeff=0., updateKinematics=true)`
that's why I resort to this solution, unfortunately.

bindings/python/algorithm/expose-contact-dynamics.cpp
jcarpent: In fact, these are old commented lines. I need to remove them.

jcarpent:
@gabrielebndn You can push on this branch if you need to provide additional fixes or unit test if you have time. In my case, I won't be able to do the unit tests before tomorrow.

@cmastalli You can provide the unit tests if you are in a rush.

cmastalli:
@jcarpent I'm not sure if I can work on this today. I could definitively work on it tomorrow.

src/algorithm/contact-dynamics.hpp
gabrielebndn: You can make it so that there are no conflicts
```
impulseDynamics(model, data, q, v_before, J, r_coeff=0., inv_damping=0.)
PINOCCHIO_DEPRECATED impulseDynamics(model, data, q, v_before, J, r_coeff, updateKinematics)
```
Without default arguments for the deprecated version.
In this way, if the user calls any of
```
impulseDynamics(model, data, q, v_before, J)
impulseDynamics(model, data, q, v_before, J, r_coeff)
```
The new version will be called, achieving the same effects as the old one. While the call
```
impulseDynamics(model, data, q, v_before, J, r_coeff, inv_damping)
```
will unambiguously refer to the new version and
```
impulseDynamics(model, data, q, v_before, J, r_coeff,false)
impulseDynamics(model, data, q, v_before, J, r_coeff,true)
```
will unambiguously refer to the old version

gabrielebndn:
> @gabrielebndn You can push on this branch if you need to provide additional fixes or unit test if you have time. In my case, I won't be able to do the unit tests before tomorrow.
> 
> @cmastalli You can provide the unit tests if you are in a rush.

Not sure I have time, but if I do, I will try to implement my proposition of switching `r_coeff` and `inv_damping` (see comment above on how to do it)

src/algorithm/contact-dynamics.hpp
jcarpent: I can handle it today. It will be fast to do during two meetings ;)
Thanks for this nice proposition.

gabrielebndn: Ok, I leave it to you, then. If you want further feedback and/or work on the unit test, try to let me know before 4PM, because I will not be able to work after 5PM and after that I will be away for two weeks :)

gabrielebndn: The Python unit test for contact dynamics already exists, so if you want to work on it you can just modify it: [unittest/python/bindings_dynamics.py](https://github.com/stack-of-tasks/pinocchio/blob/devel/unittest/python/bindings_dynamics.py)

bindings/python/pinocchio/deprecated.py
jcarpent: Seems working well.

src/algorithm/contact-dynamics.hpp
jcarpent: Done

src/algorithm/contact-dynamics.hpp
jcarpent: Sounds good.

jcarpent:
Thanks a lot @gabrielebndn for the very helpful feedback.

PR #877, opened 2019-09-17
python: fix missing pin.

paLeziart:
Fix a missing "pin." when calling computeJointJacobians(model,data,q)


jcarpent:
@paLeziart If you end up with this issue, it also means that the tutorials that you're following are not up to date.
It would be nice to do it. Can you handle it?

paLeziart:
@jcarpent It would be tedious for me to do that because I'm just a beginner using Pinocchio and I don't even have the corrections of the different exercises. 
I managed to do the two first exercises, the third one cannot be done because the source of the mobile robot is gone. For the fourth one I have a hard time doing question 4 because I don't see how to minimise (\ddot q - \ddot q_0)^T M (\ddot q - \ddot q_0) using solve_qp. For now I use it to minimize x^T M x and then add \ddot q_0 to the result but it's not working very well. Maybe there are updated tutorials somewhere that could be used to replace these ones.

PR #887, opened 2019-09-23
Make silent the warning from EigenPy >= 1.6.0

jcarpent:
Assuming that EigenPy has not been loaded before.


bindings/python/pinocchio/__init__.py
```diff
@@ -3,6 +3,12 @@
 #
 
 import numpy as np
+
+import sys
+if 'eigenpy' not in sys.modules:
+    import eigenpy
+    eigenpy.switchToNumpyMatrix() # forced default conversion towards. Silent the warning since EigenPy >= 1.6.0
```
jmirabel: Is Pinocchio compatible with both matrices and arrays ?
If yes, I think you shouldn't make the warning silent in a module. The warning should reach the users otherwise they will not be warned about the EigenPy future changes.

jcarpent: Not the tutorials and RobotWrapper if I remember well.

jcarpent:
@jmirabel The idea of fixing the warning here is for beginners who might not be confident with such messages. Is it fine for you if I proceed to the merge?

jmirabel:
This will apply to every one, not only to beginners. The warning **must** reach the users, including the beginners, otherwise the procedure is useless.

For beginners, the best you can do is to include the line that disables the warning to all tutorial and example scripts.

We can also change the warning message in eigenpy if it is not clear enough.

jcarpent:
I will merge this PR. Finally, I did not force the removal of the warning in the pinocchio library directly.

jmirabel:
Thank you @jcarpent !

PR #888, opened 2019-09-24
Fix usage of row major matrices + fix doc.

jmirabel:


unittest/explog.cpp
```diff
@@ -335,7 +335,7 @@ BOOST_AUTO_TEST_CASE(Jexplog6)
   Motion v(Motion::Random());
   
   SE3 M (exp6(v));
-  SE3::Matrix6 Jexp, Jlog;
+  Eigen::Matrix<double, 6, 6, Eigen::RowMajor> Jexp, Jlog;
```
jcarpent: I think you should try both, Row and Col Major

jmirabel: A column-major matrix is used in the test above.

jcarpent: Ok. Sounds good. By the way, you have the macro in Pinocchio to get the equivalent type with Eigen::RowMajor option. It is: PINOCCHIO_EIGEN_PLAIN_ROW_MAJOR_TYPE.

jmirabel: Ah, it only contains Jlog6, and not Jexp6.

Issue #849, opened 2019-08-06
Build error when porting from v2.1.3 to v2.1.6


fabioexavier:
When upgrading from v2.1.3 to v2.1.6, this existing block of code fails to build:

```c++
void switchJoints(pinocchio::Model *modelInOut, uint32_t const& firstJointId, uint32_t const& secondJointId)
{
....
     pinocchio::JointModel jointTemp = modelInOut->joints[firstJointId];  // Utilities.cc line 326
....
}
``` 

```
/opt/install/GCC/include/pinocchio/multibody/joint/joint-generic.hpp:129:10: required from 'struct pinocchio::JointModelTpl<double>'
Utilities.cc:326:39: required from here
/opt/install/GCC/include/boost/variant/detail/make_variant_list.hpp:40:46: error: wrong number of template arguments (21, should be at least 0)
typedef typename mpl::list< T... >::type type;
^~~~
In file included from /opt/install/GCC/include/boost/mpl/aux_/include_preprocessed.hpp:37:0,
from /opt/install/GCC/include/boost/mpl/list.hpp:46,
from /opt/install/GCC/include/boost/math/policies/policy.hpp:9,
from /opt/install/GCC/include/boost/math/special_functions/math_fwd.hpp:29,
from /opt/install/GCC/include/boost/math/special_functions/sign.hpp:17,
from /opt/install/GCC/include/boost/lexical_cast/detail/inf_nan.hpp:34,
from /opt/install/GCC/include/boost/lexical_cast/detail/converter_lexical_streams.hpp:63,
from /opt/install/GCC/include/boost/lexical_cast/detail/converter_lexical.hpp:54,
from /opt/install/GCC/include/boost/lexical_cast/try_lexical_convert.hpp:42,
from /opt/install/GCC/include/boost/lexical_cast.hpp:32,
from /opt/install/GCC/include/urdf_model/pose.h:45,
from /opt/install/GCC/include/urdf_model/joint.h:44,
from /opt/install/GCC/include/urdf_model/link.h:46,
from /opt/install/GCC/include/urdf_model/model.h:43,
from /opt/install/GCC/include/urdf_parser/urdf_parser.h:44,
from Utilities.cc:10:
[...]

PR #889, opened 2019-09-24
Fix setting of BOOST_MPL_LIMIT_LIST_SIZE

jcarpent:
Related to #849


jcarpent:
@jmirabel It seems that checking C++11 is not sufficient. I don't have time to investigate further, but I think the current PR is sufficient to fix your issue. 
I will wait for your feedback before merging.

jmirabel:
Removing the error if the compilation flag is set should be sufficient indeed. Thank you.

PR #890, opened 2019-09-25
[Doc] Add a cheat sheet

jmirabel:
I propose to add a cheat sheet of notations used in Pinocchio. This is achieved by adding a doxygen command cheatsheet. Each function having a *remarkable identity* calls this command. This produces two things:
- in the menu on the left, there is a new page called *Cheat sheet*, containing something like:
image

- in function that have a *Remarkable identity*:
image

At the moment, I do not know how to somehow sort the elements of the cheat sheet.


jmirabel:
A note: recent versions of doxygen include the possibility to define latex commands (for latex, html, mathjax, ... outputs): https://github.com/doxygen/doxygen/pull/7230

PR #902, opened 2019-10-03
Fix Python impulseDynamics and modify dignature of forwardDynamics

gabrielebndn:
The previous implementation of `impulseDynamics` in Python did not allow to use signatures `impulseDynamics(model,data,v_before,J[,r_coeff[,inv_damping]])` directly and it required to set `q` to `None` in order to use it, as in `impulseDynamics(model,data,None,v_before,J,r_coeff,inv_damping)`.

Also, the argument overload was handled in `deprecated.py`, which is not the right place for this (it would have stopped working if the deprecated version had simply been removed).

I fixed it. Now all following 6 signatures are legal and recommended:
```
impulseDynamics(model,data,q,v_before,J)
impulseDynamics(model,data,q,v_before,J,r_coeff)
impulseDynamics(model,data,q,v_before,J,r_coeff,inv_damping)
impulseDynamics(model,data,v_before,J)
impulseDynamics(model,data,v_before,J,r_coeff)
impulseDynamics(model,data,v_before,J,r_coeff,inv_damping)
```
while the following signature is deprecated:
```
impulseDynamics(model,data,q,v_before,J,r_coeff,updateKinematics)
```
and I am checking in the unit test that it raises a `pinocchio.DeprecatedWarning`.

I recommend that a similar scheme be implemented for `forwardDynamics`

### DIFF
--- a/bindings/python/algorithm/expose-contact-dynamics.cpp
+++ b/bindings/python/algorithm/expose-contact-dynamics.cpp
@@ -30,21 +30,25 @@ namespace pinocchio
                                                        const Eigen::VectorXd & q,
                                                        const Eigen::VectorXd & v_before,
                                                        const Eigen::MatrixXd & J,
-                                                       const double r_coeff,
-                                                       const double inv_damping)
+                                                       const double r_coeff = 0.,
+                                                       const double inv_damping = 0.)
     {
       return impulseDynamics(model, data, q, v_before, J, r_coeff, inv_damping);
     }
+
+    BOOST_PYTHON_FUNCTION_OVERLOADS(impulseDynamics_overloads, impulseDynamics_proxy, 5, 7)
     
     static const Eigen::VectorXd impulseDynamics_proxy_no_q(const Model & model,
                                                             Data & data,
                                                             const Eigen::VectorXd & v_before,
                                                             const Eigen::MatrixXd & J,
-                                                            const double r_coeff,
-                                                            const double inv_damping)
+                                                            const double r_coeff = 0.,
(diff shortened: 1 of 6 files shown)


jcarpent:
Looks fine for me.

jcarpent:
You can proceed as such for forwardDynamics. Can you handle it?

gabrielebndn:
Note I did not maintain compatibility with `impulseDynamics(model,data,None,v_before,J[,r_coeff[,inv_damping]])` (i.e. `q` equals to `None`) because I figured nobody uses it as it is a very recent change (in the last two weeks, it is only on v2.1.8, which is not even on the apt repoyet ) and I figured nobody uses it.

gabrielebndn:
> You can proceed as such for forwardDynamics. Can you handle it?

Sure. It will take me no time.

gabrielebndn:
> You can proceed as such for forwardDynamics. Can you handle it?

Done

gabrielebndn:
For info: now the following signatures of `forwardDynamics` are available, both in Python and C++:
```
forwardDynamics(model,data,tau,J,gamma[,inv_damping])
forwardDynamics(model,data,q,v,tau,J,gamma[,inv_damping])
```
i.e. the one which does not update the kinematics does not use neither `q` nor `v`, while the following signature
```
forwardDynamics(model,data,q,v,tau,J,gamma,inv_damping,updateKinematics)
```
is deprecated

jcarpent:
Very nice work @gabrielebndn. Always clear and clean.
Thanks a lot!

Issue #867, opened 2019-08-29
Kinematics Hessian -- Jacobian derivatives


Leph:
First, I would like to thank you all a lot for your great open source work and implementation.

The analytical derivatives of the dynamics are a very valuable feature of the library.
Is there a way using the currently implemented Pinnochio primitives to efficiently compute the kinematics Hessian -- the partial derivatives of a Jacobian of a frame ; which is a tensor of 
(6 x nv x nv) ?

Thanks very much. 

Best,

jcarpent:
@Leph Thanks @Leph for your interests in our project.

There is no yet a public interface for it.
I plan to add in the future. If it is important for your applications, I can implement it in the very next days.

Otherwise, you can still use the automatic differentiation layer with CppAD to do that from the classic Jacobians. Are you familiar with such things?

Leph:
@jcarpent, thanks for your answer.

I have already prototyped my code with classic numerical finite differences.
Thanks to the already implemented partial derivatives, the Hessian is the last piece of the puzzle
preventing me to reach the 1ms computation time.

This is not highly critical, but I will be happy to experiment it as soon as it will be included.

Thank you again.

jcarpent:
Ok. I may implement them soon.

nmansard:
Thanks @traversaro for referring to this discussion in the idyntree project. 
@Leph: are you sure you need H directly, and not for computing a quantity that we be easier to compute directly. I don't know the context of this question, but for an example: if you want H to finally compute H.v_q (the product of the tensor with the robot velocity), then we may already have this quantity computed. And it would likely be cheaper to compute it that way.

Leph:
@nmansard, I am differentiating the equation of motion under contacts and using it within a QP formulation. You are right, I am actually looking for a kind of Hessian-vector product. It would have been indeed much simplier, but unfortunately, the differentation ends up with a "H.lambda" tensor product against the *first* tensor dimention instead of the classical last one (resulting with an (nv x nv) matrix instead of a (6 x nv)).

This is equivalent to compute nv times the regular matrix-product: (dJ/dq_i)^T.lambda

nmansard:
Do you have some maths to describe what you are doing, maybe in a scientific paper?
How is that different from equation (12) of this paper https://hal.archives-ouvertes.fr/hal-01851596v2/document

jcarpent:
[...]

PR #908, opened 2019-10-15
Implement Hessian of the kinematics

jcarpent:
This is related to discussions in #867.


jcarpent:
@markusgft I've implemented a first draft of the Hessian of the kinematics. It is only in C++ and I need to update EigenPy to handle Eigen::Tensor module. 
Do you need it rapidly?

jcarpent:
@markusgft By the way, would you mind if this feature is only accessible in C++11 and more?

gabrielebndn:
It seems compilation fails on 16.04 because of a `constexpr` in Eigen.
I am wondering, is it wise to use `unsupported` features of Eigen so deep in Pinocchio? Is it wise to #include it in Data? Wouldn't it be better to isolate these things in a separate part of Pinocchio?

jcarpent:
@gabrielebndn This is because Eigen::Tensor relies on C++11. I have two solutions:
- make the code only existing in C++11 with an extra-macro to allow the inclusion of Eigen::Tensor
- make the code working for both C++98 and C++11, and then just copying the data inside an Eigen::VectorXd for alignment issues, and then providing some helper functions to treat them as Eigen::Tensor in C++11 through the Map.

jcarpent:
The only thing is that I'm much more in favor of solution 1., because it will be much more compliant with the future of Pinocchio and most platforms, which will use C++11 compilation flag by default.

jcarpent:
@gabrielebndn By the way, the naming unsupported used in Eigen is just because this is not the main feature of Eigen. But for sure, the Eigen::Tensor will remain supported for a while because it is at the heart of TensorFlow and PyTorch.

gabrielebndn:
> @gabrielebndn This is because Eigen::Tensor relies on C++11. I have two solutions:
> 
> * make the code only existing in C++11 with an extra-macro to allow the inclusion of Eigen::Tensor
> * make the code working for both C++98 and C++11, and then just copying the data inside an Eigen::VectorXd for alignment issues, and then providing some helper functions to treat them as Eigen::Tensor in C++11 through the Map.

The first solution is much better, I agree.

gabrielebndn:
> @gabrielebndn By the way, the naming unsupported used in Eigen is just because this is not the main feature of Eigen. But for sure, the Eigen::Tensor will remain supported for a while because it is at the heart of TensorFlow and PyTorch.

This was my main worry. Then I guess if TensorFlow is doing it, Pinocchio can too

wxmerkt:
Thank you very much for providing this excellent feature! Since we just had a related conversation at Crocoddyl, my question is:

If this feature is only active when compiled with C++11, it won't be part of the robotpkg binaries, even for systems where C++11/14 is default? From the discussion at Crocoddyl it appears that "Pinocchio compiled with C++98 as distributed is incompatible with C++11 in a dependent project".

cmastalli:
I am favor of option 1. After 18.04, we should forget about c++98, and together along c++11.

markusgft:
> 
> 
> @markusgft I've implemented a first draft of the Hessian of the kinematics. It is only in C++ and I need to update EigenPy to handle Eigen::Tensor module.
> Do you need it rapidly?

Thank you @jcarpent for the update and implementing the feature! I think it is going to be very useful to a lot of people. As of now, I do not need the feature too rapidly, as I can also continue using the above mentioned workaround in the meantime. So please take your time!

Regarding the ensuing discussion: I do not know the deployment details and strategy of pinocchio, but I share the opinion that fixing C++11 as baseline requirement may be reasonnable.

jcarpent:
@markusgft To update you, I finally implemented a minimal Tensor class (using snippets from Eigen). This should work on all the platforms now.

jcarpent:
I will merge this PR as it is mature enough. I will provide other features soon.

Issue #909, opened 2019-10-16
Safety checks in Python


gabrielebndn:
Currently, safety checks like checking the length of the array or verifying that the `Data` object is consistent with the `Model` are only done through asserts, i.e. only in debug mode.

This leaves the users completely on their own when using the Release mode, but it is consistent with what Eigen does, and I think it should stay this way: if users want, they can still compile in Debug mode.

However, users are left completely powerless in Python, if they are using the released versions and they are not compiling it themselves.
One solution would be to always enable the checks in Python, but I believe the best way is to add a global switch allowing to enable or disable the checks at will, something like `pinocchio.setAssertionsEnabled(bool)`. This might also be useful in the future if we want to implement a shared library.

If we decide to go this way, then all safety `assert` statements should be replaced by a macro, defined as something like this:
```
#ifdef PINOCCHIO_RUNTIME_ASSERT
#  define PINOCCHIO_ASSERT(cond,msg) \
if(pinocchio::isAssertionsEnabled && !(cond)) { \
  throw std::runtime_error(msg); \
}
#else
#  define PINOCCHIO_ASSERT(cond,msg) assert((cond) && msg)
#endif 
```

with `pinocchio::isAssertionsEnabled` initially set to false, plus some interface to properly set it.

The main motivation for this is Python, but it would be available in C++ too. Python would be compiled by default with `PINOCCHIO_RUNTIME_ASSERT` defined.

What do you think?

gabrielebndn:
Oh, I see Python actually does perform the checks! My bad, I did not remember that, I should have verified before. Still, what do you think of setting a switch?

gabrielebndn:
Also, the checks are performed, but since they are made through asserts, this actually crashes the whole program and it quits the Python interpreter. I think this is not a desirable behavior in Python. Exception would be better, in my opinion.

gabrielebndn:
> Oh, I see Python actually does perform the checks! My bad, I did not remember that, I should have verified before. Still, what do you think of setting a switch?

Actually, this is what happens: with the apt-downloaded version (v2.1.8), it seems the asserts are enabled, but with the manually compiled version (v2.1.9, but I do not think it changed) they are not (`NDEBUG` is defined).
I am I getting it right? What are the Python compilation options for the apt version? @nim65s, do you know anything about it? If I am getting it right, I find it confusing. The manually-compiled Release versions should be identical to the one actually release in apt, I think

wxmerkt:
[...]

PR #910, opened 2019-10-16
Use if() and throw exception on everything apart from model.check(data)

wxmerkt:
PR to show the diff for the example given in https://github.com/stack-of-tasks/pinocchio/issues/909#issuecomment-542824324

This PR is _not_ meant to be merged. It is solely to show the proposed difference (to be cleaned up):

1. Replace all `assert` with `if (!(cond)) { throw std::runtime_exception("Wrong size");`
2. Keeps the `assert` for `model.check(data)` as this one is expensive.

@jcarpent Can you benchmark/review this on your machine please? The numbers in my comment in the issue are from Ubuntu 18.04 on an i9-9900K.


src/macros.hpp
```diff
@@ -19,6 +19,13 @@
 
 #define PINOCCHIO_STRING_LITERAL(string) #string
 
+#include <exception>
+#include <cassert>
+
+#undef assert
```
jcarpent: You should remove this line

wxmerkt:
> @wxmerkt Thanks a lot for this PR.
> But I would not change the behavior in Spatial classes. This is for internal debug only.
> I think we should only throw in algorithms, which is the main interface with the end-user.

I agree. This is not meant to be merged as-is. This is test to see whether there is a performance penalty, as had been hypothesised in #909. It shows that even if every single `assert` is replaced with a `if` check (also internal!), **there is no penalty**. As such, we can be more selective of course and revert `spatial` to `assert`, but there isn't a performance penalty with this PR. However, I'd like to get other people's sanity checks on other machines whether this translates or is reproducible as negligible.
[On the contrary, counter-intuitively, there is a tiny speed-up (~2-3%) - I'd disregard that though as fluctuation]

wxmerkt:
> Another important thing is that you are not sending back to the user the message (if it exists).

Yes, and that should be changed, of course. The messages are really helpful. If the agreement is that we are going ahead with the suggestion, I am happy to amend and make sure all messages get passed through.

wxmerkt:
@jcarpent Thank you for the feedback. I have updated this PR:

(a) Removed the new macro from `spatial`
(b) Added a second macro to output the message

These are the new numbers:
```
 ./benchmark/timings
nq = 36
RNEA = 		12.5178 us
NLE = 		12.7267 us
NLE via RNEA = 		12.484 us
CRBA = 		24.2447 us
CRBA minimal = 		16.86 us
computeAllTerms = 		50.4631 us
Branch Induced Sparsity Cholesky = 	9.28067 us
Dense Eigen Cholesky = 	9.12935 us
Jacobian = 	4.68044 us
Jacobian Time Variation = 	10.5916 us
COM+Jcom = 	6.64457 us
COM+vCOM+aCOM = 	12.086 us
Zero Order Kinematics = 	4.2412 us
First Order Kinematics = 	6.72347 us
Second Order Kinematics = 	9.73503 us
CCRBA = 	9.90023 us
ABA = 	31.0626 us
Empty Forward Pass = 	0.24295 us
Coriolis Matrix = 	56.6241 us
Minv = 	47.7608 us
--
```

which compare with https://github.com/stack-of-tasks/pinocchio/issues/909#issuecomment-542824324 - again no penalty.

src/macros.hpp
```diff
@@ -19,6 +19,12 @@
 
 #define PINOCCHIO_STRING_LITERAL(string) #string
 
+#include <exception>
+
+/// \brief Macro to check an assert-like condition and throw a runtime error exception (with a message) if violated.
+#define PINOCCHIO_ASSERT_THROW_AT_RUNTIME(cond) if (!(cond)) { throw std::runtime_error("Wrong argument size"); }
```
jcarpent: You should change the message here, putting something more generic.

wxmerkt: Yes, sounds good. Do you have a suggestion?

jcarpent: I will take it in charge in a future PR if you don't mind, in order to provide helpful feedback to the user.

jcarpent:
> @jcarpent: Of course very happy to restrict the area where this change takes place. Would it be okay if I just revert the changes within `src/multibody/` or are `algorithm` and `parser` the only user-facing components? Part of this PR was to show that there is no performance penalty, even if applied across the full codebase.

I think we can maintain your change for: `algorithm`, `parser` and `serialization`. Thanks a lot for the change.

gabrielebndn:
I haven't checked for timings, but I've tested it. It compiles and it behaves exactly as it should in Python: an exception is thrown and it is catchable as a `RutimeError`.

Nice job!

gabrielebndn:
Just a few suggestions to improve on this PR, feel free to choose:
- Remove the one-argument macro
- Give a less verbose name to the macro, such as simply `PINOCCHIO_ASSERT_THROW`
- Use a more specific exception type. I suggest `std::invalid_argument`. I've just tested, and it shows up in Python as a `ValueError`. In this case, the macro name should probably reflect the exception type, such as `PINOCCHIO_ASSERT_ARGUMENT`

wxmerkt:
Thank you very much for the great feedback @jcarpent and @gabrielebndn. I am currently traveling with limited access to an Ubuntu machine - so it may take me a bit to address these suggestions.

I just reverted changes to multibody and codegen. I left the changes to `math` for now as we had issues with the quaternion normalization check before. Or is it safe to have that be a regular assert and ignore if not normalised? (We have these issues when applying damping to JMinvJt for impulse dynamics in Crocoddyl - it throws on that line)

jcarpent:
@wxmerkt You should also revert the folder `math`

jcarpent:
Or something like `PINOCCHIO_CHECK_ARGUMENT`.

wxmerkt:
Thank you for making the changes to fix the PR. Is the only outstanding points to change the name of the macro?

jcarpent:
@wxmerkt I will take it in charge with variadic macro.

gabrielebndn:
I really think the exception type should not be `std::runtime_error`, but `std::invalid_argument`. As a rule of thumb, it is always better if the exception thrown is as specific as possible.

jcarpent:
> I really think the exception type should not be `std::runtime_error`, but `std::invalid_argument`. As a rule of thumb, it is always better if the exception thrown is as specific as possible.

Don't worry @gabrielebndn, I'm taking this remark in future commits. Hopefully today.

jcarpent:
@wxmerkt @gabrielebndn You may provide some helpful comments on the new commits that I've just pushed.

jcarpent:
All the lights are green from my side.

jcarpent:
The mechanism does work. It is just because of the constructor and we cannot avoid it.
For the variadic, I will keep it to simplify the change and through anyway even if we do not have currently all the messages written.
I will take your recommandation into account.

src/algorithm/center-of-mass.hxx
```diff
@@ -105,7 +105,7 @@ namespace pinocchio
                            const bool computeSubtreeComs)
   {
     assert(model.check(data) && "data is not consistent with model.");
-    assert(LEVEL >= 0);
+    PINOCCHIO_CHECK_INPUT_ARGUMENT(LEVEL >= 0);
```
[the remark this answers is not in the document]
jcarpent: Done in the new commits.

wxmerkt:
Thank you very much for improving and merging this @jcarpent.

PR #920, opened 2019-10-31
Gitlab: update ci

jcarpent:


jcarpent:
@nim65s Do you know how to exclude CMakeFiles from the gcov? It appears that the solution that I've implemented is not working and there are still some coverage done on https://gepettoweb.laas.fr/doc/stack-of-tasks/pinocchio/master/coverage/

nim65s:
@jcarpent Your link shows `2019-10-08 13:04:46` ; so I think this is once again an issue on my side, I'll fix that soon.

I checked this PR, and the exclusion of CMakeFiles works well.

jcarpent:
Thanks @nim65s for your quick feedback. I will merge this PR then.

PR #921, opened 2019-10-31
Fix propagation of NAN values for exp/log computations

jcarpent:


jcarpent:
@ylabbe This PR should fix your issue

src/spatial/explog.hpp
```diff
@@ -114,7 +105,16 @@ namespace pinocchio
...
+      res(2) = (R (1, 0) > R (0, 1) ? Scalar(1) : Scalar(-1)) * (tmp[2] > Scalar(0) ? sqrt(tmp[2]) : Scalar(0));    }
+    else
+    {
+      const Scalar t = ((theta > TaylorSeriesExpansion<Scalar>::template precision<3>())
+                        ? theta / sin(theta)
+                        : Scalar(1)) / Scalar(2);
+      res(0) = t * (R (2, 1) - R (1, 2));
+      res(1) = t * (R (0, 2) - R (2, 0));
+      res(2) = t * (R (1, 0) - R (0, 1));
+
```
jmirabel: The changes in this file are cosmetic changes only, right ?

jcarpent: They just change the order of the if statement. It then allows propagating NAN correctly in the computational flow, without setting 0 values as done previously throw the if/else statements.

jmirabel: I see. Thanks for the explanation.

PR #925, opened 2019-11-01
Change ordering of macros

jcarpent:
This allows to avoid some borring warning from GCC


jcarpent:
@wxmerkt This PR should remove some anoying warnings of Variadic macros.

PR #927, opened 2019-11-04
Add buildReducedModel function

jcarpent:
Related to #905.
@markusgft and @alexswerner This feature is for you


markusgft:
Thank you @jcarpent for your effort!
One quick question: would it be possible to add an example or a unit test which covers exactly the reverse behaviour, i.e. keeping some given joints "unlocked", and locking all others?

jcarpent:
@markusgft Yes, for sure. I will do that today.

jcarpent:
@markusgft I have provided an example in examples/build-reduced-model.cpp. Hope it will be useful.

PR #933, opened 2019-11-06
cmake: fix pkg-config version constraints

sbarthelemy:
to make them compatible wih perl's implementation [1] of pkg-config.

See [2] for more details.

[1] https://metacpan.org/pod/PkgConfig
[2] https://github.com/jrl-umi3218/jrl-cmakemodules/pull/265


wxmerkt:
This breaks on Ubuntu 18.04 for me:

```bash
$ pkg-config --cflags pinocchio 
Package eigen3>=3.0.5 was not found in the pkg-config search path.
Perhaps you should add the directory containing `eigen3>=3.0.5.pc'
to the PKG_CONFIG_PATH environment variable
Package 'eigen3>=3.0.5', required by 'pinocchio', not found
```

sbarthelemy:
Hello @wxmerkt , sorry for the breakage and thank you for the detailed report.

I understood what is going on:
jrl-cmakemodules uses the requires string in two ways:

* to pass to PKG_CHEK_MODULES & co, which expects no space around the operator
* to generate the .pc, where a space around the operator are requires.

I'm working on a fix.

PR #936, opened 2019-11-07
doc: fix list of supported C++ versions

sbarthelemy:
C++98 and 03 are currently supported


jcarpent:
@sbarthelemy Thanks a lot!

PR #938, opened 2019-11-07
math/tensor: add missing aligned operator new

jcarpent:


jcarpent:
@wxmerkt This should fix your bug on Crocoddyl

wxmerkt:
Thank you for the quick action and fix, I shall try it!

Issue #939, opened 2019-11-08
updating only the momentum quantities.


proyan:
We need a function in crocoddyl, that updates the centroidal momentum hg, without additional computations Currently the version that we have (computeCentroidalDynamics) computes the kinematics as well, and thus is not as efficient.

I'm willing to implement it here, but I wish to affirm that there is an appetite for such a function in Pinocchio. What do you think?

jcarpent:
This is fine for me. I can also handle it before doing a new release. What do you think?

cmastalli:
There is two ways of doing, I'm not sure what it's the Pinocchio way of doing:
 1. `updateCentroidalDynamics(model, data)`
 2. `computeCentroidalDynamics(model, data) #we don't pass q and qd`

Which option should we pursue @jcarpent ?

jcarpent:
I think the best naming is `updateCentroidalDynamics` as done in other places in Pinocchio.

proyan:
If you can handle it, I'm not going to say no :smiley:

gabrielebndn:
My bit: I think `computeCentroidalDynamics(model,data)`. Most methods (think of `forwardDynamics`) can work with or without the configuration.
This way we could establish a nice pattern in all methods: if the configuration is passed, then it is used to compute the kinematics, otherwise the kinematics is not computed

jcarpent:
@gabrielebndn Not for `forwardDynamics`, because you don't know if you want the zero, first or second kinematics. The same applied for `computeCentroidalDynamics`. One has to understand the update by `update from the current value in Data`.

cmastalli:
> If you can handle it, I'm not going to say no smiley

Unfortunately, I don't have time. Thanks for making this effort @proyan :smiley:

jcarpent:
@cmastalli Rohan was talking to me ;)

proyan:
@jcarpent Worth the try :D
No problem, I'll make the PR within next week. I'll prefer `updateCentroidalDynamics`.
By the way, I don't think that I can do it with simply `updateCentroidalDynamics(model, data)`. Correct me if I am wrong, we'll need `v` in order to update the centroidal dynamics. And just inputting `v`, without `q` seems bad.

gabrielebndn:
For `forwardDynamics` and the like, we assume that the forward kinematics has been computed up to the first time derivative. So I do not think you need to take `v` into account. As long as the `data.v[i]` are filled with the correct values, I do not think you need anything else

[...]

PR #944, opened 2019-11-11
Rename centroidal functions and add new signatures

jcarpent:
Following discussions in #939.


cmastalli:
Thank you very much for this fast action!

PR #946, opened 2019-11-12
Remove deprecated functions and methods

jcarpent:
This PR removes some (very) old deprecated methods and functions. The idea is to provide a new release of Pinocchio 2.2.0 soon, by removing these old lines of code.

@gabrielebndn @nim65s @jmirabel 
Could you provide feedback on this PR?


jmirabel:
There will be a view glitches in HPP but nothing hard to solve.

Do you have a timeline for version 2.2.0 ?

jcarpent:
I do not have a clear timeline in mind, I mean we can wait a bit. Which functions are still in use inside HPP?

jcarpent:
@proyan @cmastalli @wxmerkt Maybe you can also provide some feedback for Crocoddyl?

jcarpent:
@aelkhour Any feedback from your side?

nim65s:
I think removing those deprecated functions and creating a release would be a good thing in general.

This might break things in dependent projects, so I will need some time to fix those before pushing it to robotpkg, but making the release should not be an issue.

Or maybe you want me to test everything in the dependent packages from Gepetto before making this release ? This is possible, but it will hold this release for an amount of time I can't foresee.

About the version numbers, if we wanted to follow semver, we should not remove something from the API without incrementing the Major number. And if you don't want to follow semver, it would be a good thing to document what those numbers mean to you in this project.

proyan:
In crocoddyl we've been actively purging old functions based on the latest pinocchio release, so I don't think crocoddyl would suffer. There are some unkempt packages that are not that well-maintained, https://github.com/loco-3d/multicontact-api for example, that could have some issues. But I don't see this as a problem. From my end, it is okay.

jcarpent:
@nim65s I think it should be nice to test these deletions and to adjust the code of Pinocchio accordingly if needed. But if you think it will take a month, then we will need to make the release before and provides some patches after.

Concerning the release versioning, the major version number means **change of philosophy**: in the context from 1 to 2, it was the full templatization of the projects. From 2 to 3, it will be something related to full computational expression (similar to Eigen).
For the minor version number, it is removing deprecated functions, add new important functionalities.
For patch version number, it is more continuous coding: add deprecated functions, more features, and fixes.

jmirabel:
> I do not have a clear timeline in mind, I mean we can wait a bit. Which functions are still in use inside HPP?

I have in mind a function which returns a threshold for finite difference in hpp-constraints. I usually try to solve all the deprecated warnings I encounter so I expect that there are not many.

jmirabel:
This one:https://github.com/humanoid-path-planner/hpp-constraints/blob/645f97f1154834021191f2f95557c7a782212387/src/differentiable-function.cc#L38

jcarpent:
@proyan Good point. Thanks a lot.

cmastalli:
Yes, in Crocoddyl, we keep align to latest Pinocchio release.
It should be alright for us.

gabrielebndn:
I heartily support removing all those very old functions and signatures. I'll check in a few minutes, please wait

bindings/python/pinocchio/deprecated.py
```diff
@@ -22,170 +22,6 @@
 StdVect_Force = deprecated("Please use StdVec_Force")(pin.StdVec_Force)
 StdVect_Motion = deprecated("Please use StdVec_Motion")(pin.StdVec_Motion)
```
gabrielebndn: You could remove this `StdVect` classes too

bindings/python/pinocchio/deprecated.py
```diff
@@ -239,6 +75,14 @@ def forwardDynamics(model, data, *args):
 + '\n    This function signature has been deprecated and will be removed in future releases of Pinocchio.'
 )
 
+@deprecated("This function has been renamed computeJointJacobian and will be removed in future releases of Pinocchio. Please change for new computeJointJacobian.")
+def jointJacobian(model, data, q, jointId):
+  return pin.computeJointJacobian(model,data,q,jointId)
+  
+@deprecated("This function has been renamed computeFrameJacobian and will be removed in future releases of Pinocchio. Please change for new computeFrameJacobian.")
+def frameJacobian(model, data, q, jointId):
```
gabrielebndn: It should be `frameId` instead of `jointId`, I think. Same in next line.

gabrielebndn:
Seems all right!

jcarpent:
@gabrielebndn Thanks a lot for the review. I included your feedback.

proyan:
I didn't spend much time in the review, but from a superficial glance it looks okay. I trust @gabrielebndn to have already perused the code properly :)

gabrielebndn:
Ah! I've just noticed something. You might double-check: make sure you delete not only the *declaration* of functions, but also their *definitions*: you did not do it for the sample models, for instance. If you do not remove the definitions, the functions will effectively still be there

jcarpent:
You're totally right. I will miss you soon ;)

gabrielebndn:
There is the definition `getFrameJacobian(model, data, frame_id, J)` too (without `ReferenceFrame`)

bindings/python/pinocchio/robot_wrapper.py
```diff
@@ -139,28 +127,12 @@ def frameClassicAcceleration(self, index):
...
-            return pin.jointJacobian(self.model, self.data, q, index)
-        else: # use deprecated signature (19 Feb 2019)
-            update_kinematics = True if len(args)==1 else args[1]
-            rf = args[0]
-            return pin.jointJacobian(self.model, self.data, q, index, rf_frame, update_kinematics)
+    def jointJacobian(self, q, index):
+        return pin.jointJacobian(self.model, self.data, q, index)
 
     def getJointJacobian(self, index, rf_frame=pin.ReferenceFrame.LOCAL):
         return pin.getFrameJacobian(self.model, self.data, index, rf_frame)
```
gabrielebndn: There is a bug here! It should call `getJointJacobian`! Apparently, it was there before! And, judging from `git blame`, it was *me* doing this mistake!

gabrielebndn:
In  [src/algorithm/frames.hxx](https://github.com/stack-of-tasks/pinocchio/blob/d7ca7d5d81e47bd1b79d9edeabdd08fec02bcf14/src/algorithm/frames.hxx) there is the definition of `framesForwardKinematics(model,data)`. This should have been removed long ago, because ti was replaced by `updateFramePlacements(model,data)`, but we forgot the definition. So, there *might* be code still using it, as it would not raise warnings. I do not know whether it should be removed right away of moved to [src/algorithm/frames.hpp](https://github.com/stack-of-tasks/pinocchio/blob/d7ca7d5d81e47bd1b79d9edeabdd08fec02bcf14/src/algorithm/frames.hpp) as deprecated. You choose.

As for the rest, I've double-checked, it seems ok.

jcarpent:
I have included all the comments. Thanks a lot to @proyan and @gabrielebndn for very helpful feedback. 
I will now merge this PR allowing people to test it on their current devel branches.
I won't make the release now but certainly new week, letting some time for testing and welcome feedback.

PR #948, opened 2019-11-13
Update readme content

jcarpent:
This PR adds the logo of Pinocchio + some plot of the performances.


jmirabel:
The performance figure is nice. What are the three elements which are benchmarked ? It may be worth to add a legend on the X axis.

README.md
```diff
@@ -32,17 +38,26 @@ If you want to learn more on **Pinocchio** internal behaviors and main features,
...
    - C++ 98/03/11/14/17/20 compliant.
 
+**Pinocchio** is extensible.  
 **Pinocchio** is multi-thread friendly.  
 **Pinocchio** is reliable and extensively tested (unit-tests, simulations and real robotics applications).  
 **Pinocchio** is supported on Mac OS X and many Linux distributions ([see build status here](http://robotpkg.openrobots.org/rbulk/robotpkg/math/pinocchio/index.html)).
 
+## Performances
+
+**Pinocchio** exploits at best the sparsity induced by the kinematic tree of robotics systems. Thanks to modern programming language paradigmes, **Pinocchio** is able to unroll most of the computations, allowing to achieve impressive performances for a large range of robots, as illustrated by the plot below.
```
gabrielebndn: paradigmes -> paradigms

jcarpent:
@gabrielebndn @jmirabel Thanks for your feedback. I took them into account.

gabrielebndn:
I don't get it. If I go on [your devel branch](https://github.com/jcarpent/pinocchio/tree/devel), the preview of the README still shows the old version of pinocchio-performances.png, without the algorithm info on the x-axis. Then I click on it and it sends me to the corrected version

gabrielebndn:
Some minimal details on the system on which these performances where obtained might be informative, such as "... as illustrated by the plot below, obtained on a standard Macbook Pro mounting Mac OS X High Sierra" (if that's your system)

jmirabel:
> Some minimal details on the system on which these performances where obtained might be informative, such as "... as illustrated by the plot below, obtained on a standard Macbook Pro mounting Mac OS X High Sierra" (if that's your system)

and may be the number of DoF of each robots ?

gabrielebndn:
> I don't get it. If I go on [your devel branch](https://github.com/jcarpent/pinocchio/tree/devel), the preview of the README still shows the old version of pinocchio-performances.png, without the algorithm info on the x-axis. Then I click on it and it sends me to the corrected version

Ok, now I see it correctly, according to @jmirabel it was because of my browser's cache, that's probably right

gabrielebndn:
> > Some minimal details on the system on which these performances where obtained might be informative, such as "... as illustrated by the plot below, obtained on a standard Macbook Pro mounting Mac OS X High Sierra" (if that's your system)
> 
> and may be the number of DoF of each robots ?

Anybody can go and look them up for themselves if they are really interested, I think, this is just for illustration purposes and there is no need to make the image heavier.
But the system on which the computations where carried out is important, I think

jcarpent:
I do not remember which OS and CPU were used, but I just remembered that @nim65s and @gabrielebndn you did together these timings.

cmastalli:
I think it's very cool the image with the benchmark. In addition to it, you should add a tiny note that describes the PC used for such benchmark.

Another detail is the Pinocchio logo, I meant it's not Pinocchio logo, it's Gepetto one.
It is definitively better if you create its own logo.

Finally, in the acknowledgement, you might add the SLMC group from Edinburgh (http://wcms.inf.ed.ac.uk/ipab/slmc).

jcarpent:
On my computer, the timings are much more lower. But I did have time yet to update the figure :(

gabrielebndn:
Oh, it is the figure from the paper? Then the answer is in there: "standard laptop equipped
with an Intel Core I7 CPU @2.4 GHz"

gabrielebndn:
@jcarpent, from the `git log` of the Pinocchio paper, I can confirm it was you including this image, so I guess it was generated on your laptop. Indeed, the description I posted above, taken from the paper, seems to be consistent with this. This also corresponds to what I think to recollect, @nim65s and I did some work on generating the benchmark values, but then you took over and produced the final image.
So, if you are convinced by my explanation, I would just copy-paste the above description to the README.

jcarpent:
@gabrielebndn Do agree with you. Thanks a lot for the push and the corrections ;)
@jmirabel Thanks also for the mistakes corrections.

README.md
```diff
@@ -121,7 +121,7 @@ You have a question or an issue? You may either directly open a [new issue](http
 
 ## Credits
 
-The following people have been involved in the development of **Pinocchio**:
+The following people have been involved in the development of **Pinocchio** and are warmly thank for their contributions:
```
gabrielebndn: "and are warmly *thanked*" or "we warmly thank them"

PR #954, opened 2019-11-15
[URDF][Experimental] Enable convex representation of meshes for collision detection

jmirabel:
This is an experimental features so I prefer not to document it.

It is experimental for the following reasons:
- it isn't tested a lot,
- it isn't benchmarked, i.e. for a collision pair (BVH with convex representation, Other), I don't know in which case the convex representation is preferable wrt the BVH representation.
- the API is not perfect yet and will likely change in the future.

### DIFF
--- a/src/parsers/urdf/geometry.hxx
+++ b/src/parsers/urdf/geometry.hxx
@@ -56,8 +56,8 @@ namespace pinocchio
           } // BOOST_FOREACH
         }
         
-        bool replaceCylinderByCapsule (const std::string & linkName,
-                                       const std::string & geomName) const
+        bool isCapsule (const std::string & linkName,
+                        const std::string & geomName) const
         {
           LinkMap_t::const_iterator _link = links_.find(linkName);
           assert (_link != links_.end());
@@ -189,7 +189,7 @@ namespace pinocchio
         // Use FCL capsules for cylinders
         else if (urdf_geometry->type == ::urdf::Geometry::CYLINDER)
         {
-          bool capsule = tree.replaceCylinderByCapsule(linkName, geomName);
+          bool capsule = tree.isCapsule(linkName, geomName);
           meshScale << 1,1,1;
           const ::urdf::CylinderSharedPtr collisionGeometry = ::urdf::dynamic_pointer_cast< ::urdf::Cylinder> (urdf_geometry);
           
--- a/src/parsers/urdf/geometry.hxx
+++ b/src/parsers/urdf/geometry.hxx
@@ -66,8 +66,13 @@ namespace pinocchio
             return false;
           BOOST_FOREACH(const ptree::value_type & cc, link.get_child("collision_checking")) {
             if (cc.first == "capsule") {
+#ifdef PINOCCHIO_URDFDOM_COLLISION_WITH_GROUP_NAME
+              std::cerr << "Warning: support for tag link/collision_checking/capsule"
+                " is not available for URDFDOM < 0.3.0" << std::endl;
+#else
               std::string name = cc.second.get<std::string>("<xmlattr>.name");
               if (geomName == name) return true;
+#endif
             }
           } // BOOST_FOREACH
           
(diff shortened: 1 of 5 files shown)


src/parsers/urdf/geometry.hxx
```diff
@@ -93,7 +111,13 @@ namespace pinocchio
         mesh->scale.z;
       }
 
-#ifdef PINOCCHIO_WITH_HPP_FCL      
+#ifdef PINOCCHIO_WITH_HPP_FCL
+# if ( HPP_FCL_MAJOR_VERSION>1 || ( HPP_FCL_MAJOR_VERSION==1 && \
+      ( HPP_FCL_MINOR_VERSION>1 || ( HPP_FCL_MINOR_VERSION==1 && \
+                                     HPP_FCL_PATCH_VERSION>3))))
+#  define PINOCCHIO_HPP_FCL_SUPERIOR_TO_1_1_3
```
nim65s: We usually say `GREATER` here

jmirabel:
To test this feature, you must use at least the tag v1.1.4-rc1 of hpp-fcl.

src/parsers/urdf/geometry.hxx
jmirabel: Is this really important ? This variable is only used locally and its meaning is clear.

nim65s: It's not :)

nim65s:
So, you are modifying the return object of a function only based on the version of HPP, without knowing if this is useful ?

I think that an experimental feature as you describe it should be turned off by default.

jmirabel:
The behavior changes if one changes a URDF from

    <collision>
      <geometry>
        <mesh filename="package://simple_humanoid_description/box.stl" />
      </geometry>
    </collision>

to

    <collision name="box">
      <geometry>
        <mesh filename="package://simple_humanoid_description/box.stl" />
      </geometry>
    </collision>
    <collision_checking>
      <convex name="box"/>
    </collision_checking>

nim65s:
I see thanks, then it's fine for me.

jmirabel:
CI failed for a reason I don't understand. The cylinder should have been treated as a capsule but it seems it wasn't...

jmirabel:
I wasn't able to reproduce the bug in the docker `eur0c.laas.fr:5000/stack-of-tasks/pinocchio:14.04`. I think this docker is not up-to-date.
@nim65s Can you confirm ? Do you have a working docker file for this version ?

jcarpent:
> I wasn't able to reproduce the bug in the docker `eur0c.laas.fr:5000/stack-of-tasks/pinocchio:14.04`. I think this docker is not up-to-date.
> @nim65s Can you confirm ? Do you have a working docker file for this version ?

You may have a look at gepetto/utils:pinocchio-14.04

jmirabel:
> gepetto/utils:pinocchio-14.04

Could you be more precise ? docker image ? a repo somewhere ? I couldn't find anything on gepgitlab.

jcarpent:
In my case, I'm using: `docker run -it gepetto/utils:pinocchio-14.04 bash`

nim65s:
I am not building 14.04 images automatically anymore, I'll do it manually.

nim65s:
The image `eur0c.laas.fr:5000/stack-of-tasks/pinocchio/py-pinocchio:14.04` aka `gepetto/utils:pinocchio-14.04`is up-to-date & available :)

nim65s:
But looking at https://github.com/Gepetto/robotpkg/commit/4bca3f8b3286a1838347448abb31c9a78a42e3c4, I'll say that if you want to use that, you will have to remove the `DEPEND_ABI.boost-headers` line and replace `c++11.mk` by `c++.mk`.

jmirabel:
I am surprised the bug didn't appear earlier. When PINOCCHIO_URDFDOM_COLLISION_WITH_GROUP_NAME is defined, the geometry name is obtained from the old deprecated attribute `urdf::Collision::group_name`. This attribute has always value 'default'...

I disabled both features when urdfdom version is too low.

jmirabel:
> I would much more deactivate the functions in the corpus of the parser, rather than changing the behavior of the code of each function internally. @jmirabel Could you apply this change?

But then it's hard to easily give a warning message...

jcarpent:
@jmirabel Thanks a lot. I plan to make the release 2.2.0 this week.

PR #959, opened 2019-11-18
[Doc] Add more remarkable identities.

jmirabel:


jcarpent:
@jmirabel Thanks for your sense of detail.

PR #967, opened 2019-11-26
[examples] Improve and test overview-urdf.py

gabrielebndn:
Python example [examples/overview-urdf.py](https://github.com/stack-of-tasks/pinocchio/blob/6fa1c7d5502629ee126f84f1a05471815fba30f4/examples/overview-urdf.py) is now executed by make test, using the in-repo ur5 by default.
I grabbed the occasion to make it more similar to the corresponding C++ file concerning the comments and the format of the outputs


gabrielebndn:
Update: I modified example [examples/i-inverse-kinematics.cpp](https://github.com/stack-of-tasks/pinocchio/blob/6fa1c7d5502629ee126f84f1a05471815fba30f4/examples/i-inverse-kinematics.cpp) to use `JacobiSVD` instead of `BDCSVD`. This way it is compatible with all Eigen versions and it can be executed on all systems, with no need of checking it in the CMakeLists

PR #968, opened 2019-11-27
fix examples

jcarpent:


gabrielebndn:
I think it would be best to automatically test these examples.
In general, we should try to test all examples in [examples/python](https://github.com/stack-of-tasks/pinocchio/tree/d925b4adcf7aece9227f3c9ce46777f4c3cfe6e8/examples/python).
Indeed, we might even remove [examples/python](https://github.com/stack-of-tasks/pinocchio/tree/d925b4adcf7aece9227f3c9ce46777f4c3cfe6e8/examples/python) and move everything into [examples](https://github.com/stack-of-tasks/pinocchio/tree/d925b4adcf7aece9227f3c9ce46777f4c3cfe6e8/examples).
A few comments about this:
- [gepetto-viewer.py](https://github.com/stack-of-tasks/pinocchio/blob/d925b4adcf7aece9227f3c9ce46777f4c3cfe6e8/examples/python/gepetto-viewer.py), [meshcat-viewer.py](https://github.com/stack-of-tasks/pinocchio/blob/d925b4adcf7aece9227f3c9ce46777f4c3cfe6e8/examples/python/meshcat-viewer.py) and [robot-wrapper-viewer.py](https://github.com/stack-of-tasks/pinocchio/blob/d925b4adcf7aece9227f3c9ce46777f4c3cfe6e8/examples/python/robot-wrapper-viewer.py) require a viewer to be installed and active. Do you think it would be possible to unit test them, at least partially?
-  [collision-detection.py](https://github.com/stack-of-tasks/pinocchio/blob/d925b4adcf7aece9227f3c9ce46777f4c3cfe6e8/examples/python/collision-detection.py) and [load-urdf.py](https://github.com/stack-of-tasks/pinocchio/blob/d925b4adcf7aece9227f3c9ce46777f4c3cfe6e8/examples/python/load-urdf.py) could definitely be unit tested, I think.
- Do we need [load-romeo.py](https://github.com/stack-of-tasks/pinocchio/blob/d925b4adcf7aece9227f3c9ce46777f4c3cfe6e8/examples/python/load-romeo.py)?

jmirabel:
You may want to use https://en.wikipedia.org/wiki/Xvfb or a similar tool to the connection to the viewer (without checking if the actual display is correct).

jcarpent:
@gabrielebndn I do agree with all your points. I will transfer all the Python examples to examples directly, with some minor renaming. I will not test at this time the viewer, as it seems a bit hard to set up the solution proposed by @jmirabel.

gabrielebndn:
@jcarpent I pushed a solution allowing to partially test `gepetto-viewer` and `meshcat-viewer`.
I simply catch the exceptions raised by `initViewer()` and `loadViewerModel()`.
This allows to execute the two scripts "correctly" even if the viewers are not installed or launched.
This will at least check that everything up to that point (in particular the URDF parsing) works well, and that there are not blatant syntax errors in the rest of the file either.
If you do not like the addition, feel free to roll it back

jcarpent:
Very nice @gabrielebndn. Thanks for pushing.

jcarpent:
@gabrielebndn Thanks a lot for your contribution.

PR #979, opened 2019-12-06
Fix FCL packaging w.r.t hpp-fcl 1.2.0

jcarpent:


bindings/python/pinocchio/__init__.py
```diff
@@ -13,3 +13,9 @@
 from .libpinocchio_pywrap import *
 from .deprecated import *
 from .shortcuts import *
+
+try:
+  import hppfcl as fcl
+  def WITH_FCL_BINDINGS(): return True
```
nim65s: A constant would not work ?

```python
WITH_FCL_BINDINGS = True
```

Also, as a general style rule, you should avoid putting multiple statements on one line.

jcarpent: Yes, shoud work too.

jcarpent: Doen.

jcarpent:
@nim65s Thanks for your feedback ;)

bindings/python/pinocchio/__init__.py
```diff
@@ -13,3 +13,13 @@
 from .libpinocchio_pywrap import *
 from .deprecated import *
 from .shortcuts import *
+
+if pin.WITH_FCL:
+  try:
+    import hppfcl as fcl
```
jmirabel: This may be confusing people with bindings of the original FCL library. I suggest to keep hppfcl.

bindings/python/multibody/fcl/expose-fcl.cpp
```diff
@@ -18,18 +18,84 @@ namespace pinocchio
...
+        .value("GEOM_CAPSULE",::hpp::fcl::GEOM_CAPSULE)
+        .value("GEOM_CONE",::hpp::fcl::GEOM_CONE)
+        .value("GEOM_CYLINDER",::hpp::fcl::GEOM_CYLINDER)
+        .value("GEOM_CONVEX",::hpp::fcl::GEOM_CONVEX)
+        .value("GEOM_PLANE",::hpp::fcl::GEOM_PLANE)
+        .value("GEOM_HALFSPACE",::hpp::fcl::GEOM_HALFSPACE)
+        .value("GEOM_TRIANGLE",::hpp::fcl::GEOM_TRIANGLE)
+        .value("GEOM_OCTREE",::hpp::fcl::GEOM_OCTREE)
+        .value("NODE_COUNT",::hpp::fcl::NODE_COUNT)
+        ;
```
jmirabel: Do you need to re-expose them ?

jmirabel: I may not understand what `eigenpy::register_symbolic_link_to_registered_type` does.

bindings/python/pinocchio/__init__.py
jcarpent: I will change it.

bindings/python/multibody/fcl/expose-fcl.cpp
jcarpent: When you have implemented the Python bindings inside hpp-fcl, I don't know how much you realized that some part was already in Pinocchio.
So, the idea is to keep these bindings inside Pinocchio for a certain fixed duration (and to remove them in 6 months for instance). To avoid the duplication of symbols, I have implemented in EigenPy register_symbolic_link_to_registered_type<Type> which:
- check if the Type has been registered
- if it has been registered, then just put a symlink in the current Python scope
- otherwise, return False.
In the present context, if the result is False, it will just expose the enum as it was done before.
Is it OK for you?

jmirabel: It's ok. Thanks.

bindings/python/pinocchio/__init__.py
jmirabel: Thanks.

PR #981, opened 2019-12-08
python: fix naming of the variable

jcarpent:
@jmirabel This PR fixes the issues revealed in #979.

### DIFF
--- a/bindings/python/pinocchio/__init__.py
+++ b/bindings/python/pinocchio/__init__.py
@@ -16,7 +16,7 @@ from .shortcuts import *
 
 if pin.WITH_HPP_FCL:
   try:
-    import hppfcl as fcl
+    import hppfcl
     WITH_HPP_FCL_BINDINGS = True
   except ImportError:
     WITH_HPP_FCL_BINDINGS = False
--- a/bindings/python/pinocchio/__init__.py
+++ b/bindings/python/pinocchio/__init__.py
@@ -2,11 +2,10 @@
 # Copyright (c) 2015-2019 CNRS INRIA
 #
 
-import numpy as np
+import numpy
 from .robot_wrapper import RobotWrapper
 from .libpinocchio_pywrap import __version__, __raw_version__
 
-from . import libpinocchio_pywrap as pin
 from . import utils
 from . import visualize
 from .explog import exp, log
--- a/bindings/python/pinocchio/__init__.py
+++ b/bindings/python/pinocchio/__init__.py
@@ -14,12 +14,12 @@ from .libpinocchio_pywrap import *
 from .deprecated import *
 from .shortcuts import *
 
-if pin.WITH_FCL:
+if pin.WITH_HPP_FCL:
   try:
     import hppfcl as fcl
-    WITH_FCL_BINDINGS = True
+    WITH_HPP_FCL_BINDINGS = True
   except ImportError:
-    WITH_FCL_BINDINGS = False
+    WITH_HPP_FCL_BINDINGS = False
 else:
-  WITH_FCL_BINDINGS = False
+  WITH_HPP_FCL_BINDINGS = False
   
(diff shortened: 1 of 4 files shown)


bindings/python/pinocchio/__init__.py
```diff
@@ -14,12 +14,12 @@
 from .deprecated import *
 from .shortcuts import *
 
-if pin.WITH_FCL:
+if pin.WITH_HPP_FCL:
   try:
     import hppfcl as fcl
```
gabrielebndn: I suggest to replace `import hppfcl as fcl` to just `import hppfcl` too. I thought this is what @jmirabel meant

gabrielebndn: Also, I would like to point out that this breaks existing Python code using the old Pinocchio-integrated HPP-FCL bindings: `pinocchio.CollisionGeometry` will not work, for instance.
In order to avoid that, you should import in Pinocchio the old classes one by one. If you want to keep this behavior forever, you can just leave it like that. Otherwise, you can easily make the pinocchio-integrated version deprecated

gabrielebndn:
If your plan is to remove the Pinocchio-integrated FCL bindings altogether (which I support), wouldn't it be better to issue some kind of warning when the import error is caught?

jcarpent:
> If your plan is to remove the Pinocchio-integrated FCL bindings altogether (which I support), wouldn't it be better to issue some kind of warning when the import error is caught?

Not now, because hppfcl Python bindings are not yet available on most existing package managers that support Pinocchio and hpp-fcl.

bindings/python/pinocchio/__init__.py
jcarpent: Done!

gabrielebndn:
> > If your plan is to remove the Pinocchio-integrated FCL bindings altogether (which I support), wouldn't it be better to issue some kind of warning when the import error is caught?
> 
> Not now, because hppfcl Python bindings are not yet available on most existing package managers that support Pinocchio and hpp-fcl.

Right. But it should be done hpp-fcl bindings are fully supported everywhere

gabrielebndn:
Also, an observation: I believe variable `WITH_HPP_FCL_BINDINGS` is redundant. If module `hppfcl` was imported successfully, then `hasattr(pinocchio, "hppfcl")` will be true, otherwise it will be not. Indeed, this expression is equivalent to `WITH_HPP_FCL_BINDINGS`, and it will always work, unless the user defines something called `hppfcl` in the Pinocchio module, which they should not do

jcarpent:
@gabrielebndn This is a question of simplicity and uniformization. Adding the present flag comes with no cost, and it is much more nicer than `hasattr(pinocchio, "hppfcl")` I would say.

gabrielebndn:
> @gabrielebndn This is a question of simplicity and uniformization. Adding the present flag comes with no cost, and it is much more nicer than `hasattr(pinocchio, "hppfcl")` I would say.

It is nicer,  but now I've been thinking: if you add the variable `WITH_HPP_FCL_BINDINGS`, there will be code using it.
One day, you will remove Pinocchio support for the Python bindings, and you will likely force a dependency on the hpp-fcl Python bindings. Then variable `WITH_HPP_FCL_BINDINGS` will be useless, because it will coincide with `WITH_HPP_FCL`, and you will have to remove it, breaking some code, possibly passing through deprecation in case external code is actually employing it.
Instead, `hasattr(pinocchio, "hppfcl")` is much more portable, it will never break, and it will require no management at all.
I do not have strong feelings about it, so do as you please, but think about it

jcarpent:
> One day, you will remove Pinocchio support for the Python bindings

I don't get your point here at all.

gabrielebndn:
> > One day, you will remove Pinocchio support for the Python bindings
> 
> I don't get your point here at all.

I mean, you will remove the implementation of the Python bindings from within Pinocchio.
But never mind, it is a minor point

PR #982, opened 2019-12-08
Examples

gabrielebndn:
I worked a little bit on the examples.
The main changes:
- I removed the prefix letter from all code examples
- The inverse kinematics example now works in 6D
- The example overview-urdf now also appears in the example list to the left of the doxygen-generated doc. The goal is to comment it more extensively than in the main page, later
- I included the viewer examples in the examples visible in the HTML-generated doc
- I added a new example on loading the geometry models

I did not comment on the examples and I do not have time to do that


examples/inverse-kinematics.cpp
```diff
@@ -39,9 +41,12 @@ int main(int /* argc */, char ** /* argv */)
       break;
     }
     pinocchio::computeJointJacobian(model,data,q,JOINT_ID,J);
-    const Eigen::VectorXd v     = - svd.compute(J.topRows<3>()).solve(err);
+    pinocchio::SE3::Matrix6 Jlog;
+    pinocchio::Jlog6(dMi, Jlog);
+    v = - svd.compute(Jlog * J).solve(err);
```
jcarpent: You should change the way you've defined the error to not use the Jlog in your expression.
Indeed, using Jlog may trouble people that are only familiar with standard IK.

So, the error should be:
```
     const pinocchio::SE3 dMi = data.oMi[JOINT_ID].actInv(oMdes); 
```
giving rise:
```
v = + svd.compute(J).solve(err);
```
If I'm not making mistakes ;)

jcarpent:
@gabrielebndn Thanks a lot for all those fixes. Very nice and helpful!
Could you change Romeo on the new examples to use TALOS if possible?

examples/inverse-kinematics.cpp
gabrielebndn: @jcarpent, then how would you convert `dMi` into a vector? The purpose of `log` here is performing this operation.
Indeed, I thought that using log was a plus. I, in the first place, was not familiar with the use of log in the inverse kinematics example, and I had to find out myself how it all worked when I had to implement an inverse kinematics of my own; hence the idea of putting it in the example. I would have appreciated the existence of this example as it is the way I am presenting it, and so I think other people will.
Of course we are not obliged to pass through the log, there are other ways. Examples can be found in Sec. 3.7.3 of Oriolo's book "Robotics Modelling, Planning and Control" ("Orientation error"), subsections "Angle axis" and "Unit Quaternion" (I can send you a link to the book if you do not have it and you wish to take a look). But it would involve lengthier computations and possibly defining dedicated functions for the error. Instead, here, I am showing how everything can be done in Pinocchio using just a few lines, employing only Pinocchio functions. I think this makes the example both more compact and more significant. I am explaining the use of the log in the text (that example was already commented for position-only error, I corrected the relevant parts).

jcarpent: In you example, as you can notice, and keep the error as:
```
err = pinocchio::log6(dMi).toVector()
```

gabrielebndn:
> If you can handle it, it would be nice, and most users will better understand it.
> @gabrielebndn Thanks in advance.

So, basically:
- inverse kinematics was already commented, I modified the text for the relevant parts
- for the viewers I added a minimal intro; the code itself already has lots of comments and it is pretty clear, comment text would amount to little more than copying the comments
- for overview-urdf and geometry-models, I might add some commentary in the future

examples/inverse-kinematics.cpp
jcarpent: You just change the way the error is represented. On your side, your expressed the error with respect to the target frame. On my side, it expressed in the local frame of the end-effector. Two opposite view of the same thing ;)

gabrielebndn:
> @gabrielebndn Thanks a lot for all those fixes. Very nice and helpful!
> Could you change Romeo on the new examples to use TALOS if possible?

There are no new examples using Romeo: those are still the old ones, I just made them visible in the doxygen doc. I would rather not do it myself for now, I have to go. Also, last time I tried to load TALOS on Meshcat I had problems, they need to be fixed. And, now that I think about it neither gepetto-gui nor Meshcat work on my laptop, so I could not check the result

examples/inverse-kinematics.cpp
gabrielebndn: Oh! Ok, I understand, I will do it immediately

gabrielebndn: Actually, are you sure? I think it is easier to understand my way. In 3D, it was:
```
err = R.transpose()*(x-xdes);
v = - svd.compute(J.topRows<3>()).solve(err);	
```
which I find better than
```
err = R.transpose()*(xdes-x);
v = svd.compute(J.topRows<3>()).solve(err);	
```

jcarpent: Relying on Jlog may not be easy to get for most people.

gabrielebndn:
> Relying on Jlog may not be easy to get for most people.

Ooh, now I see what you mean... You switch the two and then you can remove Jlog... You see, I am not super familiar with log, I learnt a different way of doing IK...
Are you sure it would still work? I would think that Jlog would be needed anyways...
Anyhow, I am out of home now and I cannot work on that. Could you try that? I think it is better if you handle it as you seem to know it better than me. If you do modify the code, there will be the doc to fix

examples/inverse-kinematics.cpp
gabrielebndn: I confirm that without Jlog it does not work, even changing the expression of the error. This does not surprise me, because `log(M^-1) = - log(M)`, so using `data.oMi[JOINT_ID].actInv(oMdes)` or `oMdes.actInv(data.oMi[JOINT_ID])` should not change anything.
I think we need to stuck to `Jlog`. So I would say this PR is ready

gabrielebndn:
I have no time to comment the tests now. I may do it another time. But for now, I think you can already merge

examples/inverse-kinematics.cpp
jcarpent: You don't have any model with exponential decay on your system. Inverse kinematics is different from inverse geometry !!!

gabrielebndn: This is the way I have always done and it works. I could sketch the equations, but I would rather not do it, I do not have time for that and I do not think it is needed.
With your proposed changes, instead, the example does not work anymore. And I do not see how it could be different, because if you incorporate log in the expression of the error, you need the derivative on the log.
I could switch to a Jacobian-transpose system, but I do not think it would change anything and I do not see the point of it.
I remind you that this is just an example of how to use the code with no intention of being 100% rigorous and functional.
For me, the question is settled. If you wish to introduce changes, you are welcome to do it yourself

gabrielebndn: @jcarpent, what about this?
I do not have more time to spend on this. If you do not like the changes, I can just remove the two commits involving the 6D inverse kinematics, and the example will stay in 3D. Then you can merge the rest.

jcarpent:
@gabrielebndn I've only limited times to work on it, as currently, I'm mostly fixing the leaks to make Pinocchio/EigenPy and others working smoothly on all the OSs. This is not a trivial task, and I hope you may understand my late reply.

What is not correct IMHO, is that your example is not doing inverse kinematics but rather inverse geometry. These are two different instances, using similar features. I hope you understand the difference.

Anyway, in both cases, the Jacobian of the log should appear but its influence can be removed when doing inverse kinematics thanks to the feedback term!

FINALLY, I will now merge this PR because I don't have to modify it, even if I consider THIS IS YOUR JOB as one of the main contributor. I cannot absorde all the propositions and modify them to suppress the mistakes/errors/approximations.

gabrielebndn:
> FINALLY, I will now merge this PR because I don't have to modify it, even if I consider THIS IS YOUR JOB as one of the main contributor. I cannot absorde all the propositions and modify them to suppress the mistakes/errors/approximations.

That's why I said it, if you prefer I can just remove the commits. This is a completely good-willed contribution and in my view it's fine, if you do not like it I can withdraw it

PR #986, opened 2019-12-10
Improve compatibility with respect to Eigen

jcarpent:


.gitlab-ci.yml
```diff
@@ -27,31 +27,31 @@ cache:
 
 #robotpkg-pinocchio-14.04-release:
 #  <<: *robotpkg-pinocchio
-#  image: eur0c.laas.fr:5000/stack-of-tasks/pinocchio/pinocchio:14.04
+#  image: memmo.laas.fr:5000/stack-of-tasks/pinocchio/pinocchio:14.04
```
nim65s: that's `memmos`

jcarpent: Why `s`?

nim65s: `memmo.laas.fr` was registered as a domain name for a website before we needed a computer for this project, so we could not name the computer with this name. Then, that's Memmo's `S`erver.

jcarpent: Good. Thanks.

PR #995, opened 2019-12-12
Fix Geometry module

jcarpent:
@jmirabel Can you check if the change I did are fine with you?

### DIFF
--- a/bindings/python/multibody/geometry-data.hpp
+++ b/bindings/python/multibody/geometry-data.hpp
@@ -75,7 +75,11 @@ namespace pinocchio
                       "Vector of distance results.")
         .def_readonly("collisionRequest",
                       &GeometryData::collisionRequest,
-                      "Defines which information should be computed by FCL for collision computations.")
+                      "Defines which information should be computed by FCL for collision computations.\n\n"
+                      "Note: it is possible to define a security_margin and a break_distance for a collision request.\n"
+                      "Most likely, for robotics application, these thresholds will be different for each collision pairs\n"
+                      "(e.g. the two hands can have a large security margin while the two hips cannot.\n"
+                      "It may be a good idea to have as many collision requests as collision pairs.)")
         .def_readonly("collisionResults",
                       &GeometryData::collisionResults,
                       "Vector of collision results.")
--- a/bindings/python/multibody/geometry-data.hpp
+++ b/bindings/python/multibody/geometry-data.hpp
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2015-2018 CNRS
+// Copyright (c) 2015-2019 CNRS INRIA
 //
 
 #ifndef __pinocchio_python_geometry_data_hpp__
@@ -26,14 +26,14 @@ namespace pinocchio
     struct CollisionPairPythonVisitor
     : public boost::python::def_visitor<CollisionPairPythonVisitor>
     {
-
       static void expose()
       {
         bp::class_<CollisionPair> ("CollisionPair",
                                    "Pair of ordered index defining a pair of collisions",
                                    bp::no_init)
-        .def(bp::init<const GeomIndex &, const GeomIndex &>(bp::args("co1 (index)", "co2 (index)"),
-                                                        "Initializer of collision pair."))
+        .def(bp::init<const GeomIndex &, const GeomIndex &>
+             (bp::args("self","index 1", "index 2"),
+              "Initializer of collision pair."))
         .def(PrintableVisitor<CollisionPair>())
         .def(CopyableVisitor<CollisionPair>())
         .def(bp::self == bp::self)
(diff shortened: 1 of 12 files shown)


bindings/python/multibody/geometry-data.hpp
```diff
@@ -44,67 +44,68 @@ namespace pinocchio
...
                       &GeometryData::distanceRequest,
-                      "Defines which information should be computed by fcl for distances")
+                      "Defines which information should be computed by FCL for distance computations")
         .def_readonly("distanceResults",
                       &GeometryData::distanceResults,
                       "Vector of distance results.")
         .def_readonly("collisionRequest",
                       &GeometryData::collisionRequest,
-                      "Defines which information should be computed by fcl for collisions.")
+                      "Defines which information should be computed by FCL for collision computations.")
```
jmirabel: A note for the future:

it is possible to define a *security_margin* and a *break_distance* for a collision request. Most likely, for robotics application, those threshold will be different for each collision pairs. E.g. the two hands can have a large security margin while the two hips cannot. It may be a good idea to have as many collision requests as collision pairs.

bindings/python/multibody/geometry-object.hpp
```diff
@@ -28,27 +28,44 @@ namespace pinocchio
...
                       bp::make_getter(&GeometryObject::meshColor,
                                       bp::return_value_policy<bp::return_by_value>()),
                       bp::make_setter(&GeometryObject::meshColor),
-                      "Color rgba for the mesh")
-        .def_readwrite("name", &GeometryObject::name, "Name of the GeometryObject")
-        .def_readwrite("parentJoint", &GeometryObject::parentJoint, "Index of the parent joint")
-        .def_readwrite("parentFrame", &GeometryObject::parentFrame, "Index of the parent frame")
+                      "Color rgba of the mesh.")
+        .def_readwrite("fcl", &GeometryObject::fcl,
+                       "The FCL CollisionGeometry associated to the given GeometryObject.")
```
jmirabel: Another unrelated note: maybe *geometry* would be a better attribute name.

src/multibody/fcl.hpp
```diff
@@ -92,7 +94,7 @@ struct GeometryObject
   /// \brief Index of the parent joint
   JointIndex parentJoint;
 
-  /// \brief The actual cloud of points representing the collision mesh of the object after scaling.
+  /// \brief The FCL CollisionGeometry
```
jmirabel: It might be worth saying that *meshFile + meshScale -> fcl* as a note.

src/multibody/geometry.hxx
```diff
@@ -42,10 +43,9 @@ namespace pinocchio
   GeomIndex GeometryModel::addGeometryObject(const GeometryObject & object,
                                              const ModelTpl<S2,O2,JointCollectionTpl> & model)
   {
-    assert( //TODO: reenable when relevant (object.parentFrame == -1) ||
-           (model.frames[object.parentFrame].type == pinocchio::BODY)  );
```
jmirabel: I don't remember why the added this assert. I think the rationale was to check data coherence. If it is not a body, the caller likely made a mistake.

src/multibody/geometry.hxx
```diff
@@ -42,10 +43,9 @@ namespace pinocchio
...
-    assert( //TODO: reenable when relevant (object.parentFrame == -1) ||
-           (model.frames[object.parentFrame].parent == object.parentJoint) );
+    if(object.parentFrame < (FrameIndex)model.nframes)
+      PINOCCHIO_CHECK_INPUT_ARGUMENT(model.frames[object.parentFrame].parent == object.parentJoint,
+                                     "The object joint parent and its frame joint parent do not match.");
```
jmirabel: Shouldn't you throw something if `object.parentFrame >= (FrameIndex)model.nframes` ?

bindings/python/multibody/geometry-object.hpp
gabrielebndn: I also do not like the name of this attribute. But the name `fcl` is in C++, and I do not see how we could change it without breaking somebody's code. In Python, we should think about consistency with C++ most of all

src/multibody/geometry.hxx
jcarpent: the `object.parentFrame` is optional. And as it is a `FrameIndex == size_t`, the default value cannot be `-1`.
I decided to set it to the max value.
So, we should not check this condition.

bindings/python/multibody/geometry-object.hpp
jcarpent: I think the best name for it would be geometry. @jmirabel @gabrielebndn What do you think?

src/multibody/geometry.hxx
jcarpent: Yes,  but as `parentFrame` is optional, I think this check is not relevant and may fail.

bindings/python/multibody/geometry-data.hpp
jcarpent: I will add this note for additional information.

bindings/python/multibody/geometry-object.hpp
jmirabel: Yes. We can put this in the list of changes for Pinocchio 3.

gabrielebndn: I think `geometry` is a good name. But again, I think it is difficult to change it in C++ without breaking code. If you are willing to do it, go on. Most of all, I would avoid having it called `fcl` in C++ and `geometry` in Python

jcarpent: Or we may proceed to the change even before, because fcl is a pointer

gabrielebndn: You mean, having two pointers to the same object, with the `fcl` pointer being deprecated?
Having two pointers, and then only using one of them, would break all code looking like `geometryObject.fcl = ...`, because it will not change `geometryObject.geometry`. I don't know how much code there may be out there doing something like this

src/multibody/fcl.hpp
jcarpent: What do you mean here?

jmirabel: I see two interpretation of the 3 attributes meshFile, meshScale and fcl:
1. fcl == meshFile (i.e. the scale wasn't applied when loading),
2. fcl == meshFile + meshScale (i.e. the scale was applied when loading).

For example, if meshFile contains a line of 1 meter, and meshScale is 0.1, then:
1. fcl contains a line of 1 meter,
2. fcl contains a line of 0.1 meter,

bindings/python/multibody/geometry-object.hpp
jcarpent: I will provide a solution soon ;)

jcarpent:
@jmirabel I've implemented a workaround to rename `GeometryObject::fcl` into `GeometryObject::geometry`. This should work. Could you try it on the stack HPP?

src/multibody/fcl.hpp
```diff
@@ -84,38 +86,57 @@ struct GeometryObject
...
   JointIndex parentJoint;
 
-  /// \brief The actual cloud of points representing the collision mesh of the object after scaling.
-  CollisionGeometryPtr fcl;
+  /// \brief The FCL CollisionGeometry (might be a Mesh, a Geometry Primitive, etc.)
+  CollisionGeometryPtr geometry;
+
+  /// \brief The former pointer to the FCL CollisionGeometry.
+  /// \deprecated It is now deprecated and has been renamed GeometryObject::geometry
+  PINOCCHIO_DEPRECATED CollisionGeometryPtr & fcl;
```
jmirabel: After some trial and error, this could be replaced by the safer version below.

```cpp
union {
  CollisionGeometryPtr geometry;
  PINOCCHIO_DEPRECATED CollisionGeometryPtr fcl;
}
```

I didn't check how doxygen behaves.

jcarpent: I've tried, and I got the following error:
```
union member 'geometry' has a non-trivial copy constructor
because field of type 'boost::detail::shared_count' has a user-provided copy constructor
```

jcarpent: Does the solution I've suggested work?

jmirabel: My bad. I actually tested it only with doubles...

jmirabel: I will give a try to HPP with the devel branch of pinocchio in the next few days.

jmirabel: I launched a compilation with these changes and the one in https://github.com/humanoid-path-planner/hpp-fcl/pull/119.

jcarpent: Sounds great. After your feedback, I will make a new release of hpp-fcl to sync anaconda.

jmirabel: It worked fine.

PR #998, opened 2019-12-17
WIP: add Pickling of Data

jcarpent:


src/multibody/joint/joint-free-flyer.hpp
```diff
@@ -192,7 +192,7 @@ namespace pinocchio
     : M(1)
     , U(U_t::Zero())
     , Dinv(D_t::Zero())
-    , UDinv(UD_t::Zero())
+    , UDinv(UD_t::Identity())
```
gabrielebndn: For consistency, shouldn't it be UDinv = U * Dinv ?

jcarpent: When you do the math, this quantity is always equal to Identity ;)

gabrielebndn: I understand it and I do agree with this change. What I mean is, shouldn't you change U and Dinv too, so to be consistent?

jcarpent: U and UDInv depends on the Inertia I input ;)

gabrielebndn: Oh, I see. Well, as they say, sorry for the noise

src/multibody/joint/joint-revolute.hpp
```diff
@@ -120,16 +120,16 @@ namespace pinocchio
...
           res.rotation().col(2).noalias() = m_cos * m.rotation().col(2) + m_sin * m.rotation().col(0);
+          res.rotation().col(1) = m.rotation().col(1);
           res.rotation().col(0).noalias() = res.rotation().col(1).cross(res.rotation().col(2));
           break;
         }
         case 2:
         {
-          res.rotation().col(2) = m.rotation().col(2);
           res.rotation().col(0).noalias() = m_cos * m.rotation().col(0) + m_sin * m.rotation().col(1);
           res.rotation().col(1).noalias() = res.rotation().col(2).cross(res.rotation().col(0));
```
jmirabel: It seems to me that `res.rotation().col(2)` is not initialized, is it ?

PR #1000, opened 2019-12-18
Deprecate BUILD_UNIT_TESTS. Closes #999

### LINKED ISSUE #999: BUILD_TESTING vs BUILD_UNIT_TESTS
Module CTest automatically creates a `BUILD_TESTING` option that selects whether to enable testing support, ON by default, see [here](https://cmake.org/cmake/help/latest/module/CTest.html).
In Pinocchio, this is included by the cmake submodule by [post-project.cmake](https://github.com/jrl-umi3218/jrl-cmakemodules/blob/54ece258eed16da94c7f10979588fcb47b744eb8/post-project.cmake#L44). This variable is then employed consistently by macro `ADD_UNIT_TEST`, defined in the cmake submodule in [test.cmake](https://github.com/jrl-umi3218/jrl-cmakemodules/blob/54ece258eed16da94c7f10979588fcb47b744eb8/test.cmake#L72), to control whether unit tests are built or not.

However, if one is building Pinocchio from source, the variable to control the building of unit tests is `BUILD_UNIT_TESTS`. `BUILD_TESTING` is then brutally set to `BUILD_UNIT_TESTS` in [the main CMakeLists.txt](https://github.com/stack-of-tasks/pinocchio/blob/474d0706d44286476d8956d43c41e6a7bcd2350b/CMakeLists.txt#L64). As you can see, the original value of `BUILD_TESTING` is completely ignored.

As a result, there are two variables which exist to control the compilation of the unit tests, but one of them is ignored and overridden by the other. When doing `ccmake`, I see this variable `BUILD_TESTING` which is set to `ON` and does nothing.

I think this is confusing, especially to newcomers. The worst part is that `BUILD_TESTING` is standard CMake, therefore it is naturally the variable of choice for those who who first delve into Pinocchio with some CMake experience; but setting this variable manually has no effect at all.

I think the main variable to control the unit tests should be `BUILD_TESTING` (defaulting to OFF). `BUILD_UNIT_TESTS` should be deprecated and later removed altogether. What do you think?

gabrielebndn:
This closes #999


gabrielebndn:
@nim65s, I think this will have to be handled in the gitlab CI and possibly for robotpkg

gabrielebndn:
Update: I added a small unrelated change: header bindings/python/multibody/fcl/transform.hpp is removed from `PINOCCHIO_PYTHON_HEADERS` when building without HPP-FCL (this was causing a warning).

nim65s:
> @nim65s, I think this will have to be handled in the gitlab CI and possibly for robotpkg

Indeed, I will handle that.

CMakeLists.txt
```diff
@@ -41,12 +45,16 @@ ENDIF(WIN32)
 
 # --- OPTIONS ----------------------------------------
 OPTION (BUILD_BENCHMARK "Build the benchmarks" OFF)
-OPTION (BUILD_UNIT_TESTS "Build the unitary tests" OFF)
 OPTION (BUILD_UTILS "Build the utils" OFF)
 OPTION (BUILD_PYTHON_INTERFACE "Build the python binding" ON)
 OPTION (BUILD_WITH_LUA_SUPPORT "Build the lua parser" OFF)
 OPTION (BUILD_WITH_COMMIT_VERSION "Build libraries by setting specific commit version" OFF)
 
+IF(DEFINED BUILD_UNIT_TESTS)
```
jcarpent: We do not need to set it deprecated:
- this won't impact the behavior of Pinocchio
- it will output a anoying messsage for non-dev people

CMakeLists.txt
```diff
@@ -19,6 +19,10 @@ IF(NOT EXISTS "${CMAKE_SOURCE_DIR}/cmake/base.cmake")
   MESSAGE(FATAL_ERROR "\nPlease run the following command first:\ngit submodule update --init\n")
 ENDIF()
 
+# Normally, BUILD_TESTING defaults to ON. This sets the default to OFF.
+# It needs to be included before module CTest, thus before command PROJECT
+OPTION (BUILD_TESTING "" OFF)
```
jcarpent: We should comply with the common rules. We should let this variable at its default value.

gabrielebndn: I agree. I added this so that the behavior stayed closer to the current one, but it is better to stick to the standards

CMakeLists.txt
gabrielebndn: I think a short deprecation notice is useful.
It will make it easier to handle the change for tools that automatically build Pinocchio.
Also, everyone's build folders are configured with `BUILD_UNIT_TESTS`. With no deprecation period, this setting will just stop working and everybody will have fall on `BUILD_TESTING` set to on, with no clue on why. At least this will output an explicit message. Maybe it should be even _more_ explicit and prescriptive, such as "we suggest that you delete you build folder and make a new one".
I have tested this to ensure that the change will be as smooth as possible.
The message is completely invisible to anybody making a new build folder from scratch.
For the deprecation, I took inspiration from `DISABLE_TESTS`: https://github.com/jrl-umi3218/jrl-cmakemodules/blob/54ece258eed16da94c7f10979588fcb47b744eb8/test.cmake#L35

jcarpent: OK. Fine for the explanations.

gabrielebndn:
Done

PR #1003, opened 2020-01-07
visualize: Add hppfcl primitives shapes to visualizer

rstrudel:
This PR allows to display primitive shapes into the gepetto visualizer

PR #1004, opened 2020-01-07
[visualize] Add geometric primitives

gabrielebndn:
I meant to wait for a day or two for further testing and/or additions in order to deliver these commits, but then I saw #1003, so I told myself I might as well do the PR now.
- Compared to #1003, I think my solution is slightly more robust to different install configurations (with/without hppfcl, mostly).
- Additionally, I fixed the visualization of geometric primitives for meshcat too
- I now use Talos in the visualization examples. previously this was not possible, since the model contained some primitives which made the visualizer crash.
- I retained the Romeo example in [meshcat-viewer-dae.py](https://github.com/stack-of-tasks/pinocchio/blob/a78eefc877f8dae19b5b07ddae372b5f4a520bb5/examples/meshcat-viewer-dae.py) to show the specific  problems that arise with DAE files in meshcat. I determined that there is no problem of textures, since Romeo does not contain them: the issue is that meshcat does not parse colors in DAE files correctly.
- I added color to the sample model so that it can be easily visualized
- I created a script called [sample-model-viewer.py](https://github.com/stack-of-tasks/pinocchio/blob/38c69350553da4b6b1c2e30b480b5be2e2dd502d/examples/sample-model-viewer.py) to visualize the sample humanoid
- #1003 also incudes cones
- #1003 features a script testing all geometric shapes which might be interesting to integrate

### DIFF
--- a/bindings/python/pinocchio/visualize/gepetto_visualizer.py
+++ b/bindings/python/pinocchio/visualize/gepetto_visualizer.py
@@ -69,6 +69,8 @@ class GepettoVisualizer(BaseVisualizer):
             return gui.addBox(meshName, w, h, d, npToTuple(meshColor))
         elif isinstance(geom, hppfcl.Sphere):
             return gui.addSphere(meshName, geom.radius, npToTuple(meshColor))
+        elif isinstance(geom, hppfcl.Cone):
+            return gui.addCone(meshName, geom.radius, 2. * geom.halfLength, npToTuple(meshColor))
         else:
             msg = "Unsupported geometry type for %s (%s)" % (geometry_object.name, type(geom) )
             warnings.warn(msg, category=UserWarning, stacklevel=2)
--- a/bindings/python/pinocchio/visualize/gepetto_visualizer.py
+++ b/bindings/python/pinocchio/visualize/gepetto_visualizer.py
@@ -1,7 +1,16 @@
 from .. import libpinocchio_pywrap as pin
 from ..shortcuts import buildModelsFromUrdf, createDatas
+from ..rpy import npToTuple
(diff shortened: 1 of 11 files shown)


rstrudel:
Thanks Gabriel, your PR seems indeed more complete. I think if you could add cones and the test I have written, which check if all the shapes are aligned along an axis and should have the same size, it would be great.

gabrielebndn:
@rstrudel yes, I will do it. I'll try to cherry-pick your commit

rstrudel:
Yes, for the test, cherry picking my first commit and checking out the gepetto visualizer file should do the trick. Thank you

gabrielebndn:
> For the test, I created a branch [topic/example_shapes](https://github.com/rstrudel/pinocchio/commits/topic/example_shapes) which contains only the test. Hope this can help

Indeed it helped, thanks!

gabrielebndn:
Done.
Some considerations on  [examples/display_shapes.py](https://github.com/gabrielebndn/pinocchio/blob/50329b985c311693fc33939b52ebb4ae403023a2/examples/display_shapes.py)
- I modified it in order to include it in the the CI. However, right now this addition is only nominal, since it fails importing hppfcl. I'll work on it later.
- It could be easily modified to work for meshcat, or adapted to work for both. However, I left it with Gepetto Viewer only

jcarpent:
@gabrielebndn and @rstrudel Thanks a lot for your contributions.

rstrudel:
@gabrielebndn thanks a lot for the work ! I can now use meshcat and gepetto visualizer on all of my usecases it is really appreciated :)

PR #1006, opened 2020-01-10
[examples/inverse-kinematics] Remove Jlog and use DLS instead

gabrielebndn:
Related to https://github.com/stack-of-tasks/pinocchio/pull/982#discussion_r355190193.
Following discussions with @nmansard, I found out why the example was not working for me without `Jlog`.

The problem is that the initial guess is `pinocchio.neutral(model)`, which is a singular configuration. Simple pseudo-inversion of the kinematic Jacobian kept it singular. Somehow, adding `Jlog` pulled the robot out of the singular configuration.

I have examined the problem carefully, and it does seem that `Jlog * err = err` and `pinv(Jlog) * err = err`, making the inclusion of `Jlog` useless for this particular case.
However, it seems that when the kinematic Jacobian `J` is singular, `pinv(Jlog * J) * err` and `pinv(J) * err` lead to different results. I am not sure wether this is a mathematical truth or a numerical artifact, but I do not find it surprising, since the the equality `pinv(Jlog * J) == pinv(J) * pinv(Jlog)` is not valid in general if one of the two matrices is singular.

These being the facts, I had two options for removing `Jlog`: starting from a non-singular configuration or using the damped pseudo-inverse. Both solutions work; I opted for the latter one, as it is more general. Also, I made some small cosmetic changes. I updated the doc accordingly.

My analysis can be confirmed by [this enhanced version of the Python script](https://gist.github.com/gabrielebndn/c994080a739bea79cbf33e2f8d88f2f1), in  my gists.

### DIFF
--- a/examples/inverse-kinematics.cpp
+++ b/examples/inverse-kinematics.cpp
@@ -22,7 +22,7 @@ int main(int /* argc */, char ** /* argv */)
   pinocchio::Data::Matrix6x J(6,model.nv);
   J.setZero();
 
-  bool success;
+  bool success = false;
   typedef Eigen::Matrix<double, 6, 1> Vector6d;
   Vector6d err;
   Eigen::VectorXd v(model.nv);
--- a/examples/inverse-kinematics.cpp
+++ b/examples/inverse-kinematics.cpp
@@ -17,11 +17,12 @@ int main(int /* argc */, char ** /* argv */)
   const double eps  = 1e-4;
   const int IT_MAX  = 1000;
   const double DT   = 1e-1;
+  const double damp = 1e-6;
 
-  pinocchio::Data::Matrix6x J(6,model.nv); J.setZero();
-  unsigned int svdOptions = Eigen::ComputeThinU | Eigen::ComputeThinV;
-  Eigen::JacobiSVD<pinocchio::Data::Matrix6x> svd(6, model.nv, svdOptions);
+  pinocchio::Data::Matrix6x J(6,model.nv);
+  J.setZero();
 
+  bool success;
   typedef Eigen::Matrix<double, 6, 1> Vector6d;
   Vector6d err;
   Eigen::VectorXd v(model.nv);
@@ -32,22 +33,28 @@ int main(int /* argc */, char ** /* argv */)
     err = pinocchio::log6(dMi).toVector();
     if(err.norm() < eps)
     {
-      std::cout << "Convergence achieved!" << std::endl;
+      success = true;
       break;
     }
     if (i >= IT_MAX)
     {
-      std::cout << "\nWarning: the iterative algorithm has not reached convergence to the desired precision" << std::endl;
+      success = false;
       break;
     }
     pinocchio::computeJointJacobian(model,data,q,JOINT_ID,J);
-    pinocchio::SE3::Matrix6 Jlog;
-    pinocchio::Jlog6(dMi, Jlog);
-    v = - svd.compute(Jlog * J).solve(err);
+    pinocchio::Data::Matrix6 JJt;
+    JJt.noalias() = J * J.transpose();
+    JJt.diagonal().array() += damp;
+    v.noalias() = - J.transpose() * JJt.ldlt().solve(err);
     q = pinocchio::integrate(model,q,v*DT);
     if(!(i%10))
-      std::cout << "error = " << err.transpose() << std::endl;
+      std::cout << i << ": error = " << err.transpose() << std::endl;
   }
+
+  if(success)
+    std::cout << "Convergence achieved!" << std::endl;
+  else
+    std::cout << "\nWarning: the iterative algorithm has not reached convergence to the desired precision" << std::endl;
 
   std::cout << "\nresult: " << q.transpose() << std::endl;
   std::cout << "\nfinal error: " << err.transpose() << std::endl;
--- a/doc/b-examples/i-inverse-kinematics.md
+++ b/doc/b-examples/i-inverse-kinematics.md
@@ -25,8 +25,11 @@ For a simple manipulator such as the one in this case, it simply corresponds to 
 but using this method generalizes well to more complex kinds of robots, ensuring validity.
 
 Next, we set some computation-related values
-\until DT
-corresponding to the desired position precision (we will see later what it exactly means), a maximum number of iterations (to avoid infinite looping in case the position is not reachable) and a positive "time step" defining the convergence rate.
+\until damp
+corresponding to the desired position precision (we will see later what it exactly means),
+a maximum number of iterations (to avoid infinite looping in case the position is not reachable),
+a positive "time step" defining the convergence rate,
+and a fixed damping factor for the pseudoinversion (see below).
 
 Then, we begin the iterative process.
 At each iteration, we begin by computing the forward kinematics:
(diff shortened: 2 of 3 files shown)


examples/inverse-kinematics.cpp
```diff
@@ -17,11 +17,12 @@ int main(int /* argc */, char ** /* argv */)
...
   const double DT   = 1e-1;
+  const double damp = 1e-6;
 
-  pinocchio::Data::Matrix6x J(6,model.nv); J.setZero();
-  unsigned int svdOptions = Eigen::ComputeThinU | Eigen::ComputeThinV;
-  Eigen::JacobiSVD<pinocchio::Data::Matrix6x> svd(6, model.nv, svdOptions);
+  pinocchio::Data::Matrix6x J(6,model.nv);
+  J.setZero();
 
+  bool success;
```
jcarpent: Init with false is the safest way do to it.

jcarpent:
> I have examined the problem carefully, and it does seem that `Jlog * err = err` and `pinv(Jlog) * err = err`, making the inclusion of `Jlog` useless for this particular case.

I don't understand your reasoning here, because `Jlog * err = err` means nothing in terms of math.
I would vote for keeping the Jlog as the error term takes into account the log operator.
But I confirmed that the influence of it is minor, except in hard cases where it helps a lot.

gabrielebndn:
> > I have examined the problem carefully, and it does seem that `Jlog * err = err` and `pinv(Jlog) * err = err`, making the inclusion of `Jlog` useless for this particular case.
> 
> I don't understand your reasoning here, because `Jlog * err = err` means nothing in terms of math.

That is how @nmansard explained it, and I can confirm from [this script](https://gist.github.com/gabrielebndn/c994080a739bea79cbf33e2f8d88f2f1) that it seems to be true.
Apparently, `Jlog` always has two eigenvalues equal to 1, and `err` belongs to the span of the corresponding eigenvectors.
This equality is certainly made possible by the fact that `Jlog` and `err` are related through `dMi`: `Jlog = pinocchio.Jlog6(dMi)`, `err = pinocchio.log(dMi)`

> I would vote for keeping the Jlog as the error term takes into account the log operator.
> But I confirmed that the influence of it is minor, except in hard cases where it helps a lot.

I thought the whole point of https://github.com/stack-of-tasks/pinocchio/pull/982#discussion_r355190193 was *removing* `Jlog`, you seemed quite convinced it should be removed. This PR was meant to address those comments.
If you agree to keep `Jlog`, then I can simply close this PR and keep things as they were before, since the example was perfectly functional.
Or, I can use `Jlog` *and* the damped pseudo-inverse, but I do not think this would be useful.
You choose.

jmirabel:
> I don't understand your reasoning here, because Jlog * err = err means nothing in terms of math.
I would vote for keeping the Jlog as the error term takes into account the log operator.
But I confirmed that the influence of it is minor, except in hard cases where it helps a lot.

Jlog has an eigen value 1 associated to two eigen vectors. As it is computed, it seems err is in the span of this two eigen vectors. @nmansard and myself could not understand why.

jmirabel:
The result would be different if equation to solve is was `log(M) = log(target)` instead of `log(M) = 0`. In this case however, one should better solve `log(target^-1 * M) = 0`

jcarpent:
> The result would be different if equation to solve is was `log(M) = log(target)` instead of `log(M) = 0`. In this case however, one should better solve `log(target^-1 * M) = 0`

I cannot see it directly. Can you provide more details?

jcarpent:
Anyway, at convergence the Jlog is closed to identity. The influcence of this term is only at the beginning of the iterations.

jcarpent:
> I thought the whole point of [#982 (comment)](https://github.com/stack-of-tasks/pinocchio/pull/982#discussion_r355190193) was _removing_ `Jlog`, you seemed quite convinced it should be removed. This PR was meant to address those comments.
> If you agree to keep `Jlog`, then I can simply close this PR and keep things as they were before, since the example was perfectly functional.
> Or, I can use `Jlog` _and_ the damped pseudo-inverse, but I do not think it would be very useful.
> You choose.

@gabrielebndn The thing is that the motivation of changing for Jlog to Identity are not clear at all. On the contrary, adding the damping is nice for the robustness of course.

gabrielebndn:
> > I thought the whole point of [#982 (comment)](https://github.com/stack-of-tasks/pinocchio/pull/982#discussion_r355190193) was _removing_ `Jlog`, you seemed quite convinced it should be removed. This PR was meant to address those comments.
> > If you agree to keep `Jlog`, then I can simply close this PR and keep things as they were before, since the example was perfectly functional.
> > Or, I can use `Jlog` _and_ the damped pseudo-inverse, but I do not think it would be very useful.
> > You choose.
> 
> @gabrielebndn The thing is that the motivation of changing for Jlog to Identity are not clear at all. On the contrary, adding the damping is nice for the robustness of course.

I would say we took some steps in understanding the problem, especially when `J` is non-singular, but I agree that the root causes are not 100% explained and it is not clear what happens for a singular value of `J`. Anyhow, this is why I am asking. Should I
1. Close this PR
1. Keep the PR as it is, only addressing the other minor comment above
1. Further modify this PR by having both `Jlog` and damping

I can easily follow any option and all of them are fine for me, but the third one will require a little more work, obviously. Really, as you like

jcarpent:
> Jlog has an eigen value 1 associated to two eigen vectors. As it is computed, it seems err is in the span of this two eigen vectors

@jmirabel I think this is a nice property to have the error to be inside the range space of Jlog. This means that in one iteration we should be able to set the error to zero. In this case, we should encourage the use of the the Jlog. What do you think?

gabrielebndn:
> > Jlog has an eigen value 1 associated to two eigen vectors. As it is computed, it seems err is in the span of this two eigen vectors
> 
> @jmirabel I think this is a nice property to have the error to be inside the range space of Jlog. This means that in one iteration we should be able to set the error to zero. In this case, we should encourage the use of the the Jlog. What do you think?

For non-singular `J`, then `Jlog` is irrelevant:
assuming `Jlog * err = err`, then obviously `err = inv(Jlog) * err` (premultiplying both sides by `inv(Jlog)`.
Then, for non-singular `J`, `pinv(Jlog * J) * err = pinv(J) * inv(Jlog) * err = pinv(J) * err`.
For singular `J`, I am not sure of the effects of `Jlog`.
Also, I do not think it is possible in general to converge in one iteration, because we still have the nonlinearity of `J`.

jcarpent:
> Also, I do not think it is possible in general to converge in one iteration, because we still have the nonlinearity of `J`.

Yes, but in the case you're almost linear, this is what you expect when you do gradient descent.

jcarpent:
> For singular `J`, I am not sure of the effects of `Jlog`.

I was indeed interested in this case, which is the more important one. The other was indeed trivial.

jcarpent:
For me, the PR valid, except the initial value for the bool which should be false. @gabrielebndn Can you change it?

For the Jlog, it would be indeed interesting to validate it latter ;)

gabrielebndn:
Ok then. Thanks for the commit!

jmirabel:
Bumping again into this, I got trapped when removing Jlog6 and I guess other people will be.

When doing 6D inverse kinematics, the expression `Jlog * err = err` holds so one can omit Jlog in the derivative expression. **However, this is only true if one do 6D IK**. Omitting Jlog won't work if, for instance, one leave one rotation free. In such case, one need Jlog.

I guess we should put a warning somewhere.

PR #1020, opened 2020-01-21
Improve Python bindings + remove useless files + add macro for ALIGNED_VECTOR

jcarpent:


gabrielebndn:
Why did you delete [unittest/python/test_case.py](https://github.com/stack-of-tasks/pinocchio/blob/cf06746f36b76b5f898918a296249a442f21f12e/unittest/python/test_case.py)? I think this was a mistake.
This file contains a subclass of unittest.TestCase, which is employed in many tests. Now all most Python unit tests do not work any more.

gabrielebndn:
> Why did you delete [unittest/python/test_case.py](https://github.com/stack-of-tasks/pinocchio/blob/cf06746f36b76b5f898918a296249a442f21f12e/unittest/python/test_case.py)? I think this was a mistake.
> This file contains a subclass of `unittest.TestCase`, which is employed in many tests. Now all most Python unit tests do not work any more.

Notably, this subclass contains `assertApprox`. A few observations:
1. The fact that the subclass is itself called `TestCase` is not a good idea. It would be better to call it something else like `PinocchioTestCase`
1. `assertApprox` is nothing else than `assertTrue(isapprox(a, b))`. Probably it could be improved, for instance allowing to specify a tolerance and/or putting a custom error message.
1. if you do not want to keep this class, then you have to modify almost all Python unit tests accordingly. But I would recommend keeping it, as it is handy, especially if point 2 is implemented, probably after implementing point 1

gabrielebndn:
Also notice that if you remove [unittest/python/bindings_model.py](https://github.com/stack-of-tasks/pinocchio/blob/cf06746f36b76b5f898918a296249a442f21f12e/unittest/python/bindings_model.py) (because you move its contents elsewhere) you also need to remove its mention from [unittest/python/CMakeLists.txt](https://github.com/stack-of-tasks/pinocchio/blob/cf06746f36b76b5f898918a296249a442f21f12e/unittest/python/CMakeLists.txt)

gabrielebndn:
> Why did you delete [unittest/python/test_case.py](https://github.com/stack-of-tasks/pinocchio/blob/cf06746f36b76b5f898918a296249a442f21f12e/unittest/python/test_case.py)? I think this was a mistake.
> This file contains a subclass of `unittest.TestCase`, which is employed in many tests. Now all most Python unit tests do not work any more.

To understand how it is used, see for instance https://github.com/stack-of-tasks/pinocchio/blob/cf06746f36b76b5f898918a296249a442f21f12e/unittest/python/bindings_model.py#L5-L7

I agree this is horrible and that is why I recommend switching to the name `PinocchioTestCase` for clarity. Also, I think it should be `from .test_case`rather than `from test_case`

jcarpent:
I've made the change. It should be fine. The thing is that before this force update, everything was working well on my side ;)

gabrielebndn:
I do not understand what you mean by "everything was working well", I do not understand how it is possible. Anyhow, you should also change the name of the class inherited from: e.g. TestModel(TestCase) -> TestModel(PinocchioTestCase)

PR #1037, opened 2020-01-28
Topic/capsule approximation

rstrudel:
This PR contains a script to approximate any mesh by an enclosing capsule with minimal volume

### DIFF
--- a/examples/capsule-approximation.py
+++ b/examples/capsule-approximation.py
@@ -4,7 +4,6 @@ Inspired from Antonio El Khoury PhD: https://tel.archives-ouvertes.fr/file/index
 Section 3.8.1 Computing minimum bounding capsules
 """
 
-
 import numpy as np
 import scipy.optimize as optimize
 
@@ -142,9 +141,15 @@ def parse_urdf(infile, outfile):
     tree.write(outfile)
 
 if __name__ == "__main__":
+    # Example for a single capsule
     #filename = "mesh.obj"
     #mesh_loader = hppfcl.MeshLoader()
     #mesh = mesh_loader.load(filename, np.ones(3))
     #vertices = mesh.vertices()
     #a, b, r = capsule_approximation(vertices)
-    parse_urdf("models/others/robots/ur_description/urdf/ur5_gripper.urdf", "test.urdf")
+    
+    # Example for a whole URDF model
+    # This path refers to Pinocchio source code but you can define your own directory here.
+    pinocchio_model_dir = join(dirname(dirname(str(abspath(__file__)))), "models")
+    urdf_filename = pinocchio_model_dir + "models/others/robots/ur_description/urdf/ur5_gripper.urdf"
+    parse_urdf(urdf_filename, "ur5_gripper_with_capsules.urdf")
--- a/examples/capsule-approximation.py
+++ b/examples/capsule-approximation.py
@@ -7,8 +7,8 @@ Section 3.8.1 Computing minimum bounding capsules
 import numpy as np
 import scipy.optimize as optimize
 
-import hppfcl, eigenpy
-eigenpy.switchToNumpyArray()
+import hppfcl
+hppfcl.switchToNumpyArray()
 
 """
 Capsule definition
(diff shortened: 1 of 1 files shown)


nim65s:
Hi,

Shouldn't this go to hpp-fcl ?

jcarpent:
@nim65s Yes, and no. I think this can be potentially encapsulated in Pinocchio with a dedicated application using it.

jmirabel:
This is an interesting feature. Thank you @rstrudel 

> I think this can be potentially encapsulated in Pinocchio with a dedicated application using it.

I do think as @nim65s : it should go into hpp-fcl. It only depends on hpp-fcl and Pinocchio users will still have access to it.

jcarpent:
The idea would be to have it in C++ directly, not relying on Python. I will write a dedicated small optimization solver for it asap.

Concerning hpp-fcl or Pinocchio, I think this module will have better visibility in Pinocchio, having in mind that we will soon support other modules for collision detection (btCollision for instance)..

So, for genericity and applications, I would prefer to keep this code internal to Pinocchio.

gabrielebndn:
This looks cool!
I would like to point out that this PR only includes an example, so it adds no functionality to Pinocchio.
Also, if you want to give it visibility: examples are _not_ installed, and they are not automatically included in the documentation. Therefore, this example will only be accessible to people who have direct access to the source code (and are curious enough to wander inside the examples).
If you want to make it more visible, I recommend adding a page to the doc which would `\include` the file. This was make the example available inside the doc, and visible online once the doc is uploaded.
If instead you want to add this functionality to Pinocchio, I agree with @jmirabel and @nim65s that it would be more appropriate to implement it inside hpp-fcl

jcarpent:
> I would like to point out that this PR only includes an example, so it adds no functionality to Pinocchio itself. Anybody wanting to use this script should copy and paste it.

This is just a raw tentative @gabrielebndn. Please see my previous comment.

jcarpent:
Other features that will be useful when looking for encapsulating Capsules:
- add the ability to serialize GeometryModel + Pickle in Python
- add the ability to export a new URDF containing the new capsules

jmirabel:
> The idea would be to have it in C++ directly, not relying on Python. I will write a dedicated small optimization solver for it asap.

I am pretty sure Eigen's Levenberg Marquartd would suffice for this.

> having in mind that we will soon support other modules for collision detection (btCollision for instance)..

Any timeline for this ?

jcarpent:
> I am pretty sure Eigen's Levenberg Marquartd would suffice for this.

Not sure. Because you have nonlinear constraint ;)

jmirabel:
Hi @jcarpent 

any advances on this ? I would be interested and could give a hand.

> add the ability to export a new URDF containing the new capsules

In Python, this is very easy to do. I guess boost XML module makes it easy in C++ too.

jcarpent:
Not from our side. I think you can directly populate this PR with the features you want.

examples/capsule-approximation.py
```diff
@@ -68,9 +68,83 @@ def capsule_approximation(vertices):
...
+
+    tree.write(outfile)
+
+if __name__ == "__main__":
+    #filename = "mesh.obj"
+    #mesh_loader = hppfcl.MeshLoader()
+    #mesh = mesh_loader.load(filename, np.ones(3))
+    #vertices = mesh.vertices()
+    #a, b, r = capsule_approximation(vertices)
+    parse_urdf("/home/jmirabel/devel/hpp/src/example-robot-data/robots/ur_description/urdf/ur5_gripper.urdf", "test.urdf")
```
jcarpent: Please provide a universal path here ;)

examples/capsule-approximation.py
```diff
@@ -8,8 +8,8 @@
 import numpy as np
 import scipy.optimize as optimize
 
-import hppfcl
-
+import hppfcl, eigenpy
```
jcarpent: you can directly proceed by:
```
hppfcl.switchToNumpyArray() 
```

examples/capsule-approximation.py
jmirabel: Well, "mesh.obj" wasn't really universal...

jcarpent: Yes, sure.

PR #1041, opened 2020-01-29
Fix bug when compiling with C++11

jcarpent:
Thanks to @aelkazdadi for raising the issue.


nim65s:
I was going to fill a bug for this, as this was preventing me from updating robotpkg's binary releases.
@proyan also got an issue with that.

Thanks for the fix !

jcarpent:
@nim65s This means that you're still compiling using C++11 by default on robotpkg?

nim65s:
Yes, I still don't have a solution to do otherwise.

jcarpent:
One solution is to not have the flag for all the OS first, I would say.

Issue #1026, opened 2020-01-24
Switching to array in Pinocchio


gabrielebndn:
Hi there!
Python2 has reached its end of life, and `numpy.matrix` is now officially deprecated.
As such, there is now a strong push to switch to `numpy.array`. In particular, we are considering this in Wandercraft.
In general, Pinocchio methods handle this through eigenpy, with its methods `switchToNumpyMatrix` and `switchToNumpyArray`.
However, there is a small subset of Pinocchio methods which are purely written in Python, such as those in [utils.py](https://github.com/stack-of-tasks/pinocchio/blob/b9d99528d2bff742d406179e8bf2cc9999f405cd/bindings/python/pinocchio/utils.py
) and [rpy.py](https://github.com/stack-of-tasks/pinocchio/blob/b9d99528d2bff742d406179e8bf2cc9999f405cd/bindings/python/pinocchio/rpy.py).
These methods were written when `matrix` was the standard, and they generally don't support `array` well: either they assume `matrix` as input, or they return `matrix`.
This should be fixed. I opened this issue to highlight the problem and to discuss about how to do it exactly. My suggestion to ensure a smooth transition:
- Modify the methods so that they work for both `matrix` and `array` inputs, analogously to what done in https://github.com/stack-of-tasks/pinocchio/pull/951, using `dot` instead of `*` (which is also an excellent example of bug caused by using `array`)
- Concerning the return type: check `eigenpy.getNumpyType()` and do the same as eigenpy

Alternatively: drop support for matrix (but I feel this is premature).

jcarpent:
Yes, you can proceed to the support of both I think, knowing that some functions are already in C++, it would be worthy to expose them instead of coding them in Python ;)

jcarpent:
@gabrielebndn What is your timeline?

gabrielebndn:
I don't have a set timeline but I am working on it. I will make a PR for the RPY module

jcarpent:
Solved by #1042.

gabrielebndn:
I am sorry, but #1042 di not solve this issue. There are still things to do in the other modules

jcarpent:
#1050 has solved many point of this issue. We can now consider this issue as solved.

gabrielebndn:
@jcarpent if you are convinced that the transition is completed, could you publish a new release, at leat a source release? It would help us switching to `numpy.array` here at Wandercraft

gabrielebndn:
[...]

PR #1042, opened 2020-01-29
Implement rpy in C++

gabrielebndn:
Related to #1026.
This implements `rpyToMatrix` and `matrixToRpy` in C++. They are available as C++ methods in the namespace `rpy`, and they are exposed in replacement to the old Python implementations.
This makes the behavior of the Python functions analogous to most other ones concerning the use of eigenpy.

I kept the old Python implementation for `npToTTuple`, `npToTuple` and `rotate`, for the following reasons:
- They already work well w.r.t. eignepy
- I don't think it makes sense to implement them in C++
- I had problems with creating the submodule `rpy` from C++
- I do not have time to do better than this right now and we want to have something acceptable as soon as possible :)

Concerning `rpyToMatrix`: I was amazed to see that a naive implementation employing the closed-form formula (first implemented in https://github.com/stack-of-tasks/pinocchio/pull/1042/commits/74723582b2b11b56097e5d54283caea1da753517) was four times _slower_ than
```cpp
(Eigen::AngleAxisd(y, Eigen::Vector3d::UnitZ())
              * Eigen::AngleAxisd(p, Eigen::Vector3d::UnitY())
              * Eigen::AngleAxisd(r, Eigen::Vector3d::UnitX())
             ).toRotationMatrix();
````
when compiled in release mode on my laptop.
I had made a little benchmark to test it, but alas, it is on my laptop, which is at home.
Therefore, I kept the latter implementation (see commit https://github.com/stack-of-tasks/pinocchio/pull/1042/commits/b0bdcd526ea44c09d2ac9ea405aca0893a74eeb8)

Concering `matrixToRpy`: eigen offers method `eulerAngles()`, but this is rarely what we want because the range of the output angles. Performance-wise, the timings of my implementation and those of `eulerAngles()` seem to be similar. I did not test it against `unsupported/Eigen/EulerAngles`, but I do not expect great difference (and I would rather avoid using that module).

### DIFF
--- a/bindings/python/math/expose-rpy.cpp
+++ b/bindings/python/math/expose-rpy.cpp
@@ -1,0 +1,59 @@
+//
+// Copyright (c) 2019 CNRS INRIA
+//
+
+#include "pinocchio/bindings/python/fwd.hpp"
+#include <boost/python.hpp>
+//#include "pinocchio/bindings/python/utils/namespace.hpp"
+#include "pinocchio/math/rpy.hpp"
+
+namespace pinocchio
+{
+  namespace python
+  {
+    namespace bp = boost::python;
+
+    Eigen::Matrix3d rpyToMatrix_proxy(const Eigen::Vector3d & rpy)
+    {
+      return pinocchio::rpy::rpyToMatrix(rpy);
+    }
+
+    Eigen::Vector3d matrixToRpy_proxy(const Eigen::Matrix3d & R)
+    {
+      return pinocchio::rpy::matrixToRpy(R);
+    }
+    
+    void exposeRpy()
+    {
+      using namespace Eigen;
+      using namespace pinocchio::rpy;
+
+      {
+        // using the rpy scope
+        //bp::scope current_scope = getOrCreatePythonNamespace("rpy");
+        
+        bp::def("_rpyToMatrix",
+                static_cast<Matrix3d (*)(const double, const double, const double)>(&rpyToMatrix),
+                bp::args("roll", "pitch", "yaw"),
+                "Given (r, p, y), the rotation is given as R = R_z(y)R_y(p)R_x(r),"
+                " where R_a(theta) denotes the rotation of theta degrees axis a");
+
+        bp::def("_rpyToMatrix",
+                &rpyToMatrix_proxy,
+                bp::arg("rpy"),
+                "Given (r, p, y), the rotation is given as R = R_z(y)R_y(p)R_x(r),"
+                " where R_a(theta) denotes the rotation of theta degrees axis a");
+
+        bp::def("_matrixToRpy",
+                &matrixToRpy_proxy,
+                bp::arg("R"),
+                "Given a rotation matrix R, the angles (r, p, y) are given so that R = R_z(y)R_y(p)R_x(r),"
+                " where R_a(theta) denotes the rotation of theta degrees axis a."
+                " The angles are guaranteed to be in the ranges: r in [-pi,pi],"
+                " p in[-pi/2,pi/2], y in [-pi,pi]");
+      }
+      
+    }
+    
+  } // namespace python
+} // namespace pinocchio
--- a/src/math/rpy.hpp
+++ b/src/math/rpy.hpp
@@ -24,7 +24,7 @@ namespace pinocchio
     /// around axis \f$\alpha\f$.
     ///
     template<typename Scalar>
-    Eigen::Matrix<Scalar,3,3> rpyToMatrix(Scalar r, Scalar p, Scalar y)
+    Eigen::Matrix<Scalar,3,3> rpyToMatrix(const Scalar r, const Scalar p, const Scalar y)
     {
       return (Eigen::AngleAxisd(y, Eigen::Vector3d::UnitZ())
               * Eigen::AngleAxisd(p, Eigen::Vector3d::UnitY())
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: src/math/rpy.hpp [hunk 1]
```diff
@@ -0,0 +1,87 @@
...
+  {
+    ///
+    /// \brief Convert from Roll, Pitch, Yaw to transformation Matrix
+    ///
+    /// Given \f$r, p, y\f$, the rotation is given as \f$ R = R_z(y)R_y(p)R_x(r) \f$,
+    /// where \f$R_{\alpha}(\theta)\f$ denotes the rotation of \f$\theta\f$ degrees
+    /// around axis \f$\alpha\f$.
+    ///
+    template<typename Scalar>
+    Eigen::Matrix<Scalar,3,3> rpyToMatrix(const Scalar r, const Scalar p, const Scalar y)
```
jcarpent: Do not use Eigen::AngleAxisd

src/math/rpy.hpp
```diff
@@ -0,0 +1,87 @@
...
+    {
+      return (Eigen::AngleAxisd(y, Eigen::Vector3d::UnitZ())
+              * Eigen::AngleAxisd(p, Eigen::Vector3d::UnitY())
+              * Eigen::AngleAxisd(r, Eigen::Vector3d::UnitX())
+             ).toRotationMatrix();
+    }
+
+    ///
+    /// \brief Convert from Roll, Pitch, Yaw to transformation Matrix
```
jcarpent: Use the term rotation matrix.

src/math/rpy.hpp
```diff
@@ -0,0 +1,87 @@
...
+    Eigen::Matrix<typename Vector3Like::Scalar,3,3,PINOCCHIO_EIGEN_PLAIN_TYPE(Vector3Like)::Options>
+    rpyToMatrix(const Eigen::MatrixBase<Vector3Like> & rpy)
+    {
+      PINOCCHIO_ASSERT_MATRIX_SPECIFIC_SIZE (Vector3Like, rpy, 3, 1);
+
+      return rpyToMatrix(rpy[0], rpy[1], rpy[2]);
+    }
+
+    ///
+    /// \brief Convert from Transformation Matrix to Roll, Pitch, Yaw
```
jcarpent: Same apply here.

bindings/python/math/expose-rpy.cpp
```diff
@@ -0,0 +1,59 @@
...
+    void exposeRpy()
+    {
+      using namespace Eigen;
+      using namespace pinocchio::rpy;
+
+      {
+        // using the rpy scope
+        //bp::scope current_scope = getOrCreatePythonNamespace("rpy");
+        
+        bp::def("_rpyToMatrix",
```
jcarpent: Why did you put some `_` as prefix?

src/math/rpy.hpp
gabrielebndn: O.O You are right!

src/math/rpy.hpp
gabrielebndn: Correct

src/math/rpy.hpp
gabrielebndn: Right

bindings/python/math/expose-rpy.cpp
gabrielebndn: I did not want them to appear in the main `pinocchio` module. I manually import them in the `rpy` submodule

gabrielebndn:
I'll make the requested fixed in a couple of hours

bindings/python/math/expose-rpy.cpp
jcarpent: It changes nothing to directly import them,no?

gabrielebndn: The current solution keeps them hidden in the main Pinocchio module but displays them in the `pinocchio.rpy` submodule instead.
I think I could do `from pin import _rpyToMatrix as rpyToMatrix`, it would be the same.
I had unexpected errors creating the submodule from the C++ bindings.
Also, I wanted to keep the Python implementation for `npToTTuple`, `npToTuple` and `rotate` and I did not how the C++-created module would interact with the Python module.
This is the quickest solution I found. I may look for something else, but I think it is a waste of time, there are more urgent things

gabrielebndn:
Fixed

src/math/rpy.hpp
```diff
@@ -26,14 +26,15 @@ namespace pinocchio
     template<typename Scalar>
     Eigen::Matrix<Scalar,3,3> rpyToMatrix(const Scalar r, const Scalar p, const Scalar y)
     {
-      return (Eigen::AngleAxisd(y, Eigen::Vector3d::UnitZ())
-              * Eigen::AngleAxisd(p, Eigen::Vector3d::UnitY())
-              * Eigen::AngleAxisd(r, Eigen::Vector3d::UnitX())
+      typedef Eigen::AngleAxis<Scalar> AngleAxis;
+      return (AngleAxis(y, Eigen::Vector3d::UnitZ())
```
jcarpent: Same for Eigen::Vector3d, you should use the one from the type related to AngleAxis

bindings/python/math/expose-rpy.cpp
jcarpent: > I had unexpected errors creating the submodule from the C++ bindings.

Not clear. What do you mean here?

src/math/rpy.hpp
gabrielebndn: Damn! I really did this too quickly...

bindings/python/math/expose-rpy.cpp
gabrielebndn: What I mean is: `bp::def("rpyToMatrix")` would expose the function in the main pinocchio namespace. I did not want that.
So I did `bp::def("_rpyToMatrix")` to hide it. Then I imported it as `rpyToMatrix` inside `pinocchio.rpy`.

Also: I had trouble with `getOrCreatePythonNamespace` (duplicate symbols).
Even if I _did_ manage to use `getOrCreatePythonNamespace("rpy")`, it would have conflicted with the existing Python submodule `pinocchio.rpy`, which I wanted to keep in order not to bother with the C++ implementation of `npToTTuple`, `npToTuple` and `rotate`.
Is it clearer now?

src/math/rpy.hpp
gabrielebndn: Damn! I really did this too quickly...

gabrielebndn: Fixed

bindings/python/math/expose-rpy.cpp
jcarpent: OK. I've now understood.

PR #1043, opened 2020-01-30
[python] Fix explog for input arrays

gabrielebndn:
Related to #1026.

This fixes module `explog` when a one-dimensional array is passed as input.
It also fixes `utils.isApprox` so that the following is true
```py
pinocchio.utils.isApprox(np.matrix([[0.], [1.], [2.]]), np.array([0., 1., 2.]))
```

In passing, I corrected the deprecation of `potentialEnergy` and `kineticEnergy`, and I fixed and added to the CI the unit test `bindings_fcl_transform.py`.

There are still a couple of things fo fix before closing #1026, most notably `utils.py` and `romeo_wrapper` (BTW, do we still need it?! Can't we just deprecated it?), plus a generalized double-check will be needed.

### DIFF
--- a/bindings/python/pinocchio/deprecated.py
+++ b/bindings/python/pinocchio/deprecated.py
@@ -129,40 +129,22 @@ buildGeomFromUrdf.__doc__ = (
   pin.buildGeomFromUrdf.__doc__
 ) 
 
-def potentialEnergy(model,data,q,update_kinematics=None):
-  if update_kinematics is None:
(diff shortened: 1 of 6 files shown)


gabrielebndn:
I see you have just exposed `isApprox` as `is_approx`  in https://github.com/stack-of-tasks/eigenpy/pull/148.
Well, the solution in this PR will be temporary then, I mostly needed it for a unit test.

Issue #1044, opened 2020-01-30
Linear Interpolation between two SE3s


ddliugit:
I am working on a problem which needs to insert interpolation poses between initial SE3 and final SE3 with respect to time knots.

The issue is how to interpolate the middle poses for two rotation matrices.

I am wondering if there is a library to do that in Pinocchio? such as spherical linear interpolation (SLERP)?

jcarpent:
You can use directly the following example: https://github.com/stack-of-tasks/pinocchio/blob/master/examples/interpolation-SE3.cpp

ddliugit:
I am using python library. How can I expose this API to python

jmirabel:
As far as I know, it isn't bound.

ddliugit:
how to bind that API  with python?

gabrielebndn:
I confirm it is not bound. There is a specific issue on it, but we closed it because we did not plan to work on it https://github.com/stack-of-tasks/pinocchio/issues/977, it can be found in [the Python bindings project](https://github.com/stack-of-tasks/pinocchio/projects/3).

 @ddliugit if you want to make the bindings you can take inspiration from the existing ones (https://github.com/stack-of-tasks/pinocchio/tree/master/bindings/python). We would all be very grateful if you made this addition to Pinocchio! But I warn you it requires a considerable level of expertise and it is no easy task for a newcomer

ddliugit:
Thanks. 

Converting to python is difficult and I will try to do that if I am available.

jmirabel:
Anyway, @ddliugit for your specific problem, you can workaround the issue by using something like

    M = a * exp6( u * log6(a.inverse()*b))

where a and b are your SE3 objects, and u is in [0,1].

ddliugit:
Thanks!

That's a good idea. I will try that!

rancheng:
I wrote a script to do the interpolation: 

[https://github.com/rancheng/SE3_Pose_Interp](https://github.com/rancheng/SE3_Pose_Interp)

hope it can help!

jmirabel:
I couldn't find the info of what kind of interpolation you do. Slerp between two consecutive transforms ?

ddliu365:
@rancheng thanks for providing this script. I actually used another library ([ndcurves](https://github.com/loco-3d/ndcurves)) for the final implementation. Just post here for your reference.

PR #1046, opened 2020-01-30
Add Interpolate operation in SE3

jcarpent:
Related to some discussions in #1044.


unittest/explog.cpp
```diff
@@ -421,4 +421,25 @@ BOOST_AUTO_TEST_CASE (test_basic)
...
+  
+  A_bis = SE3::Interpolate(A,A,1.);
+  BOOST_CHECK(A_bis.isApprox(A));
+  
+  B_bis = SE3::Interpolate(B,B,1.);
+  BOOST_CHECK(B_bis.isApprox(B));
+  
+  SE3 C = SE3::Interpolate(A,B,0.5);
+  SE3 D = SE3::Interpolate(B,A,0.5);
+  BOOST_CHECK(D.isApprox(C));
```
jmirabel: Should we check the coherence between SpecialEuclideanOperationTpl<3>::interpolate ?

jcarpent: We may do that indeed. I will add it now. Thanks for the suggestion.

PR #1048, opened 2020-01-31
Fix packaging

jcarpent:


gabrielebndn:
May suggest using `test-cpp` instead of `test-c++`? When I wan to manually execute a test, bash gets confued by the + and one is forced to type `test-c\+\+`, which is also annoying if you copy-paste the name. `test-cpp-` is easier to type and friendlier to the developer

PR #1054, opened 2020-02-03
Clean rpy bindings

gabrielebndn:
In #1042, I used an (ugly) workaround in order to make `rpyToMatrix` and `matrixToRpy` appear in the `pinocchio.rpy` submodule. Now, the problem is solved in a much cleaner way.
The `rpy` submodule does not exist anymore as a Python script. It is entirely realized in C++ and its functions are directly exposed.

Among other things, I implemented `rotate` in C++ and I exposed it (though I did not put it in the main C++ interface because I thought it was not needed there).
Also, I moved `npToTTuple` and `npToTuple` to the `utils` submodule and I deprecated them in the `rpy` submodule, because they have nothing to do with RPY!

Technical note: in order to do expose `pinocchio.rpy` as a proper submodule, I had to:
- inline `getOrCreatePythonNamespace` to avoid "duplicate symbol" linker errors
- add the following to the `__init__.py` of the `pinocchio` module:
```
import sys
sys.modules['pinocchio.rpy'] = rpy
sys.modules['pinocchio.cholesky'] = cholesky
```
the reason I had to do it is that, without these lines, `pinocchio.rpy.rpyToMatrix()` would work, `from pinocchio import rpy` would work, but `from pinocchio.rpy import rpyToMatrix` would not, issuing an error saying that module `rpy` does not exist, because basically `rpy` would not be recognized as a module of `pinocchio`, but only of `libpinocchio_pywrap`. As you can see, I did it for cholesky too. I came up with this solution, let me know if you can think of anything better.


gabrielebndn:
A possible caveat is that, when installing by source, the old Python `rpy` module is left there where it was installed. Thus, there may be conflicts, and it would be best to delete that file manually. However, on my system, it seems that the newly-created C++-implemented submodule prevails even if the Python module is still installed.

gabrielebndn:
I have just added two unrelated commits:
- Now `eye`, `zero` and `rand` issue a warning when used with `numpy.array`
- Now compiled C++ examples are prefixed with `example-cpp-` instead of `example-c++-`, analogously to unit tests

bindings/python/math/expose-rpy.cpp
```diff
@@ -22,35 +22,59 @@ namespace pinocchio
     {
       return pinocchio::rpy::matrixToRpy(R);
     }
-    
+
+    Eigen::Matrix3d rotate(const std::string axis, const double ang)
```
jcarpent: Use ref here.

jcarpent: And you can also take a char here, have you tried it?

gabrielebndn: You are right about the ref. A char was my first choice. However, it seems that it was bound to Python strings of any length, which I found fishy and undesirable. Example: `rotate('xwhatever', pi)` would pass. A string allows me to throw an exception if the length is not equal to 1.

gabrielebndn: Thanks for the corrections!

bindings/python/pinocchio/utils.py
```diff
@@ -34,18 +35,24 @@ def eye(n):
...
     else:
+        message = "This function is deprecated for numpy.array. Please use numpy.zeros"
+        _warnings.warn(message, category=DeprecatedWarning, stacklevel=2)
         return np.zeros(n)
 
 def rand(n):
     if pin.getNumpyType()==np.matrix:
         return np.matrix(np.random.rand(n, 1) if isinstance(n, int) else np.random.rand(n[0], n[1]))
     else:
+        "This function is deprecated for numpy.array. Please use numpy.random.rand"
```
jcarpent: A small remark here, I don't think that we should warn here. I mean, we can provide shortcut.

gabrielebndn: I think that in the case of `array` these functions are entirely redundant and in the long run they should be eliminated as the support for matrix is dropped.
However, this is just a personal feeling. If you disagree, it is a single isolated commit, you can certainly revert it if you want

jcarpent: I will revert it. I think it will provide too much pollution.

PR #1063, opened 2020-02-07
[benchmark][frames kinematics] add timings for frame kinematics

proyan:


benchmark/timings.cpp
```diff
@@ -332,6 +335,42 @@ int main(int argc, const char ** argv)
     forwardKinematics(model,data,qs[_smooth],qdots[_smooth], qddots[_smooth]);
   }
   std::cout << "Second Order Kinematics = \t"; timer.toc(std::cout,NBT);
+
+
+  timer.tic();
+  SMOOTH(NBT)
+  {
+    updateFramePlacements(model,data);
```
jcarpent: This test is not relevant, as data keeps to be the same.

benchmark/timings.cpp
```diff
@@ -332,6 +335,42 @@ int main(int argc, const char ** argv)
...
+  }
+  std::cout << "Update All Frames Placement = \t"; timer.toc(std::cout,NBT);
+
+  timer.tic();
+  SMOOTH(NBT)
+  {
+    updateFramePlacement(model,data, fids[_smooth]);
```
jcarpent: This one is not relevant nether, it would be the just the evaluation of the multiplication of two rigid transforms

benchmark/timings.cpp
```diff
@@ -332,6 +335,42 @@ int main(int argc, const char ** argv)
...
+  }
+  std::cout << "Update One Frame Placement = \t"; timer.toc(std::cout,NBT);
+
+  timer.tic();
+  SMOOTH(NBT)
+  {
+    getFrameVelocity(model ,data, fids[_smooth]);
```
jcarpent: Same here.

benchmark/timings.cpp
```diff
@@ -332,6 +335,42 @@ int main(int argc, const char ** argv)
...
+  }
+  std::cout << "Get One Frame Velocity = \t"; timer.toc(std::cout,NBT);
+
+  timer.tic();
+  SMOOTH(NBT)
+  {
+    getFrameAcceleration(model ,data, fids[_smooth]);
```
jcarpent: Same here.

benchmark/timings.cpp
```diff
@@ -332,6 +335,42 @@ int main(int argc, const char ** argv)
...
+  }
+  std::cout << "Update One Frame Acceleration = \t"; timer.toc(std::cout,NBT);
+
+  timer.tic();
+  SMOOTH(NBT)
+  {
+    framesForwardKinematics(model,data, qs[_smooth]);
```
jcarpent: This one can be the only one relevant, I would say.

benchmark/timings.cpp
proyan: I agree, I simply wanted to see the difference between doing it for a frame vs doing it for full tree. I'll remove it.

benchmark/timings.cpp
proyan: It runs on the full tree, and takes about 1us on my machine (compared to 3us for the kinematics).
This is significant. I suggest that we keep it please.

jcarpent: But, the way you did the benchmark is not correct. You have to update the content of data. Please see the other benchmarks please.

proyan: I removed it for now.

jcarpent:
@proyan I did not say you have to remove the bench, but just add a call to forwardKinematics in order to update the content of data. Otherwise, your compiler may shorten the computations.

jcarpent:
@proyan Do you want me to merge or do you want to add the `updateFramePlacements`?

proyan:
> @proyan I did not say you have to remove the bench, but just add a call to forwardKinematics in order to update the content of data. Otherwise, your compiler may shorten the computations.

So I create a benchmark for forwardKinematics+updateFramePlacements?

jcarpent:
Yes. You can.

benchmark/timings.cpp
```diff
@@ -334,13 +334,15 @@ int main(int argc, const char ** argv)
...
   std::cout << "Second Order Kinematics = \t"; timer.toc(std::cout,NBT);
 
-  timer.tic();
+  total = 0.;
   SMOOTH(NBT)
   {
     forwardKinematics(model,data,qs[_smooth]);
+    timer.tic();	  
     updateFramePlacements(model, data);
+    total += timer.toc(timer.DEFAULT_UNIT);
```
proyan: I see what you meant. Thanks.

Issue #1069, opened 2020-02-16
mobilerobot source code unavailable


mohamdev:
Hello, i'm trying to follow the tutorial "Drag and drop (aka inverse kinematics)", but I can't access to the mobilerobot module. In fact, in the tutorial, it is specified that the module is available on this link : https://gepettoweb.laas.fr/doc/stack-of-tasks/pinocchio/master/doxygen-html/mobilerobot_8py_source.html

But when I click on the link I have a 404 error. I guess I can't perform the tutorial witheout this module. Thank you for your feedback :)

jcarpent:
@gabrielebndn I think this issue issue is coming from your side. Could you provide some help please?

gabrielebndn:
@jcarpent he is actually referring to https://gepettoweb.laas.fr/doc/stack-of-tasks/pinocchio/master/doxygen-html/md_doc_d-practical-exercises_3-invkine.html, which is not my work. There is currently some confusion and overlap between the different tutorials and examples etc.

The link is inside the webpage, but it is currently not working. I think the best person to answer this is @nmansard. Either him or @nim65s

jcarpent:
@gabrielebndn I found the bug and I think you did want it, but it seems your made the modifications that lead to the current behavior ;) Please see https://github.com/stack-of-tasks/pinocchio/commit/68bed7b27b95e9e9569474bf02a4b420309cfd8f
I will provide a fix.

gabrielebndn:
Oooh, I see... I will comment on the PR.
Speaking of which, wouldn't it be time to manually update the uploaded doc, at least partially addressing #1007 before an automatic solution is put back in place?

gabrielebndn:
Meanwhile, @mohamdev, while waiting for the fix, all Python sources for the preactical exercises can be found in the Pinocchio repo, in the subfolder [doc/d-practical-exercises/src](https://github.com/stack-of-tasks/pinocchio/tree/e038c7bf283b1df56a35014455e0e2d6f36e03ac/doc/d-practical-exercises/src). You can either copy and paste the files one by one from Github when you need them, or you can clone the repo. This is the most convenient solution, I think: you don't even need to build it, just clone the repo and and then copy the files somewhere else

jcarpent:
Solved by #1071.
@mohamdev We will provide a new release of the doc soon. Thanks a lot for raising this very helpful issue.

mohamdev:
That's fine, thank you very much for your feedback :)

PR #1070, opened 2020-02-16
doc: fix generation of Python files

jcarpent:


gabrielebndn:
Fixes #1069. This reverts a change introduced in the commit 68bed7b27b95e9e9569474bf02a4b420309cfd8f.
The reason I had introduced this change, If I remember correctly, is that the classes and functions defined for the practical exercises where documented inside the main doc and I did not want that.

Could you check whether the classes appear in the doc?

Regardless: instead of deleting the whole block, may I suggest commenting the line? In case it becomes useful again

gabrielebndn:
> Fixes #1069. This reverts a change introduced in the commit [68bed7b](https://github.com/stack-of-tasks/pinocchio/commit/68bed7b27b95e9e9569474bf02a4b420309cfd8f).
> The reason I had introduced this change, If I remember correctly, is that the classes and functions defined for the practical exercises where documented inside the main doc and I did not want that.
> 
> Could you check whether the classes appear in the doc?

I have just compiled the doc with this commit, and this is an example of what I am talking about.
These are all namespaces defined in the practical exercises, and I believe they should not be here:
Screen Shot 2020-02-16 at 23 50 47

Instead, this section should look more like this:
Screen Shot 2020-02-16 at 23 52 03

gabrielebndn:
Hey, in #1071 I provided an alternative solution, which takes my above comments into account. If you agree with me, you may close this PR and accept mine instead

Issue #1076, opened 2020-02-19
[Tests] Missing link to urdfdom


nim65s:
Hi,

https://github.com/stack-of-tasks/pinocchio/blob/7765909f1aed41dd9471baaf8e492be5fcc99523/unittest/multiprecision.cpp#L75

On my system, this code lacks link to urdfdom:
```
/usr/bin/ld: CMakeFiles/test-cpp-multiprecision.dir/multiprecision.cpp.o: undefined reference to symbol '_ZN4urdf13parseURDFFileERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE'
/usr/bin/ld: /usr/lib/liburdfdom_model.so.1.0: error adding symbols: DSO missing from command line
```
(demangled: `urdf::parseURDFFile()`)

This can be fixed there :
https://github.com/stack-of-tasks/pinocchio/blob/7765909f1aed41dd9471baaf8e492be5fcc99523/unittest/CMakeLists.txt#L125

with
```
 # Multiprecision arithmetic
-ADD_PINOCCHIO_UNIT_TEST(multiprecision)
+ADD_PINOCCHIO_UNIT_TEST(multiprecision urdfdom)
 ```

Am I missing anything, or should I just open a PR for this ?

jcarpent:
I've a fix for it.

jcarpent:
Solved by #1077.

Issue #1074, opened 2020-02-19
Substantial increase in memory required to build between v2.2 and v2.3


wxmerkt:
I've noticed this on my notebook with 16 GB before and it's currently failing on the ROS buildfarm (i.e., releases beyond 2.2.1 may not work):
```
virtual memory exhausted: Operation not permitted
```

The changes between 2.2.1 and 2.3.0 increased memory usage substantially. E.g., with 16 GB memory I can only build with a maximum of 2 threads without exhausting the built-in memory. Using 16 threads I needed ~40-50 GB memory (for Crocoddyl).

Do you have any insights what caused the increase in memory requirement to build between 2.2 and 2.3 and is there the possibility to reduce it?

jcarpent:
There is no apparent reason I would say, except for a change in Boost. This is happening for the Python bindings or the unit tests?
One possibility is the serialization of Data and Model, which requires to stear the whole mechanism behind Boost.Variant.

jcarpent:
On your side, can you compile the Python bindings without exposition of Data and Model?

jcarpent:
Or by just commenting the lines:
https://github.com/stack-of-tasks/pinocchio/blob/7765909f1aed41dd9471baaf8e492be5fcc99523/bindings/python/multibody/model.hpp#L331
and:
https://github.com/stack-of-tasks/pinocchio/blob/7765909f1aed41dd9471baaf8e492be5fcc99523/bindings/python/multibody/model.hpp#L328

wxmerkt:
I am using the default CMake settings as if you were to run `mkdir build ; cd build ; cmake .. ; make`. I can compile (on two cores max without excessive swap) on my notebook but I have no possibility to debug the buildfarm directly.

jcarpent:
By the way, the bindings of Crocoddyl are not done correctly. The exposition is not made in individual files, requiring a lot of memory usage.

wxmerkt:
Sorry, I should have been clearer:

a) We see the increase in compiling just Pinocchio with its default CMake settings. This is what causes the ROS buildfarm to fail. It also makes the notebook ache/swap a lot.

b) In Crocoddyl, we see peaks of 40-50GB RAM usage when using 16 threads to compile.

I am mainly interested in solving (a) right now.

jcarpent:
I think this is coming then from the serialization. I may find a trick to avoid this burden in memory.

wxmerkt:
This is an example output in case it helps debugging:
[...]

PR #1077, opened 2020-02-19
Fix multiprecision test

jcarpent:
And only compiles if it is asked.


jcarpent:
Related to #1076 and #1074.

.github/workflows/ccpp.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         export MAKEFLAGS="-j2"
         mkdir build
         cd build
-        cmake .. -DCMAKE_BUILD_TYPE=Debug -DBUILD_WITH_COLLISION_SUPPORT=ON -DPYTHON_EXECUTABLE=$(which python2)
+        cmake .. -DCMAKE_BUILD_TYPE=Debug -DBUILD_WITH_COLLISION_SUPPORT=ON -DPYTHON_EXECUTABLE=$(which python2) -DBUILD_ADVANCED_TESTING=ON
```
wxmerkt: We should also remove the `-DPYTHON_EXECUTABLE` flag from CI to test how most systems will work by default. This will result in the failure reported upstream in jrl cmake modules.

jcarpent: What do you mean? By default, it will choose the Python3 I think or the first one find in the system.

wxmerkt: Yeah and that's an issue that I've opened upstream. Since that change to the jrl cmake modules we have broken code _unless_ people also explicitly specify the Python version to be the system default. This was a change in the behaviour and requires manual user specification (it auto-configured to the system default before!).

jcarpent: You're right. This should only happened for OS with CMake >= 3.12, where Python3 should be the standard. But I can provide a fix to first check for Python2 if nothing is provided. Is the rule you want to get?

jcarpent: I will provide the fix.

wxmerkt: I think I'd like to see something that uses whatever is configured for `python` or the system standard. E.g. Ubuntu 18.04 has CMake >= 3.12 but python default is 2. If someone changes their default it should pick that one up (rather than a switch based on CMake version)

PR #1083, opened 2020-02-21
[Minor] Change default material of collision object.

jmirabel:
The only required element is that alpha which should not be 0.


jcarpent:
Thanks a lot @jmirabel

PR #1085, opened 2020-02-21
[CMake] Fix pc file.

jmirabel:


jmirabel:
Without this change, the PC file would not get the WITH_HPP_FCL and URDF flags.

PR #1095, opened 2020-02-27
Multiple code-gen related fixes.

proyan:
This PR introduces multiple fixes related to code-gen in pinocchio.
Most of the fixes are self-explanatory. The one which might require explanation is 82fac6182ff4367c6fb1ea1fb169f68eddb9df60
When calling `constant_pi<CppAD::...<Scalar> >`, boost calls the getter function as described in this doc https://www.boost.org/doc/libs/1_72_0/libs/math/doc/html/math_toolkit/new_const.html.
With the inheritence that is implemented, the getter returns PI of type double. As a result,  I get an error saying `cannot convert type 'double' into type CppAD::...<Scalar>`

82fac6182ff4367c6fb1ea1fb169f68eddb9df60 overloades the getter to return the correct type.

### DIFF
--- a/src/autodiff/cppad.hpp
+++ b/src/autodiff/cppad.hpp
@@ -28,7 +28,14 @@ namespace boost
       namespace detail
       {
         template<typename Scalar>
-        struct constant_pi< CppAD::AD<Scalar> > : constant_pi<Scalar> {};
+        struct constant_pi< CppAD::AD<Scalar> > : constant_pi<Scalar> {
+          template <int N>
+          static inline CppAD::AD<Scalar> get(const mpl::int_<N>& n)
+          {
+            return CppAD::AD<Scalar>(constant_pi<Scalar>::get(n));
+          }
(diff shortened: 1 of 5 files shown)


src/autodiff/cppad.hpp
```diff
@@ -28,7 +28,14 @@ namespace boost
       namespace detail
       {
         template<typename Scalar>
-        struct constant_pi< CppAD::AD<Scalar> > : constant_pi<Scalar> {};
+        struct constant_pi< CppAD::AD<Scalar> > : constant_pi<Scalar> {
+          template <int N>
+          static inline CppAD::AD<Scalar> get(const mpl::int_<N>& n)
```
jcarpent: Small question: why do you need this?

proyan: When explog is called with the CppAD class, it calls for PI. This introduces an error.

Issue #1089, opened 2020-02-24
Frame tree


jmirabel:
Hi,

I think there is a bug in the URDF parser. Consider the following URDF and python script.
```python
from __future__ import print_function
import pinocchio
pinocchio.switchToNumpyArray()

model = pinocchio.buildModelFromXML("""
<?xml version="1.0" encoding="utf-8"?>
<robot name="test">
  <link name="base_link"/>
  <link name="link_1"/>
  <link name="link_2"/>

  <joint name="joint_1" type="fixed">
    <origin xyz="1 0 0"/>
    <axis xyz="0 0 1"/>
    <!-- <limit lower="0" upper="0.001" effort="100" velocity="0.01"/> -->
    <parent link="base_link"/>
    <child link="link_1"/>
  </joint>
  <joint name="joint_2" type="fixed">
    <origin xyz="0 1 0"/>
    <axis xyz="0 0 1"/>
    <!-- <limit lower="0" upper="0.001" effort="100" velocity="0.01"/> -->
    <parent link="link_1"/>
    <child link="link_2"/>
  </joint>
</robot>
""")

for i, f in enumerate(model.frames):
    print("{}: id {}, parent {}".format(f.name, i, f.previousFrame))
```
It prints
```
universe: id 0, parent 0
root_joint: id 1, parent 0
base_link: id 2, parent 1
joint_1: id 3, parent 1
link_1: id 4, parent 3
joint_2: id 5, parent 3
link_2: id 6, parent 5
```

Clearly, the parent frame of joint_2 should be link_1 and not joint_1. Do we agree ?

jcarpent:
You're right I think too. Indeed, the tree of Frames should keep the URDF tree ordering.

jcarpent:
By the way, the same applies for joint_1 parent should be the base_link too.

jcarpent:
@jmirabel Can you handle the fix?

jmirabel:
I will put that with #1078

jcarpent:
I will then close it. Thanks for handling it.

Issue #1078, opened 2020-02-19
Compiling URDF and SRDF parsers


jcarpent:
Current implementations of URDF and SRDF parsers are header-only, meaning that the private flags related to the various versions of URDFDOM are transferred for correct compilation. In addition, for each new executable, we have to compile the signatures again and again, while by default, URDF and SRDF assumes double scalar type.

I would like to pass these parsers to the existing but empty Pinocchio. This lib will only contain the parsers signatures, nothing more. @jmirabel @wxmerkt @cmastalli @nim65s @nmansard @aelkhour 
What is your opinion on this suggestion?

jmirabel:
I like the idea but I fear it is gonna be hard to achieve. In HPP, we are using a different JointModelCollection. I think this would be an issue, right ?

jcarpent:
No, because the JointCollection is also in the signature of default Model in Pinocchio. 
But, this means that we still need to export the compilations flags.

jcarpent:
Another solution would be to proceed to a cast.

jmirabel:
> No, because the JointCollection is also in the signature of default Model in Pinocchio.

That's the point. We don't use the default Model signature.

jmirabel:
We could also use a visitor design pattern.

jcarpent:
What do you mean?

jmirabel:
Add a class UrdfVisitor that has a prototype such as:
```c++
class UrdfVisitor {
virtual void addFreeFlyer(/* relevant args, excluding the model */) = 0;
virtual void addRevolute(/* relevant args, excluding the model */) = 0;
// and one more for each type of joint

virtual void addLink(/* relevant args, excluding the model */) = 0;
// and any other method that should be exposed.
};
```

Then a child class
```
template <typename Model> ChildClass
{
Model& model;
// implementation of each of the parent virtual methods.
}
```

Then, the URDF parser becomes independant of class Model.

jcarpent:
My initial goal would be to not have to export URDF related flags when compiling Pinocchio. I don't see how it will happen in your case?

jmirabel:
In the library, you will have a code that depends on URDF related flags like this:
```c++
void parse(std::string filename, UrdfVisitor* visitor)
{
UrdfTree urdf (filename);
...
// Add a joint
visitor->addRevolute (position_in_parent, axis, limits...);
}
```
[...]

PR #1097, opened 2020-03-06
[URDF] Move parser into the library.

jmirabel:
Solves #1078 and #1089


jmirabel:
Note that those changes may not be backward compatible. It will break the code of dependencies relying on Pinocchio's preprocessor variable related to URDFDOM.

jmirabel:
e.g. see humanoid-path-planner/hpp-pinocchio@d33104c057aa041efd12042fb7af414243957771

jcarpent:
> Note that those changes are not be backward compatible. It will break the code of dependencies relying on Pinocchio's preprocessor variable related to URDFDOM.

I don't really get the point. Can you explain more?

jmirabel:
`PINOCCHIO_URDF_SHARED_PTR` is not available anymore.

src/parsers/urdf/utils.hpp
```diff
@@ -1,10 +1,10 @@
 //
-// Copyright (c) 2015-2016 CNRS
+// Copyright (c) 2015-2020 CNRS INRIA
 //
 
 #ifndef __pinocchio_parsers_urdf_utils_hpp__
 #define __pinocchio_parsers_urdf_utils_hpp__
 
-#warning "This file is not deprecated and will be removed in future releases. There is no replacement."
+#warning "This file is now deprecated and will be removed in future releases. There is no replacement."
```
jmirabel: Thanks !

jmirabel:
Not related: I suggest we make it clear what should be considered API and what is internal.
- currently, there are the two namespaces *internal* and *details*. One of them is sufficient.
- for C++ macros, we can use a Doxygen macro to create a list of stable macros and a list of internal macros.

jcarpent:
I think internal naming is better.
Do agree for Doxygen too.

PR #1102, opened 2020-03-11
Fix URDF passing

### LINKED ISSUE #1101: Questions on adding fixed model and finding jacobian for contact points
Hi,
First of all, thanks for providing us with this great library!

I'm trying to use this library to model robot dynamics with contacts. I'm having some questions and couldn't find the solution myself.

1. What is the right way to add a static object into the model? Referring to #844 and the provided code: https://github.com/humanoid-path-planner/hpp-pinocchio/blob/77296185e895b9f6e42fa1408b44bbbf89a8cdf3/src/urdf/util.cc#L195, I try to add a static object to the robot model:
```
pinocchio::urdf::buildModel(robot_urdf_file, model);
pinocchio::urdf::buildModel(box_urdf_file, model);
```
where the ground_urdf_file is just a box with no joint.
What I want to achieve is building the robot model, and then adding a box model with a fixed root joint so it is not movable. When I'm compiling this I got the error

>terminate called after throwing an instance of 'std::invalid_argument'
  what():  Fixed joint root_joint could not be added.

2. Is there a straightforward way to get the Jacobian for contact points? I know that I can get the Jacobian of the end-link frame and I can also get the contact point position in world frame from FCL. By finding the contact point position in the end-link frame, I can find the Jacobian of contact point by screw transform. But this feels a little bit complex so I'm curious if there's an existing function or more straightforward way that I didn't find.

jcarpent:
In this PR, I have removed the adding of joints when no root joint is provided. It only appends the inertia of the fixed body to the universe.

Fix #1101.


jcarpent:
In addition, I've fixed a bug related to URDFDOM >= 1.0.0, which need C++11 (for std::shared_ptr) to work properly.

CMakeLists.txt
```diff
@@ -100,6 +100,12 @@ IF(BUILD_WITH_URDF_SUPPORT)
   LIST(APPEND CFLAGS_DEPENDENCIES "-DPINOCCHIO_WITH_URDFDOM")
 ENDIF(BUILD_WITH_URDF_SUPPORT)
 
+IF(${URDFDOM_VERSION} VERSION_GREATER "0.4.2")
+  SET(CMAKE_CXX_STANDARD 11)
```
wxmerkt: Which platforms are still based on urdfdom < 1.0?

jcarpent: at least 16.04

wxmerkt: So it's fair to say that from 18.04 onwards, Pinocchio will essentially be C++11 by default (URDF is on by default I believe).

jcarpent: Yes.

src/multibody/model.hxx
```diff
@@ -251,16 +251,21 @@ namespace pinocchio
   inline int ModelTpl<Scalar,Options,JointCollectionTpl>::
   addFrame(const Frame & frame)
   {
-    if(!existFrame(frame.name, frame.type))
+    // Check if the frame.name exists
+    if(existFrame(frame.name))
+    {
+      FrameIndex frame_id = getFrameId(frame.name);
+      if(frames[frame_id].type == frame.type)
```
jmirabel: Is there any reason to check that the frame type matches and not the other values (placement, parent joint and parent frame, etc) ?

src/parsers/urdf/model.hxx
```diff
@@ -107,9 +107,9 @@ namespace pinocchio
             model.name = name;
           }
 
-          virtual void addRootJoint (const Inertia& Y, const std::string & body_name)
+          virtual void addRootJoint(const Inertia& Y, const std::string & body_name)
           {
-            addFixedJointAndBody(0, SE3::Identity(), "root_joint", Y, body_name);
+            appendBodyToJoint(0,Y,SE3::Identity(),body_name);
```
jmirabel: This is not backward compatible.

src/parsers/urdf/model.hxx
```diff
@@ -218,15 +218,13 @@ namespace pinocchio
           {
             const Frame & frame = model.frames[fid];
             const SE3 & p = frame.placement * placement;
-            if(frame.parent > 0 && Y.mass() > Eigen::NumTraits<Scalar>::epsilon())
-            {
+            if(frame.parent >= 0 && !Y.isZero(Scalar(0)))
```
jmirabel:
```suggestion
            assert(frame.parent >= 0);
            if(!Y.isZero(Scalar(0)))
```

src/parsers/urdf/model.hxx
```diff
@@ -218,15 +218,13 @@ namespace pinocchio
...
               model.appendBodyToJoint(frame.parent, Y, p);
-            }
 
             model.addBodyFrame(body_name, frame.parent, p, (int)fid);
             // Reference to model.frames[fid] can has changed because the vector
             // may have been reallocated.
-            if (model.frames[fid].parent > 0)
+            if (model.frames[fid].parent >= 0)
```
jmirabel:
```suggestion
            assert (model.frames[fid].parent >= 0)
```

src/multibody/model.hxx
jcarpent: Maybe not. I can change it if you think we need to encompass a broader range of behaviors.
I think the type is only informative, no more.

src/parsers/urdf/model.hxx
jcarpent: But this was not the expected behavior and much more a bug I would say.

jcarpent: There was no reason to append a joint here.

jmirabel:
This PR addresses the issue of collisions between root frame name. I see several issues with the proposed changes:
- they are not backward compatible.
- they introduce a behavioral difference between a model built without root joint and one with root joint. This may impact negatively code that tries to be agnostic to this. Typically, I think I will have to update hpp-pinocchio.
- it only solves the issue of name collision when adding a model with no root joint.

I see two options to work around the name collision while preserving backward compatibility:
- pass the name of the root frame and root joint as argument, with default value "root_joint",
- pass an argument that prefixes all the names (joints, frames, geometries).

src/parsers/urdf/model.hxx
jmirabel: I am not against removing this in the future. However, as it is not backward compatible, I would postpone it to next major release. It can be considered both as a bug and a feature. Some people may rely on this (without even realizing it).

Nonetheless, I find it very convenient, when creating a model, to be sure that there is a frame called root_joint, no matter how the model was build. But I can easily modify my code to create a frame before loading the model.

src/multibody/model.hxx
jmirabel: I guess we need to check `frame.placement` and `frame.previousFrame`. If one of this two is different, then probably we aren't doing something correct...

src/parsers/urdf/model.hxx
jcarpent: For me this is a bug, we are forcing the creation of a joint while there is no creation of joint asked by the user.

jcarpent:
I do agree that we should allow the user to choose the root_joint naming.

src/multibody/model.hxx
jcarpent: So calling Frame::operator= should do the job then.

src/parsers/urdf/model.hxx
```diff
@@ -107,9 +107,12 @@ namespace pinocchio
             model.name = name;
           }
 
-          virtual void addRootJoint (const Inertia& Y, const std::string & body_name)
+          virtual void addRootJoint(const Inertia& Y, const std::string & body_name)
           {
             addFixedJointAndBody(0, SE3::Identity(), "root_joint", Y, body_name);
+//            appendBodyToJoint(0,Y,SE3::Identity(),body_name); TODO: change for the correct behavior, see https://github.com/stack-of-tasks/pinocchio/pull/1102 for discussions on the topic
```
jmirabel: Thank you for this.

PR #1113, opened 2020-03-22
[Doc] fix documentation.

jmirabel:


src/algorithm/geometry.hpp
```diff
@@ -91,6 +91,20 @@ namespace pinocchio
...
+  /// \param[in] geom_model: geometry model (const)
+  /// \param[out] geom_data: corresponding geometry data (nonconst) where collisions are computed
+  /// \param[in] stopAtFirstCollision if true, stop the loop on pairs after the first collision.
+  ///
+  /// \warning if stopAtFirstcollision = true, then the collisions vector will
+  /// not be entirely fulfilled (of course).
+  ///
+  inline bool computeCollisions(const GeometryModel & geom_model,
+                                GeometryData & geom_data,
+                                const bool stopAtFirstCollision = true)
```
jmirabel: @jcarpent I noticed that the two prototype of computeCollisions do not use the same default value for *stopAtFirstCollision*. Is this on purpose ?

jcarpent: What is the second prototype?

jcarpent: I see. No, this is not expected.

jcarpent: I think both should be true by default. No?

jmirabel: I would say so. For homogeneity, this should be changed. However, this may be extremely surprising for users and create a *hard to find* bug...

jcarpent: Yes, indeed. I changed it, knowing we can revert it if needed. But the conservatism of default value true will pop up quickly I would say. Or we can use the default to be false, meaning that the behaviour will be correct for all the cases.

jmirabel:
I think I have overwritten your commit by force-pushing to correct a silly error.

src/algorithm/geometry.hpp
jmirabel: Indeed, *false* ensures a correct behavior for all cases. It is safer. People caring for computation time should be more cautious.

PR #1084, opened 2020-02-21
[Python] Remove duplication with hpp-fcl.

jmirabel:
Follows https://github.com/humanoid-path-planner/hpp-fcl/pull/133

Should help reducing compilation time too (e.g. https://github.com/stack-of-tasks/pinocchio/issues/1074)

PR #1114, opened 2020-03-23
[Python] Fix usage of WITH_HPP_FCL_BINDINGS

jmirabel:
otherwise, `verbose = True` generates:
```python
AttributeError: 'module' object has no attribute 'WITH_HPP_FCL_BINDINGS'
```

### DIFF
--- a/bindings/python/pinocchio/__init__.py
+++ b/bindings/python/pinocchio/__init__.py
@@ -3,22 +3,17 @@
 #
 
 import numpy
-from .robot_wrapper import RobotWrapper
-from .libpinocchio_pywrap import __version__, __raw_version__
+from .libpinocchio_pywrap import *
 
 from . import utils
-from . import visualize
 from .explog import exp, log
-from .libpinocchio_pywrap import *
-from .deprecated import *
-from .shortcuts import *
 
 # Manually register submodules
 import sys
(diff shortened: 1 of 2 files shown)


jmirabel:
CI failed to install a dependency because of a public key not available...

gabrielebndn:
I think there is no need to change `pin.WITH_HPP_FCL` is fine, because this value is exposed from C++: 
https://github.com/stack-of-tasks/pinocchio/blob/4bc1b289713c46284e927064c24e845acdf29a62/bindings/python/utils/dependencies.cpp#L16-L21

Concerning `pin.WITH_HPP_FCL_BINDINGS`, you are right, that's a bug. But I think this quantity may have been rendered redundant and equivalent to `pin.WITH_HPP_FCL` from https://github.com/stack-of-tasks/pinocchio/pull/1084. Isn't that so?

jmirabel:
> I think there is no need to change `pin.WITH_HPP_FCL`

There is indeed no need. I did it for homogeneity.

jcarpent:
@jmirabel Apparently, your fix has failed on travis. In fact, I dot know how it can works and do not understand the error you got.

jmirabel:
Again, I forgot that you pushed on my branch and overrode it by forced-pushing. I have re-committed the changes.

PR #1118, opened 2020-03-25
Specialize log for autodiff

proyan:
Currenty Work in progress


proyan:
Go ahead. That is why I pushed the wip version.

proyan:
For now, I have an integrated test in crocoddyl that uses log3 and log6. Both are good according to the test. I'm preparing the unittests for pinocchio.

Meanwhile, could you please have a look at the implementation, and see if you have any comments?

proyan:
Currently, the template arguments `ADMatrix3Like, _Scalar` for log3Algo are redundant. We only need the argument `Scalar`. However, I don't see how to obtain something like `ADMatrix3Like` from the `Scalar` type.
Do you have an insight?

jcarpent:
I will make a pass after your push the tests. It will be easier in this way to check the modif.

PR #1121, opened 2020-03-26
Remove many warnings

gabrielebndn:


gabrielebndn:
This removes lots of warnings, though not all of them.
- ignored some deprecation warnings where nothing else was possible
- Manually define copy constructor and assignment operator for ForceTpl and InertiaTpl (I got `Wdeprecated-copy`)
- Specialize a unit test for joints requiring an axis (I got `Wmaybe-uninitialized`)

gabrielebndn:
I think I should specify that I was compiling for c++11. I think the `Wdeprecated-copy` do not appear for lower versions of c++

jcarpent:
@gabrielebndn Thanks a lot!

PR #1124, opened 2020-03-27
Remove more warnings + fix Python bugs

gabrielebndn:
This removes some more warnings.

Among other things, I removed the usage of `numpy.matrix` from Python unit tests.
A first commit (7518a66 ) replaces `pin.switchToNumpyMatrix` with `pin.switchToNumpyArray`, removes all occurrences of `matrix`, and fixes everything as needed.
Then, a second commit (b208d8b) removes `pin.switchToNumpyArray` too, raising the minimum required eigenpy version to 2. I think it's time to do it, but I don't know whether you like this change, because it basically forces pinocchio users to upgrade eigenpy just so that Pinocchio can save one line in the unit tests, plus I am not sure whether eigenpy2 is already well supported by all systems. If you prefer, this commit can be easily removed.

Also, I set `setGeometryMeshScales` to deprecated without any replacement. It was giving me a warning, I thought I may as well remove it: it had been introduced in #795 to solve the problem of DAE scaling in meshcat, then its use was rejected, but the function had stayed.

While looking for warnings, I discovered some of them were actually pointing to real bugs in the Python bindings, in particular for `addFrame` and probably `appendModel` too.


bindings/python/multibody/model.hpp
```diff
@@ -243,8 +243,7 @@ namespace pinocchio
         
         .def("existFrame",&Model::existFrame,existFrame_overload(bp::args("name","type"),"Returns true if the frame given by its name exists inside the Model with the given type."))
         
-        .def("addFrame",(bool (Model::*)(const std::string &,const JointIndex, const FrameIndex, const SE3 &,const FrameType &)) &Model::addFrame,bp::args("name","parent_id","placement","type"),"Add a frame to the vector of frames. See also Frame for more details. Returns False if the frame already exists.")
```
jcarpent: Smalll question before merging, why did you remove this signature?

jcarpent: I see.

PR #1126, opened 2020-03-27
Improve robustness computations of log3

jcarpent:


src/spatial/log.hxx
```diff
@@ -40,9 +40,9 @@ namespace pinocchio
         // using explicit formula. However, the precision of this method
         // is the square root of the precision with the antisymmetric
         // method (Nominal case).
-        const Scalar cphi = cos(theta - PI_value);
+        const Scalar cphi = -(tr - Scalar(1))/Scalar(2);
```
jmirabel: Mind that *tr* was not threshold-ed, thus *cphi* may have a norm greater than 1. No idea whether this can have an impact.

jcarpent: For these computations, tr is in for sure ]-1,3[

jmirabel: It is far from 3 but close to -1 and it can be slightly below -1 because of numerical imprecision.

jmirabel: If tr < -1, then line 30 sets theta to Pi so the condition of line 37 is true.

PR #1131, opened 2020-03-30
Support Collision pairs even when hpp-fcl is not present

jcarpent:


src/multibody/geometry.hpp
```diff
@@ -223,43 +224,16 @@ namespace pinocchio
...
-    /// 
-    /// The map lists the collision GeometryObjects associated to a given joint Id.
-    ///  Inner objects can be seen as geometry objects that directly move when the associated joint moves
-    std::map < JointIndex, GeomIndexList >  innerObjects;
-
-    /// \brief A list of associated collision GeometryObjects to a given joint Id
-    ///
-    /// Outer objects can be seen as geometry objects that may often be
-    /// obstacles to the Inner objects of given joint
-    std::map < JointIndex, GeomIndexList >  outerObjects;
```
jmirabel: This is used in several places in HPP. I check the code and there are other ways of doing but it will require some time.

Can we make them deprecated instead, with a doxygen deprecation warning such as (for inner objects):

    \deprecated to iterate over the list of inner objects for joint i, use
    \code
    for (GeomIndex j = 0; j < geomModel.ngeoms; ++j {
      if (geomModel.geometryObjects[j].parentJoint != j) continue;
      // Geometry object j is an inner object of joint i
    }
    \endcode

For outer objects, it is a bit more tricky because of the non duplication mentioned in the documentation below.

jcarpent: I will keep it for the moment and will delete it later. Thanks for the feedback.

Issue #1096, opened 2020-03-04
kinetic release of 2.3.1 is failing on 32bit architectures 


tfoote:
http://build.ros.org/job/Ksrc_uX__pinocchio__ubuntu_xenial__source/

Example failing build http://build.ros.org/job/Kbin_uX32__pinocchio__ubuntu_xenial_i386__binary/55/consoleFull

jcarpent:
@tfoote Thanks a lot for this report. We have discussed a bit with @wxmerkt, and it seems that the errors are coming from the build infrastructure, as mentioned by the error:
```
virtual memory exhausted: Operation not permitted
```
@tfoote Do you think we can increase available resources to avoid such issues in the future?

jcarpent:
Seems to be solved by #1141. @tfoote Feel free to reopen this issue if needed.

wxmerkt:
Sadly, it's not:

![image](https://user-images.githubusercontent.com/1664508/78879087-52e38600-7a4b-11ea-95d8-f14e30e60425.png)

Cf. http://build.ros.org/job/Ksrc_uX__pinocchio__ubuntu_xenial__source/ (i386, first in the list)

jcarpent:
This does not correspond to my last PR #1141 where the name of the test has been changed.

wxmerkt:
Yes, sorry I realised this requires a new release. Could you please make a new tag?

jcarpent:
I think everything is now fine. @wxmerkt Could you update https://github.com/ipab-slmc/pinocchio_catkin-release ?

wxmerkt:
@jcarpent, yes as soon as eigenpy is passing on xenial (I updated eigenpy for kinetic last night). Despite the same patches as on bionic, it is failing.

jcarpent:
Could you provide the link for the current status?

wxmerkt:
I fixed it with a manual patch and need to wait for release - the issue was the `CONFIG_INSTALL_DIR` location in `package-config.cmake`. As it differs from the regular workflow, we are patching it during the release.

- Failing build status of dependent package: http://build.ros.org/job/Kbin_uX32__moveit_ros_planning_interface__ubuntu_xenial_i386__binary/41/
- Updated PR with fix: https://github.com/ros/rosdistro/pull/24433

jcarpent:
OK. Thank you very much. Where is the pacth you've provided?

wxmerkt:
[...]

PR #1141, opened 2020-04-08
Split test files and update submodule CMake

jcarpent:


jcarpent:
Fixing #1096.

wxmerkt:
Awesome! Did you check if that reduces the memory to less than 2GB?

jcarpent:
No, but for sure it will reduce the overhead. The other test files contain larger compilations units ;)

wxmerkt:
Ok, just checking if it was fully fixing/closing the mentioned issue :-)

Issue #1146, opened 2020-04-09
Clipping values in randomConfiguration


jcarpent:
Currently, `randomConfiguration(model,lb,ub)` may return a non clipped vector for SO(3) segment of the configuration. One suggestion that emerges is to clip the return values inside the bounds provided by the user. 

@jmirabel and others, what is your feeling?

florent-lamiraux:
What do you mean by clipped ?

jcarpent:
A vector which is outside the bounds for the rotational values.

florent-lamiraux:
You mean outside [1-eps, 1+eps], with eps small ?

jcarpent:
Yes.

florent-lamiraux:
I have never met this problem with eps sufficiently big.

cmastalli:
> Currently, `randomConfiguration(model,lb,ub)` may return a non clipped vector for SO(3) segment of the configuration. One suggestion that emerges is to clip the return values inside the bounds provided by the user.

if you want to do this, then it might be useful to print a warning message.

jmirabel:
If you encounter a problem (which I have never encountered so far), I think a more elegant fix is to make the function `quaternion::uniformRandom` return a unitary quaternion.

I am really surprised that you actually end up with coefficients strictly superior to zero. I looked at the formula and it only multiplies numbers which should all be inferior to ones. I don't think numerical imprecision can lead to a*b > 1 when a <= 1 and b <= 1 (but I may be wrong).

gabrielebndn:
I'm not sure I understand. You mean that `randomConfiguration(model,lb,ub)` does not respect `lb` and `ub` for the quaternion part? Is this what you mean? This is a known issue, I remember talking about to somebody a while ago, @jmirabel I think. The reason why it does it is obvious reading the code: the algorithm employs `quaternion::uniformRandom` to return a random rotation (this also realtes to the above comment by @jmirabel), and simply ignores the bounds. This can be seen here:

https://github.com/stack-of-tasks/pinocchio/blob/9c03aa82d5c27dc69faaa35fe2c717d6a417f4fe/src/multibody/liegroup/special-orthogonal.hpp#L416-L430

and here:

https://github.com/stack-of-tasks/pinocchio/blob/9c03aa82d5c27dc69faaa35fe2c717d6a417f4fe/src/multibody/liegroup/special-euclidean.hpp#L392

[...]

PR #1148, opened 2020-04-10
[test] test quaternion::uniformRandom

jmirabel:
Related to https://github.com/stack-of-tasks/pinocchio/issues/1146

### DIFF
--- a/unittest/quaternion.cpp
+++ b/unittest/quaternion.cpp
@@ -28,5 +28,19 @@ BOOST_AUTO_TEST_CASE(test_assignQuaternion)
   }
 }
 
+BOOST_AUTO_TEST_CASE(test_uniformRandom)
+{
+  srand(0);
+
+  using namespace pinocchio;
+  Eigen::Quaternion<double> q;
+
+  for (int i = 0; i < (1 << 10); ++i) {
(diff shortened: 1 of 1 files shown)


unittest/quaternion.cpp
```diff
@@ -28,5 +28,19 @@ BOOST_AUTO_TEST_CASE(test_assignQuaternion)
...
 }
 
+BOOST_AUTO_TEST_CASE(test_uniformRandom)
+{
+  srand(0);
+
+  using namespace pinocchio;
+  Eigen::Quaternion<double> q;
+
+  for (int i = 0; i < 1 << 10; ++i) {
```
jcarpent, line 38: What is `i < 1 << 10`?

jmirabel: I find it more readable than `1 < 10000000000`.

jmirabel: I will change it to `i < (1 << 10)`. I don't know what operator between `<` and `<<` has precedence.

jcarpent: OK. Thanks for the reply.

jcarpent: Next time, you can simply use 1e10.

Issue #1159, opened 2020-04-17
Cannot turn BUILD_WITH_URDF_SUPPORT to OFF


proyan:
Hi,
URDF_SUPPORT cannot be turned off in the current devel. If I turn it off, either there is a simple cmake error in `CMakeLists.txt:L104`, `URDFDOM_VERSION doesn't exist`, or if I fix it, there are errors in `parsers/urdf/model.cpp`, given by:
```
[ 50%] Building CXX object src/CMakeFiles/pinocchio.dir/parsers/urdf/model.cpp.o
cd /local/rbudhira/devel/build/sot/pinocchio/_build-RELEASE/src && /usr/bin/c++   -DBOOST_MPL_LIMIT_LIST_SIZE=30 -DBOOST_MPL_LIMIT_VECTOR_SIZE=30 -DPINOCCHIO_CPPAD_REQUIRES_MATRIX_BASE_PLUGIN -DPINOCCHIO_WITH_CASADI_SUPPORT -DPINOCCHIO_WITH_CPPADCG_SUPPORT -DPINOCCHIO_WITH_CPPAD_SUPPORT -DPINOCCHIO_WITH_HPP_FCL -Dpinocchio_EXPORTS -I/local/rbudhira/devel/build/sot/pinocchio/_build-RELEASE -I/local/rbudhira/devel/build/sot/pinocchio/_build-RELEASE/include -I/home/rbudhira/devel/src/sot/pinocchio/include -isystem /usr/include/eigen3 -isystem /local/rbudhira/devel/install/sot/include -isystem /opt/ros/kinetic/include -isystem /usr/include/assimp  -pedantic -Wno-long-long -Wall -Wextra -Wcast-align -Wcast-qual -Wformat -Wwrite-strings -Wconversion  -O3 -DNDEBUG -fPIC   -DHPP_FCL_HAVE_OCTOMAP -DFCL_HAVE_OCTOMAP -DOCTOMAP_MAJOR_VERSION=1 -DOCTOMAP_MINOR_VERSION=8 -DOCTOMAP_PATCH_VERSION=1 -o CMakeFiles/pinocchio.dir/parsers/urdf/model.cpp.o -c /home/rbudhira/devel/src/sot/pinocchio/src/parsers/urdf/model.cpp
[...]

PR #1160, opened 2020-04-17
Fix .cpp include if URDF is missing

jcarpent:
Solving #1159


proyan:
https://github.com/stack-of-tasks/pinocchio/blob/master/CMakeLists.txt#L104 should also go inside the IF(BUIILD_URDF_WITH_SUPPORT) conditional :)

proyan:
Thanks!

PR #1168, opened 2020-04-20
Fix memory allocation in Python and improve the documentation

jcarpent:


jcarpent:
Related to https://github.com/stack-of-tasks/eigenpy/issues/156

jcarpent:
I did not handle all the doc yet. This will be done in future work.

Issue #1163, opened 2020-04-18
dintegrate with the assignment operators ADDTO, SETTO, MULWITH


proyan:
I'm planning to create a new `dIntegrate` signature that uses the assignmentoperators `ADDTO`(out += res), `SETTO` (out = res), `MULWITH`(out *= res).

The default setting would be SETTO, so it doesn't effect existing code.

@jcarpent What do you think?

jcarpent:
I'm fine, even if in the future, I think this should be made as a class operator and not through a function call.

Please rename `MULWITH`, with `APPLY_ON_THE_LEFT` and for the same price, it might be also useful to add `APPLY_ON_THE_RIGHT`.

jcarpent:
@proyan When do handle this task? Depending on your answer, it might be requested to perform a new release before this one, independendant from the status of this action.

proyan:
Extending to the multiplication operators might take some time and ingenuity. But what I can do is push the ADDTO and RMTO operators today, and then you could make the release. Is that okay?

jcarpent:
Yep. This is fine for me. Thanks a lot ;)

jcarpent:
Partially solved by #1169. The rest will be handled latter if needed.

PR #1172, opened 2020-04-20
dIntegrateTransport algorithm.

proyan:
This PR adds the signature corresponding to transport of an input matrix along the manifold defined by the dIntegrate, as discussed in https://github.com/stack-of-tasks/pinocchio/issues/1163

### DIFF
--- a/src/algorithm/joint-configuration.hpp
+++ b/src/algorithm/joint-configuration.hpp
@@ -5,9 +5,7 @@
 #ifndef __pinocchio_algorithm_joint_configuration_hpp__
 #define __pinocchio_algorithm_joint_configuration_hpp__
 
-#include "pinocchio/multibody/fwd.hpp"
 #include "pinocchio/multibody/model.hpp"
-
 #include "pinocchio/multibody/liegroup/liegroup.hpp"
 
 namespace pinocchio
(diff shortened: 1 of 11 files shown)


src/multibody/liegroup/liegroup-base.hpp
```diff
@@ -154,6 +154,63 @@ PINOCCHIO_LIE_GROUP_PUBLIC_INTERFACE_GENERIC(Derived,typename)
                        const Eigen::MatrixBase<JacobianOut_t> & J,
                        const AssignmentOperatorType op) const;
 
+
+    /**
+     * @brief   Transport an input matrix to the manifold defined by the dIntegrate computation.
+     *
+     * @details This input and output has to be interpreted in terms of Lie group, not vector space:
+     *          Thus, dIntegrate(q, v, J, arg) creates a manifold manifold M given by a small variation of the configuration vector or the tangent vector into the tangent space at identity.
```
jmirabel, line 162: I find this documentation a bit obscure. `dIntegrate(q, v, J, arg)` does not create a manifold. It creates the tangent space (a vector space) of a manifold at a given configuration. What does this do exactly ?

src/algorithm/joint-configuration.hpp
```diff
@@ -345,7 +343,109 @@ namespace pinocchio
...
     dIntegrate<LieGroupMap,Scalar,Options,JointCollectionTpl,ConfigVectorType,TangentVectorType,JacobianMatrixType>(model, q.derived(), v.derived(), PINOCCHIO_EIGEN_CONST_CAST(JacobianMatrixType,J),arg,op);
   }
-  
+
+  /**
+   *
+   * @brief   Transport a matrix from the terminal to the originate tangent space of the integrate operation.
+   *
+   * @details This function performs the parallel transportation of an input matrix whose columns are expressed in the tangent space of the integrated element \f$ q \oplus v \f$,
+   *          to the tangent space of \f$ q \f$.
```
jmirabel, line 352: The terminology is *the tangent space __at__ \f$ q \f$ __of__ manifold ...*.

May be it would be good to explain this in term of velocity expressed in different frames.

src/multibody/liegroup/liegroup-base.hpp
jcarpent: I've changed and I do agree. The previous documentation was not correct.

src/algorithm/joint-configuration.hpp
jcarpent: Done.

jmirabel: The term *parallel transportation* is a term of differential geometry, which most users won't understand. I have myself only a vague notion of what it is. From the documentation, I am not able to know what the function do.

jcarpent:
Related to #1163.

src/algorithm/joint-configuration.hpp
jcarpent: I've updated the doc. Please provide additional suggestions to make it better.

jmirabel: FYI: doxygen command \copydoc can avoid duplication.

For SE(3), it corresponds to (*something like*) transforming a velocity expressed in q+v into q. As it is, the documentation is already much better.

jcarpent: Yep.

cmastalli:
These functions are cool to have it in Pinocchio :)

jcarpent:
Next stage will be the Python bindings of these new functions.

proyan:
For the pyrene robot, with dimensions 38, and an input matrix of dimension 38X76, the timings on my `Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz` cpu with a `clang++-10` compiler were the following:

```
pin::dIntegrateTransport with aliasing(in us):	1.08329 us
pin::dIntegrateTransport w/o aliasing(in us):	2.53075 us
```

The first bench corresponds to the inPlace version of dIntegrateTransport with aliasing inside the SE3 lie-group. 

The second benchmark corresponds to the non-Inplace version of the dIntegrateTransport, without any aliasing, but with a matrix copy operation inside the vector space.

PR #1176, opened 2020-04-21
fix bug introduced by default scalar type for SE3 in explog

proyan:


jcarpent:
Could you add a test on it for CppAD an Casadi?

src/multibody/liegroup/special-orthogonal.hpp
```diff
@@ -475,7 +475,8 @@ namespace pinocchio
                                      const Eigen::MatrixBase<JacobianIn_t> & Jin,
                                      const Eigen::MatrixBase<JacobianOut_t> & J_out) const
     {
-      typedef typename SE3::Matrix3 Matrix3;
+      typedef typename Config_t::Scalar Scalar;
+      typedef typename SE3Tpl<Scalar,PINOCCHIO_EIGEN_PLAIN_TYPE(Config_t)::Options>::Matrix3 Matrix3;
```
jmirabel, line 479: The class this function belongs to is templatized by a scalar. Why don't we typedef a SE3 at the class level instead of in each function ?

The original typedef came from the non-scalar templetized first version I wrote.

proyan: Makes sense. Thanks for the insight.

proyan:
@jcarpent I can add a few basic tests based on cppad. However, extensive testing, and testing with casadi might take some time.

jcarpent:
OK. So just proceed with CppAD then. We will handle Casadi later

proyan:
I'm adding a difference as well.

unittest/cppadcg-joint-configurations.cpp
```diff
@@ -0,0 +1,77 @@
...
+  addJointAndBody(model,JointModelFreeFlyer(),model.getJointId("universe"),SE3::Identity(),"freeflyer",Inertia::Random());
+  addJointAndBody(model,JointModelSpherical(),model.getJointId("freeflyer_joint"),SE3::Identity(),"spherical",Inertia::Random());
+  addJointAndBody(model,JointModelPlanar(),model.getJointId("spherical_joint"),SE3::Identity(),"planar",Inertia::Random());
+  addJointAndBody(model,JointModelRX(),model.getJointId("planar_joint"),SE3::Identity(),"rx",Inertia::Random());
+  addJointAndBody(model,JointModelPX(),model.getJointId("rx_joint"),SE3::Identity(),"px",Inertia::Random());
+  addJointAndBody(model,JointModelPrismaticUnaligned(SE3::Vector3(1,0,0)),model.getJointId("px_joint"),SE3::Identity(),"pu",Inertia::Random());
+  addJointAndBody(model,JointModelRevoluteUnaligned(SE3::Vector3(0,0,1)),model.getJointId("pu_joint"),SE3::Identity(),"ru",Inertia::Random());
+  addJointAndBody(model,JointModelSphericalZYX(),model.getJointId("ru_joint"),SE3::Identity(),"sphericalZYX",Inertia::Random());
+  addJointAndBody(model,JointModelTranslation(),model.getJointId("sphericalZYX_joint"),SE3::Identity(),"translation",Inertia::Random());
+}
```
jmirabel, line 46: @jcarpent there isn't a sample model that does this ?

proyan:
@jcarpent I've added the difference test. However, it fails because of the if comparison operation here: 
1) https://github.com/stack-of-tasks/pinocchio/blob/master/src/multibody/liegroup/special-orthogonal.hpp#L118
2) https://github.com/stack-of-tasks/pinocchio/blob/master/src/multibody/liegroup/special-orthogonal.hpp#L352
3) https://github.com/stack-of-tasks/pinocchio/blob/master/src/multibody/liegroup/special-euclidean.hpp#L536
4) https://github.com/stack-of-tasks/pinocchio/blob/master/src/multibody/liegroup/special-euclidean.hpp#L224

Do you think you could have a look?

unittest/cppadcg-joint-configurations.cpp
jcarpent: No. We should add it.

jcarpent:
> @jcarpent I've added the difference test. However, it fails because of the if comparison operation here:
> 
>     1. https://github.com/stack-of-tasks/pinocchio/blob/master/src/multibody/liegroup/special-orthogonal.hpp#L118
> 
>     2. https://github.com/stack-of-tasks/pinocchio/blob/master/src/multibody/liegroup/special-orthogonal.hpp#L352
> 
>     3. https://github.com/stack-of-tasks/pinocchio/blob/master/src/multibody/liegroup/special-euclidean.hpp#L536
> 
>     4. https://github.com/stack-of-tasks/pinocchio/blob/master/src/multibody/liegroup/special-euclidean.hpp#L224
> 
> 
> Do you think you could have a look?

I will handle it.

proyan:
Thanks. I'm adding dintegrate test in the meantime.

unittest/cppadcg-joint-configurations.cpp
jcarpent: Solved

proyan:
I've added a somewhat messy implementation of dDifference with the argumenttype enums. The good thing with this is that the API remains consistent. The bad thing is that the implementation is somewhat less comely.
For now I'm finished with the changes.

unittest/liegroups.cpp
```diff
@@ -92,9 +92,9 @@ void test_lie_group_methods (T & jmodel, typename T::JointDataDerived &)
 
   // Check that difference between two equal configuration is exactly 0
   TangentVector_t zero = LieGroupType().difference(q1,q1);
-  BOOST_CHECK_MESSAGE (zero.isZero (0), std::string ("Error: difference between two equal configurations is not 0."));
+  BOOST_CHECK_MESSAGE (zero.isZero(), std::string ("Error: difference between two equal configurations is not 0."));
```
jmirabel, line 95: Isn't this expected to be exactly zero ?

jmirabel: Hum, after double thinking, no, it isn't.

jcarpent:
> I've added a somewhat messy implementation of dDifference with the argumenttype enums. The good thing with this is that the API remains consistent. The bad thing is that the implementation is somewhat less comely.
> For now I'm finished with the changes.

I will change it to make it smoother, but I think we need to rethink it later.

proyan:
Could we please have another release as soon as this is merged?

PR #1177, opened 2020-04-22
Python bindings: add Panda3D based viewer

ikalevatykh:
Simple and efficient cross-platform visualizer based on [panda3d_viewer](https://github.com/ikalevatykh/panda3d_viewer).

This feature requires panda3d_viewer to be installed, this can be done using
`pip install --user panda3d_viewer`

### DIFF
--- a/bindings/python/pinocchio/visualize/panda3d_visualizer.py
+++ b/bindings/python/pinocchio/visualize/panda3d_visualizer.py
@@ -1,4 +1,3 @@
-import multiprocessing
 import numpy as np
 import os
 import sys
--- a/bindings/python/pinocchio/visualize/panda3d_visualizer.py
+++ b/bindings/python/pinocchio/visualize/panda3d_visualizer.py
@@ -16,8 +16,10 @@ __all__ = ['Panda3dVisualizer', 'Panda3dViewer', 'ViewerClosedError']
 
 
 class Panda3dVisualizer(BaseVisualizer):
-    """A Pinocchio display using panda3d engine."""
 
+    """
+    A Pinocchio display using panda3d engine.
+    """
 
     def initViewer(self, viewer=None, load_model=False): # pylint: disable=arguments-differ
         """Init the viewer by attaching to / creating a GUI viewer."""
--- a/bindings/python/pinocchio/visualize/panda3d_visualizer.py
+++ b/bindings/python/pinocchio/visualize/panda3d_visualizer.py
@@ -17,7 +17,7 @@ __all__ = ['Panda3dVisualizer', 'Panda3dViewer', 'ViewerClosedError']
 
 class Panda3dVisualizer(BaseVisualizer):
     """A Pinocchio display using panda3d engine."""
-    
+
 
     def initViewer(self, viewer=None, load_model=False): # pylint: disable=arguments-differ
         """Init the viewer by attaching to / creating a GUI viewer."""
--- a/bindings/python/pinocchio/visualize/panda3d_visualizer.py
+++ b/bindings/python/pinocchio/visualize/panda3d_visualizer.py
@@ -33,14 +33,10 @@ class Panda3dVisualizer(BaseVisualizer):
             self.viewer = Panda3dViewer(window_title="python-pinocchio")
 
         if load_model:
-            self.loadViewerModel()
+            self.loadViewerModel(group_name=self.model.name)
 
-    def loadViewerModel(self, group_name='', color=None):
-        """Create the scene displaying the robot meshes in the viewer."""
-        # generate a unique group name if not specified
-        if not group_name:
-            group_name = str(uuid.uuid4())
-
+    def loadViewerModel(self, group_name, color=None):
+        """Create a group of nodes displaying the robot meshes in the viewer"""
         self.visual_group = group_name + "/visuals"
         self.collision_group = group_name + "/collisions"
 
--- a/bindings/python/pinocchio/visualize/panda3d_visualizer.py
+++ b/bindings/python/pinocchio/visualize/panda3d_visualizer.py
@@ -18,7 +18,7 @@ __all__ = ['Panda3dVisualizer', 'Panda3dViewer', 'ViewerClosedError']
 class Panda3dVisualizer(BaseVisualizer):
     """A Pinocchio display using panda3d engine."""
 
-    def initViewer(self, viewer=None, load_model=False):
+    def initViewer(self, viewer=None, load_model=False): # pylint: disable=arguments-differ
         """Init the viewer by attaching to / creating a GUI viewer."""
         self.visual_group = None
         self.collision_group = None
@@ -32,7 +32,7 @@ class Panda3dVisualizer(BaseVisualizer):
         if load_model:
             self.loadViewerModel(group_name=self.model.name)
 
-    def loadViewerModel(self, group_name, color=None):
+    def loadViewerModel(self, group_name, color=None): # pylint: disable=arguments-differ
         """Create a group of nodes displaying the robot meshes in the viewer."""
         self.visual_group = group_name + "/visuals"
         self.collision_group = group_name + "/collisions"
--- a/bindings/python/pinocchio/visualize/panda3d_visualizer.py
+++ b/bindings/python/pinocchio/visualize/panda3d_visualizer.py
@@ -17,6 +17,7 @@ __all__ = ['Panda3dVisualizer', 'Panda3dViewer', 'ViewerClosedError']
 
 class Panda3dVisualizer(BaseVisualizer):
     """A Pinocchio display using panda3d engine."""
+    
 
     def initViewer(self, viewer=None, load_model=False): # pylint: disable=arguments-differ
         """Init the viewer by attaching to / creating a GUI viewer."""
--- a/bindings/python/pinocchio/visualize/panda3d_visualizer.py
+++ b/bindings/python/pinocchio/visualize/panda3d_visualizer.py
@@ -1,12 +1,9 @@
 import numpy as np
-import os
-import sys
-import uuid
 import warnings
 
-from pinocchio import libpinocchio_pywrap as pin
-from pinocchio.visualize.base_visualizer import BaseVisualizer
-from pinocchio.utils import npToTuple
+from .. import libpinocchio_pywrap as pin
+from ..utils import npToTuple
+from .base_visualizer import BaseVisualizer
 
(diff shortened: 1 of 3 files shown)


cmastalli:
@ikalevatykh thank you for this contribution.

Could you let me know the benefits of this viewer with respect to Gepetto and Meshcat?

Probably you could share a picture of the rendering that you have written in the examples.

jcarpent:
@cmastalli Panda3d is a much more advance viewer than Gepetto-gui and Meshcat. It has been developed for a while by CMU.
Based on it, we can easily install Pinocchio bindings with Panda3d support for tutorials and dissemination.
The next appealing ideas would be to use Pinocchio + Panda3d for fast rendering and so on.

cmastalli:
@jcarpent could we use Panda in Jupyter Notebooks?

ikalevatykh:
Hi @cmastalli, yes you can run panda3d_viewer from Jupyter Notebook. But it always open an external window, so you can't have an interactive 3D scene inside notebook. However you can start panda3d_viewer with a flag window_type='offscreen' (without a visible window, but still need a X-server running on a machine).  Then you can render to a memory buffer and show this images in a notebook with a standart methods. 

You can try in your notebook:

```python
from panda3d_viewer import Viewer

win = Viewer(window_title='jupyter')

win.append_group('root')
win.append_box('root', 'box_node', (1, 1, 1))
```

cmastalli:
@ikalevatykh thank you so much for the info. I will try it soon.

bindings/python/pinocchio/visualize/panda3d_visualizer.py
```diff
@@ -0,0 +1,129 @@
+import multiprocessing
+import numpy as np
+import os
+import sys
+import uuid
+import warnings
+
+from pinocchio import libpinocchio_pywrap as pin
```
gabrielebndn, line 8: You can replace this with
```
from .. import libpinocchio_pywrap as pin
```

bindings/python/pinocchio/visualize/panda3d_visualizer.py
```diff
@@ -0,0 +1,129 @@
...
+from pinocchio.visualize.base_visualizer import BaseVisualizer
```
gabrielebndn, line 9: You can replace this with
```
from . import BaseVisualizer
```

bindings/python/pinocchio/visualize/panda3d_visualizer.py
```diff
@@ -0,0 +1,129 @@
...
+from pinocchio.utils import npToTuple
```
gabrielebndn, line 10: You can replace this with
```
from ..utils import npToTuple
```

gabrielebndn:
Really nice, I left a few minor comments :)

jmirabel:
Thank for this work. It is promising. How mature do think the bindings are ?

@jcarpent I remember discussions a loooong time ago (5 years ?) at LAAS about using panda3d (for Pinocchio and other software of the team). Panda3D was mentioned a year ago in https://github.com/stack-of-tasks/pinocchio/issues/657. Do you remember why we didn't take this step at that time ?

bindings/python/pinocchio/visualize/panda3d_visualizer.py
```diff
@@ -0,0 +1,129 @@
+import multiprocessing
+import numpy as np
+import os
+import sys
+import uuid
+import warnings
```
jmirabel, line 6:
```suggestion
import numpy as np
import os
import sys
import warnings
```

bindings/python/pinocchio/visualize/panda3d_visualizer.py
```diff
@@ -0,0 +1,129 @@
...
+            self.viewer = Panda3dViewer(window_title="python-pinocchio")
+
+        if load_model:
+            self.loadViewerModel()
+
+    def loadViewerModel(self, group_name='', color=None):
+        """Create the scene displaying the robot meshes in the viewer."""
+        # generate a unique group name if not specified
+        if not group_name:
+            group_name = str(uuid.uuid4())
```
jmirabel, line 43: This makes the API of this viewer slightly different from the Gepetto viewer visualizer (I don't know for MeshCat). The original design was that, if you re-run a script, it doesn't duplicate the scene and reuse what has already been done. This random string will break this.

Do you have a strong motivation for this ?

bindings/python/pinocchio/visualize/panda3d_visualizer.py
jmirabel: `uuid` may also be removed (depending on the  other comment).

bindings/python/pinocchio/visualize/panda3d_visualizer.py
ikalevatykh: Hi @jmirabel , thanks for the comments! It is questionnable. My motivation is examples with  several robots, like this one: [examples/panda3d-viewer.py](https://github.com/stack-of-tasks/pinocchio/pull/1177/files#diff-c5a5325b610b9a6df51974e70d99f40c). Imagine if several robots have the same link name (e.g. right_arm_link_0).  In a default logic where all robots are attaching to a common root, a user who don't know about group_name parameter will get strange unobvious bugs: missing links, strange links movements.
I understand the default motivation for a Gepetto visualizer which is a remote and you have to cleanup previous scene each time, but Panda3D visualizer is a local one and starts with a clear scene.
An other option is to make group_name parameter required, to prevent users from getting confused.

jmirabel: I see advantages to both a static default name and a random default one. I would follow those principles:
- the logic when you have several robots should appear where you have several robots.
- randomness prevents to know in advance what the name will be. This might be problematic.
- what happens if you re-create an instance of this viewer class in the same Python interpreter ?

I don't have a strong opinion, but I think the static name is better.

The default value may be attribute name of `Model`. This may be a *false good* (i.e. bad) idea because I don't know how often this is actually filled.

I don't mind it being required.

bindings/python/pinocchio/visualize/panda3d_visualizer.py
ikalevatykh: Hi @gabrielebndn , thanks for the comments. The PEP8 standart says [Absolute imports are recommended, as they are usually more readable and tend to be better behaved...](https://www.python.org/dev/peps/pep-0008/#imports). Do you have a special motivation of using relative imports in this case?

bindings/python/pinocchio/visualize/panda3d_visualizer.py
ikalevatykh: @jmirabel what about using a hash of a model joints to avoid randomness:

```python
if not group_name:
    group_name = "model_%0X" % hash(
        (obj.name for obj in self.visual_model.geometryObjects))
```

It seems a good solution in all cases.

bindings/python/pinocchio/visualize/panda3d_visualizer.py
gabrielebndn: Using the absolute import here may cause weird conflicts if there is more than one version of Pinocchio installed, or if you are doing `make test` (which works on a local copy in the `build` folder, not on the installed version). Therefore, in Pinocchio's Python code we are using explicit relative imports.

PEP8 also says [However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose](https://www.python.org/dev/peps/pep-0008/#imports). I would say this is our case.

I really think relative imports are better in this context, but maybe @nim65s can give us more insight

bindings/python/pinocchio/visualize/panda3d_visualizer.py
jmirabel: :) I had not thought about that.

As it is very hard to predict, I would consider it as random (I agree it is repeatable, so better than random). I think I still prefer the name to be required but I won't fight too much and let others express their opinion.

bindings/python/pinocchio/visualize/panda3d_visualizer.py
ikalevatykh: Ok, switched to relative imports

bindings/python/pinocchio/visualize/panda3d_visualizer.py
ikalevatykh: ok, fixed

bindings/python/pinocchio/visualize/panda3d_visualizer.py
ikalevatykh: OK, the group_name is a required parameter now.

bindings/python/pinocchio/visualize/panda3d_visualizer.py
ikalevatykh: fixed

bindings/python/pinocchio/visualize/panda3d_visualizer.py
ikalevatykh: fixed

bindings/python/pinocchio/visualize/panda3d_visualizer.py
ikalevatykh: fixed

bindings/python/pinocchio/visualize/panda3d_visualizer.py
ikalevatykh: fixed

jcarpent:
> @jcarpent I remember discussions a loooong time ago (5 years ?) at LAAS about using panda3d (for Pinocchio and other software of the team). Panda3D was mentioned a year ago in #657. Do you remember why we didn't take this step at that time ?

I think at this time we did not have the human resources to achieve this task and Meshcat was a good option. But with the time running, it appears that Meshcat still has some flews that it will be hard to handle on our own.
At the same time also, it appears that the @willowsierra teams of Inria are developing research that requires fast and good quality rendering for Deep Learning applications, and @ikalevatykh is in charge of the development of such a feature.
After internal discussions @Inria, we decided to rely on Panda3d for that and it was the occasion to use it for Pinocchio too.

PS: at the same time, I'm trying to push gepetto-gui on Conda. But, I don't think this is a long term option, as Panda3d seems more flexible for education and fast prototyping.

jcarpent:
@jmirabel Do you have other remarks or comments? I will include it inside the new release in such a way people can test it and report bugs if any.

Issue #1181, opened 2020-04-23
isUnitary assert in dDifference requires high precision, does not assert on 0-vector


wxmerkt:
Since 0d0e1e4999335da660de6bf63393448fa4d7302d was merged, we are having the `isUnitary` assert triggered regularly when passing states in where the quaternion is not explicitly normalised right before we pass it to `dDifference`. The off-axis terms are often in the range of 1e-8 - but they still trigger the `isUnitary` check since no precision is specified.

However, it still does not trigger when I pass a full zero-vector in (which also shows as isUnitary 0):
```
dDifference: q1 = 0.00000000000000000000e+00 7.07106781186535915396e-01 0.00000000000000000000e+00 7.07106781186559008034e-01, norm=9.99999999999999777955e-01, unitary? 1
dDifference: q2 = 0.00000000000000000000e+00 0.00000000000000000000e+00 0.00000000000000000000e+00 0.00000000000000000000e+00, norm=0.00000000000000000000e+00, unitary? 0
```
q2 here does not trigger (but should).

When q1 was slightly off (1e-8 terms, when initialised from a quaternion with `0 0.7071068 0 0.7071068`), it triggered.

**In short:**
1) The default precision for `isUnitary` is not set and requires a really high precision, or else it asserts. We could set a precision ourselves.
2) It does not throw on a Zero(7) vector, and I believe it should (?)

jmirabel:
> * It does not throw on a Zero(7) vector, and I believe it should (?)

I think so too. Eigen states that the conversion from quaternion to rotation matrix yield an undefined result if the quaternion is not normalized. I have checked what it returns when given a zero quaterion. 

This can only be fixed by checking the quaternion norm.

jcarpent:
We should maybe provide a isUnitary function dedicated to Quaternions. 
The remarks made by @wxmerkt mean that we are not checking sufficient cases inside unit testing.
It also means that at some point in Croco, re-normalization of the state vector should be enforced to avoid numerical drifts.

jcarpent:
I've implemented a tiny example to check it, but I think the asserts are correct, in the sense that they were not present before, because we were not checking them, this was a mistake from the Pinocchio side.

One solution would be to allow the users to relax this constraint, by defining his/her own tolerance value. @wxmerkt This something you would use?

Here is the script compilation command:
```
g++ $(pkg-config --cflags eigen3) -g  test.cpp -o test
```

[...]

PR #1182, opened 2020-04-24
Add helper for checking Quaternions

jcarpent:
@jmirabel @wxmerkt Related to #1181, I've added some helpers to check whether the quaternions are normalized. In the Lie groups operations, I've set the default value to 1e-8.

I would like to go a bit further and let the final user choose these values by himself/herself thanks to global variables which would be registered inside Singleton. In this way, the user can change this value online, and it will only affect the Debug mode.
What do you think?


unittest/quaternion.cpp
```diff
@@ -42,5 +42,24 @@ BOOST_AUTO_TEST_CASE(test_uniformRandom)
...
+  typedef Quaternion::Coefficients Vector4;
+  
+  for(int i = 0; i < 1e6; ++i)
+  {
+    Quaternion q;
+    q.coeffs() = Vector4::Random() + Vector4::Constant(2);
+    BOOST_CHECK(!quaternion::isNormalized(q));
+    
+    q.normalize();
+    BOOST_CHECK(quaternion::isNormalized(q));
```
wxmerkt, line 60: Looks great - thank you for the quick fix. We could also add an explicit test against Vector4d::Zero()

jmirabel:
> I would like to go a bit further and let the final user choose these values by himself/herself thanks to global variables which would be registered inside Singleton. In this way, the user can change this value online, and it will only affect the Debug mode.
> What do you think?

No sure it is necessary, but if you do it, you may also considered a preprocessor variable.

jcarpent:
At which level do we need a preprocessor variable?

jmirabel:
To set the default precision on the norm of unit norm vectors.

unittest/quaternion.cpp
jcarpent: Done

PR #1183, opened 2020-04-24
Fix panda3d installation and example

cmastalli:
The title is self-explanatory :)

### DIFF
--- a/bindings/python/pinocchio/visualize/panda3d_visualizer.py
+++ b/bindings/python/pinocchio/visualize/panda3d_visualizer.py
@@ -10,11 +10,6 @@ try:
 except ImportError:
     WITH_HPP_FCL_BINDINGS = False
 
-from panda3d_viewer import Viewer as Panda3dViewer, ViewerClosedError
-
-__all__ = ['Panda3dVisualizer', 'Panda3dViewer', 'ViewerClosedError']
-
-
 class Panda3dVisualizer(BaseVisualizer):
     """
     A Pinocchio display using panda3d engine.
@@ -28,6 +23,7 @@ class Panda3dVisualizer(BaseVisualizer):
         self.display_collisions = False
         self.viewer = viewer
 
+        from panda3d_viewer import Viewer as Panda3dViewer
         if viewer is None:
             self.viewer = Panda3dViewer(window_title="python-pinocchio")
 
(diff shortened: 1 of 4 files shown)


examples/panda3d-viewer-play.py
```diff
@@ -27,7 +26,7 @@
 def play_sample_trajectory():
     update_rate = 60
     cycle_time = 3
-    traj = np.repeat(talos.q0, cycle_time * update_rate, axis=1)
+    traj = np.repeat(np.matrix(talos.q0).T, cycle_time * update_rate, axis=1)
```
jcarpent, line 29: I don't think this is a good idea to use matrix here, as it is deprecated.

cmastalli: I agree, the code was designed using NumPy Matrix.

Do you want me to rewrite it using Array?

cmastalli: I have removed `np.matrix`.

examples/panda3d-viewer-play.py
```diff
@@ -26,7 +27,8 @@
 def play_sample_trajectory():
     update_rate = 60
     cycle_time = 3
-    traj = np.repeat(np.matrix(talos.q0).T, cycle_time * update_rate, axis=1)
+    traj = np.repeat(talos.q0.reshape((nq,1)), cycle_time * update_rate, axis=1)
+    print traj.shape
```
wxmerkt, line 31:
```python
from __future__ import print_function

print(traj.shape)
```
=> compatible with both 2 and 3

cmastalli: The printing stuff was for debugging, I removed.

bindings/python/pinocchio/visualize/panda3d_visualizer.py
```diff
@@ -121,3 +117,5 @@ def displayVisuals(self, visibility):
         """Set whether to display visual objects or not."""
         self.viewer.show_group(self.visual_group, visibility)
         self.display_visuals = visibility
+        
+__all__ = ['Panda3dVisualizer']
```
cmastalli, line 121: @jcarpent removing `Panda3dViewer` from `__all__` breaks the examples written in Pinocchio.

I am not familiar with the code, but it seems you need to put back the previous `__all___` definition.

jcarpent: @cmastalli @ikalevatykh The changes are relevant, as panda3d_viewer may not be available. Then, you cannot import the lib, otherwise, you will get an error.

ikalevatykh: @jcarpent Original variant was also OK as panda3d_visualize was not imported in visualize. _import_. Just to clarify

PR #1169, opened 2020-04-20
dIntegrate with ADDTO and RMTO assignment operators.

proyan:
This PR fixes #1163 , for the time being.

Some API stuff in this PR:
1) the `dintegrate(...., ARG, OP)` user API
2) `enum AssignmentOperatorType` is moved from the pinocchio/spatial to pinocchio/fwd (to reside along with `enum ArgumentType`)

PR #1184, opened 2020-04-25
fix API changes in #1169

nim65s:
Hi,

hpp-pinocchio doesn't compile on pinocchio 2.4.3, as `dIntegrate_dq` is called with 3 arguments, while 4 are expected.

@proyan : in #1163 you say "The default setting would be SETTO, so it doesn't affect existing code." and yet in #1169, all the methods get their prototype updated by one parameter addition without defaults.

Meaning the API has changed, doesn't it ?

And the [tests](https://github.com/stack-of-tasks/pinocchio/pull/1169/commits/17897a888a949cffad289803764d1cce0458f1e7#diff-4385ca28d5100ad9c49f0f0a6490ba24) have been updated, also meaning the API has changed. Otherwise, we should have kept the old tests with the old API, and added new tests, with the new API, side by side. Ensuring this wouldn't break existing code elsewhere, while the new features are available and working.

This PR fix this, and, in my case, hpp-pinocchio build.

### DIFF
--- a/bindings/python/algorithm/expose-joints.cpp
+++ b/bindings/python/algorithm/expose-joints.cpp
@@ -30,8 +30,8 @@ namespace pinocchio
       Eigen::MatrixXd J0(Eigen::MatrixXd::Zero(model.nv,model.nv));
       Eigen::MatrixXd J1(Eigen::MatrixXd::Zero(model.nv,model.nv));
 
-      dIntegrate(model,q,v,J0,ARG0,SETTO);
-      dIntegrate(model,q,v,J1,ARG1,SETTO);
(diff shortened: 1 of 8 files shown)


proyan:
I provided a default in the dIntegrate, but assumed that dIntgrate_dq is internal to pinocchio. Apparently that was a mistake

Thanks a lot for the fix Guilhem :)

nim65s:
As far as I know, we don't have any convention saying that one thing is internal or not. Then, everything is the public API.

jmirabel:
> As far as I know, we don't have any convention saying that one thing is internal or not.

What is in `pinocchio::details` or `pinocchio::internal` is private. Everything else is public.

PR #1186, opened 2020-04-25
Support for getting velocities and accelerations

### LINKED ISSUE #1171: Support for getting velocities and accelerations?
Hi,
traditionally, in Pinocchio, we compute `forwardKinematics()` and then we get the results directly from `data.v[i]` or `data.a[i]`. These are expressed in the LOCAL frame, and are defined according to Spatial algebra style.

However, not everybody works this way. Many people prefer to employ the "traditional" velocities, corresponding to `LOCAL_WORLD_ALIGNED`. We can see it here:
#1140
#773
#904

I wonder whether it might be interesting to add some explicit support for common transformations.

Specifically, I am thinking about the following methods:
```
getJointVelocity(model, data, jointId, referenceFrame);
getJointAcceleration(model, data, jointId, referenceFrame);
getClassicJointAcceleration(model, data, jointId, referenceFrame);
```
where the default `referenceFrame` is `LOCAL`, and also modifying
```
getFrameVelocity(model, data, jointId);
getFrameAcceleration(model, data, jointId);
```
to
```
getFrameVelocity(model, data, jointId, referenceFrame);
getFrameAcceleration(model, data, jointId, referenceFrame);
```
and adding
```
getClassicFrameAcceleration(model, data, jointId, referenceFrame);
```
This would create a nice symmetry between joints and frames.
The "classic" acceleration is inspired to an already-existing implementation inside `RobotWrapper`:
https://github.com/stack-of-tasks/pinocchio/blob/cd452efbeb3f04dc980def5fe0fc4a14edc197b7/bindings/python/pinocchio/robot_wrapper.py#L135-139

Is there any interest in this? I may contribute, it will be very quick

gabrielebndn:
Closes #1171.
I haven't really tested the wrappers in RobotWrapper, but there should not be any problem.
Let me know if you have any comments.


gabrielebndn:
There was a small problem in the code. Pretty much everywhere I was prefixing the enum names with `ReferenceFrame::`. This is apparently not strictly legal code and therefore it was not compiling on the CI, although for some reason it was perfectly fine on my computer. I guess it depends on the selected C++ standard. Now I fixed it.

As a result, all commits except the last one do not compile. If you do not like to have non-compiling commits in the commit history, you may squash this PR before accepting it

Issue #1189, opened 2020-04-28
Panda3d viewer is broken in examples


cmastalli:
For more details see https://github.com/stack-of-tasks/pinocchio/pull/1183#pullrequestreview-402233989

gabrielebndn:
The viewer isn't broken. The examples are broken. This is because they try to import `Panda3dViewer` (an alias of `Viewer`) and `ViewerClosedError` from `pinocchio.visualize.panda3d_visualize`, but they are not there any more. I think the changes in #1183 were correct and the solution to fix the examples is simple: just import `Viewer` and `ViewerClosedError` directly from `panda3d_viewer`.

cmastalli:
Thanks for the clarification. Indeed, I wasn't sure about this point.

ikalevatykh:
@cmastalli , it is not a good idea to "fix" working code ;)

cmastalli:
@cmastalli , it is not a good idea to "fix" working code ;)

@ikalevatykh I didn't modify that part of code. I just fixed the installation of the files. You could check the for history for details.

jcarpent:
Solved by #1191.

PR #1191, opened 2020-04-29
Fix of the fix of panda3d_viewer examples

ikalevatykh:
Related to #1189


cmastalli:
Thank you for handling it :smiley:

PR #1194, opened 2020-04-30
Prefer not to use collisionObjects

jmirabel:
After playing around with hpp-fcl and pinocchio, I spent quite some time trying to figure out why collision checking was not working.

It was due to the fact the I modified a collision geometry in the model, with create a new geometry data. Thus the pointer stored in the collision object of geometry data was not pointing to the same object as the collision geometry in geometry model. I don't know if I was clear...

Anyway, the API of hppfcl allow to bypass completely the collision object class. As it avoids duplicating a pointer, I believe it is better not to use it. I don't think removing it completely is a good idea neither. So I open the debate.

### DIFF
--- a/src/algorithm/geometry.hxx
+++ b/src/algorithm/geometry.hxx
@@ -58,13 +58,17 @@ namespace pinocchio
     const CollisionPair & pair = geom_model.collisionPairs[pairId];
 
     PINOCCHIO_CHECK_INPUT_ARGUMENT( pairId      < geom_data.collisionResults.size() );
-    PINOCCHIO_CHECK_INPUT_ARGUMENT( pair.first  < geom_data.collisionObjects.size() );
-    PINOCCHIO_CHECK_INPUT_ARGUMENT( pair.second < geom_data.collisionObjects.size() );
+    PINOCCHIO_CHECK_INPUT_ARGUMENT( pair.first  < geom_model.ngeoms );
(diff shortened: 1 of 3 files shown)


jmirabel:
Ok, I will remove it.

jmirabel:
Do you understand CI failure ? I can't access the logs...

jmirabel:
> As it seems not an output of the geometry algorithms, we can remove it without impacting end-users.

after double thinking, it is the output of the updateGeometryPlacements and we would be breaking the API. I have no idea whether it is actually used... This is the drawback of having only public attributes :)

jcarpent:
Maybe, before merging this PR, we may need to deprecate `collisionObjects` in GeometryData. The good thing is that this quantity has not been exposed in Python yet, making things easy. @jmirabel Can you handle it?

jmirabel:
> Maybe, before merging this PR, we may need to deprecate `collisionObjects` in GeometryData.

I tried to avoid deprecation warning (and think I succeeded) at the cost of having to overwrite the default copy constructor... Deprecating a class attribute is very painful in C++.

jmirabel:
Unrelated questions:
- isn't there a CI duplication between the github CI 16.04 (python 2.7) and one of the travis run (16.04, python 2.7) ?

jcarpent:
The Github CI is in Debug mode with CppAD tests.

Issue #1201, opened 2020-05-05
pinocchio.Data(model) fails with -march=native


cmastalli:
Hi all,

I have spent a couple of hours understanding what it is going, however, I have to admit that I don't have any clue. I have compiled Pinocchio with `march=native` flag, without support of Codegen and Collision to narrow down the issue, and this is the problem:
```python
import pinocchio

model = pinocchio.buildSampleModelHumanoid()
for i in range(100): # I used a for-loop because sometime doesn't fail in the first call
  data = pinocchio.Data(model) # This produces an error
  #data = pinocchio.Model.createData(model) # Instead this works, WHY?
```
Additionally, this is the detailed information error generated by `gdb`:
```bash
python: /usr/local/include/eigen3/Eigen/src/Core/DenseStorage.h:128: Eigen::internal::plain_array<T, Size, MatrixOrArrayOptions, 32>::plain_array() [with T = double; int Size = 36; int MatrixOrArrayOptions = 0]: Assertion `(internal::UIntPtr(eigen_unaligned_array_assert_workaround_gcc47(array)) & (31)) == 0 && "this assertion is explained here: " "http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.html" " **** READ THIS WEB PAGE !!! ****"' failed.

Thread 1 "python" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff7a24801 in __GI_abort () at abort.c:79
#2  0x00007ffff7a1439a in __assert_fail_base (fmt=0x7ffff7b9b7d8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=assertion@entry=0x7fffcacef920 "(internal::UIntPtr(eigen_unaligned_array_assert_workaround_gcc47(array)) & (31)) == 0 && \"this assertion is explained here: \" \"http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.htm"..., file=file@entry=0x7fffcacee9f8 "/usr/local/include/eigen3/Eigen/src/Core/DenseStorage.h", line=line@entry=128, 
[...]

PR #1204, opened 2020-05-05
Stored Data object as shared pointer to fix data alignment #1201

cmastalli:
Hi @jcarpent 

Let me tell you the story behind this. I though that storing the object by value might create issues with data alignment at Boost Python level. As you might know, Eigen instructs us to do not pass objects by value. Then I did a quick research on google and I found the following article: https://stackoverflow.com/questions/13177573/how-to-expose-aligned-class-with-boost-python

With this, I started to believe that makes sense my hypothesis. I decided to quickly implement it and *voila*, it is working.

I hope this solution meets Pinocchio standards.

Carlos

PD: this fixes the issue #1201.

### DIFF
--- a/bindings/python/multibody/data.hpp
+++ b/bindings/python/multibody/data.hpp
@@ -173,7 +173,7 @@ namespace pinocchio
       /* --- Expose --------------------------------------------------------- */
       static void expose()
       {
-        bp::class_<Data>("Data",
+        bp::class_<Data, boost::shared_ptr<Data> >("Data",
                          "Articulated rigid body data related to a Model.\n"
                          "It contains all the data that can be modified by the Pinocchio algorithms.",
                          bp::no_init)


jcarpent:
@cmastalli At most, we are trying to avoid relying on smart pointers for efficiency reasons.
The interpretation you make is not correct:
- Eigen allows the storage by value. But we cannot pass an argument by value to a function, otherwise, we cannot guaranty that the resulting allocated memory will be aligned.
- To have memory aligned enforced with a container, Eigen makes easy the overloading of the new operator
- The solution is that we need to force Python to hold an aligned segment in memory. One solution is to rely on Smart pointer as you did, the other one is to use template specialization.

jmirabel:
> we are trying to avoid relying on smart pointers for efficiency reasons.

Here, would they be bad ?

jmirabel:
At least, they would save your time.

jcarpent:
> At least, they would save your time.

I do agree, but we cannot apply this strategy for all the types in fact. In EigenPy for instance, the same issue will occur for Numpy array which won't be aligned on 32 bits (as suggested by the report) because the default alignment is 16 bits.
I think there is no good solution. Smart ptr will partially fix the issue, not totally. 
I'm afraid at some point we have to comply with these Python constraints. Don't you think so @cmastalli and @jmirabel ?

cmastalli:
> At least, they would save your time.

Going in this direction, I believe that pursuing the template specialization might be a nightmare. It will make the Pinocchio bindings highly depended of the boost version. 

This comment arises after analyzing the solution posted in the article (and precisely explained here: http://fhtagn.net/prog/2015/04/16/quaternion_boost_python.html). I noticed that the code of my boost version looks completely different to the one used in this article. It would safe a lot of time and headaches. And it could be not justifiable to invest our coins on it.

**I. Alternative solution: Eigen::DontAlign in bindings**

The author also proposed, as alternative solution, to enforce `Eigen::DontAlign` for the Python bindings. I initialize thought that it was simple to implement in Pinocchio Data (due to `Option` argument) but I quickly realized that *"it's not that simple*". Furthermore, this solution could create problems in libraries such as Crocoddyl which it can inherit Python code as well.

**II. The solution of ignoring this problem**

This is a plausible solution, I believe, if we are not convinced on any path. This issue can be easily patched by using `pinocchio.Model.createData(model)`. My point is we don't need to enforce the `shared_ptr` solution because we can still patch the code temporally.

Probably @nim65s and @gabrielebndn should give us a third opinion to make a conclusion about it.
@jcarpent thank you for taking this extra time on it. I promise, I won't continue on this topic :)

cmastalli:
> I do agree, but we cannot apply this strategy for all the types in fact.

Do we need to do it?

>  Smart ptr will partially fix the issue, not totally.

Could you elaborate this? Are cases where this solution won't work?
Note that this code works (I assumed that other functions as well, am I correct?):
```python
import pinocchio
import example_robot_data

robot = example_robot_data.loadTalosLegs()
model = robot.model
for i in range(100):
  data = pinocchio.Data(model)

q0 = robot.q0
pinocchio.forwardKinematics(model, data, q0)
```

jcarpent:
Just to be able to track the issue later, this is coming from this line: https://github.com/boostorg/python/blob/1727b67a5a92c8c3564d4ee244fffd758e54d7cc/include/boost/python/object/make_holder.hpp#L93 where the memory pointer is not guarantied to be aligned.

This constructor is in fact called by `boost::python::init<...>` at this level https://github.com/boostorg/python/blob/1727b67a5a92c8c3564d4ee244fffd758e54d7cc/include/boost/python/init.hpp#L310, which is then calling: https://github.com/boostorg/python/blob/1727b67a5a92c8c3564d4ee244fffd758e54d7cc/include/boost/python/detail/make_keyword_range_fn.hpp#L65

jcarpent:
> ```python
> pinocchio.forwardKinematics(model, data, q0)
> ```

This function is calling the constructor of Eigen::VectorXd, which is performing dynamic allocation of an aligned segment. So, you won't get any issue at this level. You will get trouble if you pass for instance a statically allocated object, such Eigen::Vector4d or Eigen::Quaterniond

jcarpent:
Whatever, the main question is: do we really alignment of over 16-bits in Python?  My personal answer is no, and I don't think @cmastalli is really looking to this feature neither. It was a good occasion to understand the mechanism behind Boost.Python alignment issues at least. But personally, I won't push towards this direction. 
We can keep this in mind for future.

jmirabel:
> Just to be able to track the issue later, this is coming from this line: https://github.com/boostorg/python/blob/1727b67a5a92c8c3564d4ee244fffd758e54d7cc/include/boost/python/object/make_holder.hpp#L93 where the memory pointer is not guarantied to be aligned.
> 
> This constructor is in fact called by `boost::python::init<...>` at this level https://github.com/boostorg/python/blob/1727b67a5a92c8c3564d4ee244fffd758e54d7cc/include/boost/python/init.hpp#L310, which is then calling: https://github.com/boostorg/python/blob/1727b67a5a92c8c3564d4ee244fffd758e54d7cc/include/boost/python/detail/make_keyword_range_fn.hpp#L65

boost does not allow to write your own holder type ?

jmirabel:
> Whatever, the main question is: do we really alignment of over 16-bits in Python? My personal answer is no, and I don't think @cmastalli is really looking to this feature neither. It was a good occasion to understand the mechanism behind Boost.Python alignment issues at least. But personally, I won't push towards this direction.
> We can keep this in mind for future.

I don't think it is necessary to have this in the bindings. However, this means that, to have alignment in C++, we should not compile (or just not use) the Python bindings (because of the library, used by both).

cmastalli:
> You will get trouble if you pass for instance a statically allocated object, such Eigen::Vector4d or Eigen::Quaterniond

Do we have this kind of functions? Or could we have it?
My point is Pinocchio data always works with elements that don't need alignment in Python (e.g. configuration point, velocity, accelerations and forces).

I thought that forces might be a problem, I conducted a simple test with this code:
```python
import pinocchio
import example_robot_data

robot = example_robot_data.loadTalosLegs()
model = robot.model
data = pinocchio.Data(model)

q = robot.q0
v = pinocchio.utils.rand(model.nv)
tau = pinocchio.utils.rand(model.nv)
fext = pinocchio.StdVec_Force()
[fext.append(pinocchio.Force.Zero()) for k in range(model.njoints)]
pinocchio.aba(model,data,q,v,tau,fext)
```
which is working.

jcarpent:
this is working because you are not calling the corrupted Python constructor

jcarpent:
> I don't think it is necessary to have this in the bindings. However, this means that, to have alignment in C++, we should not compile (or just not use) the Python bindings (because of the library, used by both).

Thanks @jmirabel for this statement. I suggest closing this issue and to come to it latter if needed.

PR #1207, opened 2020-05-07
[Geometry] collisionRequest -> collisionRequests and distanceRequest -> distanceRequests

jmirabel:
…-> distanceRequests

Together with https://github.com/humanoid-path-planner/hpp-fcl/pull/169, this enables to use the GJK initial guess by default.

### DIFF
--- a/bindings/python/multibody/geometry-data.hpp
+++ b/bindings/python/multibody/geometry-data.hpp
@@ -10,6 +10,7 @@
 
 #include "pinocchio/bindings/python/utils/printable.hpp"
 #include "pinocchio/bindings/python/utils/copyable.hpp"
+#include "pinocchio/bindings/python/utils/deprecation.hpp"
 
 EIGENPY_DEFINE_STRUCT_ALLOCATOR_SPECIALIZATION(pinocchio::GeometryData)
 
@@ -104,11 +105,13 @@ namespace pinocchio
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
         cl
-        .def_readonly("distanceRequest",
-                      &GeometryData::distanceRequest,
+        .add_property("distanceRequest",
+                      bp::make_getter(&GeometryData::distanceRequest,
+                                      deprecated_member<bp::return_internal_reference<> >()),
                       "Deprecated. Use distanceRequests attribute instead.")
-        .def_readonly("collisionRequest",
-                      &GeometryData::collisionRequest,
+        .add_property("collisionRequest",
+                      bp::make_getter(&GeometryData::collisionRequest,
+                                      deprecated_member<bp::return_internal_reference<> >()),
                       "Deprecated. Use collisionRequests attribute instead.")
         ;
 #pragma GCC diagnostic pop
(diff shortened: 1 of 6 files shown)


bindings/python/multibody/geometry-data.hpp
```diff
@@ -100,6 +99,20 @@ namespace pinocchio
              bp::args("self","pair_id"),
              "Deactivate the collsion pair pair_id in geomModel.collisionPairs if it exists.")
         ;
+
+#ifdef PINOCCHIO_WITH_HPP_FCL  
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+        cl
+        .def_readonly("distanceRequest",
```
jcarpent, line 107: This is not the good way to deprecate it in Python. One solution is to make it deprecated officially in the Python bindings using the right flags. Could you do it?

jmirabel: How do you deprecated an attribute in Python ?

jcarpent: I think we should use a wrapper in C++ to send an std::cout with a deprecated message similar to the message sent by the deprecated decorator in Python. 
The best would be to have a generic wrapper. It should not be so hard to do I think.

jcarpent: I think we should use a wrapper in C++ to send an std::cout with a deprecated message similar to the message sent by the deprecated decorator in Python. 
The best would be to have a generic wrapper. It should not be so hard to do I think.

jcarpent: I think we should use a wrapper in C++ to send an std::cout with a deprecated message similar to the message sent by the deprecated decorator in Python. 
The best would be to have a generic wrapper. It should not be so hard to do I think.

jmirabel: After double thinking, it can also be achieved with Python properties.
```python
class Foo:
  def __init__(self):
    self._deprecated_attr = None
  @property
  def attr(self):
    # warn here
    return self._deprecated_attr
  @attr.setter
  def attr(self, value):
    # warn here
    self._deprecated_attr = value
```

jcarpent: OK. Please go ahead.

jcarpent: I found another solution to do it directly in C++ using `default_call_policies` arguments of `boost::python::def`. I think it will make easier the deprecation ;)

jmirabel: Can you be more explicit ? I implemented something with a set of macros. I will push it and let you choose whatever you prefer.

jcarpent: The idea is to overload the pre_call function of the policy and to warn a deprected warning, as done in classic Python. Of course, I can have a look to your solution.

jmirabel: Yours is probably better but don't waste too much time for deprecated stuff.
I pushed it to my branch deprecation_warning. My computer is suffering for re-compiling everything. I prefer not to launch CI for nothing.

jcarpent: I've made a suggestion on top of your commits which seems to be simpler and able to handle both class members and functions.

jmirabel:
Thanks for having handled it.

Issue #1205, opened 2020-05-06
Task space controller robotic arm


grizzi:
Hi,
I have been trying to write a task space controller with no success so far. If it would work, I could add it to the list of examples, since this seems to be missing.

I cannot understand what is going wrong with the code. I have been using `pinocchio` also to simulate forward the robot. Here is the main part of the code (assume variables get initialized where needed):
```
pin::urdf::buildModel(urdf_path, model_, true);
pin::Data data_(model_);

q_ = pinocchio::neutral(model_);
dq_ = Eigen::VectorXd::Zero(model_.nv);
ddq_ = Eigen::VectorXd::Zero(model_.nv);
tau_ = Eigen::VectorXd::Zero(model_.nv);

// desired end effector pose near the pose where the ee gets initialized
pinocchio::SE3 x_des = pinocchio::SE3(x_des_R, x_des_t);

// gains
kp_ = Eigen::Matrix<double, 6, 6>::Identity()*10.0;
kd_ = Eigen::Matrix<double, 6, 6>::Identity()*1.0;

// repeat the following in a loop

// compute task space error
const pinocchio::SE3 dMi = x_des.actInv(data_.oMf[frame_id_]);
Eigen::VectorXd dx = pinocchio::log6(dMi).toVector();
pinocchio::Motion xd = getFrameVelocity(model_, data_, frame_id_);
Eigen::VectorXd dxd = xd.toVector();
xdd_des_ = kp_ * dx - kd_ * dxd;

// update jacobians
pin::computeJointJacobians(model_, data_, q_);       // this also compute the forward kinematics for joints
pin::framesForwardKinematics(model_, data_, q_);     // this update the frames placement
pin::getFrameJacobian(model_, data_, frame_id_, pin::WORLD, J_);
pin::computeJointJacobiansTimeVariation(model_, data_, q_, dq_);
pin::getFrameJacobianTimeVariation(model_, data_, frame_id_, pinocchio::WORLD, dJ_);

// compute feedforward torque
// I wrote the function to compute the pseudoinverse based on SVD 
Eigen::VectorXd qdd_des = math_tools::computePseudoinverse(J_) * (xdd_des_ - dJ_ * dq_);
pin::computeAllTerms(model_, data_, q_, dq_);

// get mass matrix
pin::crba(model_, data_, q_);
data_->M.triangularView<Eigen::StrictlyLower>() = data_->M.transpose().template triangularView<Eigen::StrictlyLower>();
[...]

PR #1220, opened 2020-05-22
Handle LOCAL_WOLRD_ALIGNED in frames algorithms

jcarpent:
Solve #1205. 
@grizzi This PR fixes the issue you mentioned so far. How do you install Pinoccho on your side?


grizzi:
Thanks! Right now I am using it with conda.

PR #1226, opened 2020-05-29
LieGroup Python bindings.

jmirabel:
I open this to start a discussion on the code. At the moment, the intended usage for the bindings is the following:
```python
import pinocchio
R3 = pinocchio.liegroups.R3()
SO3 = pinocchio.liegroups.SO3()
R3xSO3 =R3 * SO3
R3xSO3.difference(...)
```
This is currently independent of the rest of pinocchio.

Here are the points which should be addressed / discussed before merging:
- [x] the code is not tested yet
- [x] if I understand the code correctly, `dIntegrateTransport(Jin, Jout)` is equivalent to `dIntegrate(J); Jout = J * Jin;`. In HPP, we also have the inverse operation and we have equivalent functions for `dDifference`.
- [x] should we add a `LieGroupGeneric::operator==` ?
- [x] (kept for later) maybe we should provide a way of create the lie group for a model.
- [ ] `integrateCoeffWiseJacobian_impl` is not implemented. @jcarpent should it be ?
- [x] most Lie group variant visitors are not declared in `liegroup-variant-visitor.hpp` and ~~not documented~~.

Feel free to edit this list and/or to push to my branch.


jcarpent:
Thanks a lot @jmirabel for this very nice job, not finished yet, but promising.

For the second point, I would say that it would be useful to move all the features on the Parallel Transport on Lie groups inside Pinocchio. In this way, the feature would be complete and used by a broader audience. In terms of usage, your intuition is correct. The idea is to have access to the sparse multiplication in this way.

I think the operator== is always useful, and should be easily implemented thanks to current Variant API, in a similar that works for Joints.

Finally, we definitely need to have the possibility to extract a Cartesian product of the tree of joints for a given model. But, this feature might be implemented later on if needed.

On which part of the code do you need help?

jmirabel:
> On which part of the code do you need help?

The most orthogonal part is probably unit-testing.

jmirabel:
> For the second point, I would say that it would be useful to move all the features on the Parallel Transport on Lie groups inside Pinocchio. In this way, the feature would be complete and used by a broader audience. In terms of usage, your intuition is correct. The idea is to have access to the sparse multiplication in this way.

Does this really correspond to the mathematical meaning of a transport (https://en.wikipedia.org/wiki/Parallel_transport) ?
If no, then, I think the name is a bit misleading.

src/multibody/liegroup/liegroup-variant-visitors.hxx
```diff
@@ -39,6 +39,24 @@ namespace pinocchio
...
     };
+
+    struct LieGroupEqual : public boost::static_visitor<bool>
+    {
+      template <typename T, typename U>
+      bool operator()( const T &, const U & ) const
+      {
+        // Should we handle
+        // - dynamic size vector space versus static size vector space
+        // - dynamic versus static cartesian product
```
jmirabel, line 50: Please, pay attention to this comment.

jmirabel: I think we should handle the first. For the second, I don't think it is useful.

jcarpent: I think everything should be dynamic, as a Model is dynamic by construction.

jcarpent:
> > For the second point, I would say that it would be useful to move all the features on the Parallel Transport on Lie groups inside Pinocchio. In this way, the feature would be complete and used by a broader audience. In terms of usage, your intuition is correct. The idea is to have access to the sparse multiplication in this way.
> 
> Does this really correspond to the mathematical meaning of a transport (https://en.wikipedia.org/wiki/Parallel_transport) ?
> If no, then, I think the name is a bit misleading.

In case of Lie groups, this is fine.

src/multibody/liegroup/liegroup-variant-visitors.hxx
jmirabel: I think it is relevant to have R1, R3, R3xSO3 and R2xSO2 static.

jcarpent: Yes, of course.

jmirabel: So both questions make sense:
- dynamic(R^n, 3) == R^3 ?
- CartesianProductVariant(R^3, SO3) == R3xSO3 ?

jcarpent: I would say yes.

jcarpent: But, for instance, for JointModelVariants, we are doing this, but we should while it is a bit more complex.

jmirabel: I don't know for JointModelVariant but here, I don't think it is very complex.

jcarpent: Yes, I do agree.

jmirabel:
@jcarpent I don't have the rights to stop CI on travis. I think I used to have it...

jcarpent:
me neither.

jmirabel:
How should we call the operations:
- dIntegrate * A: dIntegrateTransport(A)
- A * dIntegrate: ?
- dDifference * A: ?
- A * dDifference: ?

jcarpent:
Maybe, we should add an argument like applyOnTheRight, applyOnTheLeft.

jmirabel:
I have always found it confusing because I never know what is applied on the left or right of what. In HPP, I added [this enum](https://github.com/humanoid-path-planner/hpp-pinocchio/blob/be8fbbb8d66167ff97bd0be9dcddc2b55a25f821/include/hpp/pinocchio/liegroup-space.hh#L69) which is more explicit

jmirabel:
The API in HPP is `space.dIntegrate_d[qv]<DerivativeTimesInput>(...)` (and same for `space.dDifference_dq[01]`).

jcarpent:
Yes, I do agree.
Or something like: `dIntegrateRightAction` and `dIntegrateLeftAction`.

jmirabel:
Another possibility is to do the following.

- Let dIntegrate return a temporary expression object of class `DIntegrateReturnType` that stores all the required information to compute the derivative.
- Add two operators for `matrix * DIntegrateReturnType` and `DIntegrateReturnType * matrix`, that returns a DIntegrateProductReturnType`
- Add method `DIntegrateReturnType::evalTo(matrix_t& result)`
- Add method `DIntegrateProductReturnType::evalTo(matrix_t& result)`

This leads to the following API:
```cpp
lg.dIntegrate_dq(...).evalTo(J);
( lg.dIntegrate_dq(...) * A ).evalTo(J);
( A * lg.dIntegrate_dq(...) ).evalTo(J);
```

jmirabel:
Obviously, it is not as elegant as making `DIntegrateReturnType` and `DIntegrateProductReturnType` Eigen expressions, but it is far much easier (especially when dealing with several versions of Eigen).

jmirabel:
Another notable point: I don't know how to nicely bind this API in Python.

jcarpent:
The resulting struct will need a lot of information to be relevant. I was thinking of such a solution, but due to the variant and so on, I would say this might be very painful. 
Maybe something like dIntegrate(SELF,A) and dIntegrate(A,SELF) introducing the SELF empty struct to mention the position of the operands.

jmirabel:
> The resulting struct will need a lot of information to be relevant.

If you implement it only for CartesianProductOperationVariantTpl, then it is not so verbose. I made a quick prototype. It looks like (with class CartesianProductOperationVariantTpl):
```cpp
    template<typename Config_t, typename Tangent_t>
    struct DIntegrateReturnType {
      CartesianProductOperationVariantTpl const& self;
      const Eigen::MatrixBase<Config_t > & q;
      const Eigen::MatrixBase<Tangent_t> & v;

      DIntegrateReturnType(const CartesianProductOperationVariantTpl& self,
          const Eigen::MatrixBase<Config_t > & q,
          const Eigen::MatrixBase<Tangent_t> & v)
        : self(self), q(q), v(v) {}

      template <typename Derived>
      Derived& evalTo(const Eigen::PlainObjectBase<Derived>& result);

      Eigen::Matrix<Scalar, Eigen::Dynamic, Eigen::Dynamic> eval()
      {
        Eigen::Matrix<Scalar, Eigen::Dynamic, Eigen::Dynamic> result;
        evalTo(result);
        return result;
      }

      template <typename Derived>
      DIntegrateProductReturnType<Config_t, Tangent_t, Derived, true>
      operator* (const Eigen::MatrixBase<Derived>& other);
    };

    template<typename Config_t, typename Tangent_t, typename MatrixType, bool dIntegrateOnLeftSide>
    struct DIntegrateProductReturnType : DIntegrateReturnType<Config_t, Tangent_t> {
      const MatrixType& other;

      DIntegrateProductReturnType(DIntegrateReturnType<Config_t, Tangent_t>& self, const Derived& other)
        : self(self), other(other) {}

      template <typename Derived>
      Derived& evalTo(const Eigen::PlainObjectBase<Derived>& result);

      Eigen::Matrix<Scalar, Eigen::Dynamic, Eigen::Dynamic> eval()
      {
        Eigen::Matrix<Scalar, Eigen::Dynamic, Eigen::Dynamic> result;
        evalTo(result);
        return result;
      }
    };
```
Then outside class CartesianProductVariantTpl, define the operator*(matrix, cartesianproduct) as
```cpp
  template<typename Derived,
    typename _Scalar, int _Options, template<typename,int> class LieGroupCollectionTpl,
    typename Config_t, typename Tangent_t>
  typename CartesianProductOperationVariantTpl<_Scalar,_Options,LieGroupCollectionTpl>::DIntegrateProductReturnType<Config_t, Tangent_t, Derived, false>
  operator* (const Eigen::MatrixBase<Derived>& lop,
             typename CartesianProductOperationVariantTpl<_Scalar,_Options,LieGroupCollectionTpl>::DIntegrateReturnType<Config_t, Tangent_t>& rop)
  {
    return typename CartesianProductOperationVariantTpl<_Scalar,_Options,LieGroupCollectionTpl>::
      DIntegrateProductReturnType<Config_t, Tangent_t, Derived, false>(rop, lop);
  }
```
Then you need to implement a visitor for each cases (as required for any other solution). We can easily make the ReturnType be templated by either *difference* or *integrate*.

This is probably more verbose than the solution you proposed, which would look like
```cpp
lg.dIntegrate_dq(q, v, SELF, A, J);
lg.dIntegrate_dq(q, v, A, SELF, J);
```

I am fine with your solution too.

jcarpent:
We may start with the first solution, producing easy Python bindings. And maybe think to the second one later. What do you think?
In the meantime, it would be nice if people like @proyan or @cmastalli can provide feedback, because, after HPP's developers, they are two more concerned people.

jmirabel:
> We may start with the first solution, producing easy Python bindings. And maybe think to the second one later. What do you think?

By first, you mean the following right ?
```cpp
lg.dIntegrate_dq(q, v, SELF, A, J);
lg.dIntegrate_dq(q, v, A, SELF, J);
```
I am fine with this.

jcarpent:
> > We may start with the first solution, producing easy Python bindings. And maybe think to the second one later. What do you think?
> 
> By first, you mean the following right ?
> 
> ```c++
> lg.dIntegrate_dq(q, v, SELF, A, J);
> lg.dIntegrate_dq(q, v, A, SELF, J);
> ```
> 
> I am fine with this.

Yes, by the simplest solution.

jmirabel:
@jcarpent I implemented what we agreed on for dIntegrate_d[qv]. Could you validate before I do the same for dDifference ?

I am a bit lost in how the argument position should be handled. Sometimes, there is a suffix _dq, sometimes an argument and sometimes a template argument... What is the reason for all those possibilities ?

src/multibody/liegroup/liegroup-variant-visitors.hxx
```diff
@@ -430,6 +430,84 @@ namespace pinocchio
...
+    template<typename LieGroupDerived>
+    static void algo(const LieGroupBase<LieGroupDerived> & lg,
+                     const M1_t& q,
+                     const M2_t& v,
+                     const M3_t& J_in,
+                     M4_t& J_out,
+                     const ArgumentPosition arg,
+                     const AssignmentOperatorType op)
+    {
+      assert((arg==ARG0||arg==ARG1) && "arg should be either ARG0 or ARG1");
```
jcarpent, line 455: Maybe you should better use PINOCCHIO_THROW

jcarpent:
This is fine for me. Just a minor remark. Thanks a lot !

jmirabel:
I think I am done now.

bindings/python/multibody/liegroups.hpp
```diff
@@ -0,0 +1,225 @@
...
+public:
+  
+  /* --- Exposing C++ API to python through the handler ----------------- */
+  template<class PyClass>
+  void visit(PyClass& cl) const
+  {
+    typedef Eigen::Matrix<double, Eigen::Dynamic, 1> ConfigVector_t;
+
+    typedef LieGroupWrapperTpl<LieGroupType> LieGroupWrapper;
+    cl
```
jcarpent, line 180: Maybe it would be nice to fill the doc here.

bindings/python/multibody/expose-liegroups.cpp
```diff
@@ -0,0 +1,62 @@
...
+{
+  LieGroupPythonVisitor<
+    CartesianProductOperationVariantTpl<double,0,LieGroupCollectionDefaultTpl>
+    >::expose("LieGroup");
+
+  bp::scope current;
+  std::string submoduleName(bp::extract<const char*>(current.attr("__name__")));
+  submoduleName.append(".liegroups");
+
+  // Create the submodule, and attach it to the current scope.
```
jcarpent, line 43: In the utils folder, there are some tools to do that automatically.

src/multibody/liegroup/liegroup-base.hxx
```diff
@@ -400,12 +530,70 @@ namespace pinocchio {
...
+      bool dIntegrateOnTheLeft,
+      const ArgumentPosition arg,
+      const AssignmentOperatorType op) const
+  {
+    Index nv_ (nv());
+    JacobianMatrix_t J (nv_, nv_);
+    dIntegrate(q, v, J, arg);
+    switch (op) {
+      case SETTO:
+        if(dIntegrateOnTheLeft) Jout = J * Jin;
```
jcarpent, line 555: Maybe, we should use the noalias() here. No?

jmirabel: This is a bit hazardous since Jout and Jin might be the same matrix. Another option would be to allow Jout to be either and Eigen::MatrixBase or a NoAlias so that users can call `dIntegrate(... Jout.noalias());`

bindings/python/multibody/expose-liegroups.cpp
jmirabel: Can you be more precise ?

jmirabel:
I don't get CI failure. It seems the failure is unrelated (or something like exhausted memory or time). I tried to re-run the jobs but I don't see them being effectively re-run...

src/multibody/liegroup/liegroup-base.hxx
jmirabel: Moreover, here, we are dealing with fairly small matrices so `noalias` won't make a big difference. I think we would gain more by specializing this function for SpecialEuclidean and CartesianProduct. I prefer to keep this for later.

bindings/python/multibody/expose-liegroups.cpp
jcarpent: See https://github.com/stack-of-tasks/pinocchio/blob/a179f5619a24000d71d748243b5068605cce4453/bindings/python/utils/namespace.hpp#L23-L33

jcarpent:
Can you rebase?

jcarpent:
@jmirabel I think this PR is ready for merging, isn't it?

jmirabel:
From my side, this is ok as a first implementation. Merging will allow people to give feedback.

jcarpent:
I will make a new release in the following hours.

PR #1234, opened 2020-06-10
Add codegen/autodiff support for eigen coeff-wise min and max operations

proyan:
This PR creates codegen/autodiff support for the Eigen `min`/`max` operations.

For that purpose, the min/max specialization needs to be loaded before Eigen/Core is loaded. Else, the eigen implementation defaults to `std::max/min`.

1. I had to separate the traits from the `fwd.hpp` so it could be loaded before `Eigen/Core`
2. The `minmax.hpp` file has to be loaded before Eigen/Core is loaded. This means loading it before `math/fwd.hpp`.
3. Since `minmax.hpp` needs to be included before `Eigen`, I had to redefine the eigen inline macros.
4. I had to refactor the header `math/fwd.hpp`, since it was loading `Eigen/Core` module, and `pinocchio/math/fwd.hpp` did not require `Eigen`. With this PR, Eigen/Core is not loaded with `pinocchio/math/fwd.hpp`.
5. I've added a unittest to check the min/max support

Point 3 might affect downstream repositories, if they only include `math/fwd.hpp` and not `fwd.hpp`.  

1. Any downstream user that is doing that would need to include `fwd.hpp` as well.
2. On the other hand, this refactoring is needed for inequality cost support in `crocoddyl`.

### DIFF
--- a/src/autodiff/cppad.hpp
+++ b/src/autodiff/cppad.hpp
@@ -61,6 +61,27 @@ namespace Eigen
       }
     };
   }
+  
+  namespace numext
+  {
+    template<typename Scalar>
+    EIGEN_DEVICE_FUNC
+    EIGEN_ALWAYS_INLINE CppAD::AD<Scalar> mini(const CppAD::AD<Scalar>& x, const CppAD::AD<Scalar>& y)
+    {
+      using ::pinocchio::internal::if_then_else;
+      using ::pinocchio::internal::LT;
+      return if_then_else(LT, y, x, y, x);
+    }
+
+    template<typename Scalar>
+    EIGEN_DEVICE_FUNC
+    EIGEN_ALWAYS_INLINE CppAD::AD<Scalar> maxi(const CppAD::AD<Scalar>& x, const CppAD::AD<Scalar>& y)
+    {
+      using ::pinocchio::internal::if_then_else;
+      using ::pinocchio::internal::LT;
+      return if_then_else(LT, x, y, y, x);
+    }
+  }
 } //namespace Eigen
 
 // Source from #include <cppad/example/cppad_eigen.hpp>
--- a/src/autodiff/cppad.hpp
+++ b/src/autodiff/cppad.hpp
@@ -5,6 +5,7 @@
 #ifndef __pinocchio_autodiff_ccpad_hpp__
 #define __pinocchio_autodiff_ccpad_hpp__
 
+#include "pinocchio/container/boost-container-limits.hpp"
 #include "pinocchio/math/fwd.hpp"
 
 // Do not include this file directly.
@@ -134,20 +135,6 @@ namespace CppAD
 } // namespace CppAD
 
 #include "pinocchio/utils/static-if.hpp"
-
-namespace Eigen
-{
-  namespace internal
-  {
-    template<typename Scalar> EIGEN_DEVICE_FUNC inline CppAD::AD<Scalar>
-    pmin(const CppAD::AD<Scalar>& a, const CppAD::AD<Scalar>& b)
-    { return numext::mini(a, b); }
-
-    template<typename Scalar> EIGEN_DEVICE_FUNC inline CppAD::AD<Scalar>
-    pmax(const CppAD::AD<Scalar>& a, const CppAD::AD<Scalar>& b)
-    { return numext::maxi(a, b); }
-  }
-}
 
 namespace pinocchio
 {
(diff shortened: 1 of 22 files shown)


src/autodiff/cppad.hpp
```diff
@@ -15,7 +16,7 @@
 #ifdef PINOCCHIO_CPPAD_REQUIRES_MATRIX_BASE_PLUGIN
   #define EIGEN_MATRIXBASE_PLUGIN <cppad/example/eigen_plugin.hpp>
 #endif
-
+#include "pinocchio/autodiff/cppad/minmax.hpp"
```
jcarpent, line 19: This file should go after the include of CppAD.

src/utils/static-if.hpp
```diff
@@ -5,7 +5,7 @@
 #ifndef __pinocchio_utils_static_if_hpp__
 #define __pinocchio_utils_static_if_hpp__
 
-#include "pinocchio/fwd.hpp"
+#include "pinocchio/traits.hpp"
```
jcarpent, line 8: Use fwd here.

unittest/joint-free-flyer.cpp
```diff
@@ -3,6 +3,7 @@
 // Copyright (c) 2015 Wandercraft, 86 rue de Paris 91400 Orsay, France.
 //
 
+#include "pinocchio/fwd.hpp"
```
jcarpent, line 6: math/fwd.hpp already includes fwd.hpp.

src/autodiff/cppad/minmax.hpp
```diff
@@ -0,0 +1,65 @@
+//
+// Copyright (c) 2020 CNRS INRIA
+//
+
+#ifndef __pinocchio_autodiff_cppad_minmax_hpp__
+#define __pinocchio_autodiff_cppad_minmax_hpp__
+
+#include "pinocchio/utils/static-if.hpp"
+
+#ifdef __CUDACC__
```
jcarpent, line 10: Do not redefine EIGEN_DEVICE_FUNC. You should use the definition provided by Eigen.

src/autodiff/cppad.hpp
```diff
@@ -5,6 +5,7 @@
 #ifndef __pinocchio_autodiff_ccpad_hpp__
 #define __pinocchio_autodiff_ccpad_hpp__
 
+#include "pinocchio/container/boost-container-limits.hpp"
```
jcarpent, line 8: This is not needed.

src/math/fwd.hpp
```diff
@@ -5,7 +5,7 @@
 #ifndef __pinocchio_math_fwd_hpp__
 #define __pinocchio_math_fwd_hpp__
 
-#include "pinocchio/fwd.hpp"
```
jcarpent, line 8: Why did you change here?

src/utils/static-if.hpp
proyan: `pinocchio/fwd` loads the `Eigen` module as well. I had to separate `traits`, since I need to load `minmax.hpp` before I load `Eigen` module. So `static-if` needs to be independent from `Eigen`, since it is being loaded with `minmax.hpp`.

jcarpent: Why do you need to load minmax.hpp before Eigen?

jcarpent: For me, there is no apparent reason.

proyan: The problem is in Line 818 at https://eigen.tuxfamily.org/dox/MathFunctions_8h_source.html.
If Eigen is loaded before, the default impementation of `std::min` is used.

jcarpent: Then the inclusion order should not change, it is just the function signature that should be specialized.

jcarpent: In addition, this won't be the std version which will be called here, but the one located in CppAD namespace.

jcarpent: The clean implementation is to define `min` and `max` in the namespace `CppAD`.

proyan: I didn't find implementation of `min/max` binary operation in CppAD. I could only find implementation of `numeric_limits::min()/max()`

jcarpent: So, you have to implement them in Pinocchio ;)

jcarpent: I would suggest to only keep the unit test and the minmax.hpp files and to restart from scratch.

proyan:
I've simplified the inclusions, so there is no change in `math/fwd.hpp`

src/utils/static-if.hpp
proyan: I tried to put that earlier, but `std::min` is seen first by Eigen. The template resolves to the default implementation

jcarpent:
But the right way to do it is to define the min max functions in the cppad namespace. It will be much more simpler and cleaner.

src/utils/static-if.hpp
jcarpent: It should not. If min/max is defined in CppAD and included, this should work fine. std::min for CppAD::AD is not defined

PR #1244, opened 2020-06-16
[CMake] fix CppAD / CppADCG case

nim65s:
Hi,

A CMake issue was reported by @proyan.

CMake is case sensitive here. Without the right case, configuration fails on ubuntu 16.04 / 18.04 with :

```
  Could not find a package configuration file provided by "cppadcg"
  (requested version 2.4.1) with any of the following names:

    cppadcgConfig.cmake
    cppadcg-config.cmake

Could not find a package configuration file provided by "cppad" (requested
  version 20180000.0) with any of the following names:

    cppadConfig.cmake
    cppad-config.cmake
```

@proyan : can you test that this fixes your issue ?


nim65s:
These are only warnings and doesn't prevent anything to work ; but you are right, the namings are misleading. I'll fix that by providing support for both cases.

PR #1260, opened 2020-07-23
Fixes tiny error in the warning messages in aba and aba-derivatives

cmastalli:


src/algorithm/aba-derivatives.hxx
```diff
@@ -410,7 +410,7 @@ namespace pinocchio
   {
     PINOCCHIO_CHECK_INPUT_ARGUMENT(q.size() == model.nq, "The joint configuration vector is not of right size");
     PINOCCHIO_CHECK_INPUT_ARGUMENT(v.size() == model.nv, "The joint velocity vector is not of right size");
-    PINOCCHIO_CHECK_INPUT_ARGUMENT(tau.size() == model.nv, "The joint acceleration vector is not of right size");
+    PINOCCHIO_CHECK_INPUT_ARGUMENT(tau.size() == model.nv, "The joint torque vector is not of right size");
     PINOCCHIO_CHECK_INPUT_ARGUMENT(fext.size() == (size_t)model.njoints, "The size of the external forces is not of right size");
```
proyan, line 414: While you are at it, this should be 
"The external forces vector is not of right size" :)

Issue #1266, opened 2020-08-03
Error when we code generate frame-derivatives stuff


cmastalli:
Dear all,

I have found a compilation error that appears in Crocoddyl when I want to code generate stuff in frame-derivatives.
This issue is not happening if we don't use the latest features in frame-derivatives.
So, I am discarding any issue from Crocoddyl side.

For more information see below:
```bash
In file included from /usr/local/include/eigen3/Eigen/Core:427:0,
                 from /usr/local/include/pinocchio/utils/cast.hpp:8,
                 from /usr/local/include/pinocchio/fwd.hpp:23,
                 from /usr/local/include/pinocchio/spatial/fwd.hpp:9,
                 from /usr/local/include/pinocchio/multibody/model.hpp:9,
                 from /usr/local/include/pinocchio/parsers/urdf.hpp:9,
                 from /home/cmastalli/devel/crocoddyl/unittest/test_codegen.cpp:9:
/usr/local/include/eigen3/Eigen/src/Core/functors/AssignmentFunctors.h: In instantiation of ‘void Eigen::internal::assign_op<DstScalar, SrcScalar>::assignCoeff(DstScalar&, const SrcScalar&) const [with DstScalar = double; SrcScalar = CppAD::AD<CppAD::cg::CG<double> >]’:
/usr/local/include/eigen3/Eigen/src/Core/AssignEvaluator.h:637:5:   required from ‘void Eigen::internal::generic_dense_assignment_kernel<DstEvaluatorTypeT, SrcEvaluatorTypeT, Functor, Version>::assignCoeff(Eigen::Index) [with DstEvaluatorTypeT = Eigen::internal::evaluator<Eigen::Matrix<double, 3, 1> >; SrcEvaluatorTypeT = Eigen::internal::evaluator<Eigen::Product<Eigen::Matrix<CppAD::AD<CppAD::cg::CG<double> >, 3, 3, 0, 3, 3>, Eigen::Matrix<CppAD::AD<CppAD::cg::CG<double> >, 3, 1, 0, 3, 1>, 0> >; Functor = Eigen::internal::assign_op<double, CppAD::AD<CppAD::cg::CG<double> > >; int Version = 0; Eigen::Index = long int]’
[...]

PR #1267, opened 2020-08-03
Fix type

jcarpent:
Related to #1266


cmastalli:
@jcarpent thank you for a prompt solution.
I confirm you that everything works perfectly from my side :)

PR #1268, opened 2020-08-04
Fix call to RNEA and ABA derivatives for symbolic frameworks

jcarpent:


jcarpent:
@nim65s I've disabled 16.04. I will wait for the fix on robotpkg to reintegrate it in the CI.

PR #1283, opened 2020-08-24
math: a more general floating point trait

aelkazdadi:


jcarpent (approved):
Nice fix. Thanks a lot for your contribution @aelkazdadi.

jcarpent (commented):
@aelkazdadi Do you know why the compilation is failing?

aelkazdadi:
```cpp
    template<typename Scalar> \
    Scalar name(const Scalar & value) \
    { using std::name; return name(value); }
```
it's a case where `Scalar` is an expression. 
```
expression<
  boost::multiprecision::detail::subtract_immediates,
  boost::multiprecision::number<
    boost::multiprecision::backends::cpp_dec_float<100>
  >,
  boost::multiprecision::number<
    boost::multiprecision::backends::cpp_dec_float<100>
  >
  ,[...]
  ,[...]
>
```
passing it through `fabs` produces another expression, which needs to be cast to the owning type `number<...>`

jcarpent:
Could provide an explicit cast?

jcarpent:
@aelkazdadi I've pushed the change we've discussed so far.

PR #1284, opened 2020-08-25
Activate the sharing of memory between C++ and Python

jcarpent:
This new feature now requires EigenPy >= 2.5.0


CMakeLists.txt
```diff
@@ -125,12 +125,14 @@ ENDIF(BUILD_WITH_CASADI_SUPPORT)
 
 SET(BOOST_REQUIRED_COMPONENTS filesystem serialization system)
 SET(BOOST_BUILD_COMPONENTS unit_test_framework)
-SET(BOOST_OPTIONAL_COMPONENTS "")
+
+ADD_PROJECT_DEPENDENCY(Boost COMPONENTS ${BOOST_REQUIRED_COMPONENTS})
+FIND_PACKAGE(Boost COMPONENTS ${BOOST_BUILD_COMPONENTS})
```
jmirabel, line 130: Shouldn't there be a *REQUIRED* somewhere here ?
I never know whether it is before or after *COMPONENTS*. I think @nim65s can shade light on this.

jcarpent: This is now done. Thanks for the comment.

CMakeLists.txt
```diff
@@ -125,12 +125,15 @@ ENDIF(BUILD_WITH_CASADI_SUPPORT)
 
 SET(BOOST_REQUIRED_COMPONENTS filesystem serialization system)
 SET(BOOST_BUILD_COMPONENTS unit_test_framework)
-SET(BOOST_OPTIONAL_COMPONENTS "")
+
+EXPORT_BOOST_DEFAULT_OPTIONS()
+ADD_PROJECT_DEPENDENCY(Boost REQUIRED COMPONENTS ${BOOST_REQUIRED_COMPONENTS})
+FIND_PACKAGE(Boost COMPONENTS ${BOOST_BUILD_COMPONENTS})
```
nim65s, line 131: I'd suggest something like
```
IF(BUILD_TESTING)
  FIND_PACKAGE(Boost REQUIRED COMPONENTS ${BOOST_BUILD_COMPONENTS})
ENDIF()
```

unittest/python/bindings_force.py
```diff
@@ -33,12 +33,18 @@ def test_set_linear(self):
         f.linear = lin
         self.assertTrue(np.allclose(f.linear, lin))
 
+        f.linear[1] = 1.
+        self.assertTrue(v.linear[1] == 1.)
```
wxmerkt, line 37: Is this meant to be `v` or `f`?

CMakeLists.txt
jcarpent: I will move it to unit tests.

unittest/python/bindings_force.py
jcarpent: You're correct

jcarpent:
As mentioned earlier, this feature requires EigenPy >= 2.5.0, which will be available on robotpkg in the following days. We need to way until then for testing the feature on all the CI.

jcarpent:
Seems things are working well. I still need to expose correctly data fields of Model and Data. I will do it tomorrow ;)

unittest/python/bindings_force.py
jcarpent: Thanks for the feedback.

PR #1285, opened 2020-08-27
existJointName/existFrame: Handle case of empty name

wxmerkt:
Noticed when we got a segmentation fault due to a wrong name (empty string). We probably don't want to allow empty strings as names...

### DIFF
--- a/src/multibody/model.hxx
+++ b/src/multibody/model.hxx
@@ -223,7 +223,7 @@ namespace pinocchio
   inline bool ModelTpl<Scalar,Options,JointCollectionTpl>::
   existJointName(const std::string & name) const
   {
-    return (names.end() != std::find(names.begin(),names.end(),name));
+    return !name.empty() && (names.end() != std::find(names.begin(),names.end(),name));
   }
 
   template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
@@ -245,7 +245,7 @@ namespace pinocchio
   inline bool ModelTpl<Scalar,Options,JointCollectionTpl>::
   existFrame(const std::string & name, const FrameType & type) const
   {
-    return std::find_if(frames.begin(), frames.end(),
+    return !name.empty() && std::find_if(frames.begin(), frames.end(),
                         details::FilterFrame(name, type)) != frames.end();
   }
 
--- a/src/parsers/urdf/geometry.hxx
+++ b/src/parsers/urdf/geometry.hxx
@@ -42,7 +42,7 @@ namespace pinocchio
         {
           if (!model.existFrame(link_name, BODY))
           {
-            throw std::invalid_argument("No link " + link_name + " in model");
+            throw std::invalid_argument("No link '" + link_name + "' in model");
           }
           fid = model.getFrameId(link_name, BODY);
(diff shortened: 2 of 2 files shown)


jcarpent (approved):
@wxmerkt Thanks for this fix. It will be deployed in Pinocchio 2.5.0.

src/multibody/model.hxx
```diff
@@ -223,7 +223,7 @@ namespace pinocchio
   inline bool ModelTpl<Scalar,Options,JointCollectionTpl>::
   existJointName(const std::string & name) const
   {
-    return (names.end() != std::find(names.begin(),names.end(),name));
+    return !name.empty() && (names.end() != std::find(names.begin(),names.end(),name));
```
jmirabel, line 226: There is no reason this SEGV for an empty string. Are you sure this is your bug ?

jmirabel (commented):
I agree it is better, as a user, not to use empty string but I don't understand why the code should prevent anyone from doing it.

src/multibody/model.hxx
jmirabel: At least, the following code runs fine:
```cpp
#include <iostream>
#include <vector>
#include <algorithm>

void test1()
{
  std::vector<std::string> names {"abc", "def"};
  std::string empty;

  if (names.end() == std::find(names.begin(), names.end(), empty))
    std::cout << "empty string not in names." << std::endl;
}

int main()
{
  test1();
  return 0;
}
```

jmirabel: Ok, re-reading the first message, I understand the SEGV is not here but later. Then, I think this is a **very bad fix**.
1. It may be surprising for some users.
2. It is incomplete. If you want to prevent empty names, then you should also handle the `add` methods. As all the attributes are public, it is not really possible to prevent it.
3. What I don't understand in your case is that, if the empty string was not in `names`, then function `exist...` returned *false* and (obviously, since you called this) handled it or assert on it. Then, how will this solve anything ?

jcarpent: I share the same opinion than @jmirabel. Could you provide @wxmerkt a simple and tiny example when the bug you mention occurs?

wxmerkt: Yes, the segmentation fault was in follow-on code. The model does _not_ contain a joint with an empty name. However, the `existJointName` did return true for the check against an empty string, and then the follow-on code used `getFrameId` to get a frame id, and then index into the `oMf` vector. That's where things broke. However, to my surprise `existJointName` did return true for an empty string. I did not debug the other code more extensively or investigate deeper why `existJointName` returned true. As the occurrence is fixed, I am happy to close this PR.

However, should we instead explicitly disallow empty names for bodies/joints on creation?

jcarpent: Yes, we might. Because we would like to have a unique identifier for each one.

jmirabel: To me, the only good reason to programmatically prevent empty string is if we were using empty string for a specific task. For instance, when adding a Frame or a Joint, if the name is empty, we may generate a unique name for this frame/joint.

Apart from this, I see no good reason to forbid empty string (as long as it is unique).

jcarpent: Currently, we have a random name generator to prevent non-unique naming.

PR #1289, opened 2020-09-07
[CMake] {HPP_FCL → hpp-fcl}_FOUND

nim65s:
since d1a7f8da4b9afd00e461be0e210d6756ae2776fd, hpp-fcl is found using the standard CMake procedure instead of pkg-config.

@jcarpent : this might help to improve the coverage rate, as this will run test-geom & test-srdf again.


jcarpent (approved):
Thanks a lot

Issue #1291, opened 2020-09-13
Are joint friction parameters from URDF loaded to the robot model?


rshum19:
Hi I have a similar question to #1147.  I'm trying to simulate a 7DOF robot arm and want to include joint friction and damping. In the URDF one can define the tag 
```
<joint>
    <dynamics damping="1.0" friction="1.0">
</joint>
```

I'm wondering if these parameters are used at all when loading the model? 

If this is not the case, then would the best way to include joint friction be by adding a joint friction model to the joint torque vector (tau) before computing the forward the dynamics like shown below?
``` 
tau = friction_model(q,v);
ddq = pinocchio::aba(model, data, q, v, tau);
```

thanks,

jcarpent:
For the damping term, which affects the torque proportionally to the joint angular velocity, your trick does the job.
For handling the friction, it is a bit harder, as you have to solve a linear complementary problem: 
`tau < friction is orthogonal to ddq > 0`

jcarpent:
What we may add, is the automatic loading of these parameters in the Model when parsing the URDF file.

rshum19:
thank you for the fast response! 

I see, so adding damping can be easily done but friction requires more effort. I will look more into friction model and adding friction. Your response clarifies my question about what gets parsed from the URDF, so I will close this issue.

gfadini:
Hi, I see that this issue has been closed, nonetheless I think that having those parameters parsed from the URDF automatically may be useful for some applications.
In my specif case I can use the coulomb friction and viscous damping by defining them though crocoddyl, but I guess that since they actually are physical parameters of the system they may be loaded in the model itself. What do you think about this, would this be relevant to other people as well?

jcarpent:
Yes, this was my remark.
The new version of Pinocchio 3.x should include them. But, even with these values, the features of Crocoddyl will not allow you to handle correctly the friction, only the damping.

gfadini:
Thanks for the answer @jcarpent , that would be a nice addition to the next version of pinocchio. I agree with you that including the static friction properly in crocoddyl won't be straightforward

cmastalli:
@jcarpent we could boost the integration is you provide a roadmap with a rough description of the API.

jcarpent:
I'm lacking of time for that. Sorry.

cmastalli:
> I'm lacking of time for that. Sorry.

No problem at all, thanks :)

Kongx231:
[...]

PR #1304, opened 2020-10-02
Add support of joint friction and dampings

jcarpent:
Related to #1291. Currently, the values are not yet exploited in dynamics algorithms.

### DIFF
--- a/bindings/python/multibody/joint/expose-joints.cpp
+++ b/bindings/python/multibody/joint/expose-joints.cpp
@@ -16,11 +16,11 @@ namespace pinocchio
     
     static void exposeVariants()
     {
-      boost::mpl::for_each<JointModelVariant::types>(ModelExposer());
+      boost::mpl::for_each<JointModelVariant::types>(JointModelExposer());
       bp::to_python_converter<pinocchio::JointModelVariant,
                               JointVariantVisitor<pinocchio::JointModelVariant > >();
 
-      boost::mpl::for_each<JointDataVariant::types>(DataExposer());
+      boost::mpl::for_each<JointDataVariant::types>(JointDataExposer());
       bp::to_python_converter<pinocchio::JointDataVariant,
                               JointVariantVisitor<pinocchio::JointDataVariant > >();
     }
--- a/bindings/python/multibody/joint/joint-derived.hpp
+++ b/bindings/python/multibody/joint/joint-derived.hpp
@@ -1,9 +1,9 @@
 //
-// Copyright (c) 2015 CNRS
+// Copyright (c) 2015-2020 CNRS INRIA
 //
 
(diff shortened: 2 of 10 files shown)


hrp2-14:
This project doesn't accept pull requests on master, please change the base branch of your pull request to devel.

jcarpent:
@cmastalli and @gfadini One step ahead.

cmastalli:
> @cmastalli and @gfadini One step ahead.

If I understood correctly, this PR is parsing the friction and damping data, right?

jcarpent:
Yes. I will exploit them in the dynamics algorithms in Pinocchio 3.x

jmirabel:
One side question: I see many code formatting changes of code which I introduced some time ago. and which are not required for this PR.

Is it your editor or do you use a code formatting tool such as clang-format ?
How can I format the code on my side so that we follow the same rules ?

jcarpent:
Yes, you're correct. I'm trying to avoid them, but I'm not removing all of them due to lack of time. 
I do think that @nim65s may be the right person to provide a common file for all the projects with a clang style. 

@nim65s Could you provide more guidance on this topic?

nim65s:
The main point about formatting is having a common thing to avoid wasting time, while acknowledging that we won't all agree on the "beauty" of the final code.

For this, I made a https://github.com/gepetto/linters repository with a configuration for some tools (including clang-format).
My gitlab-CI system use this repository to check the format of every project, eg. for this PR:
https://gitlab.laas.fr/stack-of-tasks/pinocchio/-/jobs/94343

But we still have some fights:
- the line lengths is 119 for now, but in some other SoT projects there is another configuration for 79 (this is kind of OK for the CI)
- [yapf](https://github.com/google/yapf) is still in beta, and we have to reformat everything for each new version. And since I've made that choice, it looks to me that most of the python community choose [black](https://github.com/psf/black) instead, especially since the project moved to be hosted by the PSF.

So my main suggestion is to move the discussions about formatting to that https://github.com/gepetto/linters project, and then for the projects which want to comply with it, make a big formatting commit / PR and after that ensure that the "format" job of the CI still pass for all PRs.

cmastalli:
My experience with Gepetto linter is quite good. I would suggest to follow it.
If so, you just create a PR with formatting.

In the wiki of Crocoddyl explains how to run the formatting command. I guess it would be the same for OSX.

jcarpent:
For the linting task, I will prefer to use a linter-external-user to allow the tracking of commits and change. I will apply that for Pinocchio 3.x and beyond.

jmirabel:
> linter-external-user

what is this ?

jcarpent:
Just a nickname to mention that we should use a fake user for generating the linting commit.

Issue #1308, opened 2020-10-05
Kinematics tree root and rnea()


Gregwar:
Hello,

Suppose, on an humanoid robot standing on one foot, I have the robot state `q` and `v`, and a target `a` acceleration for the joints. 

If I want to know which torques woult it require to achieve such an acceleration on the joints I can use the inverse dynamics by calling [rnea()](https://gepettoweb.laas.fr/doc/stack-of-tasks/pinocchio/master/doxygen-html/namespacepinocchio.html#a26d94ea6486d0eed1934b94767124a01).

However, the robot kinematic root (and thus the floating base) is attached to the trunk of the robot. The only external force acting on the robot is the gravity, so if the support foot was the root of the kinematic tree, `rnea()` would give a direct result.

Is there a way to change the root on the fly ?
If not, what would you do to compute this ?

jcarpent:
Hi @Gregwar,

It seems you want to solve an inverse dynamics problem, meaning that you want to find the control actuation (torque) together with the contact forces to apply on the system in order to susten its weight. This called task-space inverse dynamics in the robotics jargon. I would suggest you having a look at https://github.com/stack-of-tasks/tsid, which builds on top of Pinocchio a dedicated solver for this kind of problem.

Gregwar:
Thanks, I will have a look at that.

I do understand that in the general case (involving multiple supports) this is a complicated problem.
However, is there an answer to my above question, since in single support case it would just be a change of the robot tree root, isn't it?

jcarpent:
Yes, indeed. You're correct. It won't be so hard to implement an algorithm that change the root kinematic tree.

Gregwar:
It seems that it would be useful from an API point of view (and apparently other people are pushing in this direction)

However I perfectly understand that everything could be done without changing the root of the tree. I'm curious what would be the steps in that particular case.

jcarpent:
If you want to give a try, I may provide feedback on top of your PR.

Gregwar:
I'm not sure I want to give a try to implementing this now since I'm new to the code base

What I meant is that I'm curious how it would be to compute torques without changing the floating base. But I agree that be able to change floating base would be a good feature.

jmirabel:
You could also achieve this without changing the current code.
1. compute the centroidal inertial forces due to the acceleration you want.
2. apply newton law on the full robot (inertial forces + weight + reaction force = 0)
[...]

PR #1310, opened 2020-10-06
Start implementing changeRootJoint

jcarpent:


jcarpent:
Related to #1308.

jcarpent:
In this PR, it is still needed to inverse the joint placements, change the inertias and the frame placements according to the new root joint.

Gregwar:
One note about that (not sure if it is relevant): the degrees of freedom on the chain from (old root)  to (new root) should be "backward" if you want that the same `q` state vector to produce the same configuration

jcarpent:
> One note about that (not sure if it is relevant): the degrees of freedom on the chain from (old root) to (new root) should be "backward" if you want that the same `q` state vector to produce the same configuration

Yes this is precisely  the content of my previous comment and what I have started to implement. Did you check the code?

jmirabel:
> > One note about that (not sure if it is relevant): the degrees of freedom on the chain from (old root) to (new root) should be "backward" if you want that the same `q` state vector to produce the same configuration
> 
> Yes this is precisely the content of my previous comment and what I have started to implement. Did you check the code?

isn't this breaking compactness on which some algorithms rely ?

jcarpent:
Normally not.

jmirabel:
I am not sure to understand how it is possible to have the same configuration order for the original model and the re-rooted model, while keeping the fact that all the child joints of a joint A corresponds to the *n* DoFs following the parameter of joint A...

Anyway, I think this is a useful feature. Thanks for implementing it.

jcarpent:
No, the configuration order will change. This is the role of the user to provide the correct configuration.
For all the basic joints except the FreeFlyer, SphericalZYX and Composite this is possible.

Gregwar:
The fact that the configuration order will change should indeed be specified on the documentation, a common thing the user will want is to write some code to remap the values of its model data from the original model to the re-rooted one. My comment above is that the values themselves could be compatible (which I guess is what @jcarpent also wants).

Thanks as well for implementing this.

jcarpent:
@Gregwar It would be nice if you can provide some help on this feature. For instance, the remaping function for joints configuration, so far an so on.

Gregwar:
I am not sure what is your philosophy here: do you want the values of the joints to be compatible with the re-rooted robot or not?

As a simple example; will a revolute joint axis that is on the re-root support in the tree have its orientation inverted (`newAxis=-axis`) or not?

PR #1315, opened 2020-10-13
[algorithm/centroidal] computeCentroidalMomentum calculates vcom as well

proyan:
CentroidalMomentum takes q, v as argument, and computes only `com`. Seems to be a missed opportunity.


jcarpent (approved):
Thanks for that. Could you complete the doc for this feature?

PR #1324, opened 2020-10-20
Add isNormalized

### LINKED ISSUE #1316: Normalization of SpecialEuclidean and SpecialOrthogonal joint configuration where input has zero norm.
Normalization methods SpecialEuclidean and SpecialOrthogonal rely on Eigen's `normalize()` method. According to [the method's doc](https://eigen.tuxfamily.org/dox/classEigen_1_1MatrixBase.html#ad16303c47ba36f7a41ea264cb26bceb6), if the vector's norm is too small, then it is left unchanged. Because of this, if the input vector contains all zeroes at the place of quaternions/continuous joints, the output of Pinocchio's `normalize` is _not_ normalized.

This is unexpected behaviour. A possible solution would be normalizing to the neutral configuration for the specific joint where Eigen's normalization fails. Would you agree on this? I can handle it as soon as I have time.

Additionally, it would be helpful to add a method called `isNormalized` for joint configurations (and, more in general, `isValid`, checking normalization _and_ bounds) (for things like checking user's input and such).

gabrielebndn:
Closes #1316.

I have implemented `isNormalized` for lie groups and for the joint configuration, with bindings.

I have very limited time, so I think I will let the bounds checks go for the time being. The normalization check has much more value and it is much more difficult to implement for an inexpert user, so I think this contribution is already important per se.

### DIFF
--- a/src/algorithm/joint-configuration.hpp
+++ b/src/algorithm/joint-configuration.hpp
@@ -612,6 +612,39 @@ namespace pinocchio
 
   /**
    *
+   * @brief         Normalize a configuration vector.
+   *
+   * @param[in]     model      Model of the kinematic tree.
+   * @param[in]     q          Configuration to check (size model.nq).
+   * @param[in]     prec       Precision.
+   *
+   * @return     Whether the configuration is normalized or not, within the given precision.
+   */
+  template<typename LieGroup_t, typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType>
+  inline bool isNormalized(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                           const Eigen::MatrixBase<ConfigVectorType> & q,
+                           const Scalar prec = Eigen::NumTraits<Scalar>::dummy_precision());
+
+  /**
+   *
+   * @brief         Normalize a configuration vector.
+   *
+   * @param[in]     model      Model of the kinematic tree.
+   * @param[in]     q          Configuration to check (size model.nq).
+   * @param[in]     prec       Precision.
+   *
+   * @return     Whether the configuration is normalized or not, within the given precision.
+   */
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType>
+  inline bool isNormalized(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                           const Eigen::MatrixBase<ConfigVectorType> & q,
+                           const Scalar prec = Eigen::NumTraits<Scalar>::dummy_precision())
+  {
+    return isNormalized<LieGroupMap,Scalar,Options,JointCollectionTpl,ConfigVectorType>(model,q,prec);
+  }
+
+  /**
+   *
    * @brief         Return true if the given configurations are equivalents, within the given precision.
    * @remarks       Two configurations can be equivalent but not equally coefficient wise (e.g two quaternions with opposite coefficients give rise to the same orientation, i.e. they are equivalent.).
    *
@@ -628,7 +661,7 @@ namespace pinocchio
   isSameConfiguration(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
                       const Eigen::MatrixBase<ConfigVectorIn1> & q1,
                       const Eigen::MatrixBase<ConfigVectorIn2> & q2,
-                      const Scalar & prec);
+                      const Scalar & prec = Eigen::NumTraits<Scalar>::dummy_precision());
 
   /**
    *
--- a/src/algorithm/joint-configuration.hpp
+++ b/src/algorithm/joint-configuration.hpp
@@ -612,7 +612,7 @@ namespace pinocchio
 
   /**
    *
-   * @brief         Normalize a configuration vector.
+   * @brief         Check whether a configuration vector is normalized within the given precision provided by prec.
    *
    * @param[in]     model      Model of the kinematic tree.
    * @param[in]     q          Configuration to check (size model.nq).
@@ -627,7 +627,7 @@ namespace pinocchio
 
   /**
    *
-   * @brief         Normalize a configuration vector.
+   * @brief         Check whether a configuration vector is normalized within the given precision provided by prec.
    *
    * @param[in]     model      Model of the kinematic tree.
    * @param[in]     q          Configuration to check (size model.nq).
--- a/src/algorithm/joint-configuration.hpp
+++ b/src/algorithm/joint-configuration.hpp
@@ -623,7 +623,7 @@ namespace pinocchio
   template<typename LieGroup_t, typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType>
   inline bool isNormalized(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
                            const Eigen::MatrixBase<ConfigVectorType> & q,
-                           const Scalar prec = Eigen::NumTraits<Scalar>::dummy_precision());
+                           const Scalar& prec = Eigen::NumTraits<Scalar>::dummy_precision());
 
   /**
    *
@@ -638,7 +638,7 @@ namespace pinocchio
   template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType>
   inline bool isNormalized(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
                            const Eigen::MatrixBase<ConfigVectorType> & q,
-                           const Scalar prec = Eigen::NumTraits<Scalar>::dummy_precision())
+                           const Scalar& prec = Eigen::NumTraits<Scalar>::dummy_precision())
   {
     return isNormalized<LieGroupMap,Scalar,Options,JointCollectionTpl,ConfigVectorType>(model,q,prec);
   }
--- a/src/algorithm/joint-configuration.hxx
+++ b/src/algorithm/joint-configuration.hxx
@@ -292,6 +292,31 @@ namespace pinocchio
     }
   }
 
+  template<typename LieGroup_t, typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorIn>
+  inline bool
+  isNormalized(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+               const Eigen::MatrixBase<ConfigVectorIn> & q,
+               const Scalar prec)
+  {
+    PINOCCHIO_CHECK_ARGUMENT_SIZE(q.size(), model.nq, "The configuration vector is not of the right size");
+    PINOCCHIO_CHECK_INPUT_ARGUMENT(prec >= 0, "The precision should be positive");
+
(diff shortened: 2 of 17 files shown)


src/algorithm/joint-configuration.hpp
```diff
@@ -610,6 +610,39 @@ namespace pinocchio
...
+   * @param[in]     model      Model of the kinematic tree.
+   * @param[in]     q          Configuration to check (size model.nq).
+   * @param[in]     prec       Precision.
+   *
+   * @return     Whether the configuration is normalized or not, within the given precision.
+   */
+  template<typename LieGroup_t, typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType>
+  inline bool isNormalized(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                           const Eigen::MatrixBase<ConfigVectorType> & q,
+                           const Scalar prec = Eigen::NumTraits<Scalar>::dummy_precision());
```
jcarpent, line 626: Please use ref for prec

src/algorithm/joint-configuration.hpp
```diff
@@ -610,6 +610,39 @@ namespace pinocchio
...
+   * @param[in]     model      Model of the kinematic tree.
+   * @param[in]     q          Configuration to check (size model.nq).
+   * @param[in]     prec       Precision.
+   *
+   * @return     Whether the configuration is normalized or not, within the given precision.
+   */
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType>
+  inline bool isNormalized(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                           const Eigen::MatrixBase<ConfigVectorType> & q,
+                           const Scalar prec = Eigen::NumTraits<Scalar>::dummy_precision())
```
jcarpent, line 641: Same applies here.

src/algorithm/joint-configuration.hxx
```diff
@@ -292,6 +292,31 @@ namespace pinocchio
     }
   }
 
+  template<typename LieGroup_t, typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorIn>
+  inline bool
+  isNormalized(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+               const Eigen::MatrixBase<ConfigVectorIn> & q,
+               const Scalar prec)
```
jcarpent, line 299: Same applies here

src/multibody/liegroup/cartesian-product-variant.hxx
```diff
@@ -363,6 +363,25 @@ normalize_impl (const Eigen::MatrixBase<Config_t>& qout) const
...
+template<typename _Scalar, int _Options, template<typename,int> class LieGroupCollectionTpl>
+template <class Config_t>
+bool CartesianProductOperationVariantTpl<_Scalar,_Options,LieGroupCollectionTpl>::
+isNormalized_impl (const Eigen::MatrixBase<Config_t>& qin,
+                   const Scalar prec) const
+{
+  Index id_q = 0;
+  for(size_t k = 0; k < liegroups.size(); ++k)
+  {
+    const Index & nq = lg_nqs[k];
```
jcarpent, line 375: Don't use ref here.

src/multibody/liegroup/cartesian-product.hpp
```diff
@@ -228,6 +228,12 @@ namespace pinocchio
       lg2.normalize(Qo2(qout));
     }
 
+    template <class Config_t>
+    bool isNormalized_impl (const Eigen::MatrixBase<Config_t>& qin, const Scalar prec) const
```
jcarpent, line 232: Please use ref for prec.

src/multibody/liegroup/liegroup-base.hpp
```diff
@@ -356,11 +356,26 @@ PINOCCHIO_LIE_GROUP_PUBLIC_INTERFACE_GENERIC(Derived,typename)
      * @brief      Normalize the joint configuration given as input.
      *             For instance, the quaternion must be unitary.
      *
-     * @param[out]     qout  the normalized joint configuration.
+     * @note       If the input vector is too small (i.e., qout.norm()==0), then it is left unchanged.
```
jcarpent, line 359: This is not really what it means too small. Being more precise would help.

src/multibody/liegroup/special-euclidean.hpp
```diff
@@ -430,6 +430,14 @@ namespace pinocchio
       PINOCCHIO_EIGEN_CONST_CAST(Config_t,qout).template tail<2>().normalize();
     }
 
+    template <class Config_t>
+    static bool isNormalized_impl(const Eigen::MatrixBase<Config_t> & qin,
+                           const Scalar prec)
```
jcarpent, line 435: Missing alignment and reference.

src/multibody/liegroup/special-euclidean.hpp
```diff
@@ -783,6 +791,14 @@ namespace pinocchio
       qout_.template tail<4>().normalize();
     }
 
+    template <class Config_t>
+    static bool isNormalized_impl(const Eigen::MatrixBase<Config_t>& qin,
+                           const Scalar prec)
+    {
+      Scalar norm = Scalar(qin.template tail<4>().norm());
+      return (std::abs(norm - Scalar(1.0)) < prec );
```
jcarpent, line 799: Don't use std::abs here, but rather using std::abs. It will be helpful for Scalar overloading.

src/multibody/liegroup/special-orthogonal.hpp
```diff
@@ -305,6 +305,14 @@ namespace pinocchio
       qout_.normalize();
     }
 
+    template <class Config_t>
+    static bool isNormalized_impl (const Eigen::MatrixBase<Config_t> & qin,
+                                   const Scalar prec)
+    {
+      const Scalar norm = qin.norm();
+      return (std::abs(norm - Scalar(1.0)) < prec );
```
jcarpent, line 313: Same here

src/multibody/liegroup/special-orthogonal.hpp
```diff
@@ -574,6 +582,14 @@ namespace pinocchio
       qout_.normalize();
     }
 
+    template <class Config_t>
+    static bool isNormalized_impl (const Eigen::MatrixBase<Config_t> & qin,
+                                   const Scalar prec)
+    {
+      const Scalar norm = qin.norm();
```
jcarpent, line 589: Same here.

jcarpent (changes requested):
Thanks for this PR. I made some few comments. @gabrielebndn Could you handle them?

src/multibody/liegroup/liegroup-base.hpp
gabrielebndn: This is basically a copy-paste from Eigen's own documentation. Can't be safer and more precise than this

src/algorithm/joint-configuration.hpp
gabrielebndn: May I know why? I see this is done in `isSameConfiguration` for instance, but I actually did my best to avoid it. Pass-by-const-reference is useless for scalar types and may be counter-productive : https://stackoverflow.com/a/270435/12175820

src/multibody/liegroup/cartesian-product-variant.hxx
gabrielebndn: This was largely inspired from the other algorithms in the file, which all use a reference at this point. I agree it is useless. I can remote it from here, no problem, but then I recommend to do the same in all other algorithms

src/multibody/liegroup/special-euclidean.hpp
gabrielebndn: Ok

gabrielebndn: I am sorry, you said
```
Don't use std::abs here, but rather using std::abs
```
Not clear. I assume you meant I should just use `abs`, without `std::`?

src/algorithm/joint-configuration.hpp
jcarpent: Not in the case of Scalar for Autodiff and so on. In general, we follow the same pattern as Eigen. For isSameConfiguration, it is maybe a mistake.

src/multibody/liegroup/cartesian-product-variant.hxx
jcarpent: Yes, I will do it for the rest in Pinocchio 3.x

src/multibody/liegroup/liegroup-base.hpp
jcarpent: Do know under which precision they consider really a zero to be zero?

src/multibody/liegroup/special-euclidean.hpp
jcarpent: You should do something like:
```c++
using std::abs;
return abs(value);
```

src/algorithm/joint-configuration.hpp
gabrielebndn: No, I mean `isSameConfiguration` _is_ using ref, but I thought _that_ do be a mistake and I did differently. Ok, I'll change it

src/multibody/liegroup/liegroup-base.hpp
gabrielebndn: Absolutely no idea. Again, since we decided to accept Eigen's behavior, I just blindly copy-pasted that part of the doc, you can check it [here](https://eigen.tuxfamily.org/dox/classEigen_1_1MatrixBase.html#ad16303c47ba36f7a41ea264cb26bceb6). I agree it is very weird, I was surprised myself, in computer science you never say that a floating-point number _is_ zero, but that's verbatim from Eigen's doc and it's the best we have.

I suppose we _could_ dig into Eigen's code and find out, but it would be a mistake to document in Pinocchio's doc something which is not documented by Eigen itself and is therefore likely to change with no guarantees for us nor for the users.

src/algorithm/joint-configuration.hpp
jcarpent: @gabrielebndn When do you plan to handle the comments?

gabrielebndn: I think I can do it today, maximum by the end of the week

src/multibody/liegroup/liegroup-base.hpp
gabrielebndn: Resolving this conversation with no changes

gabrielebndn:
@jcarpent done

jcarpent (approved):

jcarpent:
OK. Thanks. Next time, it would be nice to squatch the commits that fixe the remarks I have made. They are a bit redundant I would say.

gabrielebndn:
> OK. Thanks. Next time, it would be nice to squatch the commits that fixe the remarks I have made. They are a bit redundant I would say.

Indeed they are. I don't know, I am mostly on Gitlab now, and it offers an option to automatically squash the commits, doesn't Github have the same? I think it does: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuring-commit-squashing-for-pull-requests

PR #1332, opened 2020-11-01
Add computations of kinematic regressors

jcarpent:


traversaro:
Thanks @jcarpent for adding such a quite cool feature! Do you happen to have any reference or pointer for the derivation of this regressor?

jcarpent:
@traversaro. I developed the derivatives on my side. If you think it would be worth to write a document/paper on it, we may proceed.

traversaro:
> 
> 
> @traversaro. I developed the derivatives on my side. If you think it would be worth to write a document/paper on it, we may proceed.

I have no need to look into that at the moment to be honest, do not worry. Mine was just a curiosity, as most kinematic calibration algorithm that I know typically write the geometric model in some form with a reduced number of parameters (such as Denavit Hartenberg, see for example Section 6.2 https://doi.org/10.1007/978-3-319-32552-1_6 ), while in your case if I got correctly you model the changes that each transform can experience as a 6D left-trivialized derivative (see https://github.com/stack-of-tasks/pinocchio/pull/1332/files#diff-373a0e4d5f39761421b0412c3e3a0ce767985fb27c1482f64375c801d97f63f0R68), that is something I never saw.

jcarpent:
Yes, I do agree. DH seems the most used conventions for robot
callibration. But here, working directly on SE(3) seems more convenient
I would say and I think efficient/generic.
Best,

Justin

qleonardolp:
> > @traversaro. I developed the derivatives on my side. If you think it would be worth to write a document/paper on it, we may proceed.
> 
> I have no need to look into that at the moment to be honest, do not worry. Mine was just a curiosity, as most kinematic calibration algorithm that I know typically write the geometric model in some form with a reduced number of parameters (such as Denavit Hartenberg, see for example Section 6.2 https://doi.org/10.1007/978-3-319-32552-1_6 ), while in your case if I got correctly you model the changes that each transform can experience as a 6D left-trivialized derivative (see https://github.com/stack-of-tasks/pinocchio/pull/1332/files#diff-373a0e4d5f39761421b0412c3e3a0ce767985fb27c1482f64375c801d97f63f0R68), that is something I never saw.

By the way, does pinocchio rely on the Denavit-Hartenberg standard to parse the robots _URDF_ and compute the Jacobian matrix?

PR #1339, opened 2020-11-12
Minor fixes

jcarpent:


bindings/python/multibody/model.hpp
```diff
@@ -220,8 +220,8 @@ namespace pinocchio
         .add_property("nvs",&Model::nvs)          
         .add_property("parents",&Model::parents)
         .add_property("names",&Model::names)
-        .add_property("name",&Model::name)
-        .add_property("referenceConfigurations", &Model::referenceConfigurations)
+        .def_readwrite("name",&Model::name)
+        .def_readwrite("referenceConfigurations", &Model::referenceConfigurations)
```
wxmerkt, line 224: Does directly exposing it as `def_readwrite` rather than with a getter/setter expose it to accidentally setting values for different/wrong sizes? E.g. could this lead to segfaults down the line if someone messes up the memory this way?

jcarpent: For reference configurations, behind it is a std::map, so normally, you will get some exceptions if something is not acceptable. The same applies for name (std::string).

jcarpent: Does it answer your question @wxmerkt?

wxmerkt: Say I have a robot where nq=19. Now I am accidentally writing `model.referenceConfigurations["something_new"] = np.zeros((7,))`. Will this be accepted even though with validation it would have to be nq=19?

jcarpent: This should be done at another level I would say: at the level of std::map exposer. This goes beyond this PR. If you have time, @wxmerkt you are welcome to provide a check for the issue you mention. 
Anyway, an error will be throw when using Pinocchio's algo.

wxmerkt: Ok, if the algorithm checks the sizing then it's fine - in the past I noticed there were no checks for nq/nv sizes of q and v that are passed in (and this leads to segmentation faults). It's fine in C++ because we can do Debug/Release builds but in Python it's a different story.

PR #1364, opened 2020-12-22
Pickle frame

gabrielebndn:
Improve `Frame` bindings by allowing pickling and exposing equality operator

### DIFF
--- a/bindings/python/multibody/frame.hpp
+++ b/bindings/python/multibody/frame.hpp
@@ -54,9 +54,20 @@ namespace pinocchio
         .def(FramePythonVisitor())
         .def(CopyableVisitor<Frame>())
         .def(PrintableVisitor<Frame>())
+        .def_pickle(Pickle())
         ;
       }
 
+    private:
+      struct Pickle : bp::pickle_suite
+      {
+        static
+        boost::python::tuple
+        getinitargs(const Frame& f)
+        {
+          return bp::make_tuple(f.name, f.parent, f.previousFrame, f.placement, f.type);
+        }
+      };
     };
     
 
--- a/bindings/python/multibody/frame.hpp
+++ b/bindings/python/multibody/frame.hpp
@@ -51,6 +51,7 @@ namespace pinocchio
                           "A Plucker coordinate frame related to a parent joint inside a kinematic tree.\n\n",
                           bp::no_init
                          )
+        .def(bp::self == bp::self)
         .def(FramePythonVisitor())
         .def(CopyableVisitor<Frame>())
         .def(PrintableVisitor<Frame>())
--- a/unittest/python/bindings_frame.py
+++ b/unittest/python/bindings_frame.py
@@ -44,6 +44,15 @@ class TestFrameBindings(PinocchioTestCase):
         f.placement = new_placement
         self.assertTrue(np.allclose(f.placement.homogeneous, new_placement.homogeneous))
 
+    def test_frame_equality(self):
+        M = pin.SE3.Random()
+        frame1 = pin.Frame("name", 1, 2, M, pin.OP_FRAME)
+        frame2 = pin.Frame("name", 1, 2, M, pin.OP_FRAME)
+        frame3 = pin.Frame("othername", 3, 4, pin.SE3.Random(), pin.BODY)
+
+        self.assertTrue(frame1 == frame2)
+        self.assertTrue(frame1 != frame3)
+
     def test_getters(self):
         data = self.model.createData()
         q = pin.randomConfiguration(self.model)
--- a/unittest/python/bindings_frame.py
+++ b/unittest/python/bindings_frame.py
@@ -53,6 +53,19 @@ class TestFrameBindings(PinocchioTestCase):
         self.assertTrue(frame1 == frame2)
         self.assertTrue(frame1 != frame3)
 
+    def test_pickle(self):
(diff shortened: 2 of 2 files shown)


bindings/python/multibody/frame.hpp
```diff
@@ -51,12 +51,24 @@ namespace pinocchio
                           "A Plucker coordinate frame related to a parent joint inside a kinematic tree.\n\n",
                           bp::no_init
                          )
+        .def(bp::self == bp::self)
```
jcarpent, line 54: Please move this line in the dedicated visit section of the Visitor.
Could you also add `.def(bp::self != bp::self)`

bindings/python/multibody/frame.hpp
```diff
@@ -51,12 +51,24 @@ namespace pinocchio
...
         .def(FramePythonVisitor())
         .def(CopyableVisitor<Frame>())
         .def(PrintableVisitor<Frame>())
+        .def_pickle(Pickle())
         ;
       }
 
+    private:
+      struct Pickle : bp::pickle_suite
```
jcarpent, line 63: It would be better to rely on the serialization module of Pinocchio for pickling, similar to what it is done for Model.
It is more efficient and easy to do.

jcarpent (changes requested):
Hi @gabrielebndn,

Thanks for this nice contribution. 
I made some remarks which should be easy to handle.

bindings/python/multibody/frame.hpp
gabrielebndn: I think I had not put it because there was no direct correspondent in C++, but I'll see... Anyhow, what's the point of adding `!=` explicitly? It is automatically added if `==` exists, the unit tests check that

bindings/python/multibody/frame.hpp
gabrielebndn: Yes, I had noticed it, I can certainly do it

bindings/python/multibody/frame.hpp
jcarpent: There is no reason for `!=` to be added as far as I know. Indeed, some class may not have it.

gabrielebndn: Sorry, I think it wasn't clear, what I meant was, why adding `.def(bp::self != bp::self)` explicitly? I am testing `!=` in the Python unit tests and it already works without it, which makes me think it must be defined automatically from `==`

jmirabel: > There is no reason for `!=` to be added as far as I know. Indeed, some class may not have it.

I think in Python the not-equal comparison is by default the logical not of the equal comparison, when the latter is defined. Here, the only good reason to explicitly add the not-equal comparison is if, in C++, the two operators are not the opposite of one another. I guess this will never be the case for class Frame ?

```python
from __future__ import print_function

class Foo(object):
    def __eq__(self, other):
        print("Foo.__eq__")
        return True
class Bar(object):
    pass

def check(Class):
    a, b = Class(), Class()
    print(a == b)
    print(a != b)

check(Foo)
check(Bar)
```

gabrielebndn: Well, I moved `.def(bp::self == bp::self)` to the visitor, but I did not explicitly expose `.def(bp::self != bp::self)` as it seemed unnecessary

bindings/python/multibody/frame.hpp
gabrielebndn: Done

gabrielebndn:
Codacy is complaining about using pickle, saying it's unsafe. That's the source of the only two issues. It can be ignored

jcarpent (commented):
Sorry @gabrielebndn, in fact, your previous suggestion was correct (defining by hand the Pickling as for SE3). Sorry for asking you to revert to your previous status.
Thanks in advance!
Justin

jcarpent:
By the way, I made the change ;)

jcarpent (approved):

gabrielebndn:
Ok good :)

bindings/python/multibody/frame.hpp
gabrielebndn: It seems `.def(bp::self != bp::self)` _was_ necessary for Python 2. I've pushed it, I hope it fixes everything

gabrielebndn: And `.def(bp::self != bp::self)` reuired the explicit defintion of `!=` in the C++ class too. I took care of it, hope the CI passes this time

gabrielebndn:
Well, it seems it's passing, the only unsuccessful check is something unrelated about the format

jmirabel:
> Well, it seems it's passing, the only unsuccessful check is something unrelated about the format

Format is *allowed to fail* so it should not the blocking failure. The doc-coverage check has been frozen for some time but it is now running again. Maybe it was a heavy load on the computer.

gabrielebndn:
> Format is _allowed to fail_ so it should not the blocking failure. The doc-coverage check has been frozen for some time but it is now running again.

Ah, now I understand! Yes, I had seen that the format was allowed to fail, and I had also seen that doc-coverage kept running, but the "umbrella job" gitlab-ci, including both of them, seemed to have unconditionally failed, so it was confusing. Anyhow, now all checks are fine. It was probably doc-coverage

PR #1376, opened 2021-01-20
[aba] Use data.a_gf instead of data.a

### LINKED ISSUE #1374: ABA overwrites data.a
In order to store auxiliary accelerations ABA employs `data.a`, meaning this field is modified by the algorithm and left in an inconsistent state when the algorithm has terminated (the final value corresponds to the joint acceleration due to articular velocity with zero acceleration, plus gravity)

It should probably be replaced by `data.a_gf`, as done for instance by `rnea`. Notice that `computeABADerivatives` correctly employs `data.a_gf`

Another minor potential issue is that at the end of the algorithm `data.f` is filled with values which are not consistent with the final joint acceleration, which is not taken into account

gabrielebndn:
Fix #1374

### DIFF
--- a/src/algorithm/aba.hpp
+++ b/src/algorithm/aba.hpp
@@ -25,6 +25,8 @@ namespace pinocchio
   /// \param[in] v The joint velocity vector (dim model.nv).
   /// \param[in] tau The joint torque vector (dim model.nv).
   ///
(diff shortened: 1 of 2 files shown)


jcarpent:
Thanks for this.

PR #1383, opened 2021-01-25
[Minor] Fix message in argument checking macros.

jmirabel:
Otherwise, the added unit test fails with error
> Running 1 test case...
pinocchio/unittest/macros.cpp(16): error: in "macrosTest/test_check_arguments": check expected_msg == exception.what() has failed [wrong argument size: expected 2, got 3
hint: v.size() is different from size
 != oss.str()]
pinocchio/unittest/macros.cpp(39): error: in "macrosTest/test_check_arguments": exception "std::invalid_argument" raised as expected: validation on the raised exception through predicate "check_exception_msg"
pinocchio/unittest/macros.cpp(16): error: in "macrosTest/test_check_arguments": check expected_msg == exception.what() has failed [wrong argument size: expected 2, got 3
hint: custom message with stream
 != oss.str()]
pinocchio/unittest/macros.cpp(44): error: in "macrosTest/test_check_arguments": exception "std::invalid_argument" raised as expected: validation on the raised exception through predicate "check_exception_msg"
>
> *** 4 failures are detected in the test module "macrosTest"


unittest/CMakeLists.txt
```diff
@@ -12,7 +12,13 @@ MACRO(ADD_TEST_CFLAGS target flag)
 ENDMACRO(ADD_TEST_CFLAGS)
 
 MACRO(ADD_PINOCCHIO_UNIT_TEST NAME)
-  SET(PKGS ${ARGN}) # Optional arguments
+  set(options HEADER_ONLY)
```
jcarpent, line 15: Could you uniformize the code (use SET instead of set).

jcarpent (approved):
Looks fine to me. Thanks for the update. 
Could you just handle my single comment?

jcarpent (commented):
@jmirabel Apparently, VS V142 is not happy with your changes. I don't at which level. Maybe the literal string stuff. 
Can you fix it?

jmirabel:
I made a minimal reproducible example at https://rextester.com/JSOA49209. It seems that this works
```c++
  _PINOCCHIO_CHECK_ARGUMENT_SIZE_3(v.size(), size,
      "custom message " << "with stream");
```
while this does not
```c++
  PINOCCHIO_CHECK_ARGUMENT_SIZE(v.size(), size,
      "custom message " << "with stream");
```
The compiler gets lost somewhere in the macro with variadic arguments. Does it ring a bell ?

jcarpent:
> I made a minimal reproducible example at https://rextester.com/JSOA49209. It seems that this works
> 
> ```c++
>   _PINOCCHIO_CHECK_ARGUMENT_SIZE_3(v.size(), size,
>       "custom message " << "with stream");
> ```
> 
> while this does not
> 
> ```c++
>   PINOCCHIO_CHECK_ARGUMENT_SIZE(v.size(), size,
>       "custom message " << "with stream");
> ```
> 
> The compiler gets lost somewhere in the macro with variadic arguments. Does it ring a bell ?

I think this is because you removed the `PINOCCHIO_STRING_LITERAL`.

jmirabel:
If I don't remove it, the exception message is "oss.str()" so I don't think it should be here.
Is it really important to have the variadic macro ?

src/macros.hpp
```diff
@@ -106,15 +106,15 @@ namespace pinocchio
 /// \brief Generic macro to throw an exception in Pinocchio if the condition is not met with a given input message.
 #if !defined(PINOCCHIO_NO_THROW)
   #define PINOCCHIO_THROW(condition,exception_type,message) \
-    if (!(condition)) { throw exception_type(PINOCCHIO_STRING_LITERAL(message)); }
+    if (!(condition)) { throw exception_type(message); }
```
jcarpent, line 109: PINOCCHIO_STRING_LITERAL has been removed.

src/macros.hpp
```diff
@@ -106,15 +106,15 @@ namespace pinocchio
...
 #else
   #define PINOCCHIO_THROW(condition,exception_type,message)
 #endif
 
 #define _PINOCCHIO_GET_OVERRIDE_FOR_CHECK_INPUT_ARGUMENT(_1, _2, MACRO_NAME, ...) MACRO_NAME
 
 #define _PINOCCHIO_CHECK_INPUT_ARGUMENT_2(condition, message) \
-  PINOCCHIO_THROW(condition,std::invalid_argument,PINOCCHIO_STRING_LITERAL(message))
+  PINOCCHIO_THROW(condition,std::invalid_argument,message)
```
jcarpent, line 117: PINOCCHIO_STRING_LITERAL has been removed.

jcarpent:
Yes, it is more flex for the next version of Pinocchio. Otherwise, I will have to change things.
You removed at some stages the `PINOCCHIO_STRING_LITERAL` while it was working nicely before.

jmirabel:
>  You removed at some stages the `PINOCCHIO_STRING_LITERAL` while it was working nicely before.

It does not work. Try commit ade3809 and you will see what error message you get.

jmirabel:
After digging a bit more, I think it has nothing to do with PINOCCHIO_STRING_LITERAL. The `_PINOCCHIO_GET_OVERRIDE_FOR_CHECK_INPUT_ARGUMENT` does not count correctly the arguments and evaluates to `_PINOCCHIO_CHECK_ARGUMENT_SIZE_1`.

jmirabel:
If you want to be convinced: https://rextester.com/MOKPL73574

jmirabel:
https://stackoverflow.com/questions/5134523/msvc-doesnt-expand-va-args-correctly

jcarpent:
I was reading too fastly your code.
Yes, seems to be another option: https://stackoverflow.com/a/36015150
Could you handle it?

jcarpent:
Thanks a lot for this nice and helpful fix.

PR #1385, opened 2021-02-01
[Doc] Add remarkable identities for Jlog6

jmirabel:
When rendered, this adds to doc of `Jlog6`:

image

### DIFF
--- a/src/spatial/explog.hpp
+++ b/src/spatial/explog.hpp
@@ -461,6 +461,12 @@ namespace pinocchio
    *  \f]
    *  and
    *  \f[ \beta(x)=\left(\frac{1}{x^2} - \frac{\sin x}{2x(1-\cos x)}\right) \f]
+   *
+   *  \cheatsheet For \f$ M = log_6(a \times b) \f$,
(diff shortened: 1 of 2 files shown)


jcarpent (changes requested):
You should change x which stands for the cross operator when working with Lie algebra

jmirabel:
You are correct. How should I mark the multiply. I think a mere space not very explicit.

jcarpent:
I would suggest something like:
log_6(exp_6(a)exp_6(b)) with X_b = Ad_{exp_6(b)}^-1

jmirabel:
I cannot write it with `exp_6` as I would have to write the derivatives wrt to `exp_6(a)` and not `a`. Or I did not understand what you meant. 

Here is what I came up after some re-writing.
image

jcarpent:
Could you use small letters for so(3) elements ? And Ad_A for X_A?

jmirabel:
image

jcarpent (approved):
Thanks a lot.

jmirabel:
Sorry for the misunderstanding. I wanted to come to an agreement before pushing to avoid useless CI. What you merge is the original change. I'll open a new PR.

PR #1387, opened 2021-02-03
Fix buildReducedModel

jmirabel:
The parent frame of the geometry objects was copied from the input model. As the frame order is not kept, they were erroneous.

As a side note, I think the frame order could be kept. I won't have time to dig into this.


jcarpent (commented):
Could you add a test on it? In such a way, we won't reproduce such a mistake.

jcarpent:
> As a side note, I think the frame order could be kept. I won't have time to dig into this.

Could you add a TODO?

jcarpent (approved):

PR #1399, opened 2021-03-02
Meshcat capsule visualisation

VladimirIvan:
- Adds meshcat capsule visualisation as a triangulated mesh.
- Requires meshcat-python 0.19 or newer.
- Reverts to displaying a cylinder on older versions of meshcat-python.

### DIFF
--- a/bindings/python/pinocchio/visualize/meshcat_visualizer.py
+++ b/bindings/python/pinocchio/visualize/meshcat_visualizer.py
@@ -13,6 +13,7 @@ try:
 except:
     WITH_HPP_FCL_BINDINGS = False
 
+
 def loadBVH(bvh):
     import meshcat.geometry as mg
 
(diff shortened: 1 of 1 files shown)


hrp2-14:
Hi ! This project doesn't usually accept pull requests on master. If this wasn't intentionnal, you can change the base branch of this pull request to devel (No need to close it for that). Best, a bot.

nmansard:
Hi Vlad, thanks for the contribution. Don't you think you should push the capsules in MeshCat directly?

VladimirIvan:
Yes, I'll eventually move it there but meshcat-python doesn't support python2 since v0.1.0. If I push it there I wouldn't be able to use it with ROS melodic / Ubuntu 18.04. I don't use meshcat-python outside of pinocchio so a PR here is the quickest way to get the capsules to display for now.
This is also the same way capsules are exposed in Julia (via the core geometry library rather via meshcat).

jcarpent (approved):
Thanks @VladimirIvan for this nice contribution. Would you think a vectorized version of the code is possible?
If not, I will merge it.

VladimirIvan:
I don't think that this code will be a bottleneck so I'll leave vectorizing it as a future work. Please go ahead and merge.

PR #1409, opened 2021-03-20
Enhance overall serialization + extend Geometry API

jcarpent:
This PR allows serializing data to binary containers in Python.


cmastalli (commented):
I marked a tiny detail to be modified in the Python bindings

PR #1417, opened 2021-03-31
[parser] Compute Local AABB when loading a geometry.

florent-lamiraux:
It seems that the method was not called for all the geometry types. Since the AABB is now initialized to infinity, the missing
call has visible effects.

### DIFF
--- a/src/parsers/urdf/geometry.cpp
+++ b/src/parsers/urdf/geometry.cpp
@@ -224,6 +224,7 @@ namespace pinocchio
           throw std::invalid_argument("The polyhedron retrived is empty");
         }
 
+        geometry->computeLocalAABB();
         return geometry;
       }
 #endif // PINOCCHIO_WITH_HPP_FCL


jcarpent (changes requested):
For me, this `computeLocalABBB` should go directly inside the `computeBodyRadius` as the user may add some geometries manually.

jcarpent:
I've proposed an alternative and more robust solution in 2b3c5f5.

jmirabel:
The const-cast in @jcarpent proposal feels not very clean. I agree with @jcarpent that the case of manually added geometries should be supported too.

This issue can also be fixed in hpp-fcl:
- in the constructor for primitives,
- in the model finalization for BVH.

jcarpent:
Duplicate with #1418.

>     * in the constructor for primitives,

In this case, we cannot change the object properties online.
I would rather suggest making `computeLocalAABB` const

jcarpent:
> The const-cast in @jcarpent proposal feels not very clean

I do agree. This is because the method is not marked const. But should not be a problem in general ;).
Another solution would be to provide a const method to retrieve this quantity for any `CollisionGeometry`.

florent-lamiraux:
> This issue can also be fixed in hpp-fcl:
> 
>     * in the constructor for primitives,
> 
>     * in the model finalization for BVH.
  * and in any method that modifies the geometry.
I think it would be the best solution.

jmirabel:
> In this case, we cannot change the object properties online.

Yes, it requires you to call `computeLocalAABB` or to make attributes private and modify them with class methods. But the latter is significantly more work for the developer.

> I would rather suggest making `computeLocalAABB` const

Then either the internal aabb is mutable or the it is const-cast in the class function. You don't change the fact that you break the C++ constness stuff.

jcarpent:
> Then either the internal aabb is mutable or the it is const-cast in the class function. You don't change the fact that you break the C++ constness stuff.

The minimal intervention is the current one. I let you choose, but I think these two options are the most effective one at the time being, without breaking too much the current API.

jmirabel:
Then, go for the const-cast, with a TODO note that points to this thread. Unless someone has some more time to spend on this.

jcarpent:
The fix will appear in #1418.

PR #1420, opened 2021-04-01
Enhance collision support + serialization + parallelization

jcarpent:
Exploit functors of hpp-fcl.


cmastalli (commented):
This is an interesting PR which I would like to have a deeper understanding.

The first question is:
 - *is there a way to create a compact data pool?*

 As far I understand, you are creating a vector of Pinocchio data for parallelizing the algorithms.

The second question is
- *could you report the computation time for ABA / RNEA with/out parallelization for systems which have the similar number of bodies than quadrupeds and humanoid robots?*

jcarpent:
I think the code is explicit enough to see how it works.
You are also sufficiently expert to run the bench by yourself, I have no such time on my side.

Issue #1423, opened 2021-04-07
buildReducedModel might erase collision_model and intertial information


ntorresalberto:
I'm trying to work on a reduced model of the [panda robot](https://github.com/Gepetto/example-robot-data/blob/master/robots/panda_description/urdf/panda.urdf).
I'd like to point out some quirks/problems I'm currently encountering:

1. This is more of a feature request: `RobotWrapper` could provide an API to build a reduced model. I'm not sure I can do it right now but I'd be happy to tackle a PR in the future.
2. I don't know if it actually is a problem, but `buildReducedModel` API doesn't offer a way to pass the `collision_model` from the `RobotWrapper` class. I've noticed `RobotWrapper.visual_model` and `RobotWrapper.collision_model` are both `pinocchio.pinocchio_pywrap.GeometryModel`. Would they provide the same functionality? are they interchangeable?
3. In the example below, I tried to create a `RobotWrapper` from the reduced model and it seems that it has lost the intertial information from my urdf? (or there's some other explanation to why it's now giving my `NaN` values for the gravity)

```python
import example_robot_data
import pinocchio as pin

robot = example_robot_data.load('panda')

randomconfig = pin.randomConfiguration(robot.model)
robot.forwardKinematics(randomconfig)
print('randomconfig:', randomconfig, randomconfig.shape)
g=robot.gravity(randomconfig)
print('model G:', g, g.shape)

# lock 4th joint to reduce dof
jn4_idx = robot.model.getJointId('panda_joint4')
reduced_model, reduced_visual_model = pin.buildReducedModel(
    model=robot.model,
    geom_model=robot.visual_model,
    list_of_joints_to_lock=[jn4_idx],
    reference_configuration=pin.neutral(robot.model))

robot2 = pin.RobotWrapper(model=reduced_model, visual_model=reduced_visual_model)

randomconfig2 = pin.randomConfiguration(robot2.model)
print('randomconfig2:', randomconfig2, randomconfig2.shape)
g=robot2.gravity(randomconfig2)
print('reduced model G:', g, g.shape)
```

And here's the output:
```bash
$ python reduced_panda.py 
randomconfig: [ 2.01874195 -0.3871077   1.67996741 -0.56351698  2.44279775  0.68504683
 -0.97782112  0.03072918  0.01111099] (9,)
model G: [0. 0. 0. 0. 0. 0. 0. 0. 0.] (9,)
[...]

PR #1427, opened 2021-04-08
add Robot wrapper.reduceModel function and extend buildReducedModel for multiple GeometricModel

ntorresalberto:
This was referenced in https://github.com/stack-of-tasks/pinocchio/issues/1423.

Summary:
- Extended `buildReducedModel` ([here](https://github.com/ntorresalberto/pinocchio/blob/106ebe56a879d99c1b300318dbea4a027773b132/src/algorithm/model.hxx#L498)) to take multiple GeometricModels (`std::vector<GeometricModel>`) as input. Now the old `buildReducedModel` ([here](https://github.com/ntorresalberto/pinocchio/blob/106ebe56a879d99c1b300318dbea4a027773b132/src/algorithm/model.hxx#L475)) function is redirected to this one to avoid redundant code.
- Exposed the new function as `pinocchio.buildReducedModel` in python (that takes and returns a list of GeometricModel).
- Added `RobotWrapper.reduceModel` function that uses the new `buildReducedModel`. 
I did it so `list_of_joints_to_lock` can take a mix of joint names and indexes ([here](https://github.com/ntorresalberto/pinocchio/blob/106ebe56a879d99c1b300318dbea4a027773b132/bindings/python/pinocchio/robot_wrapper.py#L192)). I'm pointing it out because this could be a source of bugs and it's not as simple to implement in C++.
- Updated the python example `build-reduced-model.py` ([here](https://github.com/ntorresalberto/pinocchio/blob/RobotWrapper.reduceModel/examples/build-reduced-model.py)).

Please let me know if there are any more changes needed.

In particular, there are some things I wasn't sure about:
- Is the `boost::python` wrapping of `std::vector<GeometricModel>` implemented correctly? ([here](https://github.com/ntorresalberto/pinocchio/blob/106ebe56a879d99c1b300318dbea4a027773b132/bindings/python/algorithm/expose-model.cpp#L82))
- Do you use some code formatter (like `clang-format` for C++ or `yapf` for python)?
- 2 tests didn't pass and I'm not sure how to check. I'm running `make test`, is there a way to see the logs?:
```
98% tests passed, 2 tests failed out of 125

Total Test time (real) = 1527.02 sec

The following tests FAILED:
	 36 - test-cpp-model (Timeout)
	 73 - test-py-rpy (Failed)
Errors while running CTest
make: *** [Makefile:130: test] Error 8
```

edit: just fixed `test-cpp-model` but I'm still clueless about `test-py-rpy`, it was failing on my machine even before my modifications and I don't know where to look.

### DIFF
--- a/bindings/python/algorithm/expose-model.cpp
+++ b/bindings/python/algorithm/expose-model.cpp
@@ -2,11 +2,10 @@
 // Copyright (c) 2019-2020 INRIA
 //
 
-#include "pinocchio/algorithm/model.hpp"
 #include "pinocchio/bindings/python/algorithm/algorithms.hpp"
-#include "pinocchio/bindings/python/serialization/serializable.hpp"
 #include "pinocchio/bindings/python/utils/list.hpp"
 #include "pinocchio/bindings/python/utils/std-vector.hpp"
+#include "pinocchio/algorithm/model.hpp"
 
 namespace pinocchio
 {
@@ -14,7 +13,7 @@ namespace pinocchio
   {
 
     namespace bp = boost::python;
-
+  
     template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
     bp::tuple appendModel_proxy(const ModelTpl<Scalar,Options,JointCollectionTpl> & modelA,
                                 const ModelTpl<Scalar,Options,JointCollectionTpl> & modelB,
@@ -31,6 +30,22 @@ namespace pinocchio
       
       return bp::make_tuple(model,geom_model);
     }
+  
+    template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType>
+    bp::tuple
+    buildReducedModel(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                      const GeometryModel & geom_model,
+                      const std::vector<JointIndex> & list_of_joints_to_lock,
+                      const Eigen::MatrixBase<ConfigVectorType> & reference_configuration)
+    {
+      typedef ModelTpl<Scalar,Options,JointCollectionTpl> Model;
+      Model reduced_model; GeometryModel reduced_geom_model;
+      
+      buildReducedModel(model,geom_model,list_of_joints_to_lock,
+                        reference_configuration,reduced_model,reduced_geom_model);
+      
+      return bp::make_tuple(reduced_model,reduced_geom_model);
+    }
 
     template <typename Scalar, int Options,
               template <typename, int> class JointCollectionTpl,
@@ -40,32 +55,13 @@ namespace pinocchio
         const std::vector<GeometryModel> &list_of_geom_models,
         const std::vector<JointIndex> &list_of_joints_to_lock,
         const Eigen::MatrixBase<ConfigVectorType> &reference_configuration) {
-      typedef ModelTpl<Scalar,Options,JointCollectionTpl> Model;
-      Model reduced_model;
+      typedef ModelTpl<Scalar, Options, JointCollectionTpl> Model;
       std::vector<GeometryModel> reduced_geom_models;
-
-      buildReducedModel(model,list_of_geom_models,list_of_joints_to_lock,
-                        reference_configuration,reduced_model,reduced_geom_models);
-      
-      return bp::make_tuple(reduced_model,reduced_geom_models);
-    }
-
-    template <typename Scalar, int Options,
-              template <typename, int> class JointCollectionTpl,
-              typename ConfigVectorType>
-    bp::tuple buildReducedModel(
-        const ModelTpl<Scalar, Options, JointCollectionTpl> &model,
-        const GeometryModel &geom_model,
-        const std::vector<JointIndex> &list_of_joints_to_lock,
-        const Eigen::MatrixBase<ConfigVectorType> &reference_configuration) {
-      typedef ModelTpl<Scalar,Options,JointCollectionTpl> Model;
       Model reduced_model;
-      std::vector<GeometryModel> geom_models{ geom_model };
-      std::vector<GeometryModel> reduced_geom_models(1); // in this case it's a single element
-      buildReducedModel(model,geom_models,list_of_joints_to_lock,
-                        reference_configuration,reduced_model,reduced_geom_models);
-      
-      return bp::make_tuple(reduced_model,reduced_geom_models.front());
+      buildReducedModel(model, list_of_geom_models, list_of_joints_to_lock,
+                          reference_configuration, reduced_model,
+                          reduced_geom_models);
+      return bp::make_tuple(reduced_model, reduced_geom_models);
     }
 
     void exposeModelAlgo()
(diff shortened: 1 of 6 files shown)


bindings/python/algorithm/expose-model.cpp
```diff
@@ -2,17 +2,19 @@
 // Copyright (c) 2019-2020 INRIA
 //
 
+#include "pinocchio/algorithm/model.hpp"
```
jcarpent, line 5: Please do not change the include order for no reason

bindings/python/algorithm/expose-model.cpp
```diff
@@ -2,17 +2,19 @@
...
 #include "pinocchio/bindings/python/algorithm/algorithms.hpp"
+#include "pinocchio/bindings/python/serialization/serializable.hpp"
```
jcarpent, line 7: Same here

bindings/python/algorithm/expose-model.cpp
```diff
@@ -2,17 +2,19 @@
...
 #include "pinocchio/bindings/python/utils/list.hpp"
-#include "pinocchio/algorithm/model.hpp"
+#include "pinocchio/bindings/python/utils/std-vector.hpp"
```
jcarpent, line 9: Same here

bindings/python/algorithm/expose-model.cpp
```diff
@@ -2,17 +2,19 @@
...
 
 namespace pinocchio
 {
   namespace python
   {
 
     namespace bp = boost::python;
-  
+
```
jcarpent, line 17: Please don’t format the coce

bindings/python/algorithm/expose-model.cpp
```diff
@@ -29,27 +31,49 @@ namespace pinocchio
...
-  
-    template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl, typename ConfigVectorType>
-    bp::tuple
-    buildReducedModel(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
-                      const GeometryModel & geom_model,
-                      const std::vector<JointIndex> & list_of_joints_to_lock,
-                      const Eigen::MatrixBase<ConfigVectorType> & reference_configuration)
-    {
+
+    template <typename Scalar, int Options,
```
jcarpent, line 35: Please don’t format the code

bindings/python/algorithm/expose-model.cpp
```diff
@@ -29,27 +31,49 @@ namespace pinocchio
...
+              template <typename, int> class JointCollectionTpl,
+              typename ConfigVectorType>
+    bp::tuple buildReducedModel(
+        const ModelTpl<Scalar, Options, JointCollectionTpl> &model,
+        const std::vector<GeometryModel> &list_of_geom_models,
```
jcarpent, line 40: You are changing the function argument. Please keep the former signature too as it will break he api otherwise

examples/build-reduced-model.py
```diff
@@ -12,12 +12,14 @@
 mesh_dir = model_path
 # You should change here to set up your own URDF file
 urdf_filename = model_path + '/ur_description/urdf/ur5_robot.urdf'
-model, collision_model, visual_model = pin.buildModelsFromUrdf(urdf_filename, mesh_dir)
+model, collision_model, visual_model = pin.buildModelsFromUrdf(
```
jcarpent, line 15: Don t change the formatting. It makes the review process longer for nothing

examples/build-reduced-model.py
```diff
@@ -31,16 +33,50 @@
         print('Warning: joint ' + str(jn) + ' does not belong to the model!')
 
 # Set initial position of both fixed and revoulte joints
-initialJointConfig = np.array([0,0,0,   # shoulder and elbow
-                                1,1,1]) # gripper)
+initialJointConfig = np.array([
```
jcarpent, line 36: Same here

src/algorithm/model.hxx
```diff
@@ -468,67 +468,98 @@ namespace pinocchio
...
+      const std::vector<JointIndex> &list_of_joints_to_lock,
+      const Eigen::MatrixBase<ConfigVectorType> &reference_configuration,
+      ModelTpl<Scalar, Options, JointCollectionTpl> &reduced_model,
+      std::vector<GeometryModel> &list_of_reduced_geom_models) {
+    typedef ModelTpl<Scalar, Options, JointCollectionTpl> Model;
+    buildReducedModel(input_model, list_of_joints_to_lock,
+                      reference_configuration, reduced_model);
+
+    // for all GeometryModels
+    for (auto &input_geom_model : list_of_geom_models) {
```
jcarpent, line 506: Don’t use auto keyword as we are still compliant with C++98

src/algorithm/model.hxx
```diff
@@ -468,67 +468,98 @@ namespace pinocchio
...
+          reduced_geom_model.addCollisionPair(cp);
+        }
 #endif
-    
+      list_of_reduced_geom_models.push_back(reduced_geom_model);
+    }
   }
 
-} // namespace pinocchio
+  } // namespace pinocchio
```
jcarpent, line 563: Don’t format the code

jcarpent (changes requested):
Thanks @ntorresalberto. You did a good job. I made remarks to be fixed.
In general, don’t change the indent or formatting of the code. I have to navigate through them making the review process a bit painful.

bindings/python/algorithm/expose-model.cpp
ntorresalberto: Sorry, my editor is doing more than I know.

bindings/python/algorithm/expose-model.cpp
ntorresalberto: I tried to keep the old signature.
I'm working on removing the formatting.
Hopefully it'll be clearer then.

One question though, I see you have a `cmake/.clang-format`, could I use that to format the C++ code?

src/algorithm/model.hxx
ntorresalberto: Noted, it should be fixed in next commit.

bindings/python/algorithm/expose-model.cpp
jcarpent: You can choose to only format your change, not everything normally. 
I will do a whole formating when switching to Pinocchio 3.x to avoid breaking the whole history.

ntorresalberto:
> Thanks @ntorresalberto. You did a good job. I made remarks to be fixed.
> In general, don’t change the indent or formatting of the code. I have to navigate through them making the review process a bit painful.

Sorry about the formatting, it should be reverted now.

Would you consider using auto-formatting tools? it would solve this tedious issues, 'll help code uniformity and it's easy to add a check in the CI to ensure PRs are compliant.
The "auto" part might sound daunting but up until now, I haven't found any cases where it would break code.

bindings/python/algorithm/expose-model.cpp
ntorresalberto: I see.

This function signature should be reverted now.

bindings/python/pinocchio/robot_wrapper.py
```diff
@@ -175,6 +175,40 @@ def updateGeometryPlacements(self, q=None, visual=False):
     def framesForwardKinematics(self, q):
         pin.framesForwardKinematics(self.model, self.data, q)
 
+    def reduceModel(self,
```
jcarpent, line 178: Maybe, it would be nice to just return a new robot, in order to keep the current one alive.

src/algorithm/model.hxx
```diff
@@ -477,56 +477,83 @@ namespace pinocchio
...
+      const std::vector<JointIndex> &list_of_joints_to_lock,
+      const Eigen::MatrixBase<ConfigVectorType> &reference_configuration,
+      ModelTpl<Scalar, Options, JointCollectionTpl> &reduced_model,
+      std::vector<GeometryModel> &list_of_reduced_geom_models) {
+
+    typedef ModelTpl<Scalar, Options, JointCollectionTpl> Model;
+    buildReducedModel(input_model, list_of_joints_to_lock, reference_configuration, reduced_model);
+
+    // for all GeometryModels
+    for (const GeometryModel &input_geom_model : list_of_geom_models) {
```
jcarpent, line 505: You are using a C++11 way of iterating over vectors. Use the more classic way of having an index ;)

src/algorithm/model.hpp
```diff
@@ -144,6 +144,31 @@ namespace pinocchio
...
+  template <typename Scalar, int Options,
+            template <typename, int> class JointCollectionTpl,
+            typename ConfigVectorType>
+  void buildReducedModel(
+      const ModelTpl<Scalar, Options, JointCollectionTpl> &model,
+      const std::vector<GeometryModel> &list_of_geom_models,
+      const std::vector<JointIndex> &list_of_joints_to_lock,
+      const Eigen::MatrixBase<ConfigVectorType> &reference_configuration,
+      ModelTpl<Scalar, Options, JointCollectionTpl> &reduced_model,
+      std::vector<std::reference_wrapper<GeometryModel>> &list_of_reduced_geom_models);
```
jcarpent, line 170: You should avoid `>>` for C++98

jcarpent (changes requested):

bindings/python/pinocchio/robot_wrapper.py
ntorresalberto: I don't really know all of the use cases but it wouldn't seem naturally coherent with the rest of the API to me (not that this `reduceModel` does though). This would be the only method to return the object right?

Would it maybe make sense to have a static function that takes a robot and returns a new reduced one? Sort of like `pin.RobotWrapper.BuildFromURDF`.

src/algorithm/model.hpp
ntorresalberto: Wow, I didn't know that, could you propose a solution in that case?

BTW, it could be useful to [enforce compliance via the compiler](https://stackoverflow.com/a/12606757) to avoid having to review this quirks.

bindings/python/pinocchio/robot_wrapper.py
jcarpent: But it means that you will need to copy twice everything if you want to keep the original robot.
I would suggest rather renaming this method `buildReducedRobot` which returns a new robot, as done for `buildReducedModel`.

src/algorithm/model.hpp
jcarpent: The CI is not checking this compliance, we may think to add it, but I don't have time for that.
you should add a space `> >`

ntorresalberto: > you should add a space > >

I didn't realize that was the problem haha

ntorresalberto: should be fixed now

src/algorithm/model.hxx
ntorresalberto: should be fixed now

bindings/python/pinocchio/robot_wrapper.py
ntorresalberto: > But it means that you will need to copy twice everything if you want to keep the original robot.
> I would suggest rather renaming this method `buildReducedRobot` which returns a new robot, as done for `buildReducedModel`.

I think I might be misunderstanding something: In order to return a new robot I would need to copy everything anyway right?

To be clear:
```python
reduced_robot = robot.buildReducedRobot(...)
```

In this case `reduced_robot` would need to have copies of every data structure in `robot`

jcarpent: Yes, but you keep the orginal model for futur use

src/algorithm/model.hxx
```diff
@@ -502,7 +502,8 @@ namespace pinocchio
     buildReducedModel(input_model, list_of_joints_to_lock, reference_configuration, reduced_model);
 
     // for all GeometryModels
-    for (const GeometryModel &input_geom_model : list_of_geom_models) {
+    for (unsigned int gmi = 0; gmi < list_of_geom_models.size(); ++gmi) {
```
jcarpent, line 505: Use size_t here.

bindings/python/pinocchio/robot_wrapper.py
ntorresalberto: I've added the change.

jcarpent (approved):
This looks nice to me. I've rebased on the current devel branch and also extend the support for aligned vectors and I will add a missing unittest.

ntorresalberto:
I've fixed some typos in the python api docs.
Some were not so obvious so they might still be wrong, feel free to revert them.

jcarpent:
Thanks a lot for the fixes. I will merge.

PR #1438, opened 2021-04-21
[Minor] Bugfix + API improvement in lie group.

jmirabel:


jcarpent (approved):
Thanks for the PR.

src/multibody/liegroup/liegroup-variant-visitors.hpp
```diff
@@ -155,23 +155,21 @@ namespace pinocchio
                    const Eigen::MatrixBase<JacobianOut_t> & J,
                    const ArgumentPosition arg);
 
-  template<typename LieGroupCollection, class ConfigL_t, class ConfigR_t, class JacobianIn_t, class JacobianOut_t>
+  template<ArgumentPosition arg, typename LieGroupCollection, class ConfigL_t, class ConfigR_t, class JacobianIn_t, class JacobianOut_t>
```
jcarpent, line 158: Do you think that they might be some impact on the end-user by changing the API?

jmirabel: It shouldn't. I only made the declaration fit with the definition.

Note that I couldn't find the definition of the previously declared function but I may have missed it.

jmirabel:
No idea why gitlab-ci failed.

PR #1440, opened 2021-04-27
Fix bug in urdf parsing (win32+debug)

francois-keith:
+ revert commit made to debug the issue.

### DIFF
--- a/src/parsers/urdf/model.hxx
+++ b/src/parsers/urdf/model.hxx
@@ -202,12 +202,13 @@ namespace pinocchio
               const std::string & body_name)
           {
             const Frame & parent_frame = model.frames[parent_frame_id];
+            const JointIndex parent_frame_parent = parent_frame.parent;
 
             const SE3 placement = parent_frame.placement * joint_placement;
             FrameIndex fid = model.addFrame(Frame(joint_name, parent_frame.parent, parent_frame_id,
                                                   placement, FIXED_JOINT, Y));
 
-            model.addBodyFrame(body_name, parent_frame.parent, placement, (int)fid);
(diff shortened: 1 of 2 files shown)


jcarpent (commented):
What is the bug? Hard to guess from your commits.
Thanks anyway @francois-keith to suggest a fix.

francois-keith:
The bug is that the unit test fails (in fact, any urdf import fails) with the error

`unknown location(0): fatal error: in "urdfTest/build_model": class std::invalid_argument: The index of the parent frame is not valid`

This is due to the fact that the value of `parent_frame.parent` changes from its initial value (0) to an undefined one (xxxxx) after the operation `model.addFrame`.

This change is likely to be caused by the push_back operation, that moves the whole vector to another place in memory, so the reference is no longer valid.

Here is a more  detailed log of what happens when running test `BOOST_AUTO_TEST_CASE ( check_tree_from_XML )`

In model.hxx, line 204
`const Frame & parent_frame = model.frames[parent_frame_id],`
parent_frame.parent's value is `0`, its address is 0000019DE9DEA958

Before `FrameIndex fid = model.addFrame(`, value is still `0`, address is still 0000019DE9DEA958

In the `addFrame` method:
Before `frames.push_back(frame)`, here is the content of `frames`: 
i=0 value `0` address 0000019DE9DEA958

After `frames.push_back(frame)`, here is the content of `frames`: 
i=0 value `0` address **is now 0000019DE9E34598**
i=1 value `0` address 0000019DE9E34688

Hence after model.addFrame, the (old) address 0000019DE9DEA958 points to an invalid value: 15987178197214944733, which causes the error.

jcarpent:
Very clear and very nice catch. Thanks a lot, @francois-keith.

Issue #1463, opened 2021-06-28
Loading URDFs in catkin devel mode


jmirabel:
Hi,

when loading URDF model in Pinocchio, one can use `os.env["ROS_PACKAGE_PATH"].split(':')` as list of directories for loading meshes. This however does not work when using catkin devel mode because the directory tree layout is not the same.
I don't think this is Pinocchio's fault but it makes loading URDF in this context slightly painful (in an automated way).

Has anyone a neat solution for this ?

olivier-stasse:
The main problem is that all Gepetto packages assume a clean install (i.e. everything is located in the install directory).
You can use:
`catkin config --install`
to have catkin forcing the installation for every package.

I found this to be the easiest way to avoid the devel mess.

jmirabel:
Yes, this is a workaround I am aware of. Personally, I always use this mode.

I was wondering whether there could be a easy solution. I personally do not see any. If nothing comes out in the 2-3 coming days, I'll close this.

I haven't checked the code but I guess packages are found by recursively looking for directories with a valid `package.xml` inside (BTW, if you want catkin to run in a infinite loop, add a loop with a symlink in your install folder). The only solution to this problem I can think of is to mimic this behavior. I think this shouldn't be done.

olivier-stasse:
Ok i got our point. You want this to work even for people using the devel way of doing thing.
I have a none-technical answer to this: devel does not exist anymore in ros2 with colcon.
I am not sure it is worthwhile to find a solution for this specific case in the long term.

nim65s:
Hi @jmirabel,

I tried the following:
```bash
apt install -qqy ros-noetic-pinocchio
source /opt/ros/noetic/setup.bash
mkdir -p ~/catkin_ws/src
cd ~/catkin_ws/src
git clone --recursive https://github.com/gepetto/example-robot-data
cd ..
catkin_make_isolated -DPYTHON_EXECUTABLE=/usr/bin/python3
source devel_isolated/setup.bash
```

And after that, in a python3 shell:
```python
>>> import example_robot_data
>>> example_robot_data.load_full('talos')
(<pinocchio.robot_wrapper.RobotWrapper object at 0x7ff853bbb040>, array([ 0.00000e+00,  0.00000e+00,  1.01927e+00,  0.00000e+00,
        0.00000e+00,  0.00000e+00,  1.00000e+00,  0.00000e+00,
[...]

PR #1467, opened 2021-07-09
Work-around for finding meshes of URDFs in ROS devel spaces #1463

wxmerkt:
I've also run into the issue described by @jmirabel in #1463. The following work-around resolves this issue (as postulated at in https://github.com/stack-of-tasks/pinocchio/issues/1463#issuecomment-872225739). 

Note, this is likely not Windows-compatible due to the forward slash "/". I'd love to hear some thoughts/discussion on a better solution for this issue

@jmirabel - would this PR solve your issue as well?

### DIFF
--- a/src/utils/file-explorer.cpp
+++ b/src/utils/file-explorer.cpp
@@ -57,9 +57,9 @@ namespace pinocchio
     // To support ROS devel/isolated spaces, we also need to look one package above the package.xml:
     fs::path path;
     std::vector<std::string> list_of_paths;
-    for (const auto& path_string : raw_list_of_paths) {
-      list_of_paths.push_back(path_string);
-      path = fs::path(path_string);
+    for (std::vector<std::string>::iterator it = raw_list_of_paths.begin(); it != raw_list_of_paths.end(); ++it) {
+      list_of_paths.push_back(*it);
+      path = fs::path(*it);
       if (fs::exists(path / "package.xml")) {
         list_of_paths.push_back(fs::path(path / "..").string());
       }
--- a/src/utils/file-explorer.cpp
+++ b/src/utils/file-explorer.cpp
@@ -6,18 +6,13 @@
 #include <boost/filesystem.hpp>
 #include "pinocchio/utils/file-explorer.hpp"
 
-#ifdef _WIN32
-const char delimiter = ';';
-#else
-const char delimiter = ':';
-#endif
-
 namespace fs = boost::filesystem;
 namespace pinocchio
 {
 
   void extractPathFromEnvVar(const std::string & env_var_name,
-                             std::vector<std::string> & list_of_paths)
+                             std::vector<std::string> & list_of_paths,
+                             const std::string & delimiter)
   {
     const char * env_var_value = std::getenv(env_var_name.c_str());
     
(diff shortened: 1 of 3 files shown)


jcarpent (changes requested):
Thanks a lot @wxmerkt for this fix. I let @jmirabel provide feedback.
Could you also handle Windows context?

src/utils/file-explorer.cpp
```diff
@@ -24,8 +24,12 @@ namespace pinocchio
       while(true)
       {
         size_t offset = policyStr.find_first_of(delimiter, lastOffset);
-        if (offset < policyStr.size())
+        if (offset < policyStr.size()) {
           list_of_paths.push_back(policyStr.substr(lastOffset, offset - lastOffset));
+          // Support for devel spaces: We also need to look one package above:
+          // Work-around for https://github.com/stack-of-tasks/pinocchio/issues/1463
+          list_of_paths.push_back(policyStr.substr(lastOffset, offset - lastOffset) + "/..");
+        }
```
jmirabel, line 32: This is a bit hazardous and kind of intrusive...

I would choose instead modify `rosPaths` function (and make it available in Python) as you did here or, arguably better, as:
```cpp
std::vector<std::string> new_list;
for (auto path : list_of_paths) {
  new_list.push_back(path);
  if (file_exists(path + "/package.xml"))
    new_list.push_back(path + "/..");
}
```

wxmerkt:
@jcarpent Since C++17 is not an option, is the use of [Boost.Filesystem](https://www.boost.org/doc/libs/1_76_0/libs/filesystem/doc/index.htm) ok?

jmirabel:
Since `rosPaths` uses ':' as path separator, then it wasn't Windows compatible anyway. You may modify `rosPaths` with something like
```cpp
#ifdef _WIN32
char delimiter = ';';
char pathsep = '\';
#else
char delimiter = ':';
char pathsep = '/';
#endif
```

jcarpent:
I think the solution proposed by @jmirabel looks good to me.

src/utils/file-explorer.cpp
jmirabel: Not sure whether I wasn't clear or whether you think this is a bad idea so I reformulate my proposition.

I think it is better to fix the list of ROS package paths in function `rosPaths` and not in this function so that it does not apply to anything else than ROS.
Someone later may use it's own environment variable and definitely won't expect this way of resolving paths.

wxmerkt: Thanks for the comment, I am modifying it now

jcarpent (approved):

src/utils/file-explorer.hpp
```diff
@@ -17,26 +17,22 @@ namespace pinocchio
...
    *
    * @param[in]  env_var_name The name of the environment variable.
    * @param[out] list_of_paths List of path to fill with the paths extracted from the environment variable value.
-   * @param[in]  delimiter The delimiter between two consecutive paths.
    */
   PINOCCHIO_DLLAPI void
   extractPathFromEnvVar(const std::string & env_var_name,
-                        std::vector<std::string> & list_of_paths,
-                        const std::string & delimiter = ":");
+                        std::vector<std::string> & list_of_paths);
```
jcarpent, line 35: I did not realize, but you break the API here. Would it possible to keep the former signature alive?

jcarpent (changes requested):

src/utils/file-explorer.hpp
wxmerkt: I've restored the old API and also ported the delimiter change over

jcarpent (approved):
Well done

jmirabel:
Thanks for having handled this.

PR #1472, opened 2021-07-16
meshcat: fix loading of BVH model

jcarpent:
BVH models should be loaded in a raw manner after checking if they are not related to a specific file


jcarpent:
Solving #1471.

PR #1476, opened 2021-07-27
Change 'typedef' to 'using', add curly braces

the-raspberry-pi-guy:
Simple PR to change:
- `typedef` to `using` when defining the Vector6d, as there are advantages to the `using` syntax: https://stackoverflow.com/a/32959200
- Add some curly braces to the success/failure scenarios, as adding further statements after these tripped me up in the same way as: https://stackoverflow.com/a/2125078

### DIFF
--- a/examples/inverse-kinematics.cpp
+++ b/examples/inverse-kinematics.cpp
@@ -23,7 +23,7 @@ int main(int /* argc */, char ** /* argv */)
   J.setZero();
 
   bool success = false;
-  using Vector6d = Eigen::Matrix<double, 6, 1>;
+  typedef Eigen::Matrix<double, 6, 1> Vector6d;
   Vector6d err;
   Eigen::VectorXd v(model.nv);
   for (int i=0;;i++)
--- a/examples/inverse-kinematics.cpp
+++ b/examples/inverse-kinematics.cpp
@@ -23,7 +23,7 @@ int main(int /* argc */, char ** /* argv */)
   J.setZero();
 
   bool success = false;
-  typedef Eigen::Matrix<double, 6, 1> Vector6d;
+  using Vector6d = Eigen::Matrix<double, 6, 1>;
   Vector6d err;
   Eigen::VectorXd v(model.nv);
   for (int i=0;;i++)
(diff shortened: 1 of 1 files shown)


hrp2-14:
Hi ! This project doesn't usually accept pull requests on master. If this wasn't intentionnal, you can change the base branch of this pull request to devel (No need to close it for that). Best, a bot.

proyan:
Hi @the-raspberry-pi-guy
Thanks for contributing to pinocchio :)
Unfortunately, type aliases are a C++11 feature, and we still maintain compatibility with C++98 systems.

the-raspberry-pi-guy:
Hi @proyan 

My apologies - I missed checking that :) No worries about type aliases. This becomes a very minor PR then, but I would still recommend including the curly braces as part of the beginner tutorials :)

Thanks

jcarpent:
> This becomes a very minor PR then, but I would still recommend including the curly braces as part of the beginner tutorials :)

Thanks for your contribution. Why do you say that?

examples/inverse-kinematics.cpp
```diff
@@ -23,7 +23,7 @@ int main(int /* argc */, char ** /* argv */)
   J.setZero();
 
   bool success = false;
-  typedef Eigen::Matrix<double, 6, 1> Vector6d;
+  using Vector6d = Eigen::Matrix<double, 6, 1>;
```
jcarpent, line 26:
```suggestion
  typedef Eigen::Matrix<double, 6, 1> Vector6d;
```

the-raspberry-pi-guy:
> > This becomes a very minor PR then, but I would still recommend including the curly braces as part of the beginner tutorials :)
> 
> Thanks for your contribution. Why do you say that?

It is a matter of preference, but as a matter of good coding style I think it is better to add curly braces here. I used this example and added a statement in what I thought was one of the nested structures, only to be caught out by the lack of curly braces. Curly brace use/not use should be consistent, hence just pointing it out. https://stackoverflow.com/questions/2125066/is-it-a-bad-practice-to-use-an-if-statement-without-curly-braces/2125078#2125078

At your discretion though of course.

jcarpent:
Thanks.

PR #1480, opened 2021-08-02
[visualize] captureImage() method

ManifoldFR:
This commit adds a `captureImage()` method for visualizers. It returns an RGB array. A minimal example:

```python
import matplotlib.pyplot as plt

visualizer.display(q)
img: np.ndarray = visualizer.captureImage()

# display with matplotlib
plt.imshow(img)
plt.axis('off')
plt.show()
```

For now, it is only implemented for `MeshcatVisualizer`. This requires that rdeits/meshcat-python#93 be merged first. To work, this requires the visualizer window be already opened in the browser.

I also added an optional argument `capture` to the `play()` method. It makes the method return a list of ndarrays containing each captured image, which can be fed to something like `imageio` to produce an MP4 video.

### DIFF
--- a/bindings/python/pinocchio/visualize/meshcat_visualizer.py
+++ b/bindings/python/pinocchio/visualize/meshcat_visualizer.py
@@ -283,7 +283,7 @@ class MeshcatVisualizer(BaseVisualizer):
             return img_arr
         except AttributeError:
             warnings.warn("meshcat.Visualizer does not have the get_image() method."
-                          " Check your version of meshcat.")
+                          " You need meschat >= 0.2.0 to get this feature.")
 
 
     def displayCollisions(self,visibility):
--- a/bindings/python/pinocchio/visualize/meshcat_visualizer.py
+++ b/bindings/python/pinocchio/visualize/meshcat_visualizer.py
@@ -275,6 +275,18 @@ class MeshcatVisualizer(BaseVisualizer):
             # Update viewer configuration.
             self.viewer[visual_name].set_transform(T)
 
+    def captureImage(self):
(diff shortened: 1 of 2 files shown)


bindings/python/pinocchio/visualize/meshcat_visualizer.py
```diff
@@ -275,6 +275,17 @@ def display(self, q = None):
...
             self.viewer[visual_name].set_transform(T)
 
+    def captureImage(self):
+        """Capture an image from the Meshcat viewer and return an RGB array."""
+        try:
+            img = self.viewer.get_image()
+            img_arr = np.asarray(img)
+            return img_arr
+        except AttributeError:
+            warnings.warn("meshcat.Visualizer does not have the get_image() method."
```
jcarpent, line 285: Could you directly provide the minimal version of Meshcat here (next version)?

jcarpent (changes requested):
Thanks @ManifoldFR for this nice feature.
I made a minor request.

bindings/python/pinocchio/visualize/meshcat_visualizer.py
ManifoldFR: I'll ask rdeits if he can release on PyPI after merging and I'll report the version number

ManifoldFR: Ok, he updated the version to `0.2.0`: https://github.com/rdeits/meshcat-python/commit/46fe99c3044f788c84d45d2c70c0483be38f80b6

jcarpent (approved):

PR #1482, opened 2021-08-03
[CMake] set INSTALL_RPATH for python on linux

nim65s:
To avoid the need for python users to set `LD_LIBRARY_PATH` when the `CMAKE_INSTALL_PREFIX` is not a standard one.

ref. https://github.com/stack-of-tasks/eigenpy/pull/241


jcarpent (approved):
This sounds good to me. What do you mean by non-standard `CMAKE_PREFIX_PATH`?

nim65s:
I mean that if one use `cmake -DCMAKE_INSTALL_PREFIX=/usr/local` (the default), `/usr/local/lib` don't need to be in `LD_LIBRARY_PATH` nor `RPATH`, because this path is already configured in `/etc/ld.so.conf.d/libc.conf`. Same for `/usr`, but not for eg. `cmake -DCMAKE_INSTALL_PREFIX=$HOME/devel/install`

PR #1485, opened 2021-08-04
Fix issue with Pool when FCL is missing

jcarpent:


jcarpent:
Solving issue reported in #1485

PR #1487, opened 2021-08-05
fix module name

nim65s:
Hi,

Running all tests on a clean repo `clone --recursive` currently fails with:
```cpp
120/147 Test #147: example-cpp-codegen-crba ...........................Child aborted***Exception:   0.16 sec
Opening file: /pinocchio/models/others/robots/ur_description/urdf/ur5_robot.urdf
Error:   File /pinocchio/models/others/robots/ur_description/urdf/ur5_robot.urdf does not exist
         at line 55 in /build/urdfdom-VnCcob/urdfdom-1.0.4+ds/urdf_parser/src/model.cpp
terminate called after throwing an instance of 'std::invalid_argument'
  what():  The file /pinocchio/models/others/robots/ur_description/urdf/ur5_robot.urdf does not contain a valid URDF model.
  ```

It looks like this folder was changed in https://github.com/stack-of-tasks/pinocchio/commit/8b0d1471b1fa5097acde159e9cf3146c45270cdc , but not everywhere.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,6 +4,6 @@
 [submodule ".travis"]
 	path = .travis
 	url = https://github.com/jrl-umi3218/jrl-travis
-[submodule "models/others"]
+[submodule "models/example-robot-data"]
 	path = models/example-robot-data
 	url = https://github.com/Gepetto/example-robot-data.git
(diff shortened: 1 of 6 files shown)


jcarpent (approved):
Thanks for the fix.

PR #1488, opened 2021-08-06
Add RVizVisualizer

EtienneAr:
Basic implementation to use RViz as visualizer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,10 +101,10 @@ Additionnal materials were presented during the [Memmo](https://www.memmo-projec
 
 **Pinocchio** provides support for many open-source and free visualizers:
 
-  - [Gepetto Viewer](https://github.com/Gepetto/gepetto-viewer): a C++ viewer based on [OpenSceneGraph](https://github.com/openscenegraph/OpenSceneGraph) with Python bindings and Blender export. See [here](https://github.com/stack-of-tasks/pinocchio-gepetto-viewer) for a C++ example on mixing **Pinocchio** and **Gepetto Viewer**.
-  - [Meshcat](https://github.com/rdeits/meshcat): supporting visualization in Python and which can be embeded inside any browser.
-  - [Panda3d](https://github.com/ikalevatykh/panda3d_viewer): supporting visualization in Python and which can be embeded inside any browser.
-  - [RViz](https://github.com/ros-visualization/rviz): supporting visualization in Python and which can interact with other ROS packages.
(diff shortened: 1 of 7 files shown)


jcarpent (changes requested):
Very nice job @EtienneAr. 
Could you mention the default minimal command lines to install the rviz python bindings (e.g with conda) in the Readme?

EtienneAr:
@jcarpent, I updated both README in 863d3fc760f70534aebc73e973ca77d588b6574f

jcarpent (approved):
Sounds good to me.
I will try on my computer ;)

cmastalli (commented):
The marker Rviz plugin is not the ideal solution. Instead, you might look the whole-body state rviz plugin (look in loco-3d organisation).

This plugin can display robot meshes, contact forces, CoM position and velocity, COP, friction come, etc.

This plugin also can visualize trajectories. We are heavily using it in our robots.

jcarpent:
> The marker Rviz plugin is not the ideal solution. Instead, you might look the whole-body state rviz plugin (look in loco-3d organisation).
> 
> This plugin can display robot meshes, contact forces, CoM position and velocity, COP, friction come, etc.
> 
> This plugin also can visualize trajectories. We are heavily using it in our robots.

Please make comments in the code directly @cmastalli with links to precise codes in the package you have mentioned. It will definitely help @EtienneAr to improve his PR.

jcarpent:
@cmastalli The goal of this viewer is to be flexible, easy to use and mostly written in Python for easy and flexible extensions.

cmastalli:
> @cmastalli The goal of this viewer is to be flexible, easy to use and mostly written in Python for easy and flexible extensions.

You can use the whole-body state plugin with Python code. Just need to write a publisher that contains the whole-body state msgs.

Check out the plugin here: https://github.com/loco-3d/whole_body_state_rviz_plugin

If this solution is in your interested, I could provide more details on the code

cmastalli:
Btw, this plugin is based on Pinocchio ;)

jcarpent:
The thing is that we may want to add additional features (buttons, players, loading FLC geometry on the flight, etc.) and have the flexibility to create a GUI for specific applications without having to write C++ codes, as offered by RVIZ+Qt.
Your interface seems to be too much specific and linked to locomotion mostly.

jcarpent:
> Btw, this plugin is based on Pinocchio ;)

Pinocchio cannot rely on a lib ... which also depends on Pinocchio.

cmastalli:
The plugin applies to any robot. It uses urdf loaded in the parameter server. Thus, you could load your FCL geometry (i.e., custom urdf). A code that use this plugin will depend on the whole-body state msg only (so not recursive dependency).

jcarpent:
Working totally fine on my side. Thanks again @EtienneAr for this nice work.

PR #1519, opened 2021-08-31
[python] Add pybind11 header.

jmirabel:
Follows discussion https://github.com/stack-of-tasks/pinocchio/discussions/1518

I think the minimum C++ standard of my solution is 14. Not sure it works with 11.

@wxmerkt I would be happy to see your solution.

### DIFF
--- a/bindings/python/pybind11.hpp
+++ b/bindings/python/pybind11.hpp
@@ -1,0 +1,109 @@
+#include <pybind11/pybind11.h>
+
+#include <pinocchio/multibody/model.hpp>
+
+namespace pinocchio {
+namespace cpp2pybind11 {
+namespace bp = boost::python;
+
+template<typename T> pybind11::object to(T& t)
+{
+  // Create PyObject using boost Python
+  typename bp::reference_existing_object::apply<T*>::type converter;
+  PyObject* pyobj = converter(t);
+  // Create the Pybind11 object
+  return pybind11::reinterpret_borrow<pybind11::object>(pyobj);
+}
+template <typename T>
+auto to(T* t) {
+  // Create PyObject using boost Python
+  typename bp::manage_new_object::apply<T*>::type converter;
+  PyObject* pyobj = converter(t);
+  // Create the Pybind11 object
+  return pybind11::reinterpret_steal<pybind11::object>(pyobj);
+}
+
+template <typename ReturnType>
+ReturnType& from(pybind11::object model) {
+  return boost::python::extract<ReturnType&>(model.ptr());
+}
+
+template <typename T>
+struct convert_type {
+  typedef T type;
+  static inline auto _to(T t) { return t; }
+  static inline auto _from(type t) { return t; }
+};
+template <>
+struct convert_type<void> {
+  static inline void _to() {}
+};
+
+template <typename T>
+struct convert_boost_python_object {
+  typedef pybind11::object type;
+  static inline auto _to(T t) {
+    return to<typename std::remove_pointer<typename std::remove_reference<
+                  typename std::remove_cv<T>::type>::type>::type>(t);
+  }
+  static inline T _from(type t) {
+    return from<
+        typename std::remove_cv<typename std::remove_reference<T>::type>::type>(
+        t);
+  }
+};
+
+#define ADD_CONVERT_TYPE(CLASS)         \
+  template <>                           \
+  struct convert_type<pinocchio::CLASS> \
+      : convert_boost_python_object<pinocchio::CLASS> {}
+
+ADD_CONVERT_TYPE(Model*);
+ADD_CONVERT_TYPE(Model&);
+ADD_CONVERT_TYPE(Model const&);
+
+namespace internal {
+
+template <typename R, typename... Args>
+auto call(R (*f)(Args...), typename convert_type<Args>::type... args) {
+  return convert_type<R>::_to(f(convert_type<Args>::_from(args)...));
+}
+template <typename... Args>
+void call(void (*f)(Args...), typename convert_type<Args>::type... args) {
+  return f(convert_type<Args>::_from(args)...);
+}
+
+template <typename T>
+struct function_wrapper;
+
+template <typename R, typename... Args>
+struct function_wrapper<R (*)(Args...)> {
+  static const size_t nargs = sizeof...(Args);
+
+  typedef R result_type;
+
+  template <size_t i>
+  struct arg {
+    typedef typename std::tuple_element<i, std::tuple<Args...>>::type type;
+  };
+
+  typedef R (*func_type)(Args...);
+
+  func_type f;
+
+  auto operator()(typename convert_type<Args>::type... args) {
+    // return convert_type<R>::_to(f(convert_type<Args>::_from(args)...));
+    return call(f, args...);
+  }
+};
+}  // namespace internal
+
+template <typename R, typename... Args>
+internal::function_wrapper<R (*)(Args...)> make_function(R (*func)(Args...)) {
+  internal::function_wrapper<R (*)(Args...)> wrapper;
+  wrapper.f = func;
+  return wrapper;
+}
+
+}  // namespace cpp2pybind11
+}  // namespace pinocchio
--- a/unittest/python_pybind11/CMakeLists.txt
+++ b/unittest/python_pybind11/CMakeLists.txt
@@ -1,0 +1,13 @@
+cmake_minimum_required(VERSION 3.4)
+project(test_boost_pybind11 LANGUAGES CXX)
+
+find_package(pybind11 REQUIRED)
+find_package(pinocchio REQUIRED)
+
+execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "import sys; sys.stdout.write(sys.version[:3])"
+  OUTPUT_VARIABLE PYTHON_MAJOR_MINOR_VERSION)
+set(BOOST_PYTHON python${PYTHON_MAJOR_MINOR_VERSION})
+find_package(Boost 1.70 REQUIRED COMPONENTS ${BOOST_PYTHON})
+
+pybind11_add_module(cpp2pybind11 cpp2pybind11.cpp)
+target_link_libraries(cpp2pybind11 PRIVATE pinocchio::pinocchio Boost::${BOOST_PYTHON})
--- a/unittest/python_pybind11/cpp2pybind11.cpp
+++ b/unittest/python_pybind11/cpp2pybind11.cpp
@@ -1,0 +1,65 @@
+#include <pinocchio/multibody/model.hpp>
+
+#include <pybind11/pybind11.h>
+#include <boost/python.hpp>
+
+#include <pinocchio/bindings/python/pybind11.hpp>
+
+pinocchio::Model* make_model()
+{
+  pinocchio::Model* model = new pinocchio::Model;
+  std::cout << "make_model: " << reinterpret_cast<intptr_t>(model) << std::endl;
+  return model;
+}
+
+pinocchio::Model& return_same_model(pinocchio::Model& m)
+{
+  return m;
+}
+
+intptr_t get_ptr(pinocchio::Model& model)
+{
+  return reinterpret_cast<intptr_t>(&model);
+}
+
+void test1(int i)
+{
+  std::cout << "no conversion: " << ' ' << i << std::endl;
+}
+void testModel1(pinocchio::Model& model)
+{
+  std::cout << "testModel1: " << &model << std::endl;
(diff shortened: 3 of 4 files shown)


bindings/python/pybind11.hpp
```diff
@@ -0,0 +1,109 @@
+#include <pybind11/pybind11.h>
+
+#include <pinocchio/multibody/model.hpp>
+
+namespace pinocchio {
+namespace cpp2pybind11 {
```
jcarpent, line 6: Could you use python namespace here?

bindings/python/pybind11.hpp
```diff
@@ -0,0 +1,109 @@
...
+template<typename T> pybind11::object to(T& t)
+{
+  // Create PyObject using boost Python
+  typename bp::reference_existing_object::apply<T*>::type converter;
+  PyObject* pyobj = converter(t);
+  // Create the Pybind11 object
+  return pybind11::reinterpret_borrow<pybind11::object>(pyobj);
+}
+template <typename T>
+auto to(T* t) {
```
jcarpent, line 18: Could you remove auto and use standard C++98 rule?

bindings/python/pybind11.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+template <typename ReturnType>
+ReturnType& from(pybind11::object model) {
+  return boost::python::extract<ReturnType&>(model.ptr());
+}
+
+template <typename T>
+struct convert_type {
+  typedef T type;
+  static inline auto _to(T t) { return t; }
```
jcarpent, line 34: inline is not needed here.

bindings/python/pybind11.hpp
```diff
@@ -0,0 +1,109 @@
...
+};
+
+#define ADD_CONVERT_TYPE(CLASS)         \
+  template <>                           \
+  struct convert_type<pinocchio::CLASS> \
+      : convert_boost_python_object<pinocchio::CLASS> {}
+
+ADD_CONVERT_TYPE(Model*);
+ADD_CONVERT_TYPE(Model&);
+ADD_CONVERT_TYPE(Model const&);
```
jcarpent, line 63:
```suggestion
ADD_CONVERT_TYPE(Model const&);

undef  ADD_CONVERT_TYPE
```

bindings/python/pybind11.hpp
```diff
@@ -0,0 +1,109 @@
...
+    return from<
+        typename std::remove_cv<typename std::remove_reference<T>::type>::type>(
+        t);
+  }
+};
+
+#define ADD_CONVERT_TYPE(CLASS)         \
+  template <>                           \
+  struct convert_type<pinocchio::CLASS> \
+      : convert_boost_python_object<pinocchio::CLASS> {}
```
jcarpent, line 59:
```suggestion
#define ADD_CONVERT_TYPE(class)         \
  template <>                           \
  struct convert_type<::pinocchio::class> \
      : convert_boost_python_object<::pinocchio::class> {}
```

unittest/python_pybind11/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+cmake_minimum_required(VERSION 3.4)
+project(test_boost_pybind11 LANGUAGES CXX)
+
+find_package(pybind11 REQUIRED)
+find_package(pinocchio REQUIRED)
```
jcarpent, line 5: I do think this test should be inserted within the main Pinocchio tests, by loading the Pybind11 module from github

unittest/python_pybind11/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+
+execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "import sys; sys.stdout.write(sys.version[:3])"
+  OUTPUT_VARIABLE PYTHON_MAJOR_MINOR_VERSION)
+set(BOOST_PYTHON python${PYTHON_MAJOR_MINOR_VERSION})
+find_package(Boost 1.70 REQUIRED COMPONENTS ${BOOST_PYTHON})
```
jcarpent, line 10: Why the version should be this one? 
I think this should work for any version.

jcarpent (changes requested):
Thanks @jmirabel for providing this.
I made some comments that are easy to handle.

bindings/python/pybind11.hpp
jmirabel: With pybind11, the standard is 11 at least.

jcarpent: OK.

unittest/python_pybind11/CMakeLists.txt
jcarpent: From the previous comment, these settings are not needed.

jcarpent (commented):
@jmirabel 
Could you also add Doxygen info?

bindings/python/pybind11.hpp
```diff
@@ -0,0 +1,109 @@
+#include <pybind11/pybind11.h>
```
jcarpent, line 1:
```suggestion
#include <pybind11/pybind11.h>
#include <boost/python.hpp>
```

jmirabel:
I'll wait to see @wxmerkt proposal before addressing the requested changes.

> Could you also add Doxygen info?

You mean documenting functions in Doxygen format ?

jcarpent:
> You mean documenting functions in Doxygen format ?

Yes, exactly.

bindings/python/pybind11.hpp
jcarpent: For this macro, it would be nice to change its name and to let it be exposed to the user, as it is a nice shortcut.

jmirabel: This is what I intent to do. I don't think it is wise to call `ADD_CONVERT_TYPE` with all possible types here. It is better to let users add conversions for the type they need.

jcarpent: By the way, we might also add a file that applies the typedef for most common types.
this should be related to the template instantiation context which is not yet set up. See #1246

wxmerkt:
Apologies for the late response to this PR - this week is particularly busy and I wanted to experiment with combining the nice C++->Python caster with an implicit typecaster I've been using - this would avoid having to wrap function calls. I might get to it tomorrow evening or Friday afternoon - however, please don't consider my slow response blocking. Would that be alright or would you like to move ahead before? In general, this PR is very nice and clean, particularly the re-use when going from C++ back to Python.

jmirabel:
Before going further, I would prefer to see what you have. Just the C++ file without the whole project is fine. I think it is enough for me to grasp what you do exactly.

jcarpent:
@wxmerkt @jmirabel any update on this PR?

wxmerkt:
My apologies, I went quiet on this one as I didn't find time to explore this further (it's ICRA season).

What I've been doing creates a `type_caster` inside pybind11's detail namespace - this way we don't need to wrap the functions and as soon as the header is included we can bind methods that take pinocchio types as arguments using pybind11. I realised I never finished the C++-back-to-Python part; and directly plugging in Joseph's solution for this part does not appear to work, but I haven't had time to explore this further. I also haven't checked if the `load` method of the `type_caster` results in copies.

The code snippet for Python-to-C++ without wrapping the functions is:
```cpp
#define BOOST_PYTHON_TO_PYBIND11_TYPE_CASTER(native_type, boost_python_name)           \
    namespace pybind11                                                                 \
    {                                                                                  \
    namespace detail                                                                   \
    {                                                                                  \
    template <>                                                                        \
    struct type_caster<native_type>                                                    \
    {                                                                                  \
    public:                                                                            \
        PYBIND11_TYPE_CASTER(native_type, boost_python_name);                          \
        /* Python -> C++ */                                                            \
        bool load(handle src, bool)                                                    \
        {                                                                              \
            PyObject* source = src.ptr();                                              \
            value = boost::python::extract<native_type>(source);                       \
            return !PyErr_Occurred();                                                  \
        }                                                                              \
        /* C++ -> Python */                                                            \
        static handle cast(native_type src, return_value_policy policy, handle parent) \
        {                                                                              \
            /* TODO */                                                                 \
            throw std::runtime_error("C++ to Python not yet implemented!");            \
            return PyLong_FromLong(1);                                                 \
        }                                                                              \
    };                                                                                 \
    } /* namespace detail */                                                           \
    } /* namespace pybind11 */

BOOST_PYTHON_TO_PYBIND11_TYPE_CASTER(pinocchio::Model, _("pinocchio.pinocchio_pywrap.Model"));
// etc
```

jmirabel:
That's neat. I'll give a try to @wxmerkt and update my PR.

jcarpent:
Thanks @jmirabel and @wxmerkt for your contributions.

jmirabel:
> I also haven't checked if the `load` method of the `type_caster` results in copies.

I had a look and it results in a copy in both directions. Given the definition of `PYBIND11_TYPE_CASTER`, I don't see any workaround.

I think it is fine to provide both options:
- no copy with wrapper,
- copy without wrapper (developer-friendly but likely less user friendly).

jcarpent:
> My apologies, I went quiet on this one as I didn't find time to explore this further (it's ICRA season).
> 
> What I've been doing creates a `type_caster` inside pybind11's detail namespace - this way we don't need to wrap the functions and as soon as the header is included we can bind methods that take pinocchio types as arguments using pybind11. I realised I never finished the C++-back-to-Python part; and directly plugging in Joseph's solution for this part does not appear to work, but I haven't had time to explore this further. I also haven't checked if the `load` method of the `type_caster` results in copies.
> 
> The code snippet for Python-to-C++ without wrapping the functions is:
> 
> ```c++
> #define BOOST_PYTHON_TO_PYBIND11_TYPE_CASTER(native_type, boost_python_name)           \
>     namespace pybind11                                                                 \
>     {                                                                                  \
>     namespace detail                                                                   \
>     {                                                                                  \
>     template <>                                                                        \
>     struct type_caster<native_type>                                                    \
>     {                                                                                  \
>     public:                                                                            \
>         PYBIND11_TYPE_CASTER(native_type, boost_python_name);                          \
>         /* Python -> C++ */                                                            \
>         bool load(handle src, bool)                                                    \
>         {                                                                              \
>             PyObject* source = src.ptr();                                              \
>             value = boost::python::extract<native_type>(source);                       \
>             return !PyErr_Occurred();                                                  \
>         }                                                                              \
>         /* C++ -> Python */                                                            \
>         static handle cast(native_type src, return_value_policy policy, handle parent) \
>         {                                                                              \
>             /* TODO */                                                                 \
>             throw std::runtime_error("C++ to Python not yet implemented!");            \
>             return PyLong_FromLong(1);                                                 \
>         }                                                                              \
>     };                                                                                 \
>     } /* namespace detail */                                                           \
>     } /* namespace pybind11 */
> 
> BOOST_PYTHON_TO_PYBIND11_TYPE_CASTER(pinocchio::Model, _("pinocchio.pinocchio_pywrap.Model"));
> // etc
> ```

It should be easy to implement the C++ -> Python. @wxmerkt Do you need it?

jcarpent:
@jmirabel Could you fix the CI issues? I would suggest using a CMake external project to find PyBind11 instead of relying on the one installed on the system.

jmirabel:
It's in my todo list.

> I would suggest using a CMake external project to find PyBind11

You mean let CMake download the sources at configure time ?

jmirabel:
The error on Windows looks unrelated to this PR:
>pinocchio/math/fwd.hpp(69,5): error : expected unqualified-id [cpp2pybind11.vcxproj]
pinocchio/math/fwd.hpp(57,68): message : expanded from macro 'PINOCCHIO_OVERLOAD_MATH_BINARY_OPERATOR' [cpp2pybind11.vcxproj]
pinocchio/math/fwd.hpp(69,5): error : expected ')' [cpp2pybind11.vcxproj]
pinocchio/math/fwd.hpp(57,68): message : expanded from macro 'PINOCCHIO_OVERLOAD_MATH_BINARY_OPERATOR' [cpp2pybind11.vcxproj]
pinocchio/math/fwd.hpp(69,45): message : to match this '(' [cpp2pybind11.vcxproj]
pinocchio/math/fwd.hpp(69,45): error : expected ')' [cpp2pybind11.vcxproj]
pinocchio/math/fwd.hpp(69,45): message : to match this '(' [cpp2pybind11.vcxproj]
pinocchio/math/fwd.hpp(69,45): error : expected ')' [cpp2pybind11.vcxproj]
pinocchio/math/fwd.hpp(69,45): message : to match this '(' [cpp2pybind11.vcxproj]

jcarpent:
It is weird that this is only occurring on this test. Could you check this is not an issue of include?

PR #1533, opened 2021-09-28
[python] Fix BaseVisualizer constructor.

jmirabel:


nim65s (approved):

bindings/python/pinocchio/visualize/base_visualizer.py
```diff
@@ -27,12 +27,12 @@ def __init__(self, model = pin.Model(), collision_model = None, visual_model = N
         else:
             self.data = data
 
-        if collision_data is None:
+        if collision_data is None and self.collision_model is not None:
```
jmirabel, line 30: In Pinocchio 3, we could remove the data arguments.

jcarpent: Some applications are passing the data to have complete sync between the viewer and the computations.
This is a good feature I think.

PR #1536, opened 2021-10-21
make copy of supports during model cast

rubengrandia:
It seems that the coping of the member "supports" was forgotten in the cast of a model.


hrp2-14:
Hi ! This project doesn't usually accept pull requests on master. If this wasn't intentionnal, you
can change the base branch of this pull request to devel (No need to close it for that). Best, a bot.

PR #1541, opened 2021-10-29
Turn off automatic documentation generation

wxmerkt:
The documentation is generated automatically if Doxygen is installed,
which takes a considerable amount of time. Most users are likely to use
the online-hosted Doxygen by default and so setting the default for
INSTALL_DOCUMENTATION to false should be fine (as is done in EigenPy).
For distribution/binaries, we save ~36-77 MB of installed files.

I wonder, should the default in `post-project.cmake` be OFF?

... and while at it three separate changes that require review (and reduced our compiler warnings to zero):
- Update deprecated header ("deprecated")
- Remove unused typename definition from pybind11 caster ("unused variable")
- Added pybind11 bindings converter test to set of automatically run tests


jcarpent:
> I wonder, should the default in `post-project.cmake` be OFF?

I think the current behavior is the most generic but conservative one, impacting a lot of projects. I would keep it like it is now.
I don't have any strong opinion on setting it OFF by default for Pinocchio. 
The overall idea is also to automatically test that the documentation is compiling ;)

jcarpent (approved):
@jmirabel What is your opinion?

jcarpent:
I've solved the bug related to OSX CI failures. This bug is due to PyBind11, which seems to directly link the module against Python, which should be avoided on OSX/Linux systems.
I made a manual creation of the module which seems to work well.
In addition, I've added the support of ccache for efficiency.

jmirabel (approved):
LGTM.

Issue #1492, opened 2021-08-12
Homogenize GeometryObject and Frame


jmirabel:
Hi,

GeometryObject and Frame classes both have a parent joint and parent frame attributes but with different names. I keep forgetting which uses what.

I wonder whether it is a good idea to take the opportunity of version 3 to homogenize this. It's an API change, although it would be easy for users to update their code. Any opinion on this ?

jmirabel:
I am fine with any naming (though I would say it is easier for users if only one of Frame or GeometryObject API do change).

I agree there are similar information but there are fields (type, inertia) of Frame which do not apply to GeometryObject and they are conceptually different. I would keep then separated or I would introduce a top class `TreeItem` that has parent joint and frame, name and placement.

jcarpent:
OK. `previousFrame` seems fine for me. There is no ambiguity at all. Thanks for proposing the notion of `TreeItem`. Or `ModelItem` because we are not only dealing with trees.

jcarpent:
@proyan Could you handle it?

proyan:
Fixed by #1570

PR #1570, opened 2021-12-07
Introduce ModelItem as parent of Frame and GeometryObject

proyan:
This PR follows from the discussion on https://github.com/stack-of-tasks/pinocchio/issues/1492. It introduces the parent `ModelItem` for Frame and `GeometryObject` classes, with `parentJoint`, `parentFrame`, `placement` and `name` as its elements
 It proposes the following API changes:
1) `frame.parent`                   -> `frame.parentJoint` 
2) `frame.previousFrame`      -> `frame.parentFrame`
3) The constructor for `GeometryObject` had the sequence `(...parentFrame, parentJoint...)`. This is now switched, to match the other API where `parentJoint` comes first. This could be a big source of bugs.
4) In order to avoid any inadvertent bugs like this, the placement argument in `GeometryObject` constructor has also been moved from its original location. Thus, the users of pinocchio3 would have to update their `GeometryObject` objects, and should see that `parentFrame` and `parentJoint` have switched locations.

The new constructor for `GeometryObject` now is : (from `src/multibody/fcl.hpp`)
```c++
 186   GeometryObject(const std::string & name,                                                                                                                                                                    
 187                  const JointIndex parent_joint,                                                                                                                                                               
 188                  const FrameIndex parent_frame,                                                                                                                                                               
 189                  const SE3 & placement,                                                                                                                                                                       
 190                  const CollisionGeometryPtr & collision_geometry,                                                                                                                                             
 191                  const std::string & meshPath = "",                                                                                                                                                           
 192                  const Eigen::Vector3d & meshScale = Eigen::Vector3d::Ones(),                                                                                                                                 
 193                  const bool overrideMaterial = false,                                                                                                                                                         
 194                  const Eigen::Vector4d & meshColor = Eigen::Vector4d::Zero(),                                                                                                                                 
 195                  const std::string & meshTexturePath = "")                                                                                                                                                    
```


proyan:
This closes #1492.

jcarpent:
To avoid that, we should rely on reference.

jcarpent:
As discussed in #1570, we should keep parent for the joint parent id.

jcarpent:
In overall, we should do the minimal modifications to not break the API. Could you revise this PR @proyan?

src/multibody/model-item.hpp
```diff
@@ -0,0 +1,71 @@
...
+  template<typename Derived>
+  struct ModelItem : NumericalBase<Derived>
+  {
+    EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+
+    typedef typename traits<Derived>::Scalar Scalar;
+    enum { Options = traits<Derived>::Options };
+    typedef SE3Tpl<Scalar,Options> SE3;
+
+    /// \brief Name of the geometry object
```
jcarpent, line 22: The documentation is not aligned with the content.

proyan:
Regarding the API change, I am of the view that since we are doing a version upgrade, we should try to remove some of the ambiguous nomenclature that we had before. There are two choices then, break in `Frame`, or break in `GeometryObject`

I think `Frame::parent` falls in the ambiguous naming category, since it is not clear whether we are talking about parent joint or parent frame. `GeometryObject` has the better naming system here, with names `parentJoint` and `parentFrame`.

The other option, as you are suggesting, would be to keep `Frame::parent` and `Frame::previousFrame` as the future choices. While `Frame` class is used much more extensively in the repository than `GeometryObject`, the naming of GeometryObject is the better one in my opinion.

So for posterity, I suggest that we keep the `parentJoint` and `parentFrame` naming, and suffer a break in Frame API, and not the other way around.

cmastalli:
> 

I also like the proposed naming by @proyan.
But, in any case, Pinocchio 3 should aim to break as little as possible the API.
And in the cases that this is not possible, then it would be great to have a migration documentation.

src/multibody/model-item.hpp
proyan: I don't see any difference in the alignment. What is the issue that you see?

proyan:
I've added the deprecated signature for Frame::parent and Frame::previousFrame. The only remaining API break is now the changed argument order in constructor of `GeometryObject`.

I could simply add the deprecated original constructor signature, however, since the original and the new signature differ only in the order of arguments, accepting the deprecated signature might be a source of bugs. I wonder what the best practice here would be @nim65s ?

src/multibody/model-item.hpp
jcarpent: Geometry is too specific

proyan: Missed it. Should be fixed now

jcarpent:
We should keep both signatures if they are not in conflict.

src/multibody/fcl.hpp
```diff
@@ -115,24 +115,32 @@ enum GeometryType
...
+{
+  typedef double Scalar;
+  enum { Options = 0};
+};
+
+struct GeometryObject : public ModelItem<GeometryObject>
 {
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+
+  typedef GeometryObject ModelItemDerived;
```
jcarpent, line 131: It would be nice to not add typedef for nothing.

src/multibody/fcl.hpp
```diff
@@ -168,32 +173,29 @@ PINOCCHIO_COMPILER_DIAGNOSTIC_IGNORED_DEPRECECATED_DECLARATIONS
...
   /// \param[in] meshScale Scale of the mesh [if applicable].
   /// \param[in] overrideMaterial If true, this option allows to overrite the material [if applicable].
   /// \param[in] meshColor Color of the mesh [if applicable].
   /// \param[in] meshTexturePath Path to the file containing the texture information [if applicable].
   ///
   GeometryObject(const std::string & name,
-                 const FrameIndex parent_frame,
                  const JointIndex parent_joint,
-                 const CollisionGeometryPtr & collision_geometry,
+                 const FrameIndex parent_frame,
```
jcarpent, line 188: Please keep the two signatures

src/multibody/frame.hpp
```diff
@@ -60,34 +65,75 @@ namespace pinocchio
...
     /// \param[in] parent Index of the parent joint in the kinematic tree.
-    /// \param[in] previousFrame Index of the parent frame in the kinematic tree.
     /// \param[in] frame_placement Placement of the frame wrt the parent joint frame.
     /// \param[in] type The type of the frame, see the enum FrameType.
     /// \param[in] inertia Inertia info attached to the frame.
     ///
     FrameTpl(const std::string & name,
-             const JointIndex parent,
-             const FrameIndex previousFrame,
+             const JointIndex parentJoint,
```
jcarpent, line 73: You should keep the existing signatures

jcarpent (changes requested):

jcarpent:
The philosophy I try to bring in Pinocchio is not API break if they can be avoided. The good solution is to deprecate useless objects or signatures. @proyan could you follow this rule?

src/multibody/frame.hpp
jcarpent: Done

src/multibody/fcl.hpp
proyan: Done

src/multibody/fcl.hpp
proyan: DOne

jcarpent (approved):

src/parsers/sample-models.hxx
```diff
@@ -154,9 +154,9 @@ namespace pinocchio
         
         parentFrame = model.getBodyId(pre+"wrist1_body");
         GeometryObject wristBall(pre+"wrist_object",
-                                 parentFrame, model.frames[parentFrame].parent,
-                                 boost::shared_ptr<fcl::Sphere>(new fcl::Sphere(0.05)),
+                                 model.frames[parentFrame].parentJoint, parentFrame,
                                  SE3::Identity(),
+                                 boost::shared_ptr<fcl::Sphere>(new fcl::Sphere(0.05)),
```
jmirabel, line 159: Coming here a bit late but still worth for future changes. It is better to use `hpp::fcl::shared_ptr` for hpp-fcl objects.

PR #1582, opened 2022-01-04
[computeAllTerms] Add missing noalias to avoid temporary allocation

wxmerkt:
This commit adds a missing noalias operator which resulted in tempory
allocations when running computeAllTerms. Eliminating the temporary
allocation results in a ~11% speed-up when performing computeAllTerms.

Timings from timings.cpp, compiled with Clang++12 (Ubuntu 20.04 stock-Eigen):
Before: 15.6495us
After: 13.8686us


wxmerkt:
FYI @nim65s Gitlab-CI is failing due to an expired certificate

jcarpent (approved):
Good catch @wxmerkt. Thanks a lot.

jcarpent:
@wxmerkt Could you sync your PR on the devel branch?
We will do the manual sync on our side.

PR #1585, opened 2022-01-06
Missing `noalias`

duburcqa:


jcarpent (approved):
Thanks @duburcqa for the fix.

duburcqa:
By the way, is this algorithm documented somewhere (I mean, in a paper). Because people are pretending it does not exist https://arxiv.org/pdf/2010.01033v1.pdf. I compared your implementation and their pseudo-code algorithm and it is essentially the same. Intermediary computations are slightly different but the computational complexity is the same.

jcarpent:
Thanks for this remark and observation @duburcqa. You match a good point.
I developed this algorithm in 2017 (commits are the proof) when developing the analytical derivatives.
By the way, the code was made public in 2018.

A lot of algorithms are published in fact directly in Pinocchio, without passing by the publication phase (we are lacking time and so many algorithms would then need a publication).
With Nicolas and other colleagues, we do think that making public the code (by open sourcing the algorithms is sufficient for paternity on the idea) is already an act of publication, similar to an article. It is even better, as it comes with empirical working proof and direct use. Additionally, the rigid body dynamics algorithms in Pinocchio are very clear, following the wonderful nomenclature created by Roy Featherstone.

If you look at https://arxiv.org/pdf/2105.05102.pdf, you will see that the proposed algorithms totally matched the code of Pinocchio done 4 years ago.

Then, if you need to cite a work related to this algorithm, I would suggest using this bibtex entry:
```
@inproceedings{carpentier2018analytical,
  title = {Analytical Derivatives of Rigid Body Dynamics Algorithms},
  author = {Carpentier, Justin and Mansard, Nicolas},
  booktitle = {Robotics: Science and Systems},
  year = {2018}
}
```

duburcqa:
I totally agree with you except on one specific point. I do think writing an arxiv article compiling this kind of algorithms would make them more visible for the community. Myself I was not aware it was an original contribution as it is not explicitly mentioned anywhere to my knowledge. I think the authors of the article I mentioned did an honest work and we're not able to find previous work, which I think is a quite sad. A arxiv paper without peer review would probably require much less effort. I agree it is not good for being cited, but may be very helpful for online visibility.

Note that I don't use Coriolis matrix algorithm and I don't need to cite it, it was just out of curiosity.

pwensing:
Thanks, @duburcqa, for bringing this connection to our attention. I developed the algorithm back in 2014 and apologize to Justin that I didn’t recognize the relationship with the Pinocchio code before publication — I would certainly have acknowledged it if I had. It is admittedly challenging to keep up with released code and understand its relationship with the state of the art (even for production-quality code like Pinocchio) without corresponding documentation in the literature.

Justin and I have discussed, and we agree that there was independent development on this matter. I have updated the [arXiv paper](https://arxiv.org/abs/2010.01033) to note this, and to acknowledge the relationship between the Coriolis algorithms. The paper provides several theoretical links with previous ideas in the adaptive control literature, which I hope may be of interest to others. Some of those theoretical seeds recently helped further the utility of Pinocchio for yet more users (see #1663), so thanks for setting up that positive synergy as an indirect result of your post.

PR #1606, opened 2022-02-06
allow `robot_wrapper` to share data with its `viz`

### LINKED ISSUE #1605: `BaseVisualizer` shouldn't have its own data ?
Hi everyone !

I find it unexpected that the method `rebuildData` creates a new copy of the data:
https://github.com/stack-of-tasks/pinocchio/blob/593d4d43fded997bb9aa2421f4e55294dbd233c4/bindings/python/pinocchio/visualize/base_visualizer.py#L40-L43

Inside `RobotWrapper`, if I call `rebuildData` on it, I will get 2 source of data, one inside that instance and one inside its `viz`.
https://github.com/stack-of-tasks/pinocchio/blob/593d4d43fded997bb9aa2421f4e55294dbd233c4/bindings/python/pinocchio/robot_wrapper.py#L234-L239

I think we should have only one source or am I missing something ? I could send a pull request if you want, probably just something simple like:
```python
     def rebuildData(self): 
         """Re-build the data objects. Needed if the models were modified. 
         Warning: this will delete any information stored in all data objects.""" 
         self.data, self.collision_data, self.visual_data = createDatas(self.model, self.collision_model, self.visual_model) 
         if self.viz is not None: 
             self.viz.data = self.data
             self.viz.collision_data = self.collision_data
             self.viz.visual_data = self.visual_data 
```

vnghia:
For your information, @jcarpent 
Close #1605

### DIFF
--- a/bindings/python/pinocchio/robot_wrapper.py
+++ b/bindings/python/pinocchio/robot_wrapper.py
@@ -234,9 +234,21 @@ class RobotWrapper(object):
     def rebuildData(self):
         """Re-build the data objects. Needed if the models were modified.
         Warning: this will delete any information stored in all data objects."""
-        self.data, self.collision_data, self.visual_data = createDatas(self.model, self.collision_model, self.visual_model)
+        data, collision_data, visual_data = createDatas(self.model, self.collision_model, self.visual_model)
         if self.viz is not None:
-            self.viz.rebuildData()
+            if (
+                id(self.data) == id(self.viz.data)
+                and id(self.collision_data) == id(self.viz.collision_data)
+                and id(self.visual_data) == id(self.viz.visual_data)
+            ):
+                self.viz.data = data
+                self.viz.collision_data = collision_data
+                self.viz.visual_data = visual_data
+            else:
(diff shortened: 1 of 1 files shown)


bindings/python/pinocchio/robot_wrapper.py
```diff
@@ -262,12 +274,25 @@ def getViewerNodeName(self, geometry_object, geometry_type):
         """For each geometry object, returns the corresponding name of the node in the display."""
         return self.viz.getViewerNodeName(geometry_object, geometry_type)
 
-    def initViewer(self, *args, **kwargs):
+    def initViewer(self, shared_data=True, *args, **kwargs):
```
jmirabel, line 277: I find that *share_data* is a better name.

jmirabel (approved):

bindings/python/pinocchio/robot_wrapper.py
vnghia: Done, could you check it again ?

jcarpent (approved):

PR #1629, opened 2022-04-05
add dummy .pre-commit-config.yaml

nim65s:
Hi,

This PR prepare the repository for pre-commit.
It doesn't change anything in the repo, but allow us to configure pre-commit.ci on it.

This will help us configure pre-commit.ci on the whole stack-of-tasks organization, without messing with some of our main projects which are not yet ready for it (I will propose the same PR for TSID & Crocoddyl)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,8 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.1.0
+    hooks:
+    -   id: check-json
+    -   id: check-symlinks
+    -   id: check-toml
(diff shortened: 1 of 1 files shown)


jcarpent:
Hi @nim65s,

As discussed previously, I will add this feature together with Pinocchio 3.x ;)

nim65s:
I know, but If we don't add a dummy file like in this PR, and if we enable pre-commit.ci on the whole stack-of-tasks org, pre-commit will complain on each push / PR / release on pinocchio until pinocchio 3, which is going to be very boring.

This PR is not enabling clang-format or black yet, I know you intend to to that later.

PR #1631, opened 2022-04-07
Make Pinocchio v2 compatible with HPP-FCL v2

wxmerkt:
While I am aware behind the rationale of twinning HPP-FCL v2 with Pinocchio v3 for C++11 reasons in binary builds, there does not appear to be anything holding back using the latest HPP-FCL v2 in source builds with the Pinocchio devel branch. This PR fixes the one pointer issue and enables Pinocchio v2/devel to be compatible with either HPP-FCL v1 or v2.

### DIFF
--- a/src/multibody/geometry.hxx
+++ b/src/multibody/geometry.hxx
@@ -81,6 +81,28 @@ namespace pinocchio
   , outerObjects (other.outerObjects)
   {}
 
+  inline GeometryData& GeometryData::operator=(const GeometryData & other)
+  {
+    if (this != &other)
+    {
+      oMg = other.oMg;
+      activeCollisionPairs = other.activeCollisionPairs;
+#ifdef PINOCCHIO_WITH_HPP_FCL
+      distanceRequests = other.distanceRequests;
+      distanceResults = other.distanceResults;
+      collisionRequests = other.collisionRequests;
+      collisionResults = other.collisionResults;
+      radius = other.radius;
+      collisionPairIndex = other.collisionPairIndex;
+      collision_functors = other.collision_functors;
+      distance_functors = other.distance_functors;
+#endif // PINOCCHIO_WITH_HPP_FCL
+      innerObjects = other.innerObjects;
+      outerObjects = other.outerObjects;
+    }
+    return *this;
+  }
+
   inline GeometryData::~GeometryData() {}
 
   template<typename S2, int O2, template<typename,int> class JointCollectionTpl>
--- a/src/parsers/urdf/geometry.cpp
+++ b/src/parsers/urdf/geometry.cpp
@@ -165,9 +165,18 @@ namespace pinocchio
           bool convex = tree.isMeshConvex (linkName, geomName);
           if (convex) {
             bvh->buildConvexRepresentation (false);
+#if HPP_FCL_MAJOR_VERSION < 2
             geometry = bvh->convex;
-          } else
+#else
+            geometry = boost::shared_ptr<fcl::CollisionGeometry>(bvh->convex.get(), [bvh](...) mutable { bvh->convex.reset(); });
+#endif  // HPP_FCL_MAJOR_VERSION
+          } else {
+#if HPP_FCL_MAJOR_VERSION < 2
             geometry = bvh;
+#else
+            geometry = boost::shared_ptr<fcl::CollisionGeometry>(bvh.get(), [bvh](...) mutable { bvh.reset(); });
+#endif  // HPP_FCL_MAJOR_VERSION
+          }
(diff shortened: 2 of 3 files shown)


proyan (approved):
Thanks Wolf! I was having some of these issues in the pinocchio3 parser as well (since the urdf parser is carried off completely from pinv2).
This should help

CMakeLists.txt
```diff
@@ -169,7 +169,10 @@ ENDIF(BUILD_PYTHON_INTERFACE)
 IF(BUILD_WITH_HPP_FCL_SUPPORT)
   ADD_DEFINITIONS(-DPINOCCHIO_WITH_HPP_FCL)
   LIST(APPEND CFLAGS_DEPENDENCIES "-DPINOCCHIO_WITH_HPP_FCL")
-  ADD_PROJECT_DEPENDENCY(hpp-fcl 1.7.3 REQUIRED PKG_CONFIG_REQUIRES "hpp-fcl >= 1.7.3")
+  ADD_PROJECT_DEPENDENCY(hpp-fcl REQUIRED PKG_CONFIG_REQUIRES "hpp-fcl >= 1.7.3")
+  IF(${hpp-fcl_VERSION} VERSION_LESS "1.7.3")
```
jcarpent, line 173: I don’t understand why you need this change? ADD_PROJECT_DEPENDENCY should handle it internally. @nim65s could you confirm?

src/multibody/geometry.hxx
```diff
@@ -81,6 +81,28 @@ namespace pinocchio
   , outerObjects (other.outerObjects)
   {}
 
+  inline GeometryData& GeometryData::operator=(const GeometryData & other)
```
jcarpent, line 84: Why do you need to explicit it?

jcarpent (changes requested):
Thanks @wxmerkt for this PR. I’ve made some few comments

CMakeLists.txt
nim65s: CMake version checks are really boring on this topic. If you require X.Y.Z, and you have version A.B.C, it checks for `A == X && (B > Y || (B == Y && C >= Z))`.

I couldn't find a better alternative than the one @wxmerkt is using here when we released example-robot-data v4 and it broke everything just because 4 > 3, but I think we'll have to find a better solution at some point.

nim65s: Ok, I found the issue: https://github.com/jrl-umi3218/jrl-cmakemodules/blob/412628866e3613387c58e0961a05ce242dc93993/package-config.cmake#L167

jcarpent: Nice catch

jcarpent: So, we need to remove it. No?

src/multibody/geometry.hxx
wxmerkt: Since the copy constructor is defined we also need to define the operator= (`-Wdeprecated-copy`). There are a few more examples e.g. `LieGroupBase` that I haven't fixed yet.

src/parsers/urdf/geometry.cpp
```diff
@@ -165,9 +165,18 @@ namespace pinocchio
           bool convex = tree.isMeshConvex (linkName, geomName);
           if (convex) {
             bvh->buildConvexRepresentation (false);
+#if HPP_FCL_MAJOR_VERSION < 2
             geometry = bvh->convex;
-          } else
+#else
+            geometry = boost::shared_ptr<fcl::CollisionGeometry>(bvh->convex.get(), [bvh](...) mutable { bvh->convex.reset(); });
```
jmirabel, line 171: Why don't we use on `hpp::fcl::shared_ptr` ?

wxmerkt: Because it will require an API change on Pinocchio side - `boost::shared_ptr<fcl::CollisionGeometry>` is part of the defined API of `retrieveGeometry`

jmirabel: Does it make sense to do this change for pinocchio v3 ?

nim65s: I think pinocchio v3 should required hpp-fcl >= v2

jcarpent: > Does it make sense to do this change for pinocchio v3 ?

Quick answer: no.

CMakeLists.txt
jcarpent: I've updated the cmake module to account for that properly. It will require an update of hpp-fcl with a new release (on going).

jcarpent (approved):

PR #1637, opened 2022-04-25
follow up on hpp-fcl v2

nim65s:
Hi,

Following #1631, we should be able to use hpp-fcl v2, but this raise the following issue on various linux:
```python 
133/134 Test #133: example-py-collision-with-point-clouds .............***Failed    1.27 sec
  File "…/pinocchio/examples/collision-with-point-clouds.py", line 70, in <module>
    go_panda_hand.geometry.buildConvexRepresentation(False)
TypeError: No to_python (by-value) converter found for C++ type: boost::shared_ptr<hpp::fcl::CollisionGeometry>
```

The CI can reproduce it, as shown in this PR.

### DIFF
--- a/.github/workflows/conda/conda-env.yml
+++ b/.github/workflows/conda/conda-env.yml
@@ -4,7 +4,7 @@ channels:
   - nodefaults
 dependencies:
   - eigen
-  - hpp-fcl=1.8.0
+  - hpp-fcl
   - numpy
   - boost
   - eigenpy
--- a/bindings/python/multibody/geometry-object.hpp
+++ b/bindings/python/multibody/geometry-object.hpp
@@ -83,6 +83,7 @@ namespace pinocchio
           .staticmethod("CreateCapsule")
 #endif // PINOCCHIO_WITH_HPP_FCL
         ;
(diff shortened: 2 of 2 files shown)


nim65s:
And here is a quick fix. This is probably not covering all use cases, so I'll let the multiple CI run to see.

jcarpent (approved):

jcarpent:
Is there possible memory leak?

cmastalli:
> Is there possible memory leak?

The error reported by @nim65s appears when we have not registered a pointer (or shared pointer) as part of the Python object.

jcarpent:
Yes, I understand. But just hope we are guaranted to not have a memory leack

nim65s:
I guess shared_ptr should help us not having memory leaks, yes, even with this registration.

But to be honest, guaranteeing anything in C++ is far beyond my skills x)

I'm more concern in the use case where this is already registered elsewhere and we add that, but probably boost is smart enough not to make a fuss about that.

cmastalli:
Boost is doing properly if we are using boost shared pointers, as far I remember, the registration macro doesn't work with std shared pointer.

PR #1667, opened 2022-06-01
Fix method play() of rviz_visualizer

danielcostanzi18:
Method `play()` overridden, `Qtest.qWait()` used for the delay between one display to another instead of `time.sleep()`.
In the original implementation method `play()` form base_visualizer was used.

### DIFF
--- a/bindings/python/pinocchio/visualize/base_visualizer.py
+++ b/bindings/python/pinocchio/visualize/base_visualizer.py
@@ -79,6 +79,9 @@ class BaseVisualizer(object):
         """Captures an image from the viewer and returns an RGB array."""
         pass
 
+    def sleep(self, dt):
+        time.sleep(dt)
+
     def play(self, q_trajectory, dt, capture=False):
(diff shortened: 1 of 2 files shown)


hrp2-14:
Hi ! This project doesn't usually accept pull requests on master.

jcarpent:
Thanks @danielcostanzi18 for your contribution.
I made a slight modifications.

jcarpent (approved):

nim65s:
Please follow the dry principle (Don't Repeat Yourself).
Instead of copy/pasting the whole `play` method, you can define the `def sleep(self, dt)` method in the base class, with `time.sleep(dt)` in the base case, and call that instead of `time.sleep(dt)` in the `play` method.
Then, in the RViz derived class, all you have to do is to define another `def sleep(self, dt)` method with the required `qWait(dt * 1000)` instead

jcarpent:
@danielcostanzi18 COuld you handle the change requested by @nim65s?

PR #1676, opened 2022-06-11
Missing noalias in ABA.

duburcqa:
AFAIK, `noalias` does not handle sum of products automatically, unless they are explicitly defined as lazy products `A.lazyProduct(B) + C.lazyProduct(D)`.

### DIFF
--- a/src/algorithm/aba.hxx
+++ b/src/algorithm/aba.hxx
@@ -138,16 +138,17 @@ namespace pinocchio
       typedef typename Data::Force Force;
 
       const JointIndex & i = jmodel.id();
-      const JointIndex & parent  = model.parents[i];
+      const JointIndex & parent = model.parents[i];
       typename Inertia::Matrix6 & Ia = data.Yaba[i];
 
-      jmodel.jointVelocitySelector(data.u) -= jdata.S().transpose()*data.f[i];
+      jmodel.jointVelocitySelector(data.u) -= jdata.S().transpose() * data.f[i];
       jmodel.calc_aba(jdata.derived(), Ia, parent > 0);
 
       if (parent > 0)
       {
         Force & pa = data.f[i];
-        pa.toVector() += Ia * data.a_gf[i].toVector() + jdata.UDinv() * jmodel.jointVelocitySelector(data.u);
+        pa.toVector().noalias() += Ia * data.a_gf[i].toVector();
+        pa.toVector().noalias() += jdata.UDinv() * jmodel.jointVelocitySelector(data.u);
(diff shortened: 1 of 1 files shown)


jcarpent (changes requested):
@duburcqa Thanks for the contribution.
Could you only commit lines related to the `noalias`? It will simplify the history then.
By the way, did you really observe a particular improvement?
On my side, it was not the case when I tested it when implementing ABA, but maybe with newer EIgen versions, it does matter.

duburcqa:
> By the way, did you really observe a particular improvement?

Hard to tell without proper benchmark on my side, by anyway it is not a big deal for me since ABA is not taking a huge amount of time relative to the whole program. BUT strangely it slightly changes the result. I guess due to rounding of errors. I don't know if it is better or worst, but I have some flaky RL unit tests failing after this modification. 

> On my side, it was not the case when I tested it when implementing ABA, but maybe with newer EIgen versions, it does matter.

Yes, I just thought the idea was to always use noalias if possible in Pinocchio, without further consideration. But maybe I misunderstood.

jcarpent:
> Hard to tell without proper benchmark on my side, by anyway it is not a big deal for me since ABA is not taking a huge amount of time relative to the whole program. BUT strangely it slightly changes the result. I guess due to rounding of errors. I don't know if it is better or worst, but I have some flaky RL unit tests failing after this modification.

It is really weird, because the numerical rounding should not affect so much your RL algorithms, as they play at totally different scale compared to the zero-th gradient estimates that your are making with RL.

jcarpent:
> Yes, I just thought the idea was to always use noalias if possible in Pinocchio, without further consideration. But maybe I misunderstood.

Yes and no. At some point, it is sometimes better to do a copy ... Here, all the quantities are of fixed dimensions, so there is no heap allocation ...

duburcqa:
> they play at totally different scale compared to the zero-th gradient estimates that your are making with RL.

Yes definitely, but they are already very flaky in the first place, just changing the seed is enough to make some of them fail either on Linux, windows or Macos.

> Here, all the quantities are of fixed dimensions, so there is no heap allocation ...

I heard that it was faster for small size arrays, but maybe it is only valid for dynamically allocated memory.

We can close if you think it is not worth the effort.

jcarpent:
I would certainly need to run extra benches to clearly acknowledge the benefits of this PR.
Yet, I have very little time in the next few weeks for that.
I will close and reopen it if needed.

jcarpent:
Thanks @duburcqa for opening this discussion.

PR #1702, opened 2022-07-02
SO3 diff impl use quaternion instead of rotation matrix

Toefinder:
It seems that using quaternion multiplication (`q0.conjugate()*q1`) then converting to matrix is actually more accurate than using matrix multiplication (`q0.matrix().transpose()*q1.matrix()`) in calculating the difference between `q1` and `q0`. 

When two SO3 configurations are very similar, with coordinates differing approximately by the smallest floating point division for `double` (I checked this using `numpy.spacing(value)` for the `numpy.float64` value of the quaternion coordinate), quaternion multiplication followed by converting to matrix gives a non-identity result (non-zero distance) while the matrix conversion followed by matrix multiplication gives an identity matrix.

### DIFF
--- a/src/multibody/liegroup/special-orthogonal.hpp
+++ b/src/multibody/liegroup/special-orthogonal.hpp
@@ -383,7 +383,7 @@ namespace pinocchio
       assert(quaternion::isNormalized(quat1,RealScalar(PINOCCHIO_DEFAULT_QUATERNION_NORM_TOLERANCE_VALUE)));
       
       PINOCCHIO_EIGEN_CONST_CAST(Tangent_t,d)
-        = log3((quat0.conjugate() * quat1).matrix().eval());
+        = quaternion::log3(Quaternion_t(quat0.conjugate() * quat1));
     }
 
     template <ArgumentPosition arg, class ConfigL_t, class ConfigR_t, class JacobianOut_t>
@@ -398,15 +398,16 @@ namespace pinocchio
       ConstQuaternionMap_t quat1 (q1.derived().data());
       assert(quaternion::isNormalized(quat1,RealScalar(PINOCCHIO_DEFAULT_QUATERNION_NORM_TOLERANCE_VALUE)));
       
-      const Matrix3 R = (quat0.conjugate() * quat1).matrix();
+      const Quaternion_t quat_diff = quat0.conjugate() * quat1;
 
       if (arg == ARG0) {
         JacobianMatrix_t J1;
-        Jlog3 (R, J1);
+        quaternion::Jlog3(quat_diff, J1);
+        const Matrix3 R = (quat_diff).matrix();
 
         PINOCCHIO_EIGEN_CONST_CAST(JacobianOut_t,J).noalias() = - J1 * R.transpose();
       } else if (arg == ARG1) {
-        Jlog3 (R, J);
+        quaternion::Jlog3(quat_diff, J);
       }
     }
 
--- a/src/multibody/liegroup/special-orthogonal.hpp
+++ b/src/multibody/liegroup/special-orthogonal.hpp
@@ -383,7 +383,7 @@ namespace pinocchio
       assert(quaternion::isNormalized(quat1,RealScalar(PINOCCHIO_DEFAULT_QUATERNION_NORM_TOLERANCE_VALUE)));
       
       PINOCCHIO_EIGEN_CONST_CAST(Tangent_t,d)
-        = log3((quat0.matrix().transpose() * quat1.matrix()).eval());
+        = log3((quat0.conjugate() * quat1).matrix().eval());
     }
 
     template <ArgumentPosition arg, class ConfigL_t, class ConfigR_t, class JacobianOut_t>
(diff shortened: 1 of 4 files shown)


jcarpent:
Thanks @Toefinder. Thanks for contributing to Pinocchio.
We also found that there was some numerical issues in log. In fact, this is coming from the log itself, and not really from the matrix multiplications.

In Pinocchio-3x, we have largely improved the numerical accuracies of this operation while also making it differentiable for autodiff. Could you wait for the new release of Pinocchio-3x?

Best,
Justin

Toefinder:
> Thanks @Toefinder. Thanks for contributing to Pinocchio. We also found that there was some numerical issues in log. In fact, this is coming from the log itself, and not really from the matrix multiplications.

I am surprised. When debugging, I find that the matrix multiplication results in an identity matrix and then that identity matrix is passed into the `log3` function. Using quaternion multiplication then converting into matrix does not have the same issue.

jcarpent:
Could you provide a tiny example in Python of what you are saying please?

Toefinder:
This issue was encountered when used in [hpp-core](https://github.com/humanoid-path-planner/hpp-core/issues/277#issue-1288718790) but I will give the full python example below:

```python3
In [1]: import pinocchio as pin
In [2]: import numpy as np

In [3]: x = pin.XYZQUATToSE3((0,0,0,0,0,-0.1953711450011105244,0.980729379442134
   ...: 9169))

In [4]: y = pin.XYZQUATToSE3((0,0,0,0,0,-0.19537114500111049664,0.98072937944213
   ...: 492244))

In [5]: np.matmul(y.rotation.T, x.rotation)
Out[5]: 
array([[1., 0., 0.],
       [0., 1., 0.],
       [0., 0., 1.]])

In [7]: quat_x = pin.Quaternion(0,0,-0.1953711450011105244,0.9807293794421349169
   ...: )

In [8]: quat_y = pin.Quaternion(0,0,-0.19537114500111049664,0.980729379442134922
   ...: 44)

In [9]: (quat_y.conjugate()*quat_x).matrix()
Out[9]: 
array([[ 1.00000000e+00, -0.00000000e+00,  0.00000000e+00],
       [ 0.00000000e+00,  1.00000000e+00,  5.55111512e-17],
       [-0.00000000e+00, -5.55111512e-17,  1.00000000e+00]])
```

unittest/joint-configurations.cpp
```diff
@@ -282,7 +282,7 @@ BOOST_AUTO_TEST_CASE ( integrate_difference_test )
  Eigen::VectorXd q1(randomConfiguration(model, -1 * Eigen::VectorXd::Ones(model.nq), Eigen::VectorXd::Ones(model.nq) ));
  Eigen::VectorXd qdot(Eigen::VectorXd::Random(model.nv));
 
- BOOST_CHECK_MESSAGE(isSameConfiguration(model, integrate(model, q0, difference(model, q0,q1)), q1), "Integrate (difference) - wrong results");
+ BOOST_CHECK_MESSAGE(isSameConfiguration(model, integrate(model, q0, difference(model, q0,q1)), q1, 1.5e-12), "Integrate (difference) - wrong results");
```
jcarpent, line 285: Why did you change the prec here?

jcarpent:
Now I understand your issue. We approve your fix, while I think it has been already done in Pinocchio-3x.
Could you just answer my only precision concern?

unittest/joint-configurations.cpp
Toefinder: With the new diff_impl, `integrate(q0, difference(q0,q1))` is less similar to the expected result `q1`. I am not totally sure why this is the case, but probably because the current
implementation of integrate also suffers from floating point error... The threshold for the integrate_difference_test needs to be changed slightly to accommodate the new implementation of difference.

jcarpent: It is highly surprising and this needs additional investigation for this PR to be merged. Could you look at it?

src/multibody/liegroup/special-orthogonal.hpp
```diff
@@ -383,7 +383,7 @@ namespace pinocchio
       assert(quaternion::isNormalized(quat1,RealScalar(PINOCCHIO_DEFAULT_QUATERNION_NORM_TOLERANCE_VALUE)));
       
       PINOCCHIO_EIGEN_CONST_CAST(Tangent_t,d)
-        = log3((quat0.matrix().transpose() * quat1.matrix()).eval());
+        = log3((quat0.conjugate() * quat1).matrix().eval());
```
jcarpent, line 386:
```suggestion
        = quaternion::log3(quat0.conjugate() * quat1);
```

jcarpent: This should save some precision roundings.

unittest/joint-configurations.cpp
Toefinder: I will need help for this since I am not too familiar with the mathematics :sweat_smile: For this issue I actually found the alternative implementation thanks to your `// TODO: perform first the Quaternion multiplications and then return a Rotation Matrix` note under `dDifference_impl` in `special-orthogonal.hpp`.

jcarpent: Could you see with @jmirabel? I guess you are already in contact with him.

src/multibody/liegroup/special-orthogonal.hpp
Toefinder: do we need the `.eval()`? 
like `quaternion::log3((quat0.conjugate() * quat1).eval());`

jcarpent:
I've solved the issues concerning accurary. I've in fact imported the fixes we made with @ManifoldFR on Pinocchio 3.

PR #1710, opened 2022-07-10
Add contact state in constrained dynamics

mayataka:
Dear all,

Thank you for continuing to provide a wonderful tool.

Regarding the new constrained-dynamics feature, I personally think that it may be useful if I can set the contact status.
That is, `std::vector<RigidConstraintModel>` is a collection of contact candidates and we describe which contacts are actually active through `RigidConstraintData::is_active`. 
The idea is also summarized in `test_with_inactive_contacts` in `unittest/constrained-dynamics.cpp`.
This is just my personal idea and I do not have confidence that others also agree with it.
This draft pull request is just a question of how the developers think about this.

Thank you

### DIFF
--- a/src/algorithm/constrained-dynamics.hxx
+++ b/src/algorithm/constrained-dynamics.hxx
@@ -267,143 +267,142 @@ namespace pinocchio
       const RigidConstraintModel & contact_model = contact_models[contact_id];
       RigidConstraintData & contact_data = contact_datas[contact_id];
 
-      if (contact_data.is_active) 
+      if(!contact_data.is_active) continue;
+
+      const int contact_dim = contact_model.size();
+      
+      const typename RigidConstraintModel::BaumgarteCorrectorParameters & corrector = contact_model.corrector;
+      const typename RigidConstraintData::Motion & contact_acceleration_desired = contact_data.contact_acceleration_desired;
+      typename RigidConstraintData::Motion & contact_acceleration_error = contact_data.contact_acceleration_error;
+        
+      const typename Model::JointIndex joint1_id = contact_model.joint1_id;
+      typename RigidConstraintData::SE3 & oMc1 = contact_data.oMc1;
+      typename RigidConstraintData::Motion & vc1 = contact_data.contact1_velocity;
+      typename RigidConstraintData::Motion & coriolis_centrifugal_acc1 = contact_data.contact1_acceleration_drift;
+      
+      const typename Model::JointIndex joint2_id = contact_model.joint2_id;
+      typename RigidConstraintData::SE3 & oMc2 = contact_data.oMc2;
+      typename RigidConstraintData::Motion & vc2 = contact_data.contact2_velocity;
+      typename RigidConstraintData::Motion & coriolis_centrifugal_acc2 = contact_data.contact2_acceleration_drift;
+      
+      const typename RigidConstraintData::SE3 & c1Mc2 = contact_data.c1Mc2;
+      
+      // Compute contact placement and velocities
+      if(joint1_id > 0)
+        vc1 = oMc1.actInv(data.ov[joint1_id]);
+      else
+        vc1.setZero();
+      if(joint2_id > 0)
+        vc2 = oMc2.actInv(data.ov[joint2_id]);
+      else
+        vc2.setZero();
+      
+      // Compute placement and velocity errors
+      if(contact_model.type == CONTACT_6D)
       {
-        const int contact_dim = contact_model.size();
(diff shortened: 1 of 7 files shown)


src/algorithm/constrained-dynamics.hxx
```diff
@@ -250,140 +266,144 @@ namespace pinocchio
...
-      
-      // Compute placement and velocity errors
-      if(contact_model.type == CONTACT_6D)
-      {
-        contact_data.contact_placement_error = -log6(c1Mc2);
-        contact_data.contact_velocity_error.toVector() = (vc1 - c1Mc2.act(vc2)).toVector();
-      }
-      else
+
+      if (contact_data.is_active)
```
jcarpent, line 270: Could you simply do not execute the rest of the code if the contact is not active, as follows?
```suggestion

      if (!contact_data.is_active) continue;
```

jcarpent (changes requested):
Dear @mayataka,

Thanks for contributing to Pinocchio.
Just to let you know that we have a private parallel development which is far beyond this branch, and it will simplify things.
What I don't understand is why you are not only selecting the active contact and passing them to Pinocchio. This is highly simple from a user perspective, no?

In addition, you have moved some blocks like typedef, etc. and this makes the review a bit complicated, while only minor changes can have been made to introduce your need.

Before moving further, I would need to understand why is_active is needed, as it implies some online memory allocation, which can be avoided if the user maintained on his own the current list of active contacts?

src/algorithm/constrained-dynamics.hxx
mayataka: > Just to let you know that we have a private parallel development which is far beyond this branch, and it will simplify things.

I understood. I may have to wait for the update.

> What I don't understand is why you are not only selecting the active contact and passing them to Pinocchio. This is highly simple from a user perspective, no?

It is of course possible including in my use case.
My suggestion is that it is also useful if it is supported within Pinocchio.
So I want to ask if this is worth introducing as the new feature at the expense of simplisity.

> Before moving further, I would need to understand why is_active is needed, as it implies some online memory allocation, which can be avoided if the user maintained on his own the current list of active contacts?

Regarding memory allocation, I also wonder if we can avoid memory re-allocation when changing the constraint dimension. 
That is, we first reserve the possible memory (large Eigen matrix) according to the number of contact candidates in `initConstraintDynamics`.
And in `constraintDynamics`, we use `Block` of the reserved memory.
This can make the implementation complicated, so I avoid including it in this pull request.

mayataka:
> Just to let you know that we have a private parallel development which is far beyond this branch, and it will simplify things.

I understood. I may have to wait for the update.

> What I don't understand is why you are not only selecting the active contact and passing them to Pinocchio. This is highly simple from a user perspective, no?

It is of course possible including in my use case.
My suggestion is that it is also useful if it is supported within Pinocchio.
So I want to ask if this is worth introducing as the new feature at the expense of simplisity.

> Before moving further, I would need to understand why is_active is needed, as it implies some online memory allocation, which can be avoided if the user maintained on his own the current list of active contacts?

Regarding memory allocation, I also wonder if we can avoid memory re-allocation when changing the constraint dimension. 
That is, we first reserve the possible memory (large Eigen matrix) according to the number of contact candidates in `initConstraintDynamics`.
And in `constraintDynamics`, we use `Block` of the reserved memory.
This can make the implementation complicated, so I avoid including it in this pull request.

jcarpent:
> Regarding memory allocation, I also wonder if we can avoid memory re-allocation when changing the constraint dimension.
> That is, we first reserve the possible memory (large Eigen matrix) according to the number of contact candidates in `initConstraintDynamics`.
> And in `constraintDynamics`, we use `Block` of the reserved memory.
> This can make the implementation complicated, so I avoid including it in this pull request.

Yes, of course. I suggest waiting the new release for that.

jcarpent:
> It is of course possible including in my use case.
> My suggestion is that it is also useful if it is supported within Pinocchio.
> So I want to ask if this is worth introducing as the new feature at the expense of simplisity.

We can then envisage putting your contributions in this current branch and we will rewrite it in the new development branch for the next major release. Is it fine for you?

mayataka:
> We can then envisage putting your contributions in this current branch and we will rewrite it in the new development branch for the next major release. Is it fine for you?

I would appreciate it if you kindly do so. 
But it seems troublesome for you and I want to re-consider about this after the next major release as I do not see the new implementation.

mayataka:
I'd like to close this draft as I look forward to the new major release.
Thank you for the discussion.

PR #1712, opened 2022-07-12
Add header only target

fabinsch:


nim65s:
I guess you probably need something like 
```cmake
TARGET_INCLUDE_DIRECTORIES(${PROJECT_NAME}_headers PUBLIC $<INSTALL_INTERFACE:include>)
```

Did you test this, by installing this project and trying something like `target_link_libraries(test_project PUBLIC pinocchio::pinocchio_headers)` ?

fabinsch:
Thanks @nim65s @jcarpent for your suggestions and advice. I added a new packaging unittest verifying that we can use the new pinocchio_headers interface.

jcarpent (approved):
Perfect! Thanks a lot, @fabinsch for this nice work.

PR #1716, opened 2022-07-14
cmake: relocatable package for recent CMake versions

nim65s:
Hi,

This follows https://github.com/stack-of-tasks/eigenpy/pull/291 & https://github.com/humanoid-path-planner/hpp-fcl/pull/319

Comments there also apply here.

I also removed a couple of `INCLUDE_DIRECTORIES`, but I think this won't break anything. The CI should check that :)

My editor was configured to remove end of line spaces, so this PR is best reviewed by using the "hide whitespace" feature, available [here](https://github.com/stack-of-tasks/pinocchio/pull/1716/files?w=1)


nim65s:
CI raises an issue with `test-cpp-python_parser` which requires libpython.so.
I think we should replace `BUILD_PYTHON_INTERFACE_WITH_PYPY` by an option set before `FINDPYTHON()` call, to get the right `PYTHON_COMPONENTS` there.

jcarpent:
> CI raises an issue with `test-cpp-python_parser` which requires libpython.so. I think we should replace `BUILD_PYTHON_INTERFACE_WITH_PYPY` by an option set before `FINDPYTHON()` call, to get the right `PYTHON_COMPONENTS` there.

This is find with me.

jcarpent (approved):

jcarpent:
Sounds good to me! Thanks a lot @nim65s.

Issue #1737, opened 2022-08-22
Relocatable Library


Booooooosh:
Hi, we have a use case where this library is built under one machine (A) and used on another machine (B). Unfortunately, machine A and B have different paths and machine B isn't able to properly load interface libraries since the path under `INTERFACE_INCLUDE_DIRECTORIES` are absolute path from machine A.

One potential root cause of this is during `TARGET_LINK_LIBRARIES` or `TARGET_INCLUDE_DIRECTORIES` we are getting library path from variables such as `${EIGEN3_INCLUDE_DIR}` [here](https://github.com/stack-of-tasks/pinocchio/blob/master/src/CMakeLists.txt#L56). While I'm digging into this I'm not able to find the definition of `MODERNIZE_TARGET_LINK_LIBRARIES`. Wonder if this is an known issue.

Thank you!

nim65s:
Hi,
This library should be easily relocatable.

Could you provide your:
- OS
- eigen version
- cmake version
- which absolute paths and where are they in the packages build on (A)

nim65s:
Well, clearly, if you still have `TARGET_INCLUDE_DIRECTORIES(${PROJECT_NAME}_headers SYSTEM INTERFACE ${EIGEN3_INCLUDE_DIR})` somewhere, this is can't be relocatable. You should update your version.

jcarpent:
@Booooooosh COuld check whether https://github.com/stack-of-tasks/pinocchio/pull/1738 fixes your issue?

nim65s:
If I understand correctly, #1737is already fixed by #1718, but @Booooooosh is not up to date.

jcarpent:
Not really, as the target exports were containing hard coded path. #1738 Solves this.

PR #1720, opened 2022-07-27
CMake: update to eigenpy 2.7.10

nim65s:
This will have to wait for eigenpy 2.7.10 to be available everywhere before the CI can succeed.

PR #1738, opened 2022-08-24
Fix relocalable

jcarpent:
Related to #1737.


nim65s (commented):
I don't understand the link #1737, where the issue is about eigen, and here you tweak eigenpy. 

I also think the modification of eigenpy::eigenpy link to MODERNIZE_TARGET_LINK_LIBRARIES is useless.

I guess the removal of the prefix will probably break things, but the CI seems to say otherwise.

The removal of target_link_boost_python is a good idea, I forgot it in #1720

bindings/python/CMakeLists.txt
```diff
@@ -130,19 +130,18 @@ IF(BUILD_PYTHON_INTERFACE)
...
+  MODERNIZE_TARGET_LINK_LIBRARIES(${PYWRAP} SCOPE PUBLIC
+    TARGETS eigenpy::eigenpy)
   IF(BUILD_WITH_OPENMP_SUPPORT)
     TARGET_COMPILE_OPTIONS(${PYWRAP} PRIVATE ${OpenMP_CXX_FLAGS})
     TARGET_COMPILE_DEFINITIONS(${PYWRAP} PRIVATE -DPINOCCHIO_PYTHON_INTERFACE_WITH_OPENMP)
     TARGET_INCLUDE_DIRECTORIES(${PYWRAP} SYSTEM PRIVATE ${OpenMP_CXX_INCLUDE_DIR})
     IF(LINK_PYTHON_INTERFACE_TO_OPENMP)
       TARGET_LINK_LIBRARIES(${PYWRAP} PRIVATE ${OpenMP_CXX_LIBRARIES})
     ENDIF(LINK_PYTHON_INTERFACE_TO_OPENMP)
-    MESSAGE(STATUS "OpenMP_CXX_LIBRARIES: ${OpenMP_CXX_LIBRARIES}")
```
cmastalli, line 140: It seems this got removed by mistake, right?

jcarpent: It was for my own debug purpose. This is not needed.

Issue #1741, opened 2022-09-06
Mesh fails to load when filename attribute doesn't start with package://


stephane-caron:
The description [``cf2p.urdf``](https://github.com/utiasDSL/gym-pybullet-drones/blob/master/gym_pybullet_drones/assets/cf2p.urdf) for the Crazyflie 2.0 drone does not load properly with pinocchio 2.6.3:

```
ValueError: Could not load resource ./cf2.dae
Unable to open file "./cf2.dae".
Hint: the mesh directory may be wrong.
```

I believe I have configured the mesh path properly. Here is why:

- The URDF loads the mesh file using a relative path: ``<mesh filename="./cf2.dae" scale=" 1 1 1"/>``
- If we add the ``package://`` prefix: ``<mesh filename="package://./cf2.dae" scale=" 1 1 1"/>``, then the mesh loads properly and can e.g. be visualized in MeshCat:

![image](https://user-images.githubusercontent.com/1189580/188667499-0d922224-d8ae-4172-835a-6b75cd14681a.png)

Is there indeed an inconsistency in the way pinocchio treats the optional ``package://`` prefix?

(Relevant [part of the URDF spec](https://wiki.ros.org/urdf/XML/link).)

jcarpent:
I think relative paths are not supported by the convention.
Package relates to some known placement for the ROS ecosystem.

At first glance, we should be able to robustify to also handle relative path wrt the current file. @fabinsch Could you handle this issue?

stephane-caron:
Thanks for checking :slightly_smiling_face: Unfortunately there are several URDFs out there that use relative paths, I just took that one as an example.

From the sentence you quote, what the spec says is: "**if** you want to make the path relative to a package, **then** use ``package://``". Unfortunately this is incomplete as it does not specify whether relative paths without ``package://`` are allowed, and if so to what directory they are relative to.

> However, I added the ability in Pinocchio to handle this kind of relative path in https://github.com/stack-of-tasks/pinocchio/pull/1742.

Looks perfect :+1:

jcarpent:
Solved by #1742.

PR #1742, opened 2022-09-07
Fix relative path in urdf

fabinsch:
Currently, we cannot import a urdf that specifies a relative mesh path without using `//package`, see https://github.com/stack-of-tasks/pinocchio/issues/1741.

### DIFF
--- a/src/parsers/utils.hpp
+++ b/src/parsers/utils.hpp
@@ -67,6 +67,7 @@ namespace pinocchio
 
     const std::string separator("://");
     const std::size_t pos_separator = string.find(separator);
+    bf::path string_path(string);
 
     if (pos_separator != std::string::npos)
     {
@@ -99,7 +100,7 @@ namespace pinocchio
         throw std::invalid_argument(exception_message);
       }
     }
-    else if (string.substr(0,2).compare("./") == 0)
+    else if (string_path.is_relative())
     {
       // handle the case where a relative mesh path is specified without using //package
       for (std::size_t i = 0; i < package_dirs.size(); ++i)
--- a/src/parsers/utils.hpp
+++ b/src/parsers/utils.hpp
@@ -99,6 +99,18 @@ namespace pinocchio
         throw std::invalid_argument(exception_message);
       }
     }
+    else if (string.substr(0,2).compare("./") == 0)
+    {
+      // handle the case where a relative mesh path is specified without using //package
+      for (std::size_t i = 0; i < package_dirs.size(); ++i)
+        {
+          if ( bf::exists( bf::path(package_dirs[i] + "/" + string.substr(2))))
+          {
+            result_path = std::string( package_dirs[i] + "/" + string.substr(2));
+            break;
+          }
+        }
+    }
     else // return the entry string
     {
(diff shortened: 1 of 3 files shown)


src/parsers/utils.hpp
```diff
@@ -99,6 +99,18 @@ namespace pinocchio
         throw std::invalid_argument(exception_message);
       }
     }
+    else if (string.substr(0,2).compare("./") == 0)
```
jcarpent, line 102: This seems not to be robust depending on the OS. Could you use rather Boost.Filesystem for that?
Additionally, it will be nice to add a test of the fix.

jcarpent (changes requested):

src/parsers/utils.hpp
fabinsch: thanks for the hint. I adapted it to use `Boost.Filesystem`.

jcarpent (commented):
I think we should also link the main pinocchio lib to boost.filesystem, which seems only the case for Windows. Could you confirm?

fabinsch:
> I think we should also link the main pinocchio lib to boost.filesystem, which seems only the case for Windows. Could you confirm?

It's fine, we are already linking to boost.filesystem for all platforms (https://github.com/stack-of-tasks/pinocchio/blob/master/src/CMakeLists.txt#L60).

fabinsch:
@wxmerkt we have a problem with the [ROS CI](https://github.com/stack-of-tasks/pinocchio/runs/8225315773?check_suite_focus=true#step:4:544) and to me it seems that the eigenpy version are not the good ones. Could you have a look please or give me a hint what I can do to fix it ? Thank you

wxmerkt:
> @wxmerkt we have a problem with the [ROS CI](https://github.com/stack-of-tasks/pinocchio/runs/8225315773?check_suite_focus=true#step:4:544) and to me it seems that the eigenpy version are not the good ones. Could you have a look please or give me a hint what I can do to fix it ? Thank you

The EigenPy versions on ROS Melodic/18.04 were reverted to an older release since dynamic-graph hadn't been fixed / re-released. Once dynamic-graph/SoT/TSID are ready for release, we can re-release EigenPy & HPP-FCL on Melodic.

@nim65s have you created releases to Melodic with the fixed packages? If so, I can get EigenPy & HPP-FCL back to the latest version

nim65s:
The releases are done, but now they require recent versions of eigenpy… Can you already put back the latest eigenpy versions ? Or how should we proceed ?

fabinsch:
I've added a unittest @jcarpent . I looked into using the `pinocchio::urdf::buildModelFromXML` function in order to not add a "new" urdf file but for now we are not loading geometry objects with this function. Therefore I added another urdf which is a copy of the simple humanoid and I changed to path to the mesh from `"package://simple_humanoid_description/box.stl"` to `"./simple_humanoid_description/box.stl"`. Test is passing locally.

wxmerkt:
EigenPy/HPP-FCL release tracker for 18.04/Melodic: https://github.com/ros/rosdistro/pull/34371

PR #1744, opened 2022-09-13
Sync submodule CMake

jcarpent:


jcarpent:
@wxmerkt Any hint why melodic is not able to find hpp-fcl? Thanks in advance for your feedback.

wxmerkt:
We need a new release of HPP-FCL with the fixed CMake (current devel with updated cmake submodule) and then I can run a release to have it available on Melodic as well

jcarpent:
OK. Perfect. I will do that now then. Thanks a lot for the quick feedback. Very helpful

wxmerkt:
@jcarpent Thank you for the swift release. The problem should be resolved now, I've restarted the build

Issue #1747, opened 2022-09-15
URDF relative paths that don't start with "./"


stephane-caron:
The update from https://github.com/stack-of-tasks/pinocchio/pull/1742 addressed https://github.com/stack-of-tasks/pinocchio/issues/1741, but people are inventive and there is another relative-path syntax out there :wink: 

There are currently 5 / 45 descriptions configured in [robot_descriptions.py](https://github.com/stephane-caron/robot_descriptions.py) that use relative paths:

| Description | Pinocchio 2.6.4 | Pinocchio 2.6.10 |
|--|--|--|
| `cf2_description` | ❌ | :heavy_check_mark: |
| `laikago_description` | ❌ | :heavy_check_mark: |
| `mini_cheetah_description` | ❌ | ❌ |
| `minitaur_description` | ❌ | ❌ |
| `pr2_description` | ❌ | ❌ |

https://github.com/stack-of-tasks/pinocchio/pull/1742 solved descriptions that do ``<mesh filename="./foo/bar">``, but it doesn't work on the remaining 3 because they do ``<mesh filename="foo/bar">``.

jcarpent:
@fabinsch COuld you handle it?

stephane-caron:
I don't have a Pinocchio build environment to test it right now, but the fix may be simple.

In:

```cpp
          if ( bf::exists( bf::path(package_dirs[i] + "/" + string.substr(2))))
          {
            result_path = std::string( package_dirs[i] + "/" + string.substr(2));
            break;
          }
```

Can the calls to ``substr(2)`` be removed?  Substrings like "/./" (or "/./.././../") shouldn't hurt the path.

Also note that a URDF may do ``<mesh filename="../foo/bar">``.

fabinsch:
Hey @stephane-caron , thanks for raising the issue. You are right, and the solution that you're proposing works. I created a PR #1748 which should fix your loading issues in https://github.com/stephane-caron/robot_descriptions.py/issues/6.

jcarpent:
@stephane-caron I have merged #1748. Could you check if it is now working well on your side?

stephane-caron:
> Could you check if it is now working well on your side?

Sorry this one went out of my radar. I've opened a follow up at https://github.com/stack-of-tasks/pinocchio/issues/1786

PR #1748, opened 2022-09-15
URDF: fix loading relative mesh path in urdf

fabinsch:
this solves #1747.

We are now handling not only `./rel-mesh-path` but also `rel-mesh-path` and `../rel-mesh-path`.


jcarpent (approved):

jcarpent:
Thanks a lot @fabinsch for the quick fix.

Issue #1752, opened 2022-09-26
Difference() behavior not consistent between joint types


stephane-caron:
As of Pinocchio 2.6.4, ``pin.difference`` doesn't behave the same for different joint types when infinity is involved:

### Free flyers:

```python
>>> import pinocchio as pin
>>> model = pin.Model()
>>> model.addJoint(0, pin.JointModelFreeFlyer(), pin.SE3.Identity(), "root_joint")
1
>>> model.upperPositionLimit
array([1.79769313e+308, 1.79769313e+308, 1.79769313e+308, 1.79769313e+308,
       1.79769313e+308, 1.79769313e+308, 1.79769313e+308])
>>> pin.difference(model, pin.neutral(model), model.upperPositionLimit)
array([nan, nan, nan, nan, nan, nan])
```

### Revolute joints:

```python
>>> import pinocchio as pin
>>> model = pin.Model()
>>> model.addJoint(0, pin.JointModelRZ(), pin.SE3.Identity(), "foo")
1
>>> model.upperPositionLimit
array([1.79769313e+308])
>>> pin.difference(model, pin.neutral(model), model.upperPositionLimit)
array([1.79769313e+308])
```

### Expected behavior:

I think the second behavior is the right one, as it is expected that "+infinity - finite = +infinity". (Also, that's the one that would trivialize computing joint limits in IK, which is my current use case :wink:)

cmastalli:
I believe it should report `inf` for undefined limits. Moreover, the result of `diff` should be also `inf`.

jcarpent:
If you have an unbounded joint, then you should instead use the unbounded joint type, which is done for that.
The default value for the joint limits is +inf: the results you obtained come from internal computations.
We cannot guarantee a fixed output, as it really depends on the compiler in use.

I would say that it is the role of the end-user to properly use the correct input and output of the difference operations (which act on Lie groups).

jcarpent:
For the case of FreeFlyer, the bounds on the rotation part have no meaning in general. The NaN comes from there.

stephane-caron:
OK, I see the stakes better. I agree it would make sense to have well-formed configurations as a pre-requisite for ``difference``. Then the next thing to do would be to document this assumption, e.g. with a ``@note`` stating that the function's behavior is unspecified otherwise. (Ideally the type system could actually check this assumption for us, rather than delegating the task to end users.)

By (ii), do you mean changing the type of position limits to be something more structured than a configuration vector?

[...]

PR #1756, opened 2022-09-29
Configuration limits for joints and model

fabinsch:
- every joint returns a vector of boolean of size nq describing if each of the DOF has bounds or not
- model: vector stacking the boolean vector of configurations limits of each joint in the model

now we can do:
```
In [1]: import pinocchio as pin
   ...: model = pin.Model()
   ...: model.addJoint(0, pin.JointModelFreeFlyer(), pin.SE3.Identity(), "root_joint")
   ...: model.addJoint(0, pin.JointModelRZ(), pin.SE3.Identity(), "joint1")
   ...: model.addJoint(0, pin.JointModelRUBZ(), pin.SE3.Identity(), "joint2")
   ...: model.configurationLimit.tolist()
Out[1]: [True, True, True, False, False, False, False, True, False, False]
```

which can be used to see where bounds are expected or not using e.g. `model.upperPositionLimit[model.configurationLimit.tolist()]` and check if these are meaningful joint bounds, as suggested in #1752 .

### DIFF
--- a/bindings/python/multibody/joint/joint-derived.hpp
+++ b/bindings/python/multibody/joint/joint-derived.hpp
@@ -33,7 +33,7 @@ namespace pinocchio
         .add_property("idx_v",&get_idx_v)
         .add_property("nq",&get_nq)
         .add_property("nv",&get_nv)
-        .add_property("has_configuration_limit", &JointModelDerived::hasConfigurationLimit,
+        .add_property("hasConfigurationLimit", &JointModelDerived::hasConfigurationLimit,
              "Return vector of boolean if joint has configuration limits.")
         .def("setIndexes",
              &JointModelDerived::setIndexes,
--- a/bindings/python/multibody/joint/joint-derived.hpp
+++ b/bindings/python/multibody/joint/joint-derived.hpp
@@ -33,7 +33,7 @@ namespace pinocchio
         .add_property("idx_v",&get_idx_v)
         .add_property("nq",&get_nq)
         .add_property("nv",&get_nv)
-        .add_property("configuration_limit", &JointModelDerived::hasConfigurationLimit,
+        .add_property("has_configuration_limit", &JointModelDerived::hasConfigurationLimit,
              "Return vector of boolean if joint has configuration limits.")
         .def("setIndexes",
              &JointModelDerived::setIndexes,
--- a/bindings/python/multibody/joint/joint-derived.hpp
+++ b/bindings/python/multibody/joint/joint-derived.hpp
@@ -33,8 +33,12 @@ namespace pinocchio
         .add_property("idx_v",&get_idx_v)
         .add_property("nq",&get_nq)
         .add_property("nv",&get_nv)
-        .add_property("hasConfigurationLimit", &JointModelDerived::hasConfigurationLimit,
-             "Return vector of boolean if joint has configuration limits.")
+        .add_property("hasConfigurationLimit", 
+                      &JointModelDerived::hasConfigurationLimit,
+                      "Return vector of boolean if joint has configuration limits.")
+        .add_property("hasConfigurationLimitInTangent",
+                      &JointModelDerived::hasConfigurationLimitInTangent,
+                      "Return vector of boolean if joint has configuration limits in tangent space.")
         .def("setIndexes",
              &JointModelDerived::setIndexes,
              bp::args("self","id","idx_q","idx_v"))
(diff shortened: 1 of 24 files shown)


bindings/python/multibody/joint/joint-derived.hpp
```diff
@@ -33,6 +33,8 @@ namespace pinocchio
         .add_property("idx_v",&get_idx_v)
         .add_property("nq",&get_nq)
         .add_property("nv",&get_nv)
+        .add_property("configuration_limit", &JointModelDerived::hasConfigurationLimit,
```
stephane-caron, line 36: Maybe ``"has_configuration_limit"`` so that the API names match (if not their case)?

stephane-caron:
In the IK joint-limit use case https://github.com/stack-of-tasks/pinocchio/issues/1752#issuecomment-1260773339, what we compute in the end is the same thing but in tangent rather than configuration space; something like:

```python
v_max[np.hstack([joint.has_configuration_limit_in_tangent_space.tolist() for joint in model.joints]))
```

This is beyond the scope of this PR, but I'm not sure I understand why planar (rotation), spherical or free-flyer (rotations) shouldn't have limits. In general, spherical joints can have arbitrary boundaries. For example in HRP-2's hips there was once a polygonal description ``A * [phi, theta] <= b``.

bindings/python/multibody/model.hpp
```diff
@@ -125,6 +125,7 @@ namespace pinocchio
         .add_property("nvs",&Model::nvs)          
         .add_property("parents",&Model::parents)
         .add_property("names",&Model::names)
+        .add_property("configurationLimit",&Model::configurationLimit)
```
jcarpent, line 128:
```suggestion
        .add_property("hasConfigurationLimit",&Model::hasConfigurationLimit)
```

bindings/python/multibody/joint/joint.hpp
```diff
@@ -31,6 +31,8 @@ namespace pinocchio
         .add_property("idx_v",&getIdx_v)
         .add_property("nq",&getNq)
         .add_property("nv",&getNv)
+        .def("has_configuration_limit", &JointModel::hasConfigurationLimit,
```
jcarpent, line 34:
```suggestion
        .def("hasConfigurationLimit", &JointModel::hasConfigurationLimit,
```

bindings/python/multibody/joint/joint-derived.hpp
```diff
@@ -33,6 +33,8 @@ namespace pinocchio
         .add_property("idx_v",&get_idx_v)
         .add_property("nq",&get_nq)
         .add_property("nv",&get_nv)
+        .add_property("has_configuration_limit", &JointModelDerived::hasConfigurationLimit,
```
jcarpent, line 36:
```suggestion
        .add_property("hasConfigurationLimit", &JointModelDerived::hasConfigurationLimit,
```

jcarpent (changes requested):
I think we need to uniformize the namings.
I made some suggestions. @fabinsch Could you handle the rest?

stephane-caron:
Hi guys. This is just a proposal. Since there can be joints with partial configuration limits, it would be convenient if Pinocchio also exported a tangent vector, like e.g. for the free flyer:

```cpp
    const std::vector<bool> JointModelFreeFlyerTpl::hasConfigurationLimitTangent() const
    {
      return {true, true, true, false, false, false};
    }
```

Here is the kind of things this tangent vector could be used for:

```python
has_configuration_limit = np.logical_and(  # runtime configuration limit
    model.upperPositionLimit < 1e20,
    model.upperPositionLimit > model.lowerPositionLimit + 1e-10,
)

bounded_joints = [
    joint
    for joint in model.joints
    if has_configuration_limit[
        # this slice could be partial thanks to hasConfigurationLimit
        slice(joint.idx_q, joint.idx_q + joint.nq)
    ].all()
]

is_tangent_of_bounded_joint = np.full(model.nv, False)
for joint in bounded_joints:
    is_tangent_of_bounded_joint[
        # this slice could be partial with hasConfigurationLimitTangent
        slice(joint.idx_v, joint.idx_v + joint.nv)
    ] = True
```

Alternatively, if we had a map from configuration to tangent coordinates https://github.com/stack-of-tasks/pinocchio/issues/1758, we could forget about ``hasConfigurationLimitTangent`` and simply do:

```python
is_velocity_of_bounded_joint = np.full(model.nv, False)
for i in range(model.nq):
    if has_configuration_limit[i]:
        is_velocity_of_bounded_joint[model.config_to_tangent(i)] = True
```

jcarpent (approved):
Thanks @fabinsch for this nice work.

PR #1764, opened 2022-10-09
Address remaining warnings + speed-up build for BUILD_TESTING=OFF

wxmerkt:
Speed-up when not building tests should be ~30% by conditionally compiling `cpp2pybind11` following the same logic as `add_unit_test`

### DIFF
--- a/src/multibody/liegroup/liegroup-base.hpp
+++ b/src/multibody/liegroup/liegroup-base.hpp
@@ -634,6 +634,11 @@ PINOCCHIO_LIE_GROUP_PUBLIC_INTERFACE_GENERIC(Derived,typename)
     ///
     /// Prevent the copy of derived class.
     LieGroupBase( const LieGroupBase & /*clone*/) {}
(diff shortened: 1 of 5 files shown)


src/multibody/liegroup/liegroup-base.hpp
```diff
@@ -634,6 +634,11 @@ PINOCCHIO_LIE_GROUP_PUBLIC_INTERFACE_GENERIC(Derived,typename)
     ///
     /// Prevent the copy of derived class.
     LieGroupBase( const LieGroupBase & /*clone*/) {}
+    void operator=(const LieGroupBase & /*x*/) {}
```
jcarpent, line 637: It should return `LieGroupBase &` instead of `void`.

jcarpent (commented):
Thanks for the PR ;)
Everything looks good to me except one point.
@wxmerkt Could you handle it?

jcarpent (approved):

PR #1775, opened 2022-10-24
Error when integrating SE3 with the same vector given as input and output

duburcqa:
I don't think it would impede performance much since quaternions are fixed-size but you probably know better than me.

### DIFF
--- a/unittest/liegroups.cpp
+++ b/unittest/liegroups.cpp
@@ -44,8 +44,9 @@ std::ostream& operator<< (std::ostream& os, const LieGroupGenericTpl<LieGroupCol
 template <typename T>
 void test_lie_group_methods (T & jmodel, typename T::JointDataDerived &)
 {
+  typedef typename LieGroup<T>::type LieGroupType;
   typedef double Scalar;
-  
+
   const Scalar prec = Eigen::NumTraits<Scalar>::dummy_precision();
   BOOST_TEST_MESSAGE ("Testing Joint over " << jmodel.shortname());
   typedef typename T::ConfigVector_t  ConfigVector_t;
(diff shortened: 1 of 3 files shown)


jcarpent (commented):
Before digging further, could you provide a reproducible issue example?
This is worth adding as a unit test to enhance the Pinocchio library.

duburcqa:
You can clearly see that without copying the original quaternion in a temporary the chack that is done later after update the quaternion to determine if the sign has been reversed (`const Scalar dot_product = res_quat.dot(quat);`) becomes useless as it always returns exactly `1.0` because `res_quat` and `quat` are precisely pointing to the same memory space :)

jcarpent:
Could you add such an example then in https://github.com/stack-of-tasks/pinocchio/blob/devel/unittest/joint-configurations.cpp? It will allow to not reproduce such bugs in the future.

jcarpent:
Thanks in advance @duburcqa for your help.

duburcqa:
The test [here](https://github.com/duburcqa/pinocchio/blob/d94de14421f6118710bce9e95ab574447fb596af/unittest/liegroups.cpp#L457) is already checking this. Just that random sampling is not a sufficient without many retry. I suggest modifying the test to check many random configuration to be generic. What do you think ?

jcarpent:
This is fine for me to add the randomization of the test to be sure to output a failure case.
Thanks a lot @duburcqa.

duburcqa:
I will have a look at it on sunday I think. I'm quite busy by then.

jcarpent:
@duburcqa Do you have any idea when do you plan to finish this PR?

duburcqa:
I will do it tomorrow. Sorry I have been quite busy lately.

jcarpent:
No problem, thanks a lot for the short notice.

duburcqa:
I updated the unit tests. It fails as expected without the proposed patch. I had to increase the tolerance because it was now failing otherwise due to running evaluating the tests 50 times with different samples now.

unittest/liegroups.cpp
```diff
@@ -91,19 +91,7 @@ void test_lie_group_methods (T & jmodel, typename T::JointDataDerived &)
...
-  {
-    typename T::JointDataDerived jdata = jmodel.createData();
-    jmodel.calc(jdata, qTest);
-    MTest = jdata.M;
-  }
-  {
-    typename T::JointDataDerived jdata = jmodel.createData();
-    jmodel.calc(jdata, qResult);
-    MResult = jdata.M;
-  }
```
jcarpent, line 104: It would be nice also to keep it.

jcarpent (approved):
@duburcqa Thanks for finishing this PR.
The PR looks to me, even if I'm surprised by the fact you need to largely increase the tolerance. This should be investigated further in the future.

unittest/liegroups.cpp
```diff
@@ -138,12 +126,12 @@ void test_lie_group_methods (T & jmodel, typename T::JointDataDerived &)
 
   // Check difference
```
jcarpent, line 127:
```suggestion
  // Check difference
  // TODO(jcarpent): check the increase of tolerance.
```

unittest/liegroups.cpp
jcarpent: I would say this is complementary ;) always better for coverage.

duburcqa:
I'm also concerned about the updated tolerance thresholds. I could eventually find some time to investigate this issue.

jcarpent:
I've shortly investigated the issue of this PR, and it is, in fact, solved by a more advanced formula present in Pinocchio 3x.

jcarpent (dismissed):

jcarpent (approved):

PR #1791, opened 2022-11-16
Skip collision pairs between geom on same joint in appendGeometryModel

jmirabel:
The change comes with a unit test which was failing before and which is self-explanatory.


jcarpent (approved):

jmirabel:
Should I do something about gitlab CI failure ? It seems unrelated.

fabinsch:
we have the same gitlab CI failure also for [PRs](https://github.com/stack-of-tasks/tsid/pull/178) on TSID currently, I think @nim65s is on it.

nim65s:
I though I fixed that 2 days ago, sorry for the inconvenience. It should be OK now.

PR #1796, opened 2022-11-23
Add helper functions for computing supported inertia and force by frames

EtienneAr:
Hello,
Following my discussion with @jcarpent, I propose the implementation of 3 new helper functions that compute supported inertia and force by frames.

Motivation : 
In our current robot setup, we have a force-torque sensor mounted in the arm. We would like to compute the forces that this sensor is supposed to measure.

Problem : 
This sensor is represented by a FIXED_JOINT in the urdf, thus it is not represented in the pinocchio model, and retrieving such info is not straightforward. 

Solution :
The method `pinocchio::computeFrameSupportedForce` compute this force. It takes into account inertial forces, gravity and external forces.
Computing supported inertias is an intermediary step in that process. I choose to expose it, as it may be useful for other uses.

### DIFF
--- a/src/algorithm/frames.hpp
+++ b/src/algorithm/frames.hpp
@@ -291,7 +291,7 @@ namespace pinocchio
   template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
   InertiaTpl<Scalar, Options>
   computeSupportedInertiaByFrame(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
-                                 DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                                 const DataTpl<Scalar,Options,JointCollectionTpl> & data,
                                  const FrameIndex frame_id,
                                  bool with_subtree);
 
@@ -324,8 +324,8 @@ namespace pinocchio
   template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
   ForceTpl<Scalar, Options>
   computeSupportedForceByFrame(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
-                             DataTpl<Scalar,Options,JointCollectionTpl> & data,
-                             const FrameIndex frame_id);
+                               const DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                               const FrameIndex frame_id);
 
 } // namespace pinocchio
 
--- a/src/algorithm/frames.hpp
+++ b/src/algorithm/frames.hpp
@@ -242,7 +242,7 @@ namespace pinocchio
   }
   
   ///
-  /// \brief Computes the Jacobian time variation of a specific frame (given by frame_id) expressed either in the WORLD frame (rf = WORLD) or in the LOCAL frame (rf = LOCAL) or in the LOCAL_WORLD_ALIGNED frame (rf = LOCAL_WORLD_ALIGNED).
+  /// \brief Computes the Jacobian time variation of a specific frame (given by frame_id) expressed either in the LOCAL frame.
   ///
   /// \note This jacobian is extracted from data.dJ. You have to run pinocchio::computeJointJacobiansTimeVariation before calling it.
   ///
@@ -252,7 +252,6 @@ namespace pinocchio
   /// \param[in] model The model structure of the rigid body system.
   /// \param[in] data The data structure of the rigid body system.
   /// \param[in] frameId The index of the frame.
-  /// \param[in] rf Reference frame in which the Jacobian is expressed.
   ///
   /// \param[out] dJ A reference on the Jacobian matrix where the results will be stored in (dim 6 x model.nv). You must fill dJ with zero elements, e.g. dJ.fill(0.).
   ///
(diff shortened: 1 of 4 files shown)


src/algorithm/frames.hpp
```diff
@@ -263,6 +262,87 @@ namespace pinocchio
...
+  * @tparam JointCollection Collection of Joint types.
+  *
+  * @param[in] model The model structure of the rigid body system.
+  * @param[in] frameId The index of the frame.
+  *
+  * @return The computed inertia.
+  */
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
+  InertiaTpl<Scalar, Options>
+  computeFrameSupportedInertiaInBody(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
```
jcarpent, line 284: `computeFrameSupportedInertia` seems sufficient.

src/algorithm/frames.hpp
```diff
@@ -263,6 +262,87 @@ namespace pinocchio
...
+  InertiaTpl<Scalar, Options>
+  computeFrameSupportedInertiaWithSubtree(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                                        DataTpl<Scalar,Options,JointCollectionTpl> & data,
+                                        const FrameIndex frame_id);
+
+/**
+  * @brief Computes the force supported by a specific frame (given by frame_id) expressed in the LOCAL frame.
+  *        The supported force corresponds to the sum of all the forces experienced after the given frame, i.e :
+  *         * The inertial forces and gravity (applied on the supported inertia in body)
+  *         * The forces applied by child joint
```
jcarpent, line 318:
```suggestion
  *         * The forces applied by child joints
```

src/algorithm/frames.hpp
```diff
@@ -263,6 +262,87 @@ namespace pinocchio
...
+  * @param[in] frameId The index of the frame.
+  *
+  * @return The computed force.
+  *
+  * @warning pinocchio::rnea should have been called first
+  */
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
+  ForceTpl<Scalar, Options>
+  computeFrameSupportedForce(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                             DataTpl<Scalar,Options,JointCollectionTpl> & data,
```
jcarpent, line 343: Is the data modified? If not, use const ref here.

src/algorithm/frames.hpp
jcarpent: maybe using the naming `computeInertiaSupportedByFrame` is even more explicit.

src/algorithm/frames.hpp
```diff
@@ -263,6 +262,87 @@ namespace pinocchio
...
+  * @param[in] data The data structure of the rigid body system.
+  * @param[in] frameId The index of the frame.
+  *
+  * @return The computed inertia.
+  *
+  * @warning forwardKinematics should have been called first
+  */
+  template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
+  InertiaTpl<Scalar, Options>
+  computeFrameSupportedInertiaWithSubtree(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
```
jcarpent, line 310: maybe using the naming `computeInertiaSupportedByFrame` is even more explicit.
I think merging the two functions in a single one would be much more coherent.
We just need to add an optional argument `bool with_subtree=false`

unittest/frames.cpp
```diff
@@ -7,13 +7,16 @@
 #include "pinocchio/algorithm/jacobian.hpp"
 #include "pinocchio/algorithm/frames.hpp"
 #include "pinocchio/algorithm/rnea.hpp"
+#include "pinocchio/algorithm/crba.hpp"
```
jcarpent, line 10: Pinocchio might not have been compiled with URDFDOM support, could you add `#ifndef PINOCCHIO_WITH_URDFDOM_SUPPORT` to make the code compile if urdfdom is missing?

jcarpent (changes requested):
Thanks @EtienneAr for this nice PR.
I made some slight comments. COuld you handle them?

src/algorithm/frames.hxx
```diff
@@ -261,6 +261,98 @@ namespace pinocchio
...
+    typedef InertiaTpl<Scalar, Options> Inertia;
+    typedef MotionTpl<Scalar, Options> Motion;
+    typedef ForceTpl<Scalar, Options> Force;
+
+    // Compute 'in body' forces
+    const Inertia fI = computeFrameSupportedInertiaInBody(model, frame_id);
+    pinocchio::SE3 oMf = updateFramePlacement(model, data, frame_id);
+    Motion v = getFrameVelocity(model, data, frame_id, LOCAL);
+    Motion a = getFrameAcceleration(model, data, frame_id, LOCAL);
+    Force f = fI.vxiv(v) + fI * a - fI * oMf.actInv(model.gravity);
```
jcarpent, line 337: Maybe, it is better to do:
```suggestion
    Force f = fI.vxiv(v) + fI * (a - oMf.actInv(model.gravity));
```

src/algorithm/frames.hxx
```diff
@@ -261,6 +261,98 @@ namespace pinocchio
...
+    typedef ModelTpl<Scalar,Options,JointCollectionTpl> Model;
+    typedef InertiaTpl<Scalar, Options> Inertia;
+    typedef MotionTpl<Scalar, Options> Motion;
+    typedef ForceTpl<Scalar, Options> Force;
+
+    // Compute 'in body' forces
+    const Inertia fI = computeFrameSupportedInertiaInBody(model, frame_id);
+    pinocchio::SE3 oMf = updateFramePlacement(model, data, frame_id);
+    Motion v = getFrameVelocity(model, data, frame_id, LOCAL);
+    Motion a = getFrameAcceleration(model, data, frame_id, LOCAL);
```
jcarpent, line 336:
```suggestion
    const pinocchio::SE3 oMf = updateFramePlacement(model, data, frame_id);
    const Motion v = getFrameVelocity(model, data, frame_id, LOCAL);
    const Motion a = getFrameAcceleration(model, data, frame_id, LOCAL);
```

unittest/frames.cpp
EtienneAr: In the case it has not been compiled with it, should I find another way to test or just skip this test ?

jcarpent: Using it, you won't need anymore to rely on urdfdom and just build dummy models using the current helpers in Pinocchio.

jcarpent (approved):

unittest/frames.cpp
EtienneAr: https://github.com/stack-of-tasks/pinocchio/pull/1797

PR #1797, opened 2022-11-24
Follow-up : Supported inertia by frame

EtienneAr:
Hello,

The [previous PR](https://github.com/stack-of-tasks/pinocchio/pull/1796) has been merged before I address all the remarks.
So this PR solves it : 
- It fixes a minor bug in the computation on the inertia with subtree
- It removes the dependency to urdfdom for the test (as mentioned here https://github.com/stack-of-tasks/pinocchio/pull/1796#discussion_r1030590778)


jcarpent (approved):
Thanks a lot @EtienneAr

PR #1843, opened 2023-01-24
CMake: fix discovery of python tests

nim65s:


nim65s:
Maybe we'll see those segfaults on 18.04 also on gitlab.

nim65s:
In "CI - Linux via APT / build (ubuntu-18.04)":
```
	 69 - test-py-bindings (SEGFAULT)
	 78 - test-py-bindings_inertia (SEGFAULT)
	 80 - test-py-bindings_motion (SEGFAULT)
```
All 3 are OK in gitlab-ci.

Might be an issue with either `BUILD_WITH_CASADI_SUPPORT` and/or `BUILD_WITH_OPENMP_SUPPORT`, as both are ON in the first and OFF in the second.

jcarpent:
I've cancelled the build against CasADi, and the CI is still falling.
@nim65s I think this is related to robotpkg binary releases on 18.04. Could you confirm?

jcarpent:
@nim65s I will close this PR as it seems solved by #1851

PR #1845, opened 2023-01-25
[python/visualize] Extend visualizer features and implement them for MeshcatVisualizer

ManifoldFR:
This PR adds new features to the Python visualization API, and implements them for MeshcatVisualizer

+ add visualizer methods for setting the camera target, position,
  environment background color
+ base class raises NotImplementedError when there is no implementation

- [x] set background color
- [x] video recording
- [x] captureImage : allow specifying output image dimensions (for meshcat, requires rdeits/meshcat-python#109)
- [x] enable/disable user camera control
- [x] set camera zoom
- [x] draw frame velocities
- [x] set camera targets
- [x] meshcat: can add objects of type `pin.GeometryObject` outside of the Pinocchio visual model, on-the-fly
- [x] adds an example showcasing a pose of Solo-12 onto a `hppfcl.HeightField`

*Note* This PR is a backport from changes added to Pinocchio 3.

### DIFF
--- a/bindings/python/pinocchio/visualize/base_visualizer.py
+++ b/bindings/python/pinocchio/visualize/base_visualizer.py
@@ -3,12 +3,14 @@ from ..shortcuts import buildModelsFromUrdf, createDatas
 
 import time
 import numpy as np
+from pathlib import Path
 
 class BaseVisualizer(object):
     """Pinocchio visualizers are employed to easily display a model at a given configuration.
     BaseVisualizer is not meant to be directly employed, but only to provide a uniform interface and a few common methods.
     New visualizers should extend this class and override its methods as neeeded.
     """
+    _video_writer = None
 
     def __init__(self, model = pin.Model(), collision_model = None, visual_model = None, copy_models = False, data = None, collision_data = None, visual_data = None):
         """Construct a display from the given model, collision model, and visual model.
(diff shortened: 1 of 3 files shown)


bindings/python/pinocchio/visualize/base_visualizer.py
```diff
@@ -54,48 +70,111 @@ def loadViewerModel(self, *args, **kwargs):
...
+
+    def drawFrameVelocities(self, *args, **kwargs):
+        """Draw current frame velocities."""
+        raise NotImplementedError()
+
     def sleep(self, dt):
         time.sleep(dt)
 
-    def play(self, q_trajectory, dt, capture=False):
+    def play(self, qs, dt=None, callback=None, **kwargs):
```
nim65s, line 131: This would break existing code with `capture=True`, right ?

ManifoldFR: Oh yeah, that would be a breaking change, since now whether or not we capture footage is controlled by the `VideoContext` context manager.

What we can do is check for "capture" inside `kwargs` and in that case raise a warning, maybe ?

nim65s: I would think warnings are for deprecation, here this should be an error, because the breaking change would already be activated.

To avoid breaking changes, we should also return the list of images, and not require imageio, which is a third-party library (which we would have to add on all package managers). Also imageio is not compatible with python 2.

ManifoldFR:
> @ManifoldFR What is the status of this PR?

I haven't worked on it again, but I need to take @nim65s 's comments into account

bindings/python/pinocchio/visualize/base_visualizer.py
ManifoldFR: We could make imageio optional, no problem.

Maybe make getting the images optional (for long streams of images it would be more efficient to just append them to the imageio stream).

My suggestion would be: use the new VideoContext by default when available, otherwise return the images.

ManifoldFR:
Ok, I think it's good now !

jcarpent:
@ManifoldFR Could handle the warnings from the linter?

bindings/python/pinocchio/visualize/base_visualizer.py
```diff
@@ -2,14 +2,35 @@
 from ..shortcuts import buildModelsFromUrdf, createDatas
 
 import time
+import numpy as np
+from pathlib import Path
```
nim65s, line 6: pathlib is in standard library only for python 3

nim65s (dismissed):
Thanks @ManifoldFR, that's way better for me !
Just one nit: you'll have to treat pathlib as imageio, until we drop support for python 2

bindings/python/pinocchio/visualize/base_visualizer.py
ManifoldFR: Ok, I fixed that and used `os.path` instead

jcarpent (approved):

ManifoldFR:
Hum, ROS-CI is not passing, can't really tell why...

nim65s:
rebasing on the current devel might help

ManifoldFR:
> rebasing on the current devel might help

I already rebased and no new commits came in since the last one I think

nim65s:
Ok, this must be a syntax error in python 2 then. I'll nudge gitlab to see if we have more details.

nim65s:
Oh, yes, the f-strings:
```
Exception raised:   File "/root/robotpkg/math/py-pinocchio/work/pinocchio-2.6.10/bindings/python/pinocchio/visualize/base_visualizer.py", line 182
    filename = time.strftime(f"{f_fmt}.{ext}")

**Comment by nim65s** [comment]
and the type hints:
```
Exception raised:   File "/root/robotpkg/math/py-pinocchio/work/pinocchio-2.6.10/bindings/python/pinocchio/visualize/base_visualizer.py", line 188
    def __init__(self, viz: BaseVisualizer, fps, filename, **kwargs):

ManifoldFR:
Great, should be fixed now.

ManifoldFR:
The CI fully passed. We just need approval now !

PR #1855, opened 2023-02-01
Fix issue with Python 3.6

jcarpent:
@duburcqa Could you check this fix on your side?

### DIFF
--- a/bindings/python/spatial/force.hpp
+++ b/bindings/python/spatial/force.hpp
@@ -137,7 +137,7 @@ namespace pinocchio
       
       static void expose()
       {
-#if BOOST_VERSION / 100 % 1000 < 71 && EIGENPY_VERSION_AT_LEAST(2,9,0)
+#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION == 6 && EIGENPY_VERSION_AT_LEAST(2,9,0)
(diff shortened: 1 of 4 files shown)


nim65s:
@duburcqa : could you confirm that robotpkg-py36-eigenpy 2.9.1 still present issues on your side ?

duburcqa:
Yes I will.

duburcqa:
> Could you check this fix on your side?

It was not working but it was before your last commit, so it was not checking the right Python version. Trying again now.

duburcqa:
@jcarpent It works!

duburcqa:
> @duburcqa : could you confirm if robotpkg-py36-eigenpy 2.9.1 still present issues on your side ?

Still crashing.

duburcqa:
You patch is not relevant for all Python versions ? Why only apply it to Python 3.6.9 ?

jcarpent:
No, as you can see, the CI is fine for the other Ubuntu versions.

jcarpent:
In general, I'm trying to fix what should be fixed and not apply a patch if it is not necessary elsewhere.
PS: the issue really seems to be coming from Boost.Python + Python 3.6

duburcqa:
Ok no problem I'm fine with it ! Thank you for the patch !

PR #1860, opened 2023-02-06
Second-order RNEA derivatives

jcarpent:
@shubhamsingh91 I made a complete refactoring of your code, leading to large improvements in terms of timings while reducing the number of lines ;)
The timings look now better, with a gain of more than 40% with clang.

PS: I was not able to push directly on your fork. Thus I had to open this new PR.


jcarpent:
In a future PR, it remains to:

- [ ] Expose Eigen::Tensor into eigenpy
- [ ] Expose SO derivatives of RNEA and related quantities in the Python bindings

PR #1865, opened 2023-02-12
Fix support of AVX2.

duburcqa:


duburcqa:
@jcarpent You probably have a explanation for why this patch is fixing memory alignment issues when converting `Eigen::matrix3d` from python to c++. I just noticed it was fixing it.

jcarpent:
The reason lies in Eigenpy 2.9.0 in fact. AVX support only works fully with Eigenpy 2.9.0 since I've forced the proper alignment of Eigen objects with Boost.Python. 
The fix you provided here is just calling properly ordering the call to `<eigenpy/fwd.hpp>`. AVX-2 works well with Pinocchio 3x where all the Python bindings have been cleaned to operate with extra scalar exposition like Casadi types, etc.

Thanks for the fix, which should help other people to have vectorization support with Pinocchio 2x.

jcarpent (approved):

duburcqa:
Ok make sense, thanks for the explanation. I will try to fix any similar import ordering issue I can find while running my own test suite.

PR #1868, opened 2023-02-16
Support force in pybind11

cmastalli:
This PR includes pybind11 support for spatial forces.

It would be nice if you guys can make a conda release with these bits.

### DIFF
--- a/bindings/python/pybind11-all.hpp
+++ b/bindings/python/pybind11-all.hpp
@@ -9,6 +9,7 @@
 /// \li DataTpl
 /// \li SE3Tpl
 /// \li MotionTpl
+/// \li ForceTpl
 /// \li GeometryModel
 /// \li (not yet) GeometryData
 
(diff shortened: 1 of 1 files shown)


jcarpent (changes requested):
@cmastalli Thanks for extending the support for PyBind11.
Could you remove the merge commit and perform a simple rebase? It will help us to maintain a clean git history.

jcarpent:
@wxmerkt The CI on ROS colcon is failling. Do you know why?

wxmerkt:
It's a temporary issue since EigenPy could not be retrieved for rolling. Curious to why that's the case since the most recent build was fine (https://build.ros2.org/job/Rbin_uJ64__eigenpy__ubuntu_jammy_amd64__binary/). We may want to retrigger the build after some time

cmastalli:
> @cmastalli Thanks for extending the support for PyBind11.
> Could you remove the merge commit and perform a simple rebase? It will help us to maintain a clean git history.

done!

jcarpent (approved):

cmastalli:
This PR can bind functions such as
```cpp
void set_force(std::map<std::string, pinocchio::Force> force);
```
but not functions like
```cpp
void std::map<std::string, pinocchio::Force> get_force();
```

Does somebody have a clue how to enable the latter option?
I could give it a try.

cmastalli:
> It's a temporary issue since EigenPy could not be retrieved for rolling. Curious to why that's the case since the most recent build was fine (https://build.ros2.org/job/Rbin_uJ64__eigenpy__ubuntu_jammy_amd64__binary/). We may want to retrigger the build after some time

Pinocchio and Crocoddyl can be installed smoothly in conda, same for ROS. I wonder if it makes sense to migrate the ROS CI based on conda. Is this makes sense for both of you?

jcarpent:
People are using Pinocchio via ROS only without conda. We should be able to maintain this compatibility

wxmerkt:
> Does somebody have a clue how to enable the latter option?

Check the code from the MPC implementation - I added converters for C++ -> pybind as well

cmastalli:
> Check the code from the MPC implementation - I added converters for C++ -> pybind as well

I checked this implementation, but it doesn't resolve the issue described.
It seems this approach cannot handle std containers.

For completeness, @wxmerkt suggested looking at a type caster that moves the memory as explained here:
https://yyc.solvcon.net/en/latest/writing/2021/pyboost11/pyboost11.html

I will address this later when things are clearer to me.

PR #1869, opened 2023-02-18
Fix some Python bindings signatures and add stub generation.

duburcqa:


jcarpent (commented):
Thanks @duburcqa for this PR. 
In fact, these things have been fixed in Pinocchio 3x. This is why I've developed the stubgen module support for Boost.Python.
I would prefer not to merge this PR as we are focused on Pinocchio 3x and try to maximize our efforts on it.

duburcqa:
Ok no problem. But I would suggest to at least merge the update of `bindings/python/multibody/joint/joints-variant.hpp` to at least provide valid classes.

Optionally, It would be nice to just merge the changes of docstring, and leave out the stub generation.

jcarpent:
In fact, thinking twice, we can accept it as a single commit and I remove it during the merge.
It could be helpful to people, before Pinocchio 3x is fully available.

jcarpent:
By the way, for the class name, it has been fixed in Pinocchio 3x.

duburcqa:
As you wish. I have no doubt many issues have been fixed in Pinocchio 3x, but the transition may take some time for people. I known intellisense technologies can be quite sensitive to badly formatted docstring and it would be nice to avoid crashing ipython or vscode.

You want me to add back the part responsible for generating stubs ?

jcarpent:
Yes, you can add it back please.

duburcqa:
Done.

jcarpent (approved):
Thanks a lot

PR #1877, opened 2023-02-28
Add documentation badge to the README header

stephane-caron:
Docs badges are a common and convenient way for programmers to go rapidly to the documentation of a project from GitHub. I believe this will improve convenience, over scrolling down to the Documentation subsection (which does not stand out very much).


jcarpent:
It is already there. I added it last week ;).
Thanks @stephane-caron for the suggestion.

Issue #1876, opened 2023-02-28
empty examples


v4hn:
[The examples folder](https://github.com/stack-of-tasks/pinocchio/tree/master/examples) includes various files which turn out empty except for a TODO saying someone should write them.
The files haven't been touched in years.

Please consider removing the files without content (and add explicit issues on github or in ONE file) to make this code base a bit more approachable for new users.

stephane-caron:
Started in https://github.com/stack-of-tasks/pinocchio/pull/1878.

Feel free to open a PR directly when you see an improvement like this one.

PR #1878, opened 2023-02-28
Remove empty examples

stephane-caron:
Also fix link to reduced-model example which was not included.

Closes https://github.com/stack-of-tasks/pinocchio/issues/1876


jcarpent (approved):
Thanks a lot @stephane-caron for taking care of it.

Issue #1879, opened 2023-03-02
Catkin build failure when generating python stubs


james-p-foster:
After seeing the recent PR adding the generation of python stubs here: https://github.com/stack-of-tasks/pinocchio/pull/1869 I have tried to build the `devel` branch with catkin, but I get build errors.

I have set up a new workspace with just `pinocchio` and its dependencies, `hpp-fcl` and `eigenpy`, and replicated the issue.

Concretely, when I run:
```
catkin build -j10 -DCMAKE_BUILD_TYPE=Release -DPYTHON_EXECUTABLE=/usr/bin/python3.8 -DBUILD_WITH_COLLISION_SUPPORT=ON -DGENERATE_PYTHON_STUBS=ON
```

I get the following:
```
Traceback (most recent call last):
  File "/home/jfoster/dummy-workspace/build/pinocchio/bindings/python/stubgen/src/pybind11_stubgen/__init__.py", line 1010, in <module>
    main()
  File "/home/jfoster/dummy-workspace/build/pinocchio/bindings/python/stubgen/src/pybind11_stubgen/__init__.py", line 988, in main
    _module = ModuleStubsGenerator(_module_name)
  File "/home/jfoster/dummy-workspace/build/pinocchio/bindings/python/stubgen/src/pybind11_stubgen/__init__.py", line 731, in __init__
    self.module = importlib.import_module(module_or_module_name)
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 848, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/jfoster/dummy-workspace/src/pinocchio/bindings/python/pinocchio/__init__.py", line 6, in <module>
    from .pinocchio_pywrap import *
SystemError: initialization of pinocchio_pywrap raised unreported exception
[...]

PR #1880, opened 2023-03-02
Debug #1879

wxmerkt:


jcarpent:
Should be solved with the new release 2.6.19.

PR #1882, opened 2023-03-05
Enabled copy and deepcopy

cmastalli:
Pickable is the worst solution for enabling an object to be copyable. Instead, the purpose solution is better while still keeping the pickability of the Pinocchio objects.

Additionally, the pickable solution creates segfaul in my Apple M1 chip. But, if I remember correctly, this doesn't happen in Ubuntu.

In short this PR enables us to do the following:
```python
import pinocchio
import copy

model = pinochio.Model()

m1 = copy.copy(model)
m2 = copy.deepcopy(model)
```

### DIFF
--- a/.github/workflows/macos-linux-conda.yml
+++ b/.github/workflows/macos-linux-conda.yml
@@ -8,6 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     env:
       CCACHE_DIR: ${{ matrix.CCACHE_DIR }}
+      BUILD_ADVANCED_TESTING: ${{ matrix.BUILD_ADVANCED_TESTING }}
 
     strategy:
       fail-fast: false
@@ -18,8 +19,10 @@ jobs:
         include:
           - os: ubuntu-latest
             CCACHE_DIR: /home/runner/.ccache
+            BUILD_ADVANCED_TESTING: OFF
           - os: macos-latest
             CCACHE_DIR: /Users/runner/.ccache
+            BUILD_ADVANCED_TESTING: ON
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


jcarpent:
Your issue on M1 is due to the fact that you are not using conda forge compilation suite to build Pinocchio from source

cmastalli:
> This PR looks useful. Could you please do a rebasing first and not a merge commit?

Done!

> Your issue on M1 is due to the fact that you are not using conda forge compilation suite to build Pinocchio from source

I am using conda, but in and interesting way. I compile Pinocchio and its dependencies within a conda environment, but don't install any of the dependencies using conda. I have created an environment using Python 3.9 as I need to install these packages along other ROS noetic stuff (via robostack).

I am not sure what you meant with `conda compilation suite`. Can you share a link where I can learn more about?

jcarpent (commented):
I will make a pass on this PR tomorrow as the copyright is not correct and I will clean a bit the content

cmastalli:
@jcarpent -- have you got some time for this?

If the issue is copyright only, then you could use the suggest option to quickly handle it.

cmastalli:
Hey guys! This PR is being here for a while. Is there anything preventing us to merge it? (Thanks)

jcarpent:
> Hey guys! This PR is being here for a while. Is there anything preventing us to merge it? (Thanks)

I have limited bandwidth on my side. And we have limited human resources.
I'm trying to prioritize according to the important issue. I hope you can understand it.

jcarpent:
One of the important issues is that you have copy-pasted the code from stack overflow without any reference or mentioning that the code is not yours directly.
In addition, I don't think the current deepcopy implementation is the way to go with C++ objects. This should be further investigated beyond simple copy-pasting. Don't you think so?

cmastalli:
> One of the important issues is that you have copy-pasted the code from stack overflow without any reference or mentioning that the code is not yours directly.
> In addition, I don't think the current deepcopy implementation is the way to go with C++ objects. This should be further investigated beyond simple copy-pasting. Don't you think so?

I understand! I might include relevant references, but this doesn't seem to be the main concern here. Might I ask what type of copy-pasting approach you think is better? Is there an example where this approach is bad?

(I am asking this as I could allocate time to investigating this)

cmastalli:
Thanks, @jcarpent!

PR #1885, opened 2023-03-09
Enforce BOOST_MPL_CFG_NO_PREPROCESSED_HEADERS

jcarpent:
Related to #1884


jcarpent:
@olivierrochel-inria It seems that adding the definition of `BOOST_MPL_CFG_NO_PREPROCESSED_HEADERS` for compiling is conflicting with Boost.Python as shown by the CI.
It seems the problem is internal to Boost.

I will then close this PR and your issue, as might conflict with projects relying on Pinocchio.
If yo have a simple example to show where you have to manually enforce `BOOST_MPL_CFG_NO_PREPROCESSED_HEADERS`, I would be interested to look at it and provide a fix if possible.

PR #1891, opened 2023-03-13
add static-contact-dynamics example

PepMS:
This PR adds a new python example. The example computes the static contact dynamics with Solo12.

### DIFF
--- a/examples/static-contact-dynamics.py
+++ b/examples/static-contact-dynamics.py
@@ -96,13 +96,13 @@ pin.framesForwardKinematics(model, data, q0)
 
 # Contact forces at base link frame
 ls__BL = []
-for l, id in zip(ls__F, feet_ids):
+for l, foot_id in zip(ls__F, feet_ids):
     l_sp__F = pin.Force(l, np.zeros(3))
-    l_sp__BL = data.oMf[bl_id].actInv(data.oMf[id].act(l_sp__F))
+    l_sp__BL = data.oMf[bl_id].actInv(data.oMf[foot_id].act(l_sp__F))
     ls__BL.append(np.copy(l_sp__BL.vector))
 
 print("\n--- CONTACT FORCES ---")
-for l, id, name in zip(ls__BL, feet_ids, feet_names):
+for l, foot_id, name in zip(ls__BL, feet_ids, feet_names):
     print("Contact force at foot {} expressed at the BL is: {}".format(
         name, l))
 
--- a/examples/static-contact-dynamics.py
+++ b/examples/static-contact-dynamics.py
@@ -1,0 +1,156 @@
+import numpy as np
+import pinocchio as pin
+import example_robot_data
+
+np.set_printoptions(linewidth=np.inf)
+
+# ----- PROBLEM STATEMENT ------
+#
+# We want to find the contact forces and torques required to stand still at a configuration 'q0'.
+# We assume 3D contacts at each of the feet
+#
+# The dynamic equation would look like:
+#
+# M*q_ddot + g(q) + C(q, q_dot) = tau + J^T*lambda --> (for the static case) --> g(q) = tau + Jc^T*lambda (1)
+
+# ----- SOLVING STRATEGY ------
+
+# Split the equation between the base link (_bl) joint and the rest of the joints (_j). That is,
+#
+#  | g_bl |   |  0  |   | Jc__feet_bl.T |   | l1 |
+#  | g_j  | = | tau | + | Jc__feet_j.T  | * | l2 |    (2)
+#                                           | l3 |
+#                                           | l4 |
+
+# First, find the contact forces l1, l2, l3, l4 (these are 3 dimensional) by solving for the first 6 rows of (2).
+# That is,
+#
+# g_bl   = Jc__feet_bl.T * | l1 |
+#                          | l2 |    (3)
+#                          | l3 |
+#                          | l4 |
+#
+# Thus we find the contact froces by computing the jacobian pseudoinverse,
+#
+# | l1 | = pinv(Jc__feet_bl.T) * g_bl  (4)
+# | l2 |
+# | l3 |
+# | l4 |
+#
+# Now, we can find the necessary torques using the bottom rows in (2). That is,
+#
+#                             | l1 |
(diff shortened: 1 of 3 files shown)


jcarpent (changes requested):
Thanks @PepMS for this very useful example. Could you handle the comments from codacy in the checks?
Thanks in advance,

jcarpent (approved):

PepMS:
It seems like there is a problem with the example-robot-data dependency in the CI. Should I remove the dependency on `example_robot_data` and load the robot like in https://github.com/stack-of-tasks/pinocchio/blob/b4d7a8b6240f760e824cba06c9c93c95f594dd12/examples/meshcat-viewer-dae.py#L17-L22 ?

Also, I think I've put the example in the wrong line of the CMakeLists.txt. Should it be here https://github.com/stack-of-tasks/pinocchio/blob/b4d7a8b6240f760e824cba06c9c93c95f594dd12/examples/CMakeLists.txt#L71 ?

jcarpent:
Yes, you should like the other examples, as example-robot-data depends on pinocchio.

jcarpent:
> Also, I think I've put the example in the wrong line of the CMakeLists.txt. Should it be here
> 
> https://github.com/stack-of-tasks/pinocchio/blob/b4d7a8b6240f760e824cba06c9c93c95f594dd12/examples/CMakeLists.txt#L71

Yes, indeed.

examples/static-contact-dynamics.py
```diff
@@ -0,0 +1,166 @@
...
+# Contact forces at local coordinates (at each foot coordinate)
+ls__f = np.split(ls, ncontact)
+
+pin.framesForwardKinematics(model, data, q0)
+
+# Contact forces at base link frame
+ls__bl = []
+for l__f, foot_id in zip(ls__f, feet_ids):
+    l_sp__f = pin.Force(l__f, np.zeros(3))
+    l_sp__bl = data.oMf[bl_id].actInv(data.oMf[foot_id].act(l_sp__f))
```
cmastalli, line 111: Make it simpler by using a single transform: from foot to joint.
This transform is inside the Pinocchio model, i.e.
```python
jMf = model.frames[foot_id].
```

examples/static-contact-dynamics.py
```diff
@@ -0,0 +1,166 @@
...
+# First, we set the frame for our contacts. We assume the contacts are placed at the following 4 frames and they are 3D
+feet_names = ['FL_FOOT', 'FR_FOOT', 'HL_FOOT', 'HR_FOOT']
+feet_ids = [model.getFrameId(n) for n in feet_names]
+bl_id = model.getFrameId('base_link')
+ncontact = len(feet_names)
+
+# Now, we need to find the contact Jacobians appearing in (1).
+# These are the Jacobians that relate the joint velocity  to the velocity of each feet
+Js__feet_q = [
+    np.copy(pin.computeFrameJacobian(model, data, q0, id, pin.LOCAL))
```
cmastalli, line 88: Compute the forces in "LOCAL_WORLD_ALIGNED" coordinates as this is what we typically want.
You can use the LOCAL_WORLD_ALIGNED contact Jacobian or multiply by the rotation matrix the linear force.

examples/static-contact-dynamics.py
```diff
@@ -0,0 +1,166 @@
...
+#                             | l4 |
+
+# ----- SOLUTION ------
+
+# 0. DATA
+pinocchio_model_dir = join(dirname(dirname(str(abspath(__file__)))), "models")
+
+model_path = join(pinocchio_model_dir, "example-robot-data/robots")
+mesh_dir = pinocchio_model_dir
+urdf_filename = "solo12.urdf"
```
cmastalli, line 54: I would use a humanoid robot, so we can combine 3d and 6d contacts.

examples/static-contact-dynamics.py
```diff
@@ -0,0 +1,166 @@
...
+joint_names = ['FL_KFE', 'FR_KFE', 'HL_KFE', 'HR_KFE']
+joint_ids = [model.getJointId(n) for n in joint_names]
+
+fs_ext = [pin.Force(np.zeros(6)) for _ in range(len(model.joints))]
+for idx, joint in enumerate(model.joints):
+    if joint.id in joint_ids:
+        fext__bl = pin.Force(ls__bl[joint_ids.index(joint.id)])
+        fs_ext[idx] = data.oMi[joint.id].actInv(data.oMf[bl_id].act(fext__bl))
+
+tau_rnea = pin.rnea(model, data, q0, v0, a0, fs_ext)
```
cmastalli, line 148: It is better if you use `pinocchio.computeGeneralizedGravity`.

examples/static-contact-dynamics.py
jcarpent: Working with both is fine I would say. No need to change here.

examples/static-contact-dynamics.py
jcarpent: Solo is very common as a template example. No reason to change.

examples/static-contact-dynamics.py
jcarpent: Both are fine.

PepMS: Thank you for the comments, @cmastalli 

In this line, I think I couldn't use computeGeneralizedGravity, since I was really commuting the ID with external forces, just to make sure that the torques computed manually are the same as the ones computed by the ID method.

Maybe you are referring to the computation done in L72. Since I saw that calling this function is equivalent to a call to RNEA I decided to use RNEA. Since this example aims at being a micro tutorial example, I think that it is nice to realize that you can find the gravitational terms by a simple call to the ID algorithm.

cmastalli: Yes, you're correct! I was referring to L72. Apologies for the confusion.

jcarpent (approved):
We will do a pass later on this example to aligned the notations. Thanks @PepMS for the helpful example.

PepMS:
Happy to contribute, @jcarpent . When the time comes, let me know if you want me to rewrite the example to meet any notation criteria.

Issue #1945, opened 2023-05-01
Adding a toHomogeneousMatrix function to Motion


stephane-caron:
Currently Pinocchio has `SE3::toHomogeneousMatrix` but not `Motion::toHomogeneousMatrix`. It arose from [this discussion](https://github.com/stack-of-tasks/pinocchio/discussions/1939) that it would be consistent and convenient to have both.

Note that in the meantime we can always do the conversions in Python:

```python
def to_homogeneous(twist: pin.Motion) -> np.ndarray: 
     homogeneous = np.zeros((4, 4)) 
     homogeneous[:3, :3] = twist.action[:3, :3] 
     homogeneous[:3, 3] = twist.linear 
     return homogeneous
```

I can help prepare a PR if this makes sense.

jcarpent:
OK to add this method to the Motion class, even if it is not a convenient way to use Motion quantities in terms of sparsity exploitation and general efficiency.

jcarpent:
Solved via #1946

PR #1946, opened 2023-05-02
Add Motion::toHomogeneousMatrix

stephane-caron:
Closes https://github.com/stack-of-tasks/pinocchio/issues/1945

### DIFF
--- a/src/spatial/motion-dense.hpp
+++ b/src/spatial/motion-dense.hpp
@@ -58,6 +58,15 @@ namespace pinocchio
       
       return X;
     }
+
+    HomogeneousMatrixType toHomogeneousMatrix_impl() const
+    {
+      HomogeneousMatrixType M;
+      M.template block<3,3>(0, 0) = skew(angular());
+      M.template block<3,1>(0, 3) = linear();
+      M.template block<1,4>(3, 0).setZero();
+      return M;
+    }
     
     template<typename D2>
     bool isEqual_impl(const MotionDense<D2> & other) const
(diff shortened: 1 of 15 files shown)


src/spatial/motion-dense.hpp
```diff
@@ -58,6 +58,15 @@ namespace pinocchio
       
       return X;
     }
+
+    HomogeneousMatrixType toHomogeneousMatrix_impl() const
+    {
+      HomogeneousMatrixType M;
+      M.template block<3,3>(0, 0) = skew(angular());
```
stephane-caron, line 65: I chose not to use `ANGULAR` and `LINEAR` here, as it would have been confusing to read `M(LINEAR, LINEAR) = angular` :thinking:

Since we already know the dimensions of the object we are working on anyway it would have been unnecessary concept-wrapping in my opinion.

stephane-caron:
Sanity check:

```console
$ ipython
Python 3.8.10 (default, Mar 13 2023, 10:26:41) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.22.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np

In [2]: import pinocchio as pin

In [3]: xi = pin.Motion(linear=np.array([1, 2, 3]), angular=np.array([4, 5, 6]))

In [4]: xi.homogeneous
Out[4]: 
array([[ 0., -6.,  5.,  1.],
       [ 6.,  0., -4.,  2.],
       [-5.,  4.,  0.,  3.],
       [ 0.,  0.,  0.,  0.]])
```

jcarpent (approved):

PR #1986, opened 2023-06-03
Documentation 

drewhamiltonasdf:
A handful of updates to the documentation to get rid of deprecated python commands etc., a few grammar fixes, missing packages etc.

### DIFF
--- a/doc/c-maths/a-rigid-bodies.md
+++ b/doc/c-maths/a-rigid-bodies.md
@@ -8,7 +8,7 @@ Rotation matrices form the so-called **Special Orthogonal** group \f$ SO(n) \f$.
 
 The set that brings together all the homogeneous transformations matrices is the **Special Euclidean** group \f$ SE(n) \f$. As with rotation matrices, there are two different groups, \f$ SE(3) \f$ for 3-dimensional transformations and \f$ SE(2) \f$ for 2-dimensional transformation, i.e. transformation in a plane. 
 
-### Using quaternions for a \f$ SO(3) \f$ object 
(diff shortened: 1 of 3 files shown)


jcarpent (approved):

jcarpent:
Thanks a lot @drewhamiltonasdf. This looks nice to me.
We can even go further by removing the use of `gepetto-gui` and using `meshcat` viewer instead.

drewhamiltonasdf:
> Thanks a lot @drewhamiltonasdf. This looks nice to me. We can even go further by removing the use of `gepetto-gui` and using `meshcat` viewer instead.

No problem. I can go back to some of the earlier docs and update to MeshCat. Is gepetto-gui completely out of favor now? Also wondering if RobotWrapper.py is still the preferred URDF import method for custom URDF's? I'm going to have to dig into loading custom URDF packages eventually...

jcarpent:
> Is gepetto-gui completely out of favor now?

Yes, it is no more the way to go. The generic approach relies on Meshcat which is multiplatform and more flexible.


> Also wondering if RobotWrapper.py is still the preferred URDF import method for custom URDF's?

It is a helper class. But for best performances, it is better to rely on a separate model and data, loaded via the custom URDF parser.

jcarpent:
@drewhamiltonasdf A small suggestion, it would be better to squeeze your commits when they are related to the same change. You can use `git commit --amend` to amend the last commit with the latest modfications.

drewhamiltonasdf:
> @drewhamiltonasdf A small suggestion, it would be better to squeeze your commits when they are related to the same change. You can use `git commit --amend` to amend the last commit with the latest modfications.

Will do going forward. Thanks!

jcarpent:
I'm merging this first PR.
Thanks @drewhamiltonasdf for this nice and helpful contribution. We are looking forward to further doc improvements thanks to you.

Issue #1999, opened 2023-06-20
CppAd & codegen test compilation failure


ghost:
Pinocchio: 2.6.19
Boost: 1.78
Eigen: 3.3.7
same log for codegen test
Log:
>/home/harper/miniconda3/envs/my_env/include/boost/math/constants/constants.hpp: In instantiation of ‘constexpr typename boost::math::constants::detail::constant_return<Real, Policy>::type boost::math::constants::pi() [with T = CppAD::AD<double>; Policy = boost::math::policies::policy<boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy>; typename boost::math::constants::detail::constant_return<Real, Policy>::type = CppAD::AD<double>]’:
/home/harper/miniconda3/envs/my_env/include/boost/math/constants/constants.hpp:259:3:   required from ‘constexpr typename boost::math::constants::detail::constant_return<T>::type boost::math::constants::pi() [with T = CppAD::AD<double>; typename boost::math::constants::detail::constant_return<T>::type = CppAD::AD<double>]’
/home/harper/Apps/pinocchio/build/include/pinocchio/math/fwd.hpp:27:46:   required from ‘const Scalar pinocchio::PI() [with Scalar = CppAD::AD<double>]’
/home/harper/Apps/pinocchio/build/include/pinocchio/multibody/liegroup/special-orthogonal.hpp:65:48:   required from ‘static typename Matrix2Like::Scalar pinocchio::SpecialOrthogonalOperationTpl<2, _Scalar, _Options>::log(const Eigen::MatrixBase<OtherDerived>&) [with Matrix2Like = Eigen::Matrix<CppAD::AD<double>, 2, 2, 0, 2, 2>; _Scalar = CppAD::AD<double>; int _Options = 0; typename Matrix2Like::Scalar = CppAD::AD<double>]’
[...]

PR #2000, opened 2023-06-20
Fix support of recent versions of Boost for CppAD and CppADCodeGen

jcarpent:
Solves #1999.


cmastalli:
Hi @jcarpent 

Thanks for fixing this issue!

I would like a release of this in Conda. The reason is that I would like to set up CI jobs testing Crocoddyl with CppADCodeGen support (more details here: https://github.com/loco-3d/crocoddyl/pull/1152)

Could you tell me when you're planning to make such a release?

jcarpent:
I can do one next week.

cmastalli:
> I can do one next week.

That would be fantastic. Thanks!

PR #2040, opened 2023-08-22
Fix Ycrb[0] not re-initialized in crbaMinimal.

duburcqa:


jcarpent (dismissed):
Thanks for the fix @duburcqa

Issue #2050, opened 2023-09-10
Numerical issue when computing gradient w.r.t. quaternion using casadi


modu-robotics:
## Bug description
I am trying to solve an optimization problem using pinocchio+casadi, where my cost function is a scalar valued function including `pinocchio::quaternion::log3` operator. When `casadi` attempts to evaluate gradient of my cost function at zero configuration, the gradient has NAN values at quaternion components. I assume this is numerical stability issue rather than a bug, since the gradient has a value at small non zero quaternion. Here is a minimal example to reproduce this:

```
  auto quat_casadi = casadi::SX::sym("quat", 4);  // qx, qy, qz, qw

  // eigen quaternion eats scalar first
  Eigen::Quaternion<casadi::SX> quat(quat_casadi(3), quat_casadi(0),
                                     quat_casadi(1), quat_casadi(2));
  auto so3 = pinocchio::quaternion::log3(quat);
  auto so3_casadi = casadi::SX::sym("so3", 3);
  pinocchio::casadi::copy(so3, so3_casadi);
  auto cost = casadi::SX::dot(so3_casadi, so3_casadi);

  auto grad_func =
      casadi::Function("grad", {quat_casadi}, {gradient(cost, quat_casadi)},
                       {"quat_in"}, {"grad_out"});

  auto out = grad_func(casadi::DMDict{
      {"quat_in", std::vector<double>{0, 0, 0, 1}}}); // evaluate gradient with a zero quaternion

  std::cout << out << std::endl; // This reports {"grad_out": [nan, nan, nan, 0]}

  auto out = grad_func(casadi::DMDict{
      {"quat_in", std::vector<double>{0, 0, 0.0499792, 0.9987503}}}); // evaluate gradient with a small yaw rotation (0.1 rad)
  std::cout << out << std::endl; // This reports {"grad_out": [0, 0, 0.4005, -1.66625e-05]}
```

### System

- OS: Mac OS
- Pinocchio version: source install from `pinocchio3-preview`

modu-robotics:
I tried to update `pinocchio3-preview` branch's `pinocchio::quaternion::log3` function to match `master` branch's `pinocchio::quaternion::log3` function and it resolves the issue.
Is there any plan to update this function?
If authors have no bandwidth for it, I am happy to make a PR.

jcarpent:
Thanks for the report. Please, integrate the change in the branch. Thanks in advanced.

PR #2056, opened 2023-09-16
Bug fix in quaterion::normalize and quaternion::log3

modu-robotics:
Related to #2050, this PR fixes a bug in `quaternion::normalize` and `quaternion::log3` methods.

### DIFF
--- a/src/math/quaternion.hpp
+++ b/src/math/quaternion.hpp
@@ -237,7 +237,7 @@ namespace pinocchio
     template<typename Quaternion>
     inline void normalize(const Eigen::QuaternionBase<Quaternion> & quat)
     {
-      return pinocchio::normalize(quat.const_cast_derived().coeffs());
+      return pinocchio::normalize(PINOCCHIO_EIGEN_CONST_CAST(Quaternion, quat).coeffs());
     }
   
     ///
--- a/unittest/casadi/spatial.cpp
+++ b/unittest/casadi/spatial.cpp
@@ -7,6 +7,7 @@
 #include <pinocchio/math/quaternion.hpp>
 #include <pinocchio/spatial/se3.hpp>
 #include <pinocchio/spatial/motion.hpp>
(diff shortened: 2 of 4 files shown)


unittest/casadi/spatial.cpp
```diff
@@ -89,4 +90,52 @@ BOOST_AUTO_TEST_CASE(test_quaternion)
...
+  auto jac_func = casadi::Function("jac_func", {quat_casadi}, {jac}, {"quat_in"}, {"jac_out"});
+
+  auto so3_result = so3_func(casadi::DMDict{{"quat_in", {0, 0, 0, 1}}}).at("so3_out");
+  auto jac_result = jac_func(casadi::DMDict{{"quat_in", {0, 0, 0, 1}}}).at("jac_out");
+
+  Eigen::Vector<double, 3> so3_expected(0, 0, 0);
+  Eigen::Matrix<double, 3, 4> jac_expected;
+  jac_expected << 2, 0, 0, 0,
+                  0, 2, 0, 0,
+                  0, 0, 2, 0;
```
modu-robotics, line 129: Comparing with hand derived analytic solution. Not sure if it is too trivial, but this was the failure before this PR. Jacobian value at zero quaternion was [I, 0] instead of [2*I, 0]

src/math/quaternion.hpp
```diff
@@ -237,7 +237,7 @@ namespace pinocchio
     template<typename Quaternion>
     inline void normalize(const Eigen::QuaternionBase<Quaternion> & quat)
     {
-      return pinocchio::normalize(quat.const_cast_derived().coeffs());
```
modu-robotics, line 240: `const_cast_dervied` doesn't exist in `Eigen::QuaternionBase`

modu-robotics (commented):
Left some comments below. Looks like this branch has a different CI setup than master branch.

Issue #2069, opened 2023-10-17
REGEX REPLACE in CMakeLists.txt fails for complex directory paths


Pathorse:
## Bug description

Pinocchio build fails when using complex directory paths, e.g., `/test_pinnochio_build/1.0+gitAUTOINC+8ajb23j4b/pinocchio/build`, [as  REGEX REPLACE fails to correctly amend strings before attempting to create symbolic links (or copying)](https://github.com/stack-of-tasks/pinocchio/blob/903cdb8b1fd3bfc21779e76b1c97dbefe5e690da/CMakeLists.txt#L269).

This issue was found while using pinocchio with the Yocto SDK.

## Expected behavior

The build should not fail for any directory paths given.

## Reproduction steps

1. Create a test folder:
```bash
mkdir -p test_pinnochio_build/1.0+gitAUTOINC+8ajb23j4b
```
2. Clone Pinocchio
```bash
git clone git@github.com:stack-of-tasks/pinocchio.git
```
3. Initialize submodules
```bash
cd  test_pinnochio_build/1.0+gitAUTOINC+8ajb23j4b/pinocchio
git submodule update --init
```
4. Build
```bash
mkdir build && cd build
cmake ..
```
5. See error

## Solution

The problem can be solved by replacing `REGEX REPLACE` with `REPLACE`, which given the fact that the desired string operation doesnt require the flexibility of regular expressions may make sense to move away from anyways (unless I am missing something here which might be the case as I am pretty new with Pinocchio). The swap must be done in three files: 

[1. Toplevel CMakeLists.txt](https://github.com/stack-of-tasks/pinocchio/blob/4512214cf08e765a28542c75bf6378e934622a38/CMakeLists.txt#L269)
[2. Python bindings CMakeLists.txt](https://github.com/stack-of-tasks/pinocchio/blob/4512214cf08e765a28542c75bf6378e934622a38/bindings/python/CMakeLists.txt#L26)
[3. Unittest CMakelists.txt](https://github.com/stack-of-tasks/pinocchio/blob/4512214cf08e765a28542c75bf6378e934622a38/unittest/CMakeLists.txt#L30)

### System

- OS: Ubuntu 22.04
- Pinocchio version: 2.6.20

jcarpent:
Thanks @Pathorse for the nice report. @nim65s @jorisv Could you have a look to this issue in more depth?

nim65s:
I think we should simply switch to a more standard CMake use, and:
[...]

PR #2070, opened 2023-10-19
Topic/move headers

jorisv:
Pinocchio headers in repository are not organized as installed ones.

In this merge request, includes and sources are split to have a standard CMake/C++ project layout.
Headers in the repository and installed ones have the same organization.

This will avoid to have to create symlinks and will solve #2069


bindings/python/CMakeLists.txt
```diff
@@ -70,57 +39,29 @@ ENDIF(BUILD_WITH_LIBPYTHON)
...
-    multibody/fcl/transform.hpp
+    ${SOURCE_HEADERS_ROOT}/multibody/fcl/transform.hpp
     )
   LIST(REMOVE_ITEM ${PROJECT_NAME}_PYTHON_SOURCES
     multibody/fcl/expose-fcl.cpp
     )
 ENDIF(NOT BUILD_WITH_HPP_FCL_PYTHON_BINDINGS)
 
 IF(NOT BUILD_WITH_OPENMP_SUPPORT)
-  LIST_FILTER("${${PROJECT_NAME}_PYTHON_HEADERS}" "^multibody/pool" ${PROJECT_NAME}_PYTHON_HEADERS)
```
jorisv, line 81: @nim65s I don’t understand why we use LIST_FILTER instead of LIST(FILTER ...) ?

LIST(FILTER ...) exists since CMake 3.6, and our minimal cmake version is the 3.10.

bindings/python/CMakeLists.txt
```diff
@@ -70,57 +39,29 @@ ENDIF(BUILD_WITH_LIBPYTHON)
...
-  MAKE_DIRECTORY("${${PROJECT_NAME}_BINARY_DIR}/include/pinocchio/bindings/python/algorithm/parallel")
-ENDIF(BUILD_WITH_OPENMP_SUPPORT)
-MAKE_DIRECTORY("${${PROJECT_NAME}_BINARY_DIR}/include/pinocchio/bindings/python/utils")
-SYMLINK_AND_INSTALL_HEADERS("${${PROJECT_NAME}_PYTHON_HEADERS}" "bindings/python")
-
-# Headers of the Python parser
-IF(BUILD_WITH_LIBPYTHON)
-  MAKE_DIRECTORY("${${PROJECT_NAME}_BINARY_DIR}/include/pinocchio/parsers/python")
-  SYMLINK_AND_INSTALL_HEADERS("${${PROJECT_NAME}_PARSER_PYTHON_HEADERS}" "")
-ENDIF(BUILD_WITH_LIBPYTHON)
```
jorisv, line 122: This code will install `bindings/python/parsers/python.hpp` in both `pinocchio/parsers` and `pinocchio/bindings/python/parsers`.

I understand that this allow to easily include the Python model parser in C++. But since this code need the user to link against the binding library, I find it a little misleading.

Of course, if we don’t do the same, back compatibility could be broken.
@nim65s what do you think ?

bindings/python/parsers/python/model.cpp
```diff
@@ -2,7 +2,7 @@
 // Copyright (c) 2016-2023 CNRS INRIA
 //
 
-#include "pinocchio/parsers/python.hpp"
```
jorisv, line 5: Revert if we still install python.hpp in `pinocchio/parsers`

bindings/python/CMakeLists.txt
nim65s: Because not so long ago we required CMake 3.1

bindings/python/CMakeLists.txt
nim65s: we can install the real one on the good location, and a shim in the wrong one.
The shim would raise a warning, and then include the good.

nim65s (commented):
Using `GLOB` / `GLOB_RECURSE` to find source files is a wrong CMake practice, and I think we should use this PR to fix this.

1. This glob will be expanded only on the first configure call. Therefore, when someone with an existing successful `build` dir run `git pull` and get new source files, those new sources will be ignored by CMake, leading to hard to understand bugs
2. This glob would also include unwanted temporary files or archive artifacts 

(eg., on NetBSD, with a .tar.gz created on mac osx:
```
$ wget https://github.com/stack-of-tasks/pinocchio/releases/download/v2.6.20/pinocchio-2.6.20.tar.gz
$ gzip -d pinocchio-2.6.20.tar.gz
$ pax -r < pinocchio-2.6.20.tar
$ find pinocchio-2.6.20/src/core/ -name \*.hpp                                                                               
pinocchio-2.6.20/src/core/binary-op.hpp
pinocchio-2.6.20/src/core/PaxHeader/binary-op.hpp
pinocchio-2.6.20/src/core/PaxHeader/unary-op.hpp
pinocchio-2.6.20/src/core/unary-op.hpp
```
)
3. jrl-cmakemodule install public headers automatically when they are defined in a standard way

I understand this is boring to write the list of all the files we want, but this would help in:
- not having to write regex to conditionally exclude some files
- not having to create directories / symlink files around
- not having to write a custom function to install those

refs.:
https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/header.cmake
https://cmake.org/cmake/help/latest/command/file.html#filesystem 
> Note: We do not recommend using GLOB to collect a list of source files from your source tree

bindings/python/CMakeLists.txt
jorisv: Done

jorisv:
> Using `GLOB` / `GLOB_RECURSE` to find source files is a wrong CMake practice, and I think we should use this PR to fix this.
> 
>     1. This glob will be expanded only on the first configure call. Therefore, when someone with an existing successful `build` dir run `git pull` and get new source files, those new sources will be ignored by CMake, leading to hard to understand bugs
> 
>     2. This glob would also include unwanted temporary files or archive artifacts
> 
> 
> (eg., on NetBSD, with a .tar.gz created on mac osx:
> 
> ```
> $ wget https://github.com/stack-of-tasks/pinocchio/releases/download/v2.6.20/pinocchio-2.6.20.tar.gz
> $ gzip -d pinocchio-2.6.20.tar.gz
> $ pax -r < pinocchio-2.6.20.tar
> $ find pinocchio-2.6.20/src/core/ -name \*.hpp                                                                               
> pinocchio-2.6.20/src/core/binary-op.hpp
> pinocchio-2.6.20/src/core/PaxHeader/binary-op.hpp
> pinocchio-2.6.20/src/core/PaxHeader/unary-op.hpp
> pinocchio-2.6.20/src/core/unary-op.hpp
> ```
> 
> ) 3. jrl-cmakemodule install public headers automatically when they are defined in a standard way
> 
> I understand this is boring to write the list of all the files we want, but this would help in:
> 
>     * not having to write regex to conditionally exclude some files
> 
>     * not having to create directories / symlink files around
> 
>     * not having to write a custom function to install those
> 
> 
> refs.: https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/header.cmake https://cmake.org/cmake/help/latest/command/file.html#filesystem
> 
> > Note: We do not recommend using GLOB to collect a list of source files from your source tree

ii

jorisv:
> Using `GLOB` / `GLOB_RECURSE` to find source files is a wrong CMake practice, and I think we should use this PR to fix this.
> 
>     1. This glob will be expanded only on the first configure call. Therefore, when someone with an existing successful `build` dir run `git pull` and get new source files, those new sources will be ignored by CMake, leading to hard to understand bugs
> 
>     2. This glob would also include unwanted temporary files or archive artifacts
> 
> 
> (eg., on NetBSD, with a .tar.gz created on mac osx:
> 
> ```
> $ wget https://github.com/stack-of-tasks/pinocchio/releases/download/v2.6.20/pinocchio-2.6.20.tar.gz
> $ gzip -d pinocchio-2.6.20.tar.gz
> $ pax -r < pinocchio-2.6.20.tar
> $ find pinocchio-2.6.20/src/core/ -name \*.hpp                                                                               
> pinocchio-2.6.20/src/core/binary-op.hpp
> pinocchio-2.6.20/src/core/PaxHeader/binary-op.hpp
> pinocchio-2.6.20/src/core/PaxHeader/unary-op.hpp
> pinocchio-2.6.20/src/core/unary-op.hpp
> ```
> 
> ) 3. jrl-cmakemodule install public headers automatically when they are defined in a standard way
> 
> I understand this is boring to write the list of all the files we want, but this would help in:
> 
>     * not having to write regex to conditionally exclude some files
> 
>     * not having to create directories / symlink files around
> 
>     * not having to write a custom function to install those
> 
> 
> refs.: https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/header.cmake https://cmake.org/cmake/help/latest/command/file.html#filesystem
> 
> > Note: We do not recommend using GLOB to collect a list of source files from your source tree

I Agree on GLOB/GLOB_RECURSE pitfall.
Also the way we deleting files when options are not enabled is not pleasant and safe (it's easy to forgot).

I will have a look on jrl-cmakemodule header management.

jorisv:
@nim65s It good, last commit remove GLOB/GLOB_RECURSE for sources. And it use ${PROJECT_NAME}_HEADERS to install the headers files.

include/pinocchio/parsers/python.hpp
```diff
@@ -0,0 +1,12 @@
+//
+// Copyright (c) 2015-2023 CNRS INRIA
+//
+
+#ifndef __pinocchio_parser_python_hpp__
+#define __pinocchio_parser_python_hpp__
+
+#include "pinocchio/macros.hpp"
```
nim65s, line 8: I guess if you included this file, this was for using a macro for displaying a warning, right ?

nim65s (commented):
This is really nice, thanks !
I still have a couple comments

sources.cmake
```diff
@@ -0,0 +1,393 @@
...
+SET(${PROJECT_NAME}_HPP_FCL_PUBLIC_HEADERS
+  include/pinocchio/spatial/fcl-pinocchio-conversions.hpp
+  )
+
+SET(_binary_headers_root ${${PROJECT_NAME}_BINARY_DIR}/include/pinocchio)
+SET(${PROJECT_NAME}_CORE_PRIVATE_HEADERS
+  ${_binary_headers_root}/config.hpp
+  ${_binary_headers_root}/deprecated.hpp
+  ${_binary_headers_root}/warning.hpp
+  )
```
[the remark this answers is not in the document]
jorisv: Their both : generate and private (they will not be installed)

CMakeLists.txt
```diff
@@ -204,86 +198,80 @@ IF(BUILD_WITH_HPP_FCL_SUPPORT)
   ENDIF(BUILD_PYTHON_INTERFACE)
 ENDIF(BUILD_WITH_HPP_FCL_SUPPORT)
 
+# Sources definition
+INCLUDE(sources.cmake)
+
+IF(BUILD_WITH_URDF_SUPPORT)
+  LIST(APPEND ${PROJECT_NAME}_CORE_SOURCES ${${PROJECT_NAME}_URDF_SOURCES})
+  LIST(APPEND ${PROJECT_NAME}_CORE_PUBLIC_HEADERS ${${PROJECT_NAME}_URDF_PUBLIC_HEADERS})
+ENDIF(BUILD_WITH_URDF_SUPPORT)
```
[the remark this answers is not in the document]
jorisv: Ok to postpone this.
We also plan to apply clang-format to all the code base (after pinocchio3 merge), we could apply cmake-format at the same time).

include/pinocchio/parsers/python.hpp
jorisv: Right, I used the warning macro but after talking to @jcarpent he doesn't agree to deprecate this header.

I will remove this useless include.

nim65s: I don't see the point in not deprecating it, but this is not an issue. 
In this case we can remove the include of macros.hpp

sources.cmake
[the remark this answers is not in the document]
nim65s: I think they are generated and publicly installed by the jrl cmake modules.

How do you test their installation ?

With pip they are:
```
$ pip download pin && unzip pin-*.whl && find . -name deprecated.hpp
./cmeel.prefix/include/pinocchio/deprecated.hpp
```

With robotpkg too:
```
$ apt install -qqy robotpkg-pinocchio && find /opt/openrobots/include -name deprecated.hpp
/opt/openrobots/include/pinocchio/deprecated.hpp
```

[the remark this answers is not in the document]
jorisv: Ok, my bad, I didn't see those headers were installed by jrlcmakemodules.

It's fixed.

nim65s (approved):

PR #2080, opened 2023-11-08
depend on packaging module

singingwolfboy:
Python's [`distutils` module is deprecated](https://peps.python.org/pep-0632/), and [has been removed from Python 3.12](https://docs.python.org/3/whatsnew/3.12.html#distutils). Pinocchio is importing from `distutils.version` -- the recommended upgrade path for this is to switch to the [`packaging` module](https://pypi.org/project/packaging/), which includes [utilities for version comparison](https://packaging.pypa.io/en/stable/version.html).

Without this change, Pinocchio will not run on Python 3.12.

### DIFF
--- a/.github/workflows/conda/conda-env.yml
+++ b/.github/workflows/conda/conda-env.yml
@@ -10,4 +10,5 @@ dependencies:
   - eigenpy
   - urdfdom
   - python
+  - packaging
   - mamba
--- a/bindings/python/pinocchio/visualize/meshcat_visualizer.py
+++ b/bindings/python/pinocchio/visualize/meshcat_visualizer.py
@@ -6,7 +6,7 @@ from . import BaseVisualizer
 import os
 import warnings
 import numpy as np
-from distutils.version import LooseVersion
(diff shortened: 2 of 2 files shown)


hrp2-14:

nim65s:
Hi @singingwolfboy, and thanks for this proposition.

I don't think it is worth to add a dependency on `packaging` for this. I'm not sure how different package managers and different python versions would like this (I'm especiallly thinking about ROS here)

hpp-fcl v1.7.7 is [kind of old](https://github.com/humanoid-path-planner/hpp-fcl/releases/tag/v1.7.7), so we could basically raise the required version, which is currently [1.7.3](https://github.com/humanoid-path-planner/hpp-fcl/releases/tag/v1.7.3) according to our main CMakeLists.txt.

Another option would be to trivially parse this version number, with eg. `tuple(int(i) for i in hppfcl.__version__.split('.')) >= (1, 7, 7)`

jcarpent:
cf #2086

PR #2098, opened 2023-11-25
Add frame visualization to Meshcat visualizer

sea-bass:
This PR adds frame visualization to the Meshcat visualizer -- the use case here is for some educational content I'm making where I will start with a pure model that has no visual or collision models, and I still want the frames to show up. But I also think it would generally be useful to have this available.

Frame visualization is `False` by default to not change behavior for users.

The repro case for the animation below looks roughly like this right now:

```python
viz = MeshcatVisualizer(model, collision_model, visual_model, data=data)
viz.initViewer(open=True)
viz.loadViewerModel()
viz.displayFrames(True)  # Default args
viz.displayFrames(True, frame_ids=[10, 11, 12], axis_length=0.2, axis_width=2)  # Optional args
# You can also call viz.initializeFrames() with the same keyword args as above if you want to
# reinitialize frame visualizations with new options, without toggling visibility.

q = pinocchio.randomConfiguration(model)
v = np.zeros_like(q)
tau = np.zeros_like(q)

for i in range(100):
    a1 = pinocchio.aba(model, data, q, v, tau)
    vnext = v + dt * a1
    qnext = pinocchio.integrate(model, q, dt * vnext)

    viz.display(qnext)

    q = qnext
    v = vnext
    time.sleep(dt)
```

https://github.com/stack-of-tasks/pinocchio/assets/4603398/e8bbf5e6-c0c1-414f-b0e1-7579ba1345f2


hrp2-14:

bindings/python/pinocchio/visualize/meshcat_visualizer.py
```diff
@@ -589,6 +600,39 @@ def displayVisuals(self, visibility):
...
+                mg.PointsGeometry(
+                    position=axis_length * FRAME_AXIS_POSITIONS,
+                    color=FRAME_AXIS_COLORS,
+                ),
+                mg.LineBasicMaterial(
+                    linewidth=axis_width,
+                    vertexColors=True,
+                ),
+            )
+        )
```
nim65s, line 620: Do we really need to set those objects each and every time we draw a frame ?

sea-bass: Definitely not. We could initialize them once and only update their transforms in a loop.

As long as we expose a mechanism to update the set of objects in case frames are added/removed.

sea-bass: @nim65s I just pushed an alternate implementation that does something like what you said. Updated the syntax in the PR description.

nim65s (approved):
This looks good to me, thanks for adding this useful feature !

I guess we could add parameters in `loadViewerModel` to let the user define if they want to display collisions, visuals, and/or frames or not, but this is out of the scope of this PR, and would better be consistently applied on all viewers.

jcarpent (dismissed):

jcarpent (approved):
Thanks @sea-bass for this very helpful contribution.

jcarpent (approved):

PR #2104, opened 2023-11-29
Bug fix in mesh scaling

matheecs:


jcarpent:
What is the bug here?

PR #2112, opened 2023-12-04
CMake: export CppAD finder

nim65s:
to fix https://github.com/Gepetto/example-robot-data/issues/182


jcarpent (approved):

jcarpent (commented):
I approved, but in fact you don't have to export CppAD as we do not linked against it for the core project, only for the tests.

jcarpent (approved):

Issue #2084, opened 2023-11-13
Add Material in `pinocchio::GeometryObject`


jorisv:
`pinocchio::GeometryObject` already hold renderer information (texture path, color).
Since a lot of renderer ask for Material, it can be nice to add it.

[Material definition in OpenGL](https://en.wikibooks.org/wiki/OpenGL_Programming/Intermediate/Materials)

Discussed in https://github.com/stack-of-tasks/pinocchio/discussions/2079

PR #2119, opened 2023-12-13
Add material to GeometryObject

jorisv:
This PR solve #2084.

I have added `GeometryObject::meshMaterial` attribute to store different kind of material with a boost::variant.

Right now, only two material type exists : GeometryNoMaterial and GeometryPhongMaterial.
But more complex ones can be added in the future.

The boost::variant Python binding is working well and can maybe be ported to manage the `Model::joints`.

`meshcat_visualizer.py` use this new feature BUT [meshcat-python Phong material](https://github.com/meshcat-dev/meshcat-python/blob/785bc9d5ba6f8a8bb79ee8b25f523805946c1fbd/src/meshcat/geometry.py#L185) is not implemented.
I had to patch my meshcat-python to be able to display specular et emissive light.

Before merging I have to add an example but I would like to have a technical review before going further.

### DIFF
--- a/include/pinocchio/bindings/python/multibody/geometry-object.hpp
+++ b/include/pinocchio/bindings/python/multibody/geometry-object.hpp
@@ -33,7 +33,7 @@ namespace pinocchio
         template<typename T>
         result_type operator()(T & t) const
         {
-          return boost::python::incref(boost::python::object(t).ptr());
+          return bp::incref(bp::object(t).ptr());
         }
       };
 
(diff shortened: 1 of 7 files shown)


include/pinocchio/bindings/python/multibody/geometry-object.hpp
```diff
@@ -19,26 +19,99 @@ namespace pinocchio
...
+      {
+        static result_type convert(GeometryMaterial const & gm)
+        {
+          return apply_visitor(GeometryMaterialValueToObject(), gm);
+        }
+
+        template<typename T>
+        result_type operator()(T & t) const
+        {
+          return boost::python::incref(boost::python::object(t).ptr());
```
jcarpent, line 36:
```suggestion
          return bp::incref(bp::object(t).ptr());
```

jcarpent (approved):
Looks good to me.

jcarpent (dismissed):

jcarpent:
The CI failure seems to be independent of the content of this PR. Let's move on.

PR #2124, opened 2023-12-20
crba: avoid temporaries with .noalias()

alaurenzi:
Dear maintainers, 
for a few projects we'd like to avoid allocating memory inside function running inside a real time loop.

I have found out that a few temporaries are generated inside the `crba` function. Adding a `.noalias()` here and there was enough to keep Eigen from generated temporaries during matrix multiplications.

Let me know what you think!

Arturo


hrp2-14:

alaurenzi:
Whop! Wrong base branch.. I'll re-open it from devel to avoid the "out of date branch" warning

PR #2125, opened 2023-12-20
crba: avoid temporaries with .noalias()

alaurenzi:
Dear maintainers,
for a few projects we'd like to avoid allocating memory inside function running inside a real time loop.

I have found out that a few temporaries are generated inside the crba function. Adding a .noalias() here and there was enough to keep Eigen from generating temporaries during matrix multiplications.

Let me know what you think!

Arturo

### DIFF
--- a/include/pinocchio/algorithm/crba.hxx
+++ b/include/pinocchio/algorithm/crba.hxx
@@ -74,10 +74,10 @@ namespace pinocchio
 
       /* F[1:6,i] = Y*S */
       //data.Fcrb[i].block<6,JointModel::NV>(0,jmodel.idx_v()) = data.Ycrb[i] * jdata.S();
-      jmodel.jointCols(data.Fcrb[i]) = data.Ycrb[i] * jdata.S();
+      jmodel.jointCols(data.Fcrb[i]).noalias() = data.Ycrb[i] * jdata.S();
 
       /* M[i,SUBTREE] = S'*F[1:6,SUBTREE] */
-      data.M.block(jmodel.idx_v(),jmodel.idx_v(),jmodel.nv(),data.nvSubtree[i]) 
+      data.M.block(jmodel.idx_v(),jmodel.idx_v(),jmodel.nv(),data.nvSubtree[i]).noalias()
       = jdata.S().transpose()*data.Fcrb[i].middleCols(jmodel.idx_v(),data.nvSubtree[i]);
 
       const JointIndex & parent = model.parents[i];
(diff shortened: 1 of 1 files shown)


ManifoldFR:
Hi,

You found that a malloc happened at the lines you modified ? I thought the involved matrices had fixed Eigen sizes.

alaurenzi:
For some of them you may be right.. I belive the problematic line is
```
data.M.block(jmodel.idx_v(),jmodel.idx_v(),jmodel.nv(),data.nvSubtree[i]).noalias() = ... * ...
```
where the LHS block has dynamic size.

ManifoldFR:
You're right, the LHS is a dynamic-sized Eigen expression.

Have you run the tests/benchmarks with this change and the `CHECK_RUNTIME_MALLOC` flag (I think you need a Debug build type for that)?

jcarpent:
@alaurenzi Thanks for reaching us. Could you provide a robot model that triggers allocation?
It would be indeed required to check on our side.

alaurenzi:
Hi @jcarpent , please have a look to this minimal self contained example: [pinocchio_malloc_test.zip](https://github.com/stack-of-tasks/pinocchio/files/13740137/pinocchio_malloc_test.zip)

- `wget https://github.com/stack-of-tasks/pinocchio/files/13740137/pinocchio_malloc_test.zip`
- `unzip pinocchio_malloc_test.zip`
- `cd pinocchio_malloc_test`
- `bash run.bash`
- The program will compile and execute under `gdb`. It is supposed to crash on malloc, then `bt` to inspect backtrace

On my machine the interesting line is
```
#16 0x00005555555acb3b in pinocchio::CrbaBackwardStep<double, 0, pinocchio::JointCollectionDefaultTpl>::algo<pinocchio::JointModelRevoluteUnalignedTpl<double, 0> > (jmodel=..., jdata=..., mo
del=..., data=...)
    at /home/alaurenzi/code/core_ws/install/include/pinocchio/algorithm/crba.hxx:81
```

which corresponds to  
```
/* M[i,SUBTREE] = S'*F[1:6,SUBTREE] */
      data.M.block(jmodel.idx_v(),jmodel.idx_v(),jmodel.nv(),data.nvSubtree[i]) 
      = jdata.S().transpose()*data.Fcrb[i].middleCols(jmodel.idx_v(),data.nvSubtree[i]);
```

Adding `.noalias()` fixes it.

jcarpent:
See #2126 for the final solving + test

jcarpent:
@alaurenzi I've provided a clean revision in #2126 with a unit test. Could you check on your side if it does what you expect?

Issue #2146, opened 2024-02-01
Meshcat visualizer error when collision model is undefined


stephane-caron:
## Bug description

The Meshcat visualizer tries to read from the collision model even when it is undefined (``collision_model=None``):

```
You can open the visualizer by visiting the following URL:
http://127.0.0.1:7001/static/
Traceback (most recent call last):
  File "/home/stephane-caron/src/project/run.py", line 361, in <module>
    run(scenario, args, rho, mu)
  File "/home/stephane-caron/src/project/run.py", line 189, in run
    scene = Scene(scenario, record=args.record)  # scene is not multiprocess
  File "/home/stephane-caron/src/project/project/scene.py", line 49, in __init__
    visualizer = start_meshcat_visualizer(robot)
  File "/home/stephane-caron/.micromamba/envs/project/lib/python3.10/site-packages/pink/visualization.py", line 26, in start_meshcat_visualizer
    viz.loadViewerModel()
  File "/home/stephane-caron/.micromamba/envs/project/lib/python3.10/site-packages/pinocchio/visualize/meshcat_visualizer.py", line 641, in loadViewerModel
    for collision in self.collision_model.geometryObjects:
AttributeError: 'NoneType' object has no attribute 'geometryObjects'
```

## Expected behavior

The visualizer ignores the collision model when it is not defined.

## Reproduction steps

Create a visualizer with ``collision_model=None``.

jorisv:
Fixed in #2147

PR #2147, opened 2024-02-01
Don't iterate on collision geometries when undefined

stephane-caron:
Closes https://github.com/stack-of-tasks/pinocchio/issues/2146


stephane-caron:
I have tested this change locally and it loads my collision-less model.

(I looked into the unit test directory to see if we already had a fixture for visualizers where I could add a test for this, but we don't. Maybe hard to add to the CI?)

jorisv (changes requested):
Add a changelog entry (under Fixed section)

jorisv (approved):

PR #2149, opened 2024-02-02
Accelerate CI

jorisv:
- On new push, cancel running workflow
- Avoid running workflow on some file change
- Fix ccache use in linux.yml


.github/workflows/linux.yml
```diff
@@ -1,6 +1,27 @@
...
+on:
+  push:
+    paths-ignore:
+      - 'doc/**'
+      - '.gitlab-ci.yml'
+      - '.gitignore'
+      - '*.md'
+      - 'CITATION.*'
+      - 'COPYING.LESSER'
+      - 'colcon.xml'
```
fabinsch, line 12: I do not find a file called like this, just `colcon.pkg` .

.github/workflows/linux.yml
```diff
@@ -1,6 +1,27 @@
 name: CI - Linux via APT
 
-on: [push,pull_request]
+on:
+  push:
+    paths-ignore:
```
fabinsch, line 5: maybe we could add `.pre-commit-config.yaml` ?

.github/workflows/linux.yml
```diff
@@ -12,7 +33,10 @@ jobs:
         os: [ubuntu-22.04,ubuntu-20.04]
 
     env:
-      CCACHE_DIR: /github/home/.ccache
+      CCACHE_BASEDIR: "${GITHUB_WORKSPACE}"
+      CCACHE_DIR: "${GITHUB_WORKSPACE}/.ccache"
+      CCACHE_COMPRESS: true
+      CCACHE_COMPRESSLEVEL: 6
```
fabinsch, line 39: how did you choose level 6 ?

> Ccache will by default compress all data it puts into the cache using the compression algorithm [Zstandard](http://zstd.net/) (zstd) using compression level 1. The algorithm is fast enough that there should be little reason to turn off compression to gain performance.

jorisv: I don't remember where I found the level 6.  But if you search this variable in Github you can find some project that configure it : https://github.com/search?q=CCACHE_COMPRESSLEVEL%3D&type=code

Since we don't have an infinite cache (10 Go I think) it's useful to compress to avoid cache destruction.

But I didn't do a benchmark to choose it.

> A positive value corresponds to normal Zstandard compression levels. Lower levels (e.g. 1) mean faster compression but worse compression ratio. Higher levels (e.g. 19) mean slower compression but better compression ratio. The maximum possible value depends on the libzstd version, but at least up to 19 is available for all versions. Decompression speed is essentially the same for all levels. As a rule of thumb, use level 5 or lower since higher levels may slow down compilations noticeably. Higher levels are however useful when recompressing the cache with command line option -X/--recompress.

.github/workflows/linux.yml
jorisv: You're right. In fact, we don't have any CI job that check the pre-commit. I will add it.

.github/workflows/linux.yml
jorisv: Fixed

fabinsch (approved):

fabinsch (approved):

Issue #2158, opened 2024-03-01
Parent frame of GeometryObject changes after calling BuildReducedModel


joao-pm-santos96:
## Bug description

Hello developers! Can you please help me?

When I call BuildReducedModel, the parentFrame of my GeometryObjects change to a non-existing frame. I this the issue is here:
https://github.com/stack-of-tasks/pinocchio/blob/0caf0ca4d07e63834cdc420c703993662c59e01b/include/pinocchio/algorithm/model.hxx#L569-L570

## Expected behavior

They shouldn't change.

As always, thank you for your time!

jcarpent:
Thanks @joao-pm-santos96 for raising this issue.
Could you provide a small tiny example to reproduce it locally in such a way we prepare a unit tests covering your case?

joao-pm-santos96:
Of course! In here https://github.com/stack-of-tasks/pinocchio/blob/0caf0ca4d07e63834cdc420c703993662c59e01b/unittest/model.cpp#L458-L464

I would also add

```c++
BOOST_CHECK_EQUAL(go1.parentFrame, go2.parentFrame);
```

joao-pm-santos96:
Also, building upon https://github.com/stack-of-tasks/pinocchio/blob/master/examples/build-reduced-model.py , here something that fails:

```python
import pinocchio as pin
import numpy as np

from os.path import *

# Goal: Build a reduced model from an existing URDF model by fixing the desired joints at a specified position.

# Load UR robot arm
# This path refers to Pinocchio source code but you can define your own directory here.
pinocchio_model_dir = join(dirname(dirname(str(abspath(__file__)))), "models")
model_path = pinocchio_model_dir + '/example-robot-data/robots'
mesh_dir = pinocchio_model_dir
# You should change here to set up your own URDF file
urdf_filename = model_path + '/ur_description/urdf/ur5_robot.urdf'
model, collision_model, visual_model = pin.buildModelsFromUrdf(urdf_filename, mesh_dir)

# Check dimensions of the original model
print('standard model: dim=' + str(len(model.joints)))
for jn in model.joints:
    print(jn)
print('-' * 30)

# Create a list of joints to lock
jointsToLock = ['wrist_1_joint', 'wrist_2_joint', 'wrist_3_joint']

# Get the ID of all existing joints
jointsToLockIDs = []
for jn in jointsToLock:
    if model.existJointName(jn):
        jointsToLockIDs.append(model.getJointId(jn))
    else:
[...]

PR #2160, opened 2024-03-05
Fix Frames order in reducedModel

MegMll:
Fix for frames list order when building a reducedModel. 
Fix the problem  in the geometryReduced Model mentioned in  #2158 .

### DIFF
--- a/include/pinocchio/algorithm/model.hxx
+++ b/include/pinocchio/algorithm/model.hxx
@@ -501,8 +501,7 @@ namespace pinocchio
       reduced_model.referenceConfigurations.insert(std::make_pair(config_name, reduced_config_vector));
     }
     
-    // Add all frames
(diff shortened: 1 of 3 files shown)


include/pinocchio/algorithm/model.hxx
```diff
@@ -462,8 +502,8 @@ namespace pinocchio
     }
     
     // Add all frames
-    typename Model::FrameVector::const_iterator frame_it = input_model.frames.begin();
-    for(++frame_it;frame_it != input_model.frames.end(); ++frame_it)
+    // typename Model::FrameVector::const_iterator frame_it = input_model.frames.begin();
```
jcarpent, line 505: This line should be removed if no more used.

jcarpent:
@MegMll You need to also update the changelog ;)

jcarpent (approved):

PR #2204, opened 2024-04-04
Introduced inertia/symmetric3 minus operators

cmastalli:
This PR introduces minus operators for symmetric3 and inertia classes. It includes Python bindings and unit tests.

### DIFF
--- a/include/pinocchio/spatial/inertia.hpp
+++ b/include/pinocchio/spatial/inertia.hpp
@@ -360,9 +360,11 @@ namespace pinocchio
     {
       PINOCCHIO_ASSERT_MATRIX_SPECIFIC_SIZE(Vector10Like, params, 10, 1);
 
+      const Scalar eps = ::Eigen::NumTraits<Scalar>::epsilon();
+
       const Scalar & mass = params[0];
       Vector3 lever = params.template segment<3>(1);
-      lever /= mass;
+      lever /= (mass >= 0) ? math::max(mass,eps) : math::min(mass,-eps);
 
       return InertiaTpl(mass, lever, Symmetric3(params.template segment<6>(4)) + AlphaSkewSquare(mass,lever));
     }
--- a/include/pinocchio/spatial/inertia.hpp
+++ b/include/pinocchio/spatial/inertia.hpp
@@ -363,6 +363,7 @@ namespace pinocchio
       const Scalar eps = ::Eigen::NumTraits<Scalar>::epsilon();
 
       const Scalar & mass = params[0];
+      assert(mass >= Scalar(0.));
       Vector3 lever = params.template segment<3>(1);
       lever /= (mass >= 0) ? math::max(mass,eps) : math::min(mass,-eps);
 
@@ -440,6 +441,7 @@ namespace pinocchio
       const Scalar eps = ::Eigen::NumTraits<Scalar>::epsilon();
 
       const Scalar & mab = mass()-Yb.mass();
+      assert(mab >= Scalar(0.));
       const Scalar mab_inv = (mab >= 0) ? Scalar(1)/math::max(mab,eps) : Scalar(1)/math::min(mab,-eps);
       const Vector3 & AB = (lever()-Yb.lever()).eval();
       return InertiaTpl(mab,
@@ -453,6 +455,7 @@ namespace pinocchio
       
       const InertiaTpl& Ya = *this;
       const Scalar & mab = mass()-Yb.mass();
+      assert(mab >= Scalar(0.));
       const Scalar mab_inv = (mab >= 0) ? Scalar(1)/math::max(mab,eps) : Scalar(1)/math::min(mab,-eps);
       const Vector3 & AB = (Ya.lever()-Yb.lever()).eval();
       lever() *= (mass()*mab_inv); lever() -= (Yb.mass()*mab_inv)*Yb.lever(); //c *= mab_inv;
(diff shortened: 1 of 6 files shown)


include/pinocchio/spatial/inertia.hpp
```diff
@@ -424,6 +428,37 @@ namespace pinocchio
...
+    InertiaTpl __minus__(const InertiaTpl & Yb) const
+    {
+      /* Y_{a-b} = ( m_a-m_b,
+       *             (m_a*c_a - m_b*c_b ) / (m_a - m_b),
+       *             I_a - I_b + (m_a*m_b)/(m_a-m_b) * AB_x * AB_x )
+       */
+
+      const Scalar eps = ::Eigen::NumTraits<Scalar>::epsilon();
+
+      const Scalar & mab = mass()-Yb.mass();
```
jcarpent, line 440: It would be nice to check whether the resulting inertia remains physically consistent, certainly with an assert.

include/pinocchio/spatial/symmetric3.hpp
```diff
@@ -339,6 +340,16 @@ namespace pinocchio
       m_data += s2.m_data; return *this;
     }
 
+    Symmetric3Tpl operator-(const Symmetric3Tpl & s2) const
+    {
+      return Symmetric3Tpl((m_data-s2.m_data).eval());
```
jcarpent, line 345: The eval is not needed here.
```suggestion
      return Symmetric3Tpl(m_data-s2.m_data);
```

include/pinocchio/spatial/inertia.hpp
```diff
@@ -424,6 +428,37 @@ namespace pinocchio
...
+      const Scalar mab_inv = (mass()-Yb.mass() >= 0) ? Scalar(1)/math::max((Scalar)(mass()-Yb.mass()),eps) : -Scalar(1)/math::max((Scalar)(Yb.mass()-mass()),eps);
```
jcarpent, line 441: You do not even reuse `mab` here.

include/pinocchio/spatial/inertia.hpp
```diff
@@ -424,6 +428,37 @@ namespace pinocchio
...
+                        inertia()-Yb.inertia() + (mass()*Yb.mass()*mab_inv)* typename Symmetric3::SkewSquare(AB));
+    }
+
+    InertiaTpl& __mequ__(const InertiaTpl & Yb)
+    {
+      const Scalar eps = ::Eigen::NumTraits<Scalar>::epsilon();
+      
+      const InertiaTpl& Ya = *this;
+      const Scalar & mab = mass()-Yb.mass();
+      const Scalar mab_inv = (mass()-Yb.mass() >= 0) ? Scalar(1)/math::max((Scalar)(mass()-Yb.mass()),eps) : -Scalar(1)/math::max((Scalar)(Yb.mass()-mass()),eps);
```
jcarpent, line 454: Same comments applied here.

jcarpent (changes requested):
Thanks @cmastalli for adding this operation.
I made some comments that you can easily handle to enhance the overall behavior and efficiency.

include/pinocchio/spatial/inertia.hpp
jcarpent, line 441: I don't think that the case `mab < 0` has a physical meaning.

include/pinocchio/spatial/inertia.hpp
cmastalli: I have included asserts for checking mass is positive (here and in FromDynamicParameters). However, I wonder what is the expected design for this class. Should `InertialTpl` handle fully physically consistent only? Or could it accept "partial" physical consistency or anything?

For instance, the `Random` function generates positive masses and rotational inertias with positive principal inertia components. However, this is not what we refer to as *fully physical consistency*. This means also having positive second moments of inertia, i.e., satisfying the named "triangular inequalities". In short, **Random function doesn't support full physical consistency, just a partial physical consistency**.

Moreover, `FromDynamicParameters` doesn't assert positive masses or rotation inertia triangular inequalities. In short, **FromDynamicParameters function supports any type of parameter including the ones that don't satisfy physics**.

In conclusion, I am inclined 
  1. to allow any time of parameters, 
  2. to remove asserts (as requested and introduced in this PR), and
  3. to define a "checkConsistency" function where users can trigger their checks on demand.

Please let me know what are your thoughts and how to proceed with this PR.

include/pinocchio/spatial/inertia.hpp
cmastalli: This depends on how we want to design this class. At the moment, it is unclear to me. See my comment at https://github.com/stack-of-tasks/pinocchio/pull/2204#discussion_r1552596106.

include/pinocchio/spatial/inertia.hpp
cmastalli: @jcarpent -- please let me know your thoughts. I still think we should not include these asserts as resulting operators might not necessarily satisfy fully physical consistency. The reason is this class should change much more to do that. We can arrange a meeting to discuss this.

Be aware that the unit tests are not passing because they are designed to test conditions that don't meet full physical consistency. I'll need to modify them if we want to do that.

jorisv:
Hello @cmastalli,

Do you still need this PR to be merged ? @jcarpent tell me it was to do some Inertia estimation.
Is the [Pseudo Inertia and Log Cholesky Parametrization](https://github.com/stack-of-tasks/pinocchio/pull/2296) solving your issue ?

jorisv:
@cmastalli I will take care of the rebase.

cmastalli:
> @cmastalli I will take care of the rebase.

Thanks :)

PR #2218, opened 2024-04-11
Modify model appending algorithm

florent-lamiraux:
This modification follows https://github.com/stack-of-tasks/pinocchio/discussions/2203.
  
In the new version, when appending model B to model A on frame F, joints are copied in the
following order
  - joints of model A up to the parent of F,
  - all the descendents of parent of F,
  - the remaining joints of A.

### DIFF
--- a/include/pinocchio/algorithm/model.hxx
+++ b/include/pinocchio/algorithm/model.hxx
@@ -221,6 +221,23 @@ namespace pinocchio
          model, geomModel);
   }
 
+  // Compute whether Joint child is a descendent of parent in a given model
+  // Joints are represented by their id in the model
+  template <typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
+  static bool hasAncestor(const ModelTpl<Scalar,Options,JointCollectionTpl> & model,
+                          JointIndex child, JointIndex parent)
+  {
+    typedef typename ModelTpl<Scalar,Options,JointCollectionTpl>::IndexVector IndexVector_t;
+    // Any joints has universe as an acenstor
+    assert(model.supports[child][0] == 0);
+    for (typename IndexVector_t::const_iterator it=model.supports[child].begin();
+	 it!=model.supports[child].end(); ++it)
+    {
+      if (*it == parent) return true;
+    }
+    return false;
+  }
+
   template<typename Scalar, int Options, template<typename,int> class JointCollectionTpl>
   void
   appendModel(const ModelTpl<Scalar,Options,JointCollectionTpl> & modelA,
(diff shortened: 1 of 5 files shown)


include/pinocchio/algorithm/model.hxx
```diff
@@ -221,6 +221,24 @@ namespace pinocchio
...
+    if (child == parent) return true;
+    JointIndex i = child;
+    while (i != 0)
+    {
+      i = model.parents[i];
+      if (i == parent) return true;
+    }
+    return false;
+  }
+
```
jcarpent, line 241: Could you rather use `model.supports[child]` to check whether parent is supporting joint of `child`?

include/pinocchio/algorithm/model.hxx
```diff
@@ -256,9 +274,28 @@ namespace pinocchio
...
+      if (hasAncestor(modelA, jid, frame.parent))
+      {
+        AJointsBeforeB.push_back(jid);
+      } else
+      {
+        AJointsAfterB.push_back(jid);
+      }
+    }
+    // Copy modelA joints that should come before model B
+    for (auto jid : AJointsBeforeB)
```
jcarpent, line 298: This is a C++11 feature, targeted for Pinocchio 3.

include/pinocchio/algorithm/model.hxx
```diff
@@ -275,7 +312,8 @@ namespace pinocchio
     }
 
     // Copy remaining joints of modelA
-    for (JointIndex jid = frame.parent+1; jid < modelA.joints.size(); ++jid)
+    // Copy modelA joints that should come before model B
+    for (auto jid : AJointsAfterB)
```
jcarpent, line 316: Same comment apply here.

jcarpent (changes requested):
Thanks a lot @florent-lamiraux for working on this feature.
I made some remarks to enhance the integration of this PR into Pinocchio.

florent-lamiraux:
@jcarpent : I will handle your comments after fixing the tests.

include/pinocchio/algorithm/model.hxx
florent-lamiraux: It seems to me that the documentation of `model.supports` is incorrect. I would expect:
"supports[j] corresponds to the collection of all joints located on the path between joint *j*  and the root."
Is it correct ? If so, I will fix the doc in this PR.

include/pinocchio/algorithm/model.hxx
florent-lamiraux: done

include/pinocchio/algorithm/model.hxx
florent-lamiraux: done

florent-lamiraux:
I addressed the above comments, modified test-cpp-model to test the new behavior, and improved the documentation
of Model.supports.

include/pinocchio/algorithm/model.hxx
```diff
@@ -256,12 +273,32 @@ namespace pinocchio
...
+    for (JointIndex jid = frame.parent+1; jid < modelA.joints.size(); ++jid)
+    {
+      if (hasAncestor(modelA, jid, frame.parent))
+      {
+        AJointsBeforeB.push_back(jid);
+      } else
+      {
+        AJointsAfterB.push_back(jid);
+      }
+    }
```
jmirabel, line 295: Given the compact organization of the list of joints, I would expect that you can copy all the joints of model A until the first joint whose id `id_insert` matches:
- `id_frame_parent < id_insert`: the joint is after the `frame.parent` where modelB is appended.
- `modelA.parents[id_insert] < id_frame_parent`: the joint parent is before `frame.parent`.

`id_insert` is to be added after model B.

If that is true, the code would *arguably* be more atomic.

jmirabel: Note that I'm not specifically asking you to change. I'm done with the current implementation.

jcarpent (approved):
I will merge this PR as it seems fine, even if it can be improved. Let's postpone it as a future work.

include/pinocchio/algorithm/model.hxx
```diff
@@ -256,12 +273,32 @@ namespace pinocchio
...
     details::appendUniverseToModel (modelA, geomModelA, 0, id, model, geomModel);
+    // Compute joints of A that should be added before and after joints of B
+    std::vector<JointIndex> AJointsBeforeB;
+    std::vector<JointIndex> AJointsAfterB;
+    // All joints until the parent of frameInModelA come first
     for (JointIndex jid = 1; jid <= frame.parent; ++jid)
+    {
+      AJointsBeforeB.push_back(jid);
+    }
+    // descendants of the parent of frameInModelA come also before model B
```
jcarpent, line 285:
```suggestion
    // descendants of the parent of frameInModelA come also before model B
    // TODO(jcarpent): enhancement by taking into account the compactness of the joint ordering.
```

PR #2263, opened 2024-05-30
add python test for `contactInverseDynamics`

fabinsch:
In a similar way to the [cpp test](https://github.com/stack-of-tasks/pinocchio/blob/master/unittest/contact-inverse-dynamics.cpp#L82) for the `contactInverseDynamics` function, I propose to add a python test and check if this function is usable from Python. 

the c++ function requires for the contact models, data and cones
```
contactInverseDynamics(..., 
const std::vector<RigidConstraintModelTpl<Scalar, Options>, ConstraintModelAllocator> & contact_models,
std::vector<RigidConstraintDataTpl<Scalar, Options>, ConstraintDataAllocator> & contact_datas,
const std::vector<CoulombFrictionConeTpl<Scalar>, CoulombFrictionConelAllocator> & cones,
```

and this works well when constructing them as `StdVec_RigidConstraintData()` and then appending contact data here (see test1). However, if you try to pass a list of contact data, it does not work (see test3).
This is not the case for contact models and cones, they work both as list of model/cones and as `pin.StdVec` (see test2).

The different is, that models/cones are `const std::vectors` while data is not const.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - Fix compilation issue for Boost 1.85 ([#2255](https://github.com/stack-of-tasks/pinocchio/pull/2255))
+- Fix python bindings of `contactInverseDynamics` ([#2263](https://github.com/stack-of-tasks/pinocchio/pull/2263))
 
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Fix compilation issue for Boost 1.85 ([#2255](https://github.com/stack-of-tasks/pinocchio/pull/2255))
 
+
+### Added
+
+- Python unittest for `contactInverseDynamics` function ([#2263](https://github.com/stack-of-tasks/pinocchio/pull/2263))
+
 ## [3.0.0] - 2024-05-27
 
 ### Added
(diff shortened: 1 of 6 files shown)


fabinsch:
The problem of calling the Python bindings of `contactInverseDynamics` with a list of `RigidConstraintData` objects described above is fixed with https://github.com/stack-of-tasks/pinocchio/pull/2263/commits/5dc80d65e714890bc6c78ea52d54c844cf90532a.

jorisv (dismissed):

jcarpent (approved):
Very great job @fabinsch. Thanks for providing the fix AND the precious unit test.

fabinsch:
> I don't even know if it's actually useful.

I checked and we use it [here](https://github.com/stack-of-tasks/pinocchio/blob/master/bindings/python/parsers/urdf/geometry.cpp#L57) in `bindings/python/parsers/urdf/geometry.cpp`. However, I am not sure why we also include it in `expose-model.cpp` and `expose-rnea.cpp`.  This should probably be removed there.

jorisv (approved):

jcarpent:
> > I don't even know if it's actually useful.
> 
> I checked and we use it [here](https://github.com/stack-of-tasks/pinocchio/blob/master/bindings/python/parsers/urdf/geometry.cpp#L57) in `bindings/python/parsers/urdf/geometry.cpp`. However, I am not sure why we also include it in `expose-model.cpp` and `expose-rnea.cpp`. This should probably be removed there.

We use in rnea because of `fext` arg, we might be a list.

fabinsch:
yes, you are right that the external forces can be passed as a list from python to the `rnea` bindings. 

I checked why our unittest for this [case](https://github.com/stack-of-tasks/pinocchio/blob/master/unittest/python/bindings_rnea.py#L71) is still working (see CI) without the `list.hpp` header in `expose-rnea`. The reason is, that the `StdAlignedVectorPythonVisitor` for `Force`, called [here](https://github.com/stack-of-tasks/pinocchio/blob/master/bindings/python/spatial/expose-force.cpp#L21) to expose `StdVec_Force`, implements a  `FromPythonListConverter` [here](https://github.com/stack-of-tasks/pinocchio/blob/master/include/pinocchio/bindings/python/utils/std-aligned-vector.hpp#L84). I also verified this on my machine with a quick test.

PR #2273, opened 2024-06-04
Test vector transport and its inverse with `dIntegrateTransport`

fabinsch:
Hi, I propose adding another test for the `dIntegrateTransport` function to check its "inverse" by using the reverse path. I use this function for vector transport along different manifolds.

### DIFF
--- a/unittest/liegroups.cpp
+++ b/unittest/liegroups.cpp
@@ -490,6 +490,53 @@ struct LieGroup_JintegrateJdifference
   }
 };
 
+struct LieGroup_dIntegrateTransport
+{
+  template<typename T>
+  void operator()(const T) const
+  {
+    typedef typename T::ConfigVector_t ConfigVector_t;
+    typedef typename T::TangentVector_t TangentVector_t;
+    typedef typename T::JacobianMatrix_t JacobianMatrix_t;
+
+    PINOCCHIO_COMPILER_DIAGNOSTIC_PUSH
+    PINOCCHIO_COMPILER_DIAGNOSTIC_IGNORED_MAYBE_UNINITIALIZED
+    T lg;
+    BOOST_TEST_MESSAGE(lg.name());
+    ConfigVector_t qa, qb(lg.nq());
+    qa = lg.random();
+    TangentVector_t v(lg.nv()), tvec_at_q1(lg.nv()), tvec_at_q0(lg.nv()), tvec_at_q0_(lg.nv());
+    v.setRandom();
+    lg.integrate(qa, v, qb);
+
+    // transport random tangent vector from q1 to q0
+    tvec_at_q1.setRandom();
+    lg.dIntegrateTransport(qa, v, tvec_at_q1, tvec_at_q0, ARG0);
+
+    // test opposite direction
+    ConfigVector_t qa_ = qb;
+    TangentVector_t v_ = -v; // reverse path
+    ConfigVector_t qb_ = lg.integrate(qa_, v_);
+
+    BOOST_CHECK_SMALL((qa - qb_).norm(), 1e-6); // recover init point on manifold
+
+    TangentVector_t tvec_at_q1_ = tvec_at_q0;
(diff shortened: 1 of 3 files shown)


ManifoldFR:
Great contribution @fabinsch ! Can you also test for matrices ? It should still work the same

fabinsch:
Sure, I will extend it for matrices. One point for me was to check if it actually works for a single vector (`np.array`), as the function doc explicitly speaks about "transporting a matrix". But as the doc string states correctly, only the row size matters and this was tested here.

```
In [2]: pin.dIntegrateTransport?
Docstring:
dIntegrateTransport( (Model)model, (numpy.ndarray)q, (numpy.ndarray)v, (numpy.ndarray)Jin, (ArgumentPosition)argument_position) -> numpy.ndarray :
    Takes a matrix expressed at q (+) v and uses parallel transport to express it in the tangent space at q.    This operation does the product of the matrix by the Jacobian of the integration operation, but more efficiently.Parameters:
        model: model of the kinematic tree
        q: the joint configuration vector (size model.nq)
        v: the joint velocity vector (size model.nv)
        Jin: the input matrix (row size model.nv)
```

ManifoldFR:
perfect!

ManifoldFR (dismissed):

jcarpent (commented):
Does this test already exist in C++?

fabinsch:
> Does this test already exist in C++?

we are currently testing [this](https://github.com/stack-of-tasks/pinocchio/blob/master/unittest/joint-configurations.cpp#L257) in our cpp unittests for `dIntegrateTransport` which is not exactly the same as my test I think.

stephane-caron (commented):
(Minor comment.)

unittest/python/bindings_liegroups.py
```diff
@@ -104,6 +104,42 @@ def test_dIntegrateTransport(self):
...
+            q0 = lg.random()
+            v = np.random.rand(lg.nv)
+            q1 = lg.integrate(q0, v)
+
+            # transport random tangent vector from q1 to q0
+            tvec_at_q1 = np.random.rand(lg.nv)
+            tvec_at_q0 = lg.dIntegrateTransport(q0, v, tvec_at_q1, pin.ARG0)
+
+            # test opposite direction
+            q0_ = q1.copy()
```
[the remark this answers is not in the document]
fabinsch: I renamed the variables and tried to simplify slightly, hope it is clearer now.

unittest/liegroups.cpp
```diff
@@ -490,6 +490,49 @@ struct LieGroup_JintegrateJdifference
...
+    lg.integrate(qa, v, qb);
+
+    // transport random tangent vector from q1 to q0
+    tvec_at_qb.setRandom();
+    lg.dIntegrateTransport(qa, v, tvec_at_qb, tvec_at_qa, ARG0);
+
+    // test reverse direction
+    TangentVector_t v_r = -v; // reverse path
+    ConfigVector_t qa_r = lg.integrate(qb, v_r);
+    lg.dIntegrateTransport(qa_r, v_r, tvec_at_qa, tvec_at_qa_r, ARG0);
```
stephane-caron, line 519: Shouldn't this be ``dIntegrateTransport(qb, v_r, tvec_at_qa, ...)``?

Since we check that ``qa`` is equal to ``qa_r`` below, this line amounts to integrating $v_r$ from $qa_r$, where we will end... Somewhere else than $q_a$ or $q_b$ :sweat_smile:

Meanwhile this line would fully make sense to me if we start from $q_b$: then $q_b \oplus v_r$ would land on $q_{ar} = q_a$, where we take ``tvec_at_qa``, and transport it back to $q_b$, which is why we have the equality between ``tvec_at_qb`` and ``tvec_at_qa_r`` below. (But then, ``tvec_at_qa_r`` lies in the tangent space at $q_b$, not $q_{ar}$! :stuck_out_tongue:)

@fabinsch I understand the test passed so I must have missed something. If you can double check at some point that would be :ok_hand:

stephane-caron: Meanwhile I fully agree with the Python variant.

PR #2288, opened 2024-06-18
Always suppose hpp-fcl Python binding are present when hpp-fcl C++ library is found

jorisv:
Pinocchio Python binding needs hpp-fcl Python binding for collision/contact related algorithms.

We allow with the `BUILD_WITH_HPP_FCL_PYTHON_BINDINGS` CMake variable to build Pinocchio C++ library with hpp-fcl but to ignore hpp-fcl in the Python bindings if the hpp-fcl Python bindings are not installed.

This create the following issues:
- To compute `BUILD_WITH_HPP_FCL_PYTHON_BINDINGS` we must call the Python interpreter. This create some issues when cross compiling.
- We never test this case in our CI, so we don't know if this option is working.

This PR remove this option to simplify the build.
In the case an user build Pinocchio Python binding with hpp-fcl C++ library but without hpp-fcl binding, the Python interpreter will raise an ImportError.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix compilation issue for Boost 1.85 ([#2255](https://github.com/stack-of-tasks/pinocchio/pull/2255))
 - Fix python bindings of `contactInverseDynamics` ([#2263](https://github.com/stack-of-tasks/pinocchio/pull/2263))
 - Deactivate `BUILD_WITH_LIBPYTHON` when building with PyPy ([#2274](https://github.com/stack-of-tasks/pinocchio/pull/2274))
+- Fix Python bindings cross building with `hpp-fcl` ([#2288](https://github.com/stack-of-tasks/pinocchio/pull/2288))
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


ManifoldFR:
This seems good to me, noting the issues we've had over at the conda-forge feedstock with cross-building: https://github.com/conda-forge/pinocchio-feedstock/pull/117/checks?check_run_id=26326030013.

jcarpent (approved):

nim65s (approved):
This has been on my todolist for years, thanks a lot !

jorisv:
> It would be perhaps better to check if the hpp-fcl Python bindings' CMake target exists, if we want to reintroduce a check like this in the future.

I don't think we want to reintroduce this in the future. The build issues in conda-forge show this option was broken. We already have too much options in Pinocchio, and it's really hard to test all the combination.

I also think this kind of issue should be managed by a package manager instead of CMake.

ManifoldFR:
> > It would be perhaps better to check if the hpp-fcl Python bindings' CMake target exists, if we want to reintroduce a check like this in the future.
> 
> I don't think we want to reintroduce this in the future. The build issues in conda-forge show this option was broken. We already have too much options in Pinocchio, and it's really hard to test all the combination.
> 
> I also think this kind of issue should be managed by a package manager instead of CMake.

Totally makes sense! I agree the combinatorics of all the options was too much.

Issue #2293, opened 2024-06-20
Merging models transformation issue


HarisCavkic:
Hello all,

## Issue description
I am using pinocchio to load urdf models, do collision checking and visualising with meshcat. Im currently facing following issue:
When I load ur5 robot urdf (ur5_robot.urdf from you example robot data) and I load robotiq 85 gripper model (https://github.com/Daniella1/urdf_files_dataset/blob/main/urdf_files/ros-industrial/xacro_generated/robotiq/robotiq_2f_85_gripper_visualization/urdf/robotiq_arg2f_85_model.urdf), everything works fine. 
The problem arises when I want to merge these two models using pinocchio.appendModel, concretely when there is transformation present. This means when I call the appendModel with identity matrix everything is ok, except the fact that the gripper is not correctly placed, but when I add transformation (x,y,z,r,p,y => np.array([0,0.085,0,0,0,1.5708]), that I then transform to pinocchio.SE3) and call pinocchio.appendModel part of the gripper does get correctly transformed, but other part  (robotiq_85_left_knuckle_link, robotiq_85_left_finger_link, robotiq_85_right_inner_knuckle_link, robotiq_85_right_finger_tip_link, robotiq_85_right_knuckle_link and robotiq_85_left_finger_link) stays in initial position, i.e., the same as when identity matrix is the transformation matrix.

For context this is in the SE3 object:
Rotation R:  [[-3.6732051e-06 -1.0000000e+00  0.0000000e+00]
 [ 1.0000000e+00 -3.6732051e-06  0.0000000e+00]
 [ 0.0000000e+00  0.0000000e+00  1.0000000e+00]]
Translation t:  [0.    0.085 0.   ]

## Expected behavior
Whole gripper is transformed properly.

## Reproduction steps
```
def pose_to_transformation_matrix(pose, rot_type="rad"):
    x, y, z, roll, pitch, yaw = pose
    if rot_type == "rad":
        # Create transformation matrix
        transformation_mat = np.eye(4)
        transformation_mat[:3, -1] = [x, y, z]
        # Create rotation matrix
        R = pin.rpy.rpyToMatrix(roll, pitch, yaw)
        transformation_mat[:3, :3] = R
        return transformation_mat
    else:
        raise ValueError("Only radians are supported for now.")


def merge_models(robot_model, robot_geom_model, robot_visual_model, gripper_model, gripper_collision_model,
                 gripper_visual_model, gripper_transform):
[...]

PR #2295, opened 2024-06-21
[Model] Fix appendModel for parallel joints at the base

MegMll:
Solves [#2293 ](https://github.com/stack-of-tasks/pinocchio/issues/2293)

When joints at the base are in parallel, the transformation between modelA and modelB was applied only the first joint of the list, whereas it should be applied on all joints whose parent is the base of modelB.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


jcarpent (approved):
Thanks a lot @MegMll for fixing this issue

jcarpent:
@MegMll Could you add simple unit tests to check this fix?

HarisCavkic:
Hello,

We are currently using pinocchio with pip(linux/Mac) and with conda-forge(Win). When can we expect updates for this fix?

Kind regards,
Haris

jorisv:
We will probably make a new Pinocchio release next week. The conda-forge package should be available one day after the release.

PR #2296, opened 2024-06-23
Different inertia parametrizations and Physical consistency

simeon-ned:
### Pseudo Inertia and Log Cholesky Parametrization

This PR introduces the capability to use pseudo inertia and Log Cholesky parametrization within Pinocchio. These additions are based on recent advancements in ensuring the physical consistency of inertial parameters and provide a novel approach to parameterize the pseudo inertia matrix. The new functionalities include conversions to/from pseudo inertia, Log Cholesky parametrization, and related handy functions such as the Jacobian of Log Cholesky to dynamic parameters. [Here you may find some short related notes on these parametrizations](https://colab.research.google.com/drive/1xFte2FT0nQ0ePs02BoOx4CmLLw5U-OUZ#scrollTo=YNd0xZv8Gv8y).

### Key Contributions
1. **Pseudo Inertia Parametrization:**
   - Implementation of the pseudo inertia matrix as described in the paper [Linear Matrix Inequalities for Physically-Consistent Inertial Parameter Identification: A Statistical Perspective on the Mass Distribution](https://arxiv.org/abs/1701.04395).
   - Conversion functions to/from pseudo inertia.
   - Ensuring physical consistency constraints via positive definiteness.

2. **Log Cholesky Parametrization:**
   - Incorporation of the Log Cholesky decomposition method as detailed in [Smooth Parameterization of Rigid-Body Inertia](https://ieeexplore.ieee.org/document/9690029).
   - Functions for the Log Cholesky decomposition and reparameterization of the pseudo inertia matrix.
   - Jacobian computation for Log Cholesky to dynamic parameters conversion, which may be useful for adaptive control.

This PR also partially close the problem of checking physical consistency mentioned in [#2204](https://github.com/stack-of-tasks/pinocchio/pull/2204#discussion_r1552596106)


jorisv:
@jcarpent Do you think is a good idea to create dedicated type to store pseudo inertia and log cholesky parametrization ?

- Provide dedicated constructor to easily create this matrices
- Provide getter to matrix parameter (mass, h, sigma, H_bar)
- Store related method (i.e. calculateLogCholeskyJacobian)
- Avoid to call FromPseudoInertia or FromLogCholeskyParameters with a raw matrix (and then, making mistake)

jcarpent:
Thanks @simeon-ned for this PR and thanks @jorisv for making a first feedback.
I will have time to dive into this PR next week.

simeon-ned:
@jcarpent Sure, looking forward to your reply! Tag me anytime, and I'll be glad to assist. Just a note, I'm not very proficient in C++, but will do my best :)

unittest/python/bindings_inertia.py
```diff
@@ -111,6 +111,40 @@ def test_dynamic_parameters(self):
...
+                [0, 0, 0, 1],
+            ]
+        )
+        pseudo_chol = U @ U.T
+
+        inertia = pin.Inertia.FromLogCholeskyParameters(eta)
+        pseudo = inertia.toPseudoInertia()
+        self.assertTrue(np.all(np.linalg.eigvals(pseudo) > 0))
+        self.assertApprox(pseudo, pseudo_chol)
+
```
jcarpent, line 147: @simeon-ned Could add this test in C++ too? Indeed, the Python bindings are not necessarily.

jcarpent:
To move beyond, I do agree with @jorisv suggestions.
Having dedicated parametrizations (like Eigen::Quaternion) will help us to move forward for properly differentiating along different parametrization of Inertias.
I let @jorisv provides further guidance.

> @jcarpent Do you think is a good idea to create dedicated type to store pseudo inertia and log cholesky parametrization ?
> 
>     * Provide dedicated constructor to easily create this matrices
> 
>     * Provide getter to matrix parameter (mass, h, sigma, H_bar)
> 
>     * Store related method (i.e. calculateLogCholeskyJacobian)
> 
>     * Avoid to call FromPseudoInertia or FromLogCholeskyParameters with a raw matrix (and then, making mistake)

jcarpent (commented):
Dear @simeon-ned,

Your PR provides a great enhancement to Pinocchio thanks a lot.
It would be nice to add in the doc some references to the papers you have mentioned.

Additionally, to enable proper differentiation, we might think adding dedicated types for parametrizing inertias, in such a way we can easily extract and compute Jacobians, and enable conversions between parametrization externally to the main InertiaTpl class.

Thanks again for contributing to Pinocchio, this definitely enriches the library.
Don't forget to add your name in the contributor list ;)

simeon-ned:
@jcarpent, sure, I will add more information in the docstrings, include myself in the contribution list, and work on implementing @jorisv's proposition regarding separate classes for log-Cholesky and pseudo inertia.

Meanwhile, I would like to discuss the conversion to log-Cholesky parameters from dynamic and pseudo inertia. I have drafted a Python implementation for this conversion, as shown below:

```python
def pseudo2cholesky(pseudo_inertia: np.ndarray) -> np.ndarray:
    n = pseudo_inertia.shape[0]
    indices = np.arange(n - 1, -1, -1)  # Indices to reverse the order

    # Apply the inversion using indices for rows and columns
    reversed_inertia = pseudo_inertia[indices][:, indices]

    # Perform Cholesky decomposition on the permuted matrix A'
    L_prime = np.linalg.cholesky(reversed_inertia)

    # Apply the reverse permutation to L_prime and transpose it to form U
    U = L_prime[indices][:, indices]

    return U
```

However, I am a bit confused about how to achieve the equivalent reversal of order in `reversed_inertia` and `U` using C++ and Eigen, similar to the approach in Python with NumPy. The challenge arises because the log-Cholesky parameterization requires the decomposition to be in the form \( U U^T \) instead of \( U^T U \), where \( U \) is an upper triangular matrix. This form ensures that the resulting log-Cholesky parameters are interpretable and physically coupled with dynamic parameters (such as mass lever, etc.).

I certainly can achieve this with multiplication by permutation matrices, but this may not be the most efficient approach. If you have any tips or suggestions on how to handle this in C++ using Eigen in a more efficient manner, it would be greatly appreciated.

unittest/python/bindings_inertia.py
simeon-ned: Will do!

simeon-ned: Done, the test are present in `unittest/spatial.cpp`

simeon-ned:
@jcarpent, I have reworked the code to be better aligned with @jorisv's suggestions.

For now, there are two new structures in inertia.hpp, namely PseudoInertiaTpl and LogCholeskyParametersTpl. These structures store all related methods for conversions between different inertia parametrizations, as well as constructors and getters.

InertiaTpl itself contains methods that allow building Inertia from PseudoInertia or LogCholeskyParameters and converting back to them.

I have also added a dedicated cpp test that performs conversions back and forth between different inertia parametrizations, checks physical consistency, verifies that the Jacobian of the log Cholesky parametrization is not singular, and prints the introduced implementation to the console.

On top of that, I have added some references to papers in the docstrings for the introduced types.

The only remaining issue is the Python bindings. I am not very familiar with Boost.Python and will try to implement it, but I may need some help if you would be so kind.

Please let me know if anything else is needed or if there are further improvements that can be made.

include/pinocchio/spatial/inertia.hpp
```diff
@@ -889,6 +960,366 @@ namespace pinocchio
...
+  {
+    typedef Eigen::Matrix<Scalar, 4, 4, Options> Matrix4;
+    typedef Eigen::Matrix<Scalar, 3, 1, Options> Vector3;
+    typedef Eigen::Matrix<Scalar, 3, 3, Options> Matrix3;
+    typedef Eigen::Matrix<Scalar, 10, 1, Options> Vector10;
+    typedef LogCholeskyParametersTpl<Scalar, Options> LogCholeskyParameters;
+
+    Scalar mass;    ///< Mass of the pseudo inertia
+    Vector3 h;      ///< Vector part of the pseudo inertia
+    Matrix3 Sigma;  ///< 3x3 matrix part of the pseudo inertia
```
jorisv, line 982:
```suggestion
    Matrix3 sigma;  ///< 3x3 matrix part of the pseudo inertia
```

member should be in camelCase.

include/pinocchio/spatial/inertia.hpp
```diff
@@ -889,6 +960,366 @@ namespace pinocchio
...
+    Matrix4 matrix; ///< Full 4x4 pseudo inertia matrix
```
jorisv, line 983: matrix is redundant with mass, h and sigma.
This can lead to some error: if an user modify mass, matrix is no longer valid.

I see two options:
- Only storing matrix and provide getter to obtain mass, h and sigma
- Only store mass, h, sigma and don't store matrix

I prefer the second option: less code, and the matrix is only used in the `disp_impl` function.

include/pinocchio/spatial/inertia.hpp
```diff
@@ -889,6 +960,366 @@ namespace pinocchio
...
+
+    explicit PseudoInertiaTpl(const Vector10 & dynamic_params)
+    {
+      *this = FromDynamicParameters(dynamic_params);
+    }
+
+    explicit PseudoInertiaTpl(const LogCholeskyParameters & log_cholesky)
+    {
+      *this = log_cholesky.toPseudoInertia();
+    }
```
jorisv, line 1011: In think you can remove these constructors.

They are redundant with the static functions FromInertia, FromDynamicParameters, FromMatrix.

I prefer having on constructor allowing to initialize easily all the class parameters (like this one `PseudoInertiaTpl(Scalar mass, const Vector3 & h, const Matrix3 & Sigma)`) and static function as conversion/helper constructors.

This is more explicit and avoid overloading issues (what can we do if there is two way to create a PseudInertia from a Vector10 ?).

include/pinocchio/spatial/inertia.hpp
```diff
@@ -889,6 +960,366 @@ namespace pinocchio
...
+     * @return A PseudoInertiaTpl object.
+     */
+    static PseudoInertiaTpl FromDynamicParameters(const Vector10 & dynamic_params)
+    {
+      Scalar mass = dynamic_params[0];
+      Vector3 h = dynamic_params.template segment<3>(1);
+      Matrix3 I_bar;
+      I_bar << dynamic_params[4], dynamic_params[5], dynamic_params[7], dynamic_params[5],
+        dynamic_params[6], dynamic_params[8], dynamic_params[7], dynamic_params[8],
+        dynamic_params[9];
```
jorisv, line 1052:
```suggestion
      // clang-format off
      I_bar << dynamic_params[4], dynamic_params[5], dynamic_params[7],
               dynamic_params[5], dynamic_params[6], dynamic_params[8],
               dynamic_params[7], dynamic_params[8], dynamic_params[9];
      // clang-format on
```

include/pinocchio/spatial/inertia.hpp
```diff
@@ -889,6 +960,366 @@ namespace pinocchio
...
+      return InertiaTpl<Scalar, Options>::FromDynamicParameters(dynamic_params);
+    }
+
+    /**
+     * @brief Constructs a PseudoInertiaTpl object from log Cholesky parameters.
+     * @param log_cholesky A 10-dimensional vector of log Cholesky parameters.
+     * @return A PseudoInertiaTpl object.
+     */
+    static PseudoInertiaTpl
+    FromLogCholeskyParameters(const typename LogCholeskyParameters::Vector10 & log_cholesky)
```
jorisv, line 1106: Like InetriaTpl, prefer initializing directly from LogCholeskyParameters. It's safer and you do the conversion just two line bellow.

include/pinocchio/spatial/inertia.hpp
```diff
@@ -889,6 +960,366 @@ namespace pinocchio
...
+   * IEEE Robotics and Automation Letters 7.2 (2022): 2771-2778.
+   */
+  template<typename Scalar, int Options>
+  struct LogCholeskyParametersTpl
+  {
+    typedef Eigen::Matrix<Scalar, 10, 1, Options> Vector10;
+    typedef Eigen::Matrix<Scalar, 10, 10, Options> Matrix10;
+    typedef PseudoInertiaTpl<Scalar, Options> PseudoInertia;
+
+    Vector10 parameters; ///< 10-dimensional vector of log Cholesky parameters
```
jorisv, line 1144: Same remark than for PseudoInertiaTpl::matrix.

This is redundant and can lead to some issues.

include/pinocchio/spatial/inertia.hpp
```diff
@@ -889,6 +960,366 @@ namespace pinocchio
...
+      dynamic_params[2] = t2;
+      dynamic_params[3] = t3;
+      dynamic_params[4] = s23 * s23 + t2 * t2 + t3 * t3 + exp_d2 * exp_d2 + exp_d3 * exp_d3;
+      dynamic_params[5] = -s12 * exp_d2 - s13 * s23 - t1 * t2;
+      dynamic_params[6] =
+        s12 * s12 + s13 * s13 + t1 * t1 + t3 * t3 + exp_d1 * exp_d1 + exp_d3 * exp_d3;
+      dynamic_params[7] = -s13 * exp_d3 - t1 * t3;
+      dynamic_params[8] = -s23 * exp_d3 - t2 * t3;
+      dynamic_params[9] =
+        s12 * s12 + s13 * s13 + s23 * s23 + t1 * t1 + t2 * t2 + exp_d1 * exp_d1 + exp_d2 * exp_d2;
```
jorisv, line 1197: You can use math::pow improve the readability (you can also declare `using math::pow` at the beginning of this function to have a shorter function call).

You can also use the `// clang-format off` and `// clang-format on` comment to improve the formatting.

include/pinocchio/spatial/inertia.hpp
```diff
@@ -228,9 +229,16 @@ namespace pinocchio
...
     }
+  };
 
-  }; // class InertiaBase
+  // Forward declare the PseudoInertia and LogCholeskyParameters structs
+  template<typename Scalar, int Options>
+  struct PseudoInertiaTpl;
 
+  template<typename Scalar, int Options>
+  struct LogCholeskyParametersTpl;
```
jorisv, line 239: You can move this forward declaration in `pinocchio/spatial/fwd.hpp:60` and default initialize Options with `context::Options`.

unittest/spatial.cpp
```diff
@@ -737,6 +737,83 @@ BOOST_AUTO_TEST_CASE(test_Inertia)
...
+    // s12 = eta[4], s23 = eta[5], s13 = eta[6]; double t1 = eta[7], t2 = eta[8], t3 = eta[9];
+
+    double exp_alpha = std::exp(log_cholesky.alpha);
+    double exp_d1 = std::exp(log_cholesky.d1);
+    double exp_d2 = std::exp(log_cholesky.d2);
+    double exp_d3 = std::exp(log_cholesky.d3);
+
+    Eigen::Matrix4d U;
+    U << exp_d1, log_cholesky.s12, log_cholesky.s13, log_cholesky.t1, 0, exp_d2, log_cholesky.s23,
+      log_cholesky.t2, 0, 0, exp_d3, log_cholesky.t3, 0, 0, 0, 1;
```
jorisv, line 771:
```suggestion
    // clang-format off
    U << exp_d1, log_cholesky.s12, log_cholesky.s13, log_cholesky.t1,
         0, exp_d2, log_cholesky.s23, log_cholesky.t2,
         0, 0, exp_d3, log_cholesky.t3,
         0, 0, 0, 1;
    // clang-format on
```

jorisv (changes requested):
Hello @simeon-ned,

Thanks for your contribution.
I have reviewed some of your change.

- Don't hesitate to discuss the review
- If you lake of time and agree with the review I can apply it myself
- For the Boost.Python binding:
  - I can give you some guidance
  - I can do it myself if you're not interested to learn about Boost.Python

unittest/spatial.cpp
simeon-ned: Hi @jorisv,

Thank you for your detailed review. I appreciate your feedback and will address all the related comments over the weekend. If I encounter any difficulties, I will reach out for assistance.

Regarding the Boost.Python binding, I will do my best to implement it at the beginning of next week. Since this will be my first time working with Boost.Python, I might need some additional time and possibly your guidance. If I struggle with any aspects, I will let you know and may require your help.

simeon-ned: Hi @jorisv,

Thank you for your detailed review. I appreciate your feedback and will address all the related comments over the weekend. If I encounter any difficulties, I will reach out for assistance.

Regarding the Boost.Python binding, I will do my best to implement it at the beginning of next week. Since this will be my first time working with Boost.Python, I might need some additional time and possibly your guidance. If I struggle with any aspects, I will let you know and may require your help.

simeon-ned:
Hello @jorisv,

I've made progress on addressing the issues you mentioned in the C++ part of the code and updated the related unit tests. However, I'm still facing challenges with modifying the Python bindings. Here's what I've done so far:

1. Added new classes in `include/pinocchio/bindings/python/spatial`:
   - `LogCholeskyParametersPythonVisitor`
   - `PseudoInertiaPythonVisitor`

2. Added related methods in `InertiaPythonVisitor`

The new methods are now visible in `pin.Inertia`, but it seems that `PseudoInertia` and `LogCholesky` are not being exposed properly. 

Could you please provide some help with that? Thank you!

include/pinocchio/bindings/python/spatial/inertia.hpp
```diff
@@ -293,6 +310,260 @@ namespace pinocchio
...
+      static Matrix4 __array__(const PseudoInertia & self, bp::object)
+      {
+        return self.toMatrix();
+      }
+
+      struct Pickle : bp::pickle_suite
+      {
+        static boost::python::tuple getinitargs(const PseudoInertia & pi)
+        {
+          return bp::make_tuple(pi.toMatrix());
```
jorisv, line 448: getinitargs must return argument that will be provided to the constructor.

Since we have remove the constructor taking the full matrix, we will have to return mass, h and sigma.

include/pinocchio/bindings/python/spatial/inertia.hpp
```diff
@@ -293,6 +310,260 @@ namespace pinocchio
...
+      typedef typename LogCholeskyParameters::Scalar Scalar;
+      typedef typename LogCholeskyParameters::Vector10 Vector10;
+      typedef typename LogCholeskyParameters::Matrix10 Matrix10;
+
+    public:
+      template<class PyClass>
+      void visit(PyClass & cl) const
+      {
+        PINOCCHIO_COMPILER_DIAGNOSTIC_PUSH
+        PINOCCHIO_COMPILER_DIAGNOSTIC_IGNORED_SELF_ASSIGN_OVERLOADED
```
jorisv, line 476: Same than for PseudoInertia

include/pinocchio/bindings/python/spatial/inertia.hpp
```diff
@@ -293,6 +310,260 @@ namespace pinocchio
...
+        cl.def(
+            "__init__",
+            bp::make_constructor(
+              &LogCholeskyParametersPythonVisitor::makeFromVector, bp::default_call_policies(),
+              bp::args("log_cholesky_parameters")),
+            "Initialize from a 10-dimensional vector of log Cholesky parameters.")
```
jorisv, line 482: You don't need to create a custom constructor here. `LogCholeskyParametersTpl(const Vector10 & log_cholesky)` already exists.

include/pinocchio/bindings/python/spatial/inertia.hpp
```diff
@@ -293,6 +310,260 @@ namespace pinocchio
...
+        typedef ::boost::python::detail::not_specified HolderType;
+#endif
+        bp::class_<PseudoInertia, HolderType>(
+          "PseudoInertia",
+          "This class represents a pseudo inertia matrix and it is defined by its mass, vector "
+          "part, and 3x3 matrix part.\n\n"
+          "Supported operations ...",
+          bp::no_init)
+          .def(PseudoInertiaPythonVisitor<PseudoInertia>())
+          .def(CastVisitor<PseudoInertia>())
```
jorisv, line 433: If you want to add this visitor, you need to implement a cast method as in : https://github.com/stack-of-tasks/pinocchio/blob/master/include/pinocchio/spatial/inertia.hpp#L867-L874

include/pinocchio/bindings/python/spatial/inertia.hpp
```diff
@@ -293,6 +310,260 @@ namespace pinocchio
...
+#else
+        typedef ::boost::python::detail::not_specified HolderType;
+#endif
+        bp::class_<LogCholeskyParameters, HolderType>(
+          "LogCholeskyParameters",
+          "This class represents log Cholesky parameters.\n\n"
+          "Supported operations ...",
+          bp::no_init)
+          .def(LogCholeskyParametersPythonVisitor<LogCholeskyParameters>())
+          .def(CastVisitor<LogCholeskyParameters>())
```
jorisv, line 541: Same than for PseudoInertia

include/pinocchio/spatial/inertia.hpp
```diff
@@ -889,6 +931,348 @@ namespace pinocchio
...
+    dynamic_params[3] = t3;
+    dynamic_params[4] = pow(s23, 2) + pow(t2, 2) + pow(t3, 2) + pow(exp_d2, 2) + pow(exp_d3, 2);
+    dynamic_params[5] = -s12 * exp_d2 - s13 * s23 - t1 * t2;
+    dynamic_params[6] = pow(s12, 2) + pow(s13, 2) + pow(t1, 2) + pow(t3, 2) + pow(exp_d1, 2) + pow(exp_d3, 2);
+    dynamic_params[7] = -s13 * exp_d3 - t1 * t3;
+    dynamic_params[8] = -s23 * exp_d3 - t2 * t3;
+    dynamic_params[9] = pow(s12, 2) + pow(s13, 2) + pow(s23, 2) + pow(t1, 2) + pow(t2, 2) + pow(exp_d1, 2) + pow(exp_d2, 2);
+
+    const Scalar exp_2_alpha = exp(2 * alpha);
+    dynamic_params *= exp_2_alpha;
```
jorisv, line 1139: Add two space indent.

include/pinocchio/spatial/inertia.hpp
```diff
@@ -889,6 +931,348 @@ namespace pinocchio
...
+    jacobian(8, 9) = -t2 * exp_2alpha;
+
+    jacobian(9, 0) = 2 * (pow(s12, 2) + pow(s13, 2) + pow(s23, 2) + pow(t1, 2) + pow(t2, 2) + exp_2d1 + exp_2d2) * exp_2alpha;
+    jacobian(9, 1) = 2 * exp_2alpha * exp_2d1;
+    jacobian(9, 2) = 2 * exp_2alpha * exp_2d2;
+    jacobian(9, 4) = 2 * s12 * exp_2alpha;
+    jacobian(9, 5) = 2 * s23 * exp_2alpha;
+    jacobian(9, 6) = 2 * s13 * exp_2alpha;
+    jacobian(9, 7) = 2 * t1 * exp_2alpha;
+    jacobian(9, 8) = 2 * t2 * exp_2alpha;
```
jorisv, line 1249: Add two space indent

README.md
```diff
@@ -43,18 +43,23 @@ or via pip (currently only available on Linux):
...
+- [Performances](#performances)
+- [Ongoing developments](#ongoing-developments)
+- [Installation](#installation)
+  - [ROS](#ros)
+- [Visualization](#visualization)
+- [Citing Pinocchio](#citing-pinocchio)
+- [Questions and Issues](#questions-and-issues)
+- [Credits](#credits)
+- [Open-source projects relying on Pinocchio](#open-source-projects-relying-on-pinocchio)
+- [Acknowledgments](#acknowledgments)
```
jorisv, line 62: Use old indent

jorisv (changes requested):
Hello @simeon-ned,

You did the most complicated part.
Now you can edit this file: https://github.com/stack-of-tasks/pinocchio/blob/master/bindings/python/spatial/expose-inertia.cpp

and add
```cpp
      PseudoInertiaPythonVisitor<context::PseudoInertia>::expose();
      LogCholeskyParametersPythonVisitor<context::LogCholeskyParameters>::expose();
```

Don't forget to add PseudoInertia and LogCholeskyParameters typedef in `include/pinocchio/bindings/python/context/generic.hpp`.

You can also add binding test in `unittest/python/bindings_inertia.py`.

lvjonok:
Hello @jorisv,

I will take over this PR moving forward.

The CI seems to be running, and I have addressed most of your previous comments. However, I have just discovered that the bindings are broken, I will update it shortly.

I also have a few open questions that I hope you can help me with:

- There is a specific function `FromDynamicParameters`. For `InertiaTpl`, it is defined as a template over `Vector10`, but we haven't introduced it for `PseudoInertiaTpl` and `LogCholeskyTpl`. Should this be updated to maintain consistency across these classes?
- In a previous discussion, you mentioned:
  > If you want to add this visitor, you need to implement a cast method as in: https://github.com/stack-of-tasks/pinocchio/blob/master/include/pinocchio/spatial/inertia.hpp#L867-L874
  - Could you elaborate on why we might need this visitor and in what cases it would be necessary?

Looking forward to your feedback!

Best regards,
Lev Kozlov

lvjonok:
Just a quick update. I have managed to fix the python bindings and it works well in testing. While doing that I have partially answered the first question about template for `VectorLike10`.

From my understanding, it helps to cover these casts when dealing with python bindings and eigenpy.

Other than that, I think I will just wait for the feedback from now. I believe I have completed most of the checkpoints for this PR to be ready :)

simeon-ned:
@lvjonok Thank you so much for your help with this pull request! It's been invaluable.

@jorisv @jcarpent When you have a moment, could you please take a look at our PR? I believe we've addressed most of the issues you mentioned.

include/pinocchio/bindings/python/spatial/inertia.hpp
```diff
@@ -295,6 +312,251 @@ namespace pinocchio
...
+            "Matrix part of the Pseudo Inertia.")
+
+          .def(
+            "toMatrix", &PseudoInertia::toMatrix, bp::arg("self"),
+            "Returns the pseudo inertia as a 4x4 matrix.")
+          .def(
+            "toDynamicParameters", &PseudoInertiaPythonVisitor::toDynamicParameters_proxy,
+            bp::arg("self"), "Returns the dynamic parameters representation.")
+          .def(
+            "FromDynamicParameters", &PseudoInertia::template FromDynamicParameters<VectorXs>,
```
jorisv, line 357: This code can crash.

There is no runtime check to verify that the input matrix is Vector10.
InertiaTpl suffer from the same issue and I was able to crash the interpreter with the following code

```python
pin.Inertia.FromDynamicParameters(np.array([]))
```

include/pinocchio/bindings/python/spatial/inertia.hpp
```diff
@@ -295,6 +312,251 @@ namespace pinocchio
...
+            "Returns the Pseudo Inertia representation.")
+          .def(
+            "toInertia", &LogCholeskyParameters::toInertia, bp::arg("self"),
+            "Returns the Inertia representation.")
+          .def(
+            "calculateJacobian", &LogCholeskyParametersPythonVisitor::calculateJacobian_proxy,
+            bp::arg("self"), "Calculates the Jacobian of the log Cholesky parameters.")
+
+          .def("__array__", &LogCholeskyParameters::toDynamicParameters)
+          .def("__array__", &__array__)
```
jorisv, line 494: Array definition have changed to support numpy 2 https://github.com/stack-of-tasks/pinocchio/blob/devel/include/pinocchio/bindings/python/spatial/se3.hpp#L200-L202

include/pinocchio/bindings/python/spatial/inertia.hpp
```diff
@@ -295,6 +312,251 @@ namespace pinocchio
...
+#endif
+        bp::class_<LogCholeskyParameters, HolderType>(
+          "LogCholeskyParameters",
+          "This class represents log Cholesky parameters.\n\n"
+          "Supported operations ...",
+          bp::no_init)
+          .def(LogCholeskyParametersPythonVisitor<LogCholeskyParameters>())
+          .def(CopyableVisitor<LogCholeskyParameters>())
+          .def(PrintableVisitor<LogCholeskyParameters>());
+        // .def(CastVisitor<LogCholeskyParameters>())
```
jorisv, line 536: __array__ should have the following signature static Matrix4 __array__(const PseudoInertia & self, bp::object, bp::object)

include/pinocchio/bindings/python/spatial/inertia.hpp
```diff
@@ -295,6 +312,251 @@ namespace pinocchio
...
+      typedef typename LogCholeskyParameters::Vector10 Vector10;
+      typedef typename LogCholeskyParameters::Matrix10 Matrix10;
+      typedef Eigen::Matrix<Scalar, Eigen::Dynamic, 1, Options> VectorXs;
+      typedef Eigen::Matrix<Scalar, Eigen::Dynamic, Eigen::Dynamic, Options> MatrixXs;
+
+    public:
+      template<class PyClass>
+      void visit(PyClass & cl) const
+      {
+        cl.def(bp::init<const VectorXs &>((bp::arg("self"), bp::arg("log_cholesky_parameters"))))
```
jorisv, line 472: There is no runtime check of the vector input size. This can make the interpreter crash.

jorisv (changes requested):

jorisv:
Hello @simeon-ned and @lvjonok,

> There is a specific function FromDynamicParameters. For InertiaTpl, it is defined as a template over Vector10, but we haven't introduced it for PseudoInertiaTpl and LogCholeskyTpl. Should this be updated to maintain consistency across these classes?

I don't think it's mandatory to have consistency between InertiaTpl, PseudoInertiaTpl and LogCholeskyTpl. Since we don't plan to be able to replace one of these type by another, we don't really need to have a common set of methods between them.
If having a `FromDynamicParameters` implementation is useful for you, you can implement it. Otherwise, we don't need it.

> In a previous discussion, you mentioned:
>    If you want to add this visitor, you need to implement a cast method as in: https://github.com/stack-of-tasks/pinocchio/blob/master/include/pinocchio/spatial/inertia.hpp#L867-L874
>    Could you elaborate on why we might need this visitor and in what cases it would be necessary?

The cast visitor allow to cast to a different scalar type. It's useful if you want to convert an InertiaTpl made of double to an InertiaTpl made of casadi/cppad/cppadcg/boost.mpf scalar. I think is still a good idea to support it.

jorisv:
I will work on the **crash** issue since it also affect InertiaTpl.

If you want I can also address myself the last issues since it's really minor.

lvjonok:
> I will work on the **crash** issue since it also affect InertiaTpl.
> 
> If you want I can also address myself the last issues since it's really minor.

Yes, sure. Thanks for explanations, it really matters

include/pinocchio/bindings/python/spatial/inertia.hpp
```diff
@@ -295,6 +312,251 @@ namespace pinocchio
...
+      static Matrix3 getSigma(const PseudoInertia & self)
+      {
+        return self.sigma;
+      }
+      static void setSigma(PseudoInertia & self, const Matrix3 & sigma)
+      {
+        self.sigma = sigma;
+      }
+
+      static VectorXs toDynamicParameters_proxy(const PseudoInertia & self)
```
jorisv, line 410: Don't use VectorXs if we choose to bind Vector10

jorisv: We will not bind Vector10, so it's fine.

jorisv (dismissed):

jorisv (approved):

jorisv:
All green we can merge.

@simeon-ned, @lvjonok once again, thank you for your contribution !

Issue #2321, opened 2024-07-10
Minimal cmake example to compile a C++ program from outside for Pinocchio 3 / undefined reference to parseRootTree


Cfather:
## Bug description

Main error when I use cmake to compile a program with pinocchio. The following function can not be found:
```
undefined reference to `pinocchio::urdf::details::parseRootTree(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, pinocchio::urdf::details::UrdfVisitorBaseTpl<double, 0>&)
```
The exact same CMakeLists used to work for me a few months ago, until recently I rebuilt my container and found robotpkg automatically installed Pinocchio 3 and the compilation did not work anymore.

I am currently working on a deadline. 
It would be greatly appreciated if anyone could provide a minimal CMakeLists.txt that solves this problem.

## Expected behavior

The test script should be successfully compiled.

## Reproduction steps

You can copy and paste the test script (test.cpp) and the CMakeLists.txt somewhere, then run the following commands:
1. `mkdir build`
2. `cd build`
3. `cmake ..` (this step is good to me)
4. `make` (this step raises error)

### Docker image

I won't provide my full dockerfile here but I was installing Pinocchio as advised in its [documentation page](https://stack-of-tasks.github.io/pinocchio/download.html) through robotpkg.

```
mkdir -p /etc/apt/keyrings
curl http://robotpkg.openrobots.org/packages/debian/robotpkg.asc \
    | tee /etc/apt/keyrings/robotpkg.asc
echo "deb [arch=amd64 signed-by=/etc/apt/keyrings/robotpkg.asc] http://robotpkg.openrobots.org/packages/debian/pub $(lsb_release -cs) robotpkg" \
    | tee /etc/apt/sources.list.d/robotpkg.list
apt-get update
apt install -qqy robotpkg-py3*-pinocchio

export PATH=/opt/openrobots/bin:$PATH
export PKG_CONFIG_PATH=/opt/openrobots/lib/pkgconfig:$PKG_CONFIG_PATH
export LD_LIBRARY_PATH=/opt/openrobots/lib:$LD_LIBRARY_PATH
export PYTHONPATH=/opt/openrobots/lib/python3.10/site-packages:$PYTHONPATH
export CMAKE_PREFIX_PATH=/opt/openrobots:$CMAKE_PREFIX_PATH
```

### Code

The script and the CMakeLists.txt are attached here to reproduce the issue:

[...]

PR #2322, opened 2024-07-11
Add missing dependencies into pkg-config

jorisv:
This PR add the following libraries into the pkg-config file (when built with the right option) :

- pinocchio_parsers
- pinocchio_collision
- pinocchio_extra
- pinocchio_casadi
- pinocchio_cppad
- pinocchio_cppadcg

Add missing defines.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -145,12 +145,15 @@ jobs:
           mkdir build
           cd build
           export CMAKE_PREFIX_PATH=/usr/local
+          # Test CMake module packaging
           cmake .. \
             -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
             -DPYTHON_EXECUTABLE=$(which python3)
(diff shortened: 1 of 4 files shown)


jorisv:
Fix #2321

jorisv:
CI is broken because of https://github.com/conda-forge/qhull-feedstock/pull/19

nim65s (dismissed):
LGTM, but maybe we could add code from #2321 in CI ?
I guess if this was broken for so long, it shows that none of us use pkg-config.

jorisv:
@nim65s We can maybe modify `unittest/packaging/pkgconfig` to use more dependencies.

jorisv:
@nim65s I have added a test for pinocchio_parsers + pkgconfig.

It's not relevant to test cppad, casadi, cppadcg and collision because they only contains template instantiation.

jcarpent (approved):

Issue #2329, opened 2024-07-15
newer version meshcat visualization issue for mesh


MiaoDragon:
Hello. Thanks for the software effort of the Pinocchio. I've been experiencing a visualization issue in meshcat viewer for meshes. It seems for the new version 3.1 the mesh scale is not correctly displayed when I have a mixed mesh files of STL and DAE. One visualization is shown below for my motoman robot with robotiq gripper, and the gripper has a very large scale. This is the visual geometries.
![mesh-scale-version-3 1](https://github.com/user-attachments/assets/0dd491f5-3584-4d7e-9f76-9d9f5c107ce6)
The collision geometries are also problematic.
![meshcat-collision-3 1](https://github.com/user-attachments/assets/3a3c022b-9fc2-4b21-b9d1-30361f8431df)

The scale issue of visual geometrices do not exist for version 2.7, while the collision geometries still have the same issue. The visual geometries are shown as follows:
![meshcat-visual-2 7](https://github.com/user-attachments/assets/f3d97b7a-a9ae-47e1-a2e5-8c1688f0ea2b)

The visualized model can be extracted from
https://github.com/DanManN/motoman

After the model package is installed in a catkin workspace, and the workspace is made, use the following code to visualize the results:
(the following code is saved as a file motoman_config.py)
```python
import sys
import os
import rospy
from rospkg import RosPack

class MotomanSDA10F:
    def __init__(self):
        self.ignore_collision_ee_links = [
            "motoman_right_ee",
            "left_outer_knuckle",
            "left_outer_finger",
            "left_inner_finger",
            "left_inner_finger_pad",
            "left_inner_knuckle",
            "right_outer_knuckle",
            "right_outer_finger",
            "right_inner_finger",
            "right_inner_finger_pad",
            "right_inner_knuckle",
            "robotiq_arg2f_extra_link",
            "robotiq_arg2f_base_link",
            "arm_right_link_7_t",
            "arm_right_link_6_b",
        ]
        rp = RosPack()
        self.robot_urdf = rp.get_path(
            'motoman_sda10f_moveit_config'
        ) + '/config/gazebo_motoman_sda10f.urdf'
        package_dirs = []
        for package in ['motoman_sda10f_support', 'robotiq_2f_85_gripper_visualization']:
            path = os.path.join(rp.get_path(package), '..')
            path = os.path.abspath(path)
            package_dirs.append(path)
            # package_dirs.append(rp.get_path(package))
[...]

PR #2331, opened 2024-07-16
Fix Dae support in MeshCat

jcarpent:
Solves #2329

### DIFF
--- a/bindings/python/pinocchio/visualize/meshcat_visualizer.py
+++ b/bindings/python/pinocchio/visualize/meshcat_visualizer.py
@@ -110,6 +110,7 @@ if import_meshcat_succeed:
             # Attributes to be specified by the user
             self.path = None
             self.material = None
+            self.intrinsic_transform = mg.tf.identity_matrix()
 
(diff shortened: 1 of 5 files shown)


bindings/python/pinocchio/visualize/meshcat_visualizer.py
```diff
@@ -739,7 +739,7 @@ def loadMeshFromFile(self, geometry_object):
         # Get file type from filename extension.
         _, file_extension = os.path.splitext(geometry_object.meshPath)
         if file_extension.lower() == ".dae":
-            obj = DaeMeshGeometry(geometry_object.meshPath)
+            obj = mg.DaeMeshGeometry.from_file(geometry_object.meshPath)
```
jorisv, line 742: We should remove our implementation of DaeMeshGeometry.

Also the following code can be removed :

```python
# DaeMeshGeometry
import xml.etree.ElementTree as Et
import base64
```

jorisv (approved):

Issue #2251, opened 2024-05-29
Allow to choose build options to accelerate build in PR


jorisv:
When building Pinocchio 3 with all options activated CI can last up to 4h. We should find a way to optionally activate build options.

This can be done by using label on a pull request:

- build_with_all: All options
- build_with_collision: BUILD_WITH_COLLISION_SUPPORT
- build_with_urdf: BUILD_WITH_URDF_SUPPORT
- build_with_sdf: BUILD_WITH_SDF_SUPPORT
- build_with_autodiff: BUILD_WITH_AUTODIFF_SUPPORT
- build_with_casadi: BUILD_WITH_CASADI_SUPPORT
- build_with_codegen: BUILD_WITH_CODEGEN_SUPPORT
- build_with_mpfr: BUILD_PYTHON_BINDINGS_WITH_BOOST_MPFR_SUPPORT
- build_with_openmp: BUILD_WITH_OPENMP_SUPPORT
- build_with_extra: BUILD_WITH_EXTRA_SUPPORT
- build_with_accelerate: BUILD_WITH_ACCELERATE_SUPPORT

PR should be run with minimal options, and build_with_all should be activated before merging.

PR #2334, opened 2024-07-16
CI - Compilation options based on pull request labels

MegMll:
This pr aims at providing cmake compilations options based on the labels from pull requests. When no label are added, a minimal setup is provided. 

For more informations, please see https://github.com/stack-of-tasks/pinocchio/issues/2251

### DIFF
--- a/.github/workflows/scripts/get_label.js
+++ b/.github/workflows/scripts/get_label.js
@@ -8,8 +8,7 @@ module.exports = async ({github, context, core}) => {
   
     const prNumber = context.issue.number || getPullRequestNumber(context.ref);
 
-    let cmakeFlags = '-DCMAKE_CXX_COMPILER_LAUNCHER=ccache ';
-    console.log(cmakeFlags);
+    let cmakeFlags = '';
     if(isNaN(prNumber))
     {
         core.setOutput("cmakeFlags", cmakeFlags);
--- a/.github/workflows/scripts/get_label.js
+++ b/.github/workflows/scripts/get_label.js
@@ -1,0 +1,46 @@
+module.exports = async ({github, context, core}) => {
+    const getPullRequestNumber = (ref) => {
+        core.debug(`Parsing ref: ${ref}`);
+        // This assumes that the ref is in the form of `refs/pull/:prNumber/merge`
+        const prNumber = ref.replace(/refs\/pull\/(\d+)\/merge/, '$1');
+        return parseInt(prNumber, 10);
+    };
+  
+    const prNumber = context.issue.number || getPullRequestNumber(context.ref);
+
+    let cmakeFlags = '-DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Debug -DCMAKE_TOOLCHAIN_FILE=${GITHUB_WORKSPACE}/.github/workflows/cmake/linux-debug-toolchain.cmake -DPYTHON_EXECUTABLE=$(which python3)';
+    console.log(cmakeFlags);
+    if(isNaN(prNumber))
+    {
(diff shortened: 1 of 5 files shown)


jcarpent (commented):
Very nice action and idea. Thanks to @MegMll and @jorisv

.github/workflows/scripts/get_label.js
```diff
@@ -0,0 +1,59 @@
...
+            ' -DGENERATE_PYTHON_STUBS=ON'
+        ],
+        build_collision: ' -DBUILD_WITH_COLLISION_SUPPORT=ON',
+        build_casadi: ' -DBUILD_WITH_CASADI_SUPPORT=ON',
+        build_autodiff: ' -DBUILD_WITH_AUTODIFF_SUPPORT=ON',
+        build_codegen: ' -DBUILD_WITH_CODEGEN_SUPPORT=ON',
+        build_extra: ' -DBUILD_WITH_EXTRA_SUPPORT=ON',
+        build_openmp: ' -DBUILD_WITH_OPENMP_SUPPORT=ON',
+        build_mpfr: ' -DBUILD_PYTHON_BINDINGS_WITH_BOOST_MPFR_SUPPORT=ON',
+        build_sdf: ' -DBUILD_WITH_SDF_SUPPORT=ON'
```
jorisv, line 44: - sdf: To manage sdf we have to install gazebo dependency. But this dependency can create a lot of conflict.
- Do we add build_urdf ? In this case, we have to deactivate it by default.
- We can add build_accerelate option. CMakeLists.txt set it up to FALSE on other platform than Mac so we don't have to add a test in this script.

jorisv (changes requested):
Great work !

get_label.js allow to factorize a lot of code.

.github/workflows/scripts/get_label.js
jcarpent: We need gazebo for SDF?

MegMll: For SDF, i tried and in the ned we only need libsdformat, which can be installed via conda. It's quite light, so might be worth it. 
And i added the accelerate option

jorisv: My bad, I had some issue month ago to find sdf library on conda.

jorisv:
This PR don't require a changelog entry since it only modify the CI.

jorisv (approved):

.github/workflows/scripts/get_label.js
```diff
@@ -0,0 +1,107 @@
...
+        else
+        {       
+            const reviewMessage = `👋 Hi,
+            this is a reminder message to please assign a proper label to this Pull Request.
+            The possible labels are:
+            
+            - build_collision (build pinocchio with coal support)
+            - build_casadi (build pinoochio with casadi support)
+            - build_autodiff (build pinocchio with cppad support)
+            - build_codegen
```
jorisv, line 88:
```suggestion
            - build_codegen (build pinocchio with cppadcg support)
```

.github/workflows/scripts/get_label.js
```diff
@@ -0,0 +1,107 @@
...
+            - build_extra (build pinocchio with extra algorithms)
+            - build_mpfr
```
jorisv, line 90:
```suggestion
            - build_mpfr (build pinocchio with Boost.Multiprecision support)
```

.github/workflows/scripts/get_label.js
```diff
@@ -0,0 +1,107 @@
...
+            - build_sdf (build sdf parser)
```
jorisv, line 91:
```suggestion
            - build_sdf (build pinocchio with sdf parser)
```

jorisv (changes requested):

jorisv (approved):

PR #2339, opened 2024-07-18
urdf parser: use static casts when type is known

nim65s:
Hi,

Inside those `if (urdf_geometry->type == ::urdf::Geometry::…)` conditions, we can statically be sure about the real type of `urdf_geometry`. We should therefore prefer using `static_cast` over dynamic ones.

(Moreover, using dynamic casts would require to check for nullptr before using the result, which was forgotten here.)

Therefore, at runtime, if RTTI fails for whatever reason, everything related with urdf segfaults, because we try to access members (like filename or radius) of a nullptr. This is the case at least with nix, on aarch64-darwin, with clang 16.

I have no idea why, but `typeid(::urdf::Mesh) != typeid(*urdf_geometry)` (or even `typeid(*urdf_mesh)` after a static cast to `::urdf::Mesh`…), so dynamic cast fails. `type_info::name` is correct with `urdf::Mesh`, but `type_info::hash_code` is different

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - Append pinocchio optional libraries into pkg-config file ([#2322](https://github.com/stack-of-tasks/pinocchio/pull/2322))
 - Fixed support of DAE meshes with MeshCat ([#2331](https://github.com/stack-of-tasks/pinocchio/pull/2331))
+- Fixed pointer casts in urdf parser ([#2339](https://github.com/stack-of-tasks/pinocchio/pull/2339))
 
 ### Added
 - Add getMotionAxis method to helical, prismatic, revolute and ubounded revolute joint ([#2315](https://github.com/stack-of-tasks/pinocchio/pull/2315))


jorisv (approved):

jcarpent (approved):

florent-lamiraux:
It reminds me a problem we had in the stack of tasks with typeid. From what I remember, the same type compiled in two different .so libraries do not have the same typeid but I do not remember the details.

jorisv:
@nim65s Did you get this issue by building pinocchio and running urdf tests or when running third party code that use pinocchio urdf parser ?

nim65s:
In pinocchio unittests:
```
The following tests FAILED:
         41 - test-cpp-urdf (Failed)
         44 - test-cpp-geometry-model (Failed)
         45 - test-cpp-geometry-algorithms (Failed)
         46 - test-cpp-broadphase (Failed)
         47 - test-cpp-tree-broadphase (Failed)
         48 - test-cpp-srdf (Failed)
        131 - example-cpp-geometry-models (SEGFAULT)
        135 - example-cpp-collisions (SEGFAULT)
        142 - example-py-gepetto-viewer (SEGFAULT)
        143 - example-py-build-reduced-model (SEGFAULT)
        144 - example-py-robot-wrapper-viewer (SEGFAULT)
        145 - example-py-geometry-models (SEGFAULT)
        149 - example-py-meshcat-viewer (SEGFAULT)
        150 - example-py-meshcat-viewer-dae (SEGFAULT)
        151 - example-py-collisions (SEGFAULT)
        152 - example-py-collision-with-point-clouds (SEGFAULT)
        153 - example-py-append-urdf-model-with-another-model (SEGFAULT)
        154 - example-py-simulation-contact-dynamics (SEGFAULT)
        155 - example-py-static-contact-dynamics (SEGFAULT)

jorisv:
As say @florent-lamiraux this can hide some linking issues (by using two different version of URDF in different shared library).

But this will be really painful to investigate...

PR #2340, opened 2024-07-19
Add core headers to target library

jcarpent:
This is mainly necessary when use cmake generator features

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -80,7 +80,11 @@ function(PINOCCHIO_TARGET target_name)
         PARENT_SCOPE)
   endif()
 
-  add_library(${LIB_NAME} ${LIBRARY_TYPE} ${${PROJECT_NAME}_CORE_PUBLIC_HEADERS})
+  if(NOT "${LIBRARY_TYPE}" STREQUAL "INTERFACE")
+    add_library(${LIB_NAME} ${LIBRARY_TYPE} ${${PROJECT_NAME}_CORE_PUBLIC_HEADERS})
+  else()
+    add_library(${LIB_NAME} ${LIBRARY_TYPE})
+  endif()
   target_link_libraries(${LIB_NAME} ${LIBRARY_PUBLIC_SCOPE} ${PROJECT_NAME}_headers)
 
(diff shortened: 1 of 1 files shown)


jorisv:
As I remember, this create issue with CMake 3.16.

Cf this comment :

```cmake
# On CMake 3.16, we can't target_sources(${PROJECT_NAME}_headers INTERFACE
# ${${PROJECT_NAME}_CORE_PUBLIC_HEADERS})
```

Let's see what the linux ci is saying.

src/CMakeLists.txt
```diff
@@ -80,7 +80,11 @@ function(PINOCCHIO_TARGET target_name)
         PARENT_SCOPE)
   endif()
 
-  add_library(${LIB_NAME} ${LIBRARY_TYPE})
+  if(NOT "${LIBRARY_TYPE}" STREQUAL "INTERFACE")
+    add_library(${LIB_NAME} ${LIBRARY_TYPE} ${${PROJECT_NAME}_CORE_PUBLIC_HEADERS})
+  else()
+    add_library(${LIB_NAME} ${LIBRARY_TYPE})
+  endif()
   target_link_libraries(${LIB_NAME} ${LIBRARY_PUBLIC_SCOPE} ${PROJECT_NAME}_headers)
```
jorisv, line 88: You can maybe move the target_sources command call bellow:

```cpp
  if(NOT ARGS_INTERFACE)
    target_sources(${LIB_NAME} PRIVATE ${ARGS_SOURCES})
    set_target_properties(
      ${LIB_NAME}
      PROPERTIES LINKER_LANGUAGE CXX
                 INSTALL_RPATH "\$ORIGIN"
                 VERSION ${PROJECT_VERSION})
  endif()
  ```

jcarpent: Super. Thanks for the suggestion.

jorisv: The idea was more to use the target_source command instead of moving the add_library command.
Also, the comment is still relevant. After CMake 3.16, it's possible to attach headers to an INTERFACE library.

jorisv (changes requested):
I have reopen the previous discussion

jcarpent:
> I have reopen the previous discussion

Should be solved now.

PR #2341, opened 2024-07-19
Fix urdfdom dependency

marcbone:
The package.xml relies on liburdfdom-dev, which comes from Ubuntu instead of the one that comes from ROS 2. This usually leads to a situation where you have urdfdom installed twice on your system. Once from this package as liburdfdom-dev that is installed in `/usr/lib/x86_64-linux-gnu/` and once from other ROS 2  packages on your system that use the urdfdom from ROS 2 (like the urdf package), which is installed in `/opt/ros/{ROS_DISTRO}/lib/x86_64-linux-gnu`. This will now lead to [CMake warnings in other ROS 2 project that try to use urdfdom](https://github.com/Gauss-Robotics/xbot2_interface/pull/1#issuecomment-2238603609). It still works, but it might break in the future.

 This PR changes the package.xml to rely on the urdfdom package that comes from ROS.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -21,7 +21,7 @@
   <exec_depend condition="$ROS_VERSION == 1">catkin</exec_depend>
   <depend>python3</depend>
   <depend>python3-numpy</depend>
-  <depend>liburdfdom-dev</depend>
+  <depend>urdfdom</depend>
   <depend>eigen</depend>
   <depend>boost</depend>
   <depend>eigenpy</depend>
--- a/package.xml
+++ b/package.xml
@@ -21,7 +21,8 @@
   <exec_depend condition="$ROS_VERSION == 1">catkin</exec_depend>
   <depend>python3</depend>
   <depend>python3-numpy</depend>
-  <depend>urdfdom</depend>
+  <depend condition="$ROS_VERSION == 2">urdfdom</depend>
+  <depend condition="$ROS_VERSION == 1">liburdfdom-dev</depend>
   <depend>eigen</depend>
   <depend>boost</depend>
   <depend>eigenpy</depend>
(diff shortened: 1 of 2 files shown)


jcarpent (approved):

jcarpent (commented):
@nim65s @wxmerkt Could you provide some feedback on this PR?

nim65s:
I don't understand why ROS would distribute a package which is available upstream. I also don't understand how this change could fix the situation. From what I understand, in cases you have urdfdom both in /usr and /opt/ros, this warning will always be there, right ? To me, the issue is that you have twice the same package installed from different sources.

Also, it seems easier to remove /opt/ros from your configure/build paths than removing /usr: this is not even possible on ubuntu I think.

wxmerkt:
The background to this is that in some cases upstream packages were seriously out of date, had bugs, or incompatible versions were on different versions used by a ROS distro, so packages got vendored in. I believe liburdfdom had the challenge in the past that the system-/upstream-version was quite behind. Some digging shows that this may have been related to issues with locale settings.

The reason why urdfdom is vendored into ROS2 is due to an API break by removing the TinyXML compatibility and upgrading to urdfdom 4 in Rolling and newer, cf. https://discourse.ros.org/t/upcoming-api-break-in-urdfdom/34750

For this PR, we need changes however using conditional tags:
- Noetic doesnt have `urdfdom`
- Jazzy is the first one to have `urdfdom` released at 4.x
- Iron has `urdfdom` at 3.1.x
- Humble has `urdfdom` at 3.0.x

As such, let's make the switch of the tag conditional on `ROS_VERSION` - if ROS_VERSION = 1, we keep the existing, and if ROS_VERSION = 2, you can switch.

package.xml
```diff
@@ -21,7 +21,7 @@
   <exec_depend condition="$ROS_VERSION == 1">catkin</exec_depend>
   <depend>python3</depend>
   <depend>python3-numpy</depend>
-  <depend>liburdfdom-dev</depend>
+  <depend>urdfdom</depend>
```
wxmerkt, line 24:
```suggestion
  <depend condition="$ROS_VERSION == 1">urdfdom</depend>
  <depend condition="$ROS_VERSION == 2">liburdfdom-dev</depend>
```

PR #2350, opened 2024-07-25
Separate collision_color in MeshcatVisualizer.loadViewerModel

stephane-caron:
The proposal of this PR is to add a separate ``collision_color`` argument to `MeshcatVisualizer.loadViewerModel`.

- **Before:** `color` applies to both visual and collision model
- **After:** `color` applies to visual, `collision_color` to collision model

This can be convenient to debug collision meshes alongside the visual model (with only `color`, meshes overlapped):

```py
robot.setVisualizer(MeshcatVisualizer())
robot.initViewer(open=True)
robot.loadViewerModel(collision_color=(1.0, 0.0, 0.0, 0.3))
robot.viz.displayCollisions(True)
```

image


jcarpent:
@stephane-caron could you rebase instead of merging ?

stephane-caron:
> @stephane-caron could you rebase instead of merging ?

Sure. GitHub gave me a button to resolve the conflict and I naively clicked on it :blush:

jcarpent (changes requested):
I would go for `visual_color` too. And when color is there, it should be used for both, or mark it as deprecated option.

jcarpent (approved):

PR #2351, opened 2024-07-25
doc: Fix typos and errors.

abussy-aldebaran:
I corrected typos and errors I found reading the docs : https://gepettoweb.laas.fr/doc/stack-of-tasks/pinocchio/master/doxygen-html/

Here are the less trivial ones, i.e. the ones I'm not totally sure about:

- in the intro, there is `where are the arguments...`. I added `arg1, arg2, ...`, and didn't include `model` and `data`.
- a SO(3) was used instead of SO(2).
- `\\( SO(3) \\)` is used in a title and render as is in chrome. I simply replaced with `SO(3)`
- I assume that "blocking of Cardan" refers to Gimbal Lock ?
- `the whole of all velocity vectors` means `the set of all velocity vectors` ?
- I think that a zero line is missing in the constraint matrix of the cylindrical joint.

PR #2353, opened 2024-07-26
doc - Correct ??? placeholder.

abussy-aldebaran:
These ??? where left out in #2351.


abussy-aldebaran:
In #2351, I mixed up and forgot one modification.

jcarpent (approved):
Thanks for the fix not cached before.

jcarpent (approved):

PR #2367, opened 2024-08-12
modernize python code base

nim65s:
Hi,

This update python linters and follow its suggestions. Some  changes are automatic, others are manual: ref. individual commits.

No functional changes should come from here, except in the practicals. The python files there were still using python 2 syntax, so I made some upgrades.


jcarpent:
I would like to squash the commits to simplify the history. Could squash your commits @nim65s and then update the git-commit-rev accordingly?

nim65s:
done :)

jcarpent:
Super, thanks a lot @nim65s for the quick change.

Issue #2357, opened 2024-07-31
A Few Small Problems In Example Scripts


Cfather:
Hi, I would like to report a few (really) small bugs in the example scripts.

1. [simulation-closed-kinematic-chains.py](https://github.com/stack-of-tasks/pinocchio/blob/master/examples/simulation-closed-kinematic-chains.py):
line 161 
```
constraint_model.corrector.Kd = 2.0 * np.sqrt(constraint_model.corrector.Kp)
```
doesn't work for me. I got the following error:
```
RuntimeError: The number of elements does not fit with the vector type.
```
I changed it to
```
constraint_model.corrector.Kd[:] = 2.0 * np.sqrt(constraint_model.corrector.Kp)
```
and it worked. not sure why I can't directly assign a numpy vector to it.

2. [simulation-contact-dynamics.py](https://github.com/stack-of-tasks/pinocchio/blob/master/examples/simulation-contact-dynamics.py):
This is really small... Line 16-18 seems to call
```
pinocchio_model_dir = join(dirname(dirname(str(abspath(__file__)))), "models")
```
one more time. It might be unnecessary.

3. [talos-simulation.py](https://github.com/stack-of-tasks/pinocchio/blob/master/examples/talos-simulation.py):
Line 4:
```
from example_robot_data import loadTalos
```
seems to belong to an older version of example_robot_data? At least I installed example_robot_data through conda a few weeks ago and `loadTalos` is not there anymore.
I changed it to:
```
import example_robot_data
```
and line 11 to:
```
robot = example_robot_data.load("talos")
```
and the code works.
But the visualization seems to break. When I run `gepetto-gui`, there's nothing showing up in the window (just the GUI window, I went to 127.0.0.1 and there was nothing). There was following error from `gepetto-gui`:
```
libpng warning: iCCP: known incorrect sRGB profile
libpng warning: iCCP: known incorrect sRGB profile
libpng warning: iCCP: known incorrect sRGB profile
libpng warning: iCCP: known incorrect sRGB profile
Error reading file /home/yuzhen/miniconda3/share/gepetto-viewer/fonts/arial.ttfQOpenGLContext::swapBuffers() called with non-exposed window, behavior is undefined
```
The simulation works fine but I can not see the visualization. not sure if this is just problem on my side or if I used `gepetto-gui` incorrectly.

- OS: [Ubuntu 22.04]
- Pinocchio version: [3.0.0]

Sang-SC:
[...]

PR #2392, opened 2024-08-27
Fix a few examples scripts

MegMll:
Fix the problems that were shown by https://github.com/stack-of-tasks/pinocchio/issues/2357.

Also remove a compilation warning coming from mjcf unittest.


jcarpent (approved):
Thanks a lot @MegMll for the nice fix.

PR #2403, opened 2024-09-04
Parser/mjcf : Add warning and fix root joint of mjcf models

MegMll:
This pr aims at being a bit more flexible concerning textures in mjcf model that are not supported in pinocchio.

It also fix a small bug that was introduced in https://github.com/stack-of-tasks/pinocchio/pull/2386.

### DIFF
--- a/src/parsers/mjcf/mjcf-graph.cpp
+++ b/src/parsers/mjcf/mjcf-graph.cpp
@@ -500,15 +500,29 @@ namespace pinocchio
         namespace fs = boost::filesystem;
         MjcfTexture text;
         auto file = el.get_optional<std::string>("<xmlattr>.file");
+        auto name_ = el.get_optional<std::string>("<xmlattr>.name");
+        auto type = el.get_optional<std::string>("<xmlattr>.type");
+
+        std::string name;
+        if (name_)
+          name = *name_;
+        else if (*type == "skybox")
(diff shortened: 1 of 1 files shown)


jcarpent (changes requested):
@MegMll Thanks for providing this fix.
Could you add a test to be sure that the bug will be catched in the future?

jcarpent (approved):
Super, thanks for adding this test.

src/parsers/mjcf/mjcf-graph.cpp
```diff
@@ -500,15 +500,29 @@ namespace pinocchio
...
-        std::string name = getName(el, filePath);
+        auto name_ = el.get_optional<std::string>("<xmlattr>.name");
+        auto type = el.get_optional<std::string>("<xmlattr>.type");
 
-        text.filePath =
-          updatePath(compilerInfo.strippath, compilerInfo.texturedir, modelPath, filePath).string();
+        std::string name;
+        if (name_)
+          name = *name_;
+        else if (*type == "skybox")
```
jcarpent, line 509:
```suggestion
        else if (type && *type == "skybox")
```

jcarpent: Should not be this solution (in case type is empty)?

PR #2413, opened 2024-09-11
Feature: Support for Parsing MJCF `equality/connect` Tag for Closed Chains

lvjonok:
Hello Pinocchio Team,

While working with the `buildModel` function, I noticed that when an SDF description is passed, it directly produces a vector of `RigidConstraintModel`s. However, the `mjcf` format, which also supports closed-chain descriptions, currently does not parse the `equality/connect` tag.

To address this, I have added a draft implementation to enable parsing of the `equality/connect` tag in the `mjcf` format. I believe this enhancement will ease the work for people who use it.

Please feel free to review the draft implementation and share your feedback on its applicability to the Pinocchio library. I look forward to your comments!

Best regards,
Lev Kozlov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Default visualizer can be changed with `PINOCCHIO_VIEWER` environment variable ([#2419](https://github.com/stack-of-tasks/pinocchio/pull/2419))
 - Add more Python and C++ examples related to inverse kinematics with 3d tasks ([#2428](https://github.com/stack-of-tasks/pinocchio/pull/2428))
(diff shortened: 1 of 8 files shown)


jcarpent:
Thanks @lvjonok for sharing this effort with the community. Indeed, at the time being, we only support limited features of MJCF format and we are looking for external contributions to enhance it.

We give some feedback to you in the next following days to enhance your contribution.

Thanks again for contributing to Pinocchio,
Best,

Justin

jcarpent:
@lvjonok It seems the current parsing is not deterministic according to the current issue on the mjcf.cpp test.
Do you know why?

jcarpent:
> @lvjonok It seems the current parsing is not deterministic according to the current issue on the mjcf.cpp test. Do you know why?

I found the issue, related to the use of unordered_map, which is system dependant.

jcarpent (approved):

PR #2415, opened 2024-09-11
doc - Add explanations for null space projection.

abussy-aldebaran:
This is a follow-up from #2366


jcarpent (approved):

jorisv:
@abussy-aldebaran You can add a changelog entry, something like « Add explanation about null space projection formula » in Added section.

Thanks for the PR !

PR #2416, opened 2024-09-14
contact-info: cache nv, fix #2304

### LINKED ISSUE #2304: test-cpp-contact-cholesky failure with GCC 13.3.0
Hi,

This is a preliminary bug report, I'm trying a few things to check that.

But so far, with nix, since gcc was upgraded from 13.2.0 to 13.3.0, test-cpp-contact-cholesky fail with big differences in some matrices (ie. 0.207 != 5e-17).

I suspect this is an issue in GCC, so I will deactivate this test on nix, and try to make a proper bug report to GCC.

nim65s:
Hi,

This fix the symptoms described in #2304.
We should probably dig a bit more to find the cause.

Thanks a lot @jorisv for help there !

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - Fix linkage of Boost.Serialization on Windows ([#2400](https://github.com/stack-of-tasks/pinocchio/pull/2400))
 - Fix mjcf parser appending of inertias at root joint ([#2403](https://github.com/stack-of-tasks/pinocchio/pull/2403))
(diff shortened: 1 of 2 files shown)


jcarpent (commented):
Very weird issues with GCC-13. This compiler should be banned for such mistakes, no?!

ManifoldFR:
Does GCC 14 have the same issue ?

jorisv (approved):

jorisv:
Thanks for the fix @nim65s.

I have missed that `JointModelBase::nv` default implementation call a `boost::static_visitor`.
So, regardless of the compiler bug, this fix simplify the code.

I will merge this PR but keep the issue open to try to understand if this is a know GCC bug.

PR #2443, opened 2024-10-03
Fix some examples

jorisv:
- Fix talos-simulation.py
- Fix cassie-simulation.py

### DIFF
--- a/examples/cassie-simulation.py
+++ b/examples/cassie-simulation.py
@@ -2,43 +2,10 @@ from time import sleep
 
 import numpy as np
 import pinocchio
-from example_robot_data import readParamsFromSrdf
-from pinocchio import JointModelFreeFlyer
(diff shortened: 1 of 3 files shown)


examples/cassie-simulation.py
```diff
@@ -2,43 +2,10 @@
...
-    root_joint=JointModelFreeFlyer(),
-    root_link_name="pelvis",
-    parent_guidance=sdf_parent_guidance,
-)
-
-robot.q0 = readParamsFromSrdf(
-    robot.model, srdf_path, has_rotor_parameters=False, referencePose="standing"
-)
+import example_robot_data
+from pinocchio.visualize import GepettoVisualizer
```
jcarpent, line 6: Why do you want to use GepettoVisualizer?

Issue #2454, opened 2024-10-14
Schedule a nigtly/weekly build to test all Pinocchio options


jorisv:
To accelerate the PR build time, we have added a way to avoid building Pinocchio with all options ON (https://github.com/stack-of-tasks/pinocchio/issues/2251).

This is working but reviewers can forgot to setup the right options and this can create regressions.

To overcome this issue, we can create a nightly/weekly build that will build Pinocchio with all the options ON. This will avoid to slowdown PR build time, but will allow us to catch regression.

jorisv:
Closed by #2456

PR #2456, opened 2024-10-14
Add a scheduled run on ci with build_all

MegMll:
This pr aims at programming a weekly build_all ci, as mentionned in https://github.com/stack-of-tasks/pinocchio/issues/2454


.github/workflows/scripts/get_compilation_flags.js
```diff
@@ -77,6 +64,29 @@ module.exports = async ({github, context, core}) => {
         };
     }
 
+    // Get the GitHub event name that triggered the workflow
+    const eventName = process.env.GITHUB_EVENT_NAME;
+    if (eventName == "schedule")
```
jorisv, line 69: Instead of doing this, is it possible to pass label as an input ?

One way to doing this is to pass an environment variable as input :
- https://github.com/actions/github-script?tab=readme-ov-file#use-env-as-input
- https://github.com/actions/github-script?tab=readme-ov-file#run-a-separate-file-with-an-async-function

If this environment variable is setup use it as labelNames. If the environment variable is not setup, we extract label from data.labels.

.github/workflows/windows-conda.yml
```diff
@@ -41,8 +43,20 @@ jobs:
         compiler: [cl, clang-cl]
 
     steps:
+      # Extract branch name or set 'devel' branch for scheduled events
+      - name: Get branch name or checkout devel
+        shell: pwsh  # Use PowerShell on Windows
```
jorisv, line 48: I think you can use bash as shell on Windows.

jorisv (approved):

PR #2475, opened 2024-11-08
Add pinocchio_python_parser target

### LINKED ISSUE #2474: `pinocchio::python::buildModel` is not build in the main library
`pinocchio::python::buildModel` allow to load a model from a python scripts.

This function should be build in the C++ library and not in the binding. But right now, it's builded only in the bindings and never exposed.

We should add new pinocchio_python_model library to build this code.

jorisv:
Close #2474 

- [x] Create new target
- [x] Move code in core section
- [x] Deprecate bindings header
- [x] Export symbol with dedicated config.hpp
- [x] Use clear CMake variable name

### DIFF
--- a/bindings/python/parsers/python/model.cpp
+++ b/bindings/python/parsers/python/model.cpp
@@ -4,8 +4,6 @@
 
 #include "pinocchio/bindings/python/parsers/python.hpp"
 
-#include <iostream>
-#include <Python.h>
 #include <boost/version.hpp>
 #include <boost/algorithm/string/predicate.hpp>
 
@@ -27,11 +25,6 @@ namespace pinocchio
       bp::object main_module = bp::import("__main__");
       // Get a dict for the global namespace to exec further python code with
       bp::dict globals = bp::extract<bp::dict>(main_module.attr("__dict__"));
-
-      // We need to link to the pinocchio PyWrap. We delegate the dynamic loading to the python
-      // interpreter.
-      bp::object cpp_module(
-        (bp::handle<>(bp::borrowed(PyImport_AddModule("libpinocchio_pywrap")))));
 
       // That's it, you can exec your python script, starting with a model you
       // can update as you want.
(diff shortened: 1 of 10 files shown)


bindings/python/parsers/python/model.cpp
```diff
@@ -28,11 +26,6 @@ namespace pinocchio
       // Get a dict for the global namespace to exec further python code with
       bp::dict globals = bp::extract<bp::dict>(main_module.attr("__dict__"));
 
-      // We need to link to the pinocchio PyWrap. We delegate the dynamic loading to the python
-      // interpreter.
-      bp::object cpp_module(
-        (bp::handle<>(bp::borrowed(PyImport_AddModule("libpinocchio_pywrap")))));
```
jorisv, line 34: @jcarpent  I removed this part and I don't see the point to use it. Tests are working fine on Linux and OSX (Windows is still running)

bindings/python/parsers/python/model.cpp
```diff
@@ -67,10 +60,6 @@ namespace pinocchio
 
       // close the interpreter
       // cf. https://github.com/numpy/numpy/issues/8097
-#if PY_MAJOR_VERSION < 3
-      Py_Finalize();
-#else
```
jorisv, line 72: @jcarpent I removed Python2 specific code

bindings/python/parsers/python/model.cpp
```diff
@@ -28,11 +26,6 @@ namespace pinocchio
       // Get a dict for the global namespace to exec further python code with
       bp::dict globals = bp::extract<bp::dict>(main_module.attr("__dict__"));
 
-      // We need to link to the pinocchio PyWrap. We delegate the dynamic loading to the python
```
jorisv, line 31: @jcarpent  There is some workaround for BOOST 1.58. Can I remove them ?

bindings/python/parsers/python/model.cpp
jcarpent: Perfect

bindings/python/parsers/python/model.cpp
jcarpent: Fine

bindings/python/parsers/python/model.cpp
jcarpent: Yes, sure.

jcarpent (approved):

jcarpent (approved):

PR #2476, opened 2024-11-12
Try to use pre compiled header to accelerate the build

jorisv:
I setup pch on pinocchio_pywrap_default and obtain the following result:

Build with PCH:      ninja pinocchio_pywrap_default  702.06s user 20.62s system 1337% cpu 54.044 total
Build without PCH: ninja pinocchio_pywrap_default  979.61s user 32.80s system 1565% cpu 1:04.67 total

Also, memory consumption seem to be lower. But it's hard de measure.


ManifoldFR:
That's a 28.3% decrease in terms of compilation time. That's kinda nuts.

Have you also tried this on an Apple Silicon Mac ? Like the M1 Ultra Desktop we have, or perhaps @lmontaut's laptop ?

ManifoldFR:
Also, it changes nothing about symbols etc?

jorisv:
> That's a 28.3% decrease in terms of compilation time. That's kinda nuts.
> 
> Have you also tried this on an Apple Silicon Mac ? Like the M1 Ultra Desktop we have, or perhaps @lmontaut's laptop ?

I only did the benchmark on my computer. We can compare CI time, but since we use ccache, I don't think it will be fair.

> Also, it changes nothing about symbols etc?

I don't think it change anything on the symbol. But you right, I will try to compare binary size.

MegMll:
After some testing of pch for pinocchio_default, it has been decided to only use it for the bindings. Indeed the rest of the library does not really benefit from it, compile time is pretty much the same.

Binary size is not changed when using pch, compared to no pch.

jcarpent (approved):

bindings/python/CMakeLists.txt
```diff
@@ -161,6 +161,11 @@ if(BUILD_PYTHON_INTERFACE)
   set(PINOCCHIO_PYTHON_INSTALL_DIR ${ABSOLUTE_PYTHON_SITELIB}/${PROJECT_NAME})
 
   pinocchio_python_bindings_specific_type(default)
+  # Precompile Headers are used to speed up compilation of bindings. Tests were performed also on
+  # pinocchio_default, and there was no gain in using pch for the main lib, compared to the gain for
+  # bindings. PCH are used to precompile headers of other scalar types as well.
+  target_precompile_headers(${PYWRAP}_default PRIVATE
+                            ${PROJECT_SOURCE_DIR}/include/pinocchio/bindings/python/pch.hpp)
```
ManifoldFR, line 168:
```suggestion
  # Precompiled headers are used to speed up compilation of the Python bindings. We have also tested 
  # precompiling headers for the main library's `pinocchio_default` target, but the time gain there was
  # negligible compared to the time gained in the Python bindings.
  # We also precompile headers for other scalar types as well.
  target_precompile_headers(${PYWRAP}_default PRIVATE
                            ${PROJECT_SOURCE_DIR}/include/pinocchio/bindings/python/pch.hpp)
```

ManifoldFR (commented):
Just suggesting a small change for the CMake listfile's comments, for clarity.

ManifoldFR (dismissed):

bindings/python/CMakeLists.txt
```diff
@@ -161,6 +161,12 @@ if(BUILD_PYTHON_INTERFACE)
   set(PINOCCHIO_PYTHON_INSTALL_DIR ${ABSOLUTE_PYTHON_SITELIB}/${PROJECT_NAME})
 
   pinocchio_python_bindings_specific_type(default)
+  # Precompiled headers are used to speed up compilation of the Python bindings. We have also tested 
+  # precompiling headers for the main library's `pinocchio_default` target, but the time gain there was
+  # negligible compared to the time gained in the Python bindings.
+  # We also precompile headers for other scalar types as well.
+  target_precompile_headers(${PYWRAP}_default PRIVATE
```
nim65s, line 168: This require CMake >= 3.16, so we should update the first line of the main CMakeListst.txt

ManifoldFR (approved):

nim65s (approved):

PR #2477, opened 2024-11-13
Mjcf : Fix parsing of armature parameters

MegMll:
This pr aims at correcting parsing of armature parameter available in mjcf models. As of now, mjcf provided only one value per joint, but in pinocchio armature is based on the `model.nv`.

In this pr, we modify how armature is parsed and use the pinocchio standard.

Another aim of this pr is to add support for default tag that does not have a class name. Until now, this was taken into account in the mjcf parser.


jcarpent (approved):
Very nice Megan ;)

src/parsers/mjcf/mjcf-graph.cpp
```diff
@@ -137,6 +137,7 @@ namespace pinocchio
       template<int Nq, int Nv>
       RangeJoint RangeJoint::concatenate(const RangeJoint & range) const
       {
+        typedef UrdfVisitor::Vector Vector;
```
jorisv, line 140: This typedef is never used.

jorisv (changes requested):

PR #2502, opened 2024-12-06
Fix coal integration

jorisv:
Fix some coal related issue:
- unittest/parallel-geometry.cpp fail because of this issue https://github.com/coal-library/coal/issues/636

Fix issue with quadrotor-ocp not convering on Windows:
 - [x] Find a quick fix
 - [x] Search if we can make the problem converge again
   - I can't reproduce on my computer, so it's skiped


hrp2-14:

Issue #2487, opened 2024-11-26
Example meshcat-viewer-octree.py not displaying octree


joao-pm-santos96:
Hello developers. 
I've found an issue which I would like to have your comments about. Please, let me know if I can help with anything.

## Bug description

When running examples/meshcat-viewer-octree.py, the script cannot display the octree. Here is the full error:

```bash
This example is skiped as HPP-FCL has not been compiled with octomap support.
You can open the visualizer by visiting the following URL:
http://127.0.0.1:7000/static/
/home/joao/pixi-test/.pixi/envs/default/lib/python3.12/site-packages/pinocchio/visualize/meshcat_visualizer.py:900: UserWarning: The geometry object named octree is not supported by Pinocchio/MeshCat for vizualization.
  self.loadViewerGeometryObject(
/home/joao/pixi-test/.pixi/envs/default/lib/python3.12/site-packages/pinocchio/visualize/meshcat_visualizer.py:909: UserWarning: The geometry object named octree is not supported by Pinocchio/MeshCat for vizualization.
  self.loadViewerGeometryObject(
gio: http://127.0.0.1:7000/static/: Operation not supported
```

I've also tried installing coal and updating the script to import it instead of hppfcl and the result was:

```bash
Traceback (most recent call last):
  File "/home/joao/pixi-test/../lumen/nexxus/ola.py", line 25, in <module>
    octree_object = pin.GeometryObject("octree", 0, pin.SE3.Identity(), octree)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Boost.Python.ArgumentError: Python argument types in
    GeometryObject.__init__(GeometryObject, str, int, SE3, OcTree)
did not match C++ signature:
    __init__(_object* self, pinocchio::GeometryObject otherGeometryObject)
    __init__(_object* self, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > name, unsigned long parent_joint, std::shared_ptr<hpp::fcl::CollisionGeometry> collision_geometry, pinocchio::SE3Tpl<double, 0> placement)
[...]

PR #2503, opened 2024-12-06
Fix meshcat examples

jorisv:
This PR fix meshcat examples.

- `meshcat-viewer-octree.py` used deprecated API
- `meshcat-viewer-solo.py` used example_robot_data, but this package can't be imported in Pinocchio
Some meshact examples where broken, but also, some display nothing except 
- I had to add a sleep at the end of the script to display something on my browser. Is it normal @ManifoldFR ?


jorisv:
Should fix #2487

ManifoldFR:
It's totally normal, the meshcat visualizer's startup of its server is non blocking, so your script MAY exit before the server got spun up and your browser tab opens.

You should run your script with `python -i`

jorisv:
Ok, thank you for the explanation !

jorisv:
And there is no a blocking call in the API to wait for the viewer ?

ManifoldFR:
I don't think there is, apart from the method which tries to render an image (and I'm not sure the call terminates if the viewer is not already available). We could do a PR for this but nothing is happening upstream...

jcarpent (approved):

PR #2526, opened 2024-12-22
Mjcf: Use xyz as a default value for eulerseq compiler option

JafarAbdi:
This PR initializes `mapEulerAngles` with an identity matrix by default (matching MuJoCo's default 'xyz' sequence as documented in https://mujoco.readthedocs.io/en/stable/XMLreference.html#compiler-eulerseq).

While testing robots from https://github.com/google-deepmind/mujoco_menagerie, I encountered nan values in some frame transforms. After debugging, I found out that `mapEulerAngles` was not initialized by default and returned garbage values when no eulerseq was specified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


hrp2-14:

jcarpent (approved):
Thanks a lot @JafarAbdi for this nice contribution. Thanks a lot for adding the test.

jcarpent (approved):

PR #2535, opened 2024-12-31
mjcf parser: Fix unknown size vector parsing

JafarAbdi:
I ran into `std::__ios_failure: basic_ios::clear: iostream error` errors while trying to parse [g1.xml](https://github.com/google-deepmind/mujoco_menagerie/blob/main/unitree_g1/g1.xml), after debugging it I found that it happens when parsing the keyframe's qpos attribute [here](https://github.com/google-deepmind/mujoco_menagerie/blob/main/unitree_g1/g1.xml#L347-L355).

This PR fixes it by changing the stream exception handling from failbit to badbit and simplifies vector parsing logic to better handle whitespace and newlines in MJCF files.

### DIFF
--- a/unittest/mjcf.cpp
+++ b/unittest/mjcf.cpp
@@ -8,6 +8,7 @@
 #include "pinocchio/multibody/model.hpp"
 
 #include "pinocchio/parsers/mjcf.hpp"
+#include "pinocchio/parsers/mjcf/mjcf-graph.hpp"
 #include "pinocchio/parsers/urdf.hpp"
 
 #include "pinocchio/algorithm/joint-configuration.hpp"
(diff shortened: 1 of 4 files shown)


hrp2-14:

unittest/mjcf.cpp
```diff
@@ -923,7 +924,8 @@ BOOST_AUTO_TEST_CASE(adding_keyframes)
                     <key name="test"
                     qpos="0 0 0.596
                         0.988015 0 0.154359 0
-                        0.988015 0 0.154359 0"/>
+                        0.988015 0 0.154359 0
+                        "/>
```
JafarAbdi, line 928: Should I revert this change? It's just meant as an extra test, otherwise a similar case is added in the `test_get_unknown_size_vector_from_stream` test

jcarpent:
@MegMll @jorisv Could you handle this PR?

unittest/mjcf.cpp
MegMll: No it's okay, it adds another test.

MegMll (dismissed):

MegMll (approved):

PR #2537, opened 2025-01-01
Add support for vertex-based mesh definitions in MJCF

JafarAbdi:
This PR adds support for parsing MJCF meshes that are defined directly with vertices in the XML, rather than only supporting external mesh files. This makes it possible to load https://github.com/google-deepmind/mujoco/blob/main/model/car/car.xml and https://github.com/google-deepmind/mujoco/tree/main/model/cube

### DIFF
--- a/src/parsers/mjcf/mjcf-graph.cpp
+++ b/src/parsers/mjcf/mjcf-graph.cpp
@@ -582,20 +582,51 @@ namespace pinocchio
 
         MjcfMesh mesh;
         auto file = el.get_optional<std::string>("<xmlattr>.file");
-        if (!file)
-          throw std::invalid_argument("Only meshes with files are supported");
-
-        fs::path filePath(*file);
-        std::string name = getName(el, filePath);
-
-        mesh.filePath =
-          updatePath(compilerInfo.strippath, compilerInfo.meshdir, modelPath, filePath).string();
-
         auto scale = el.get_optional<std::string>("<xmlattr>.scale");
         if (scale)
           mesh.scale = internal::getVectorFromStream<3>(*scale);
+        if (file)
+        {
+          fs::path filePath(*file);
+          std::string name = getName(el, filePath);
 
-        mapOfMeshes.insert(std::make_pair(name, mesh));
+          mesh.filePath =
+            updatePath(compilerInfo.strippath, compilerInfo.meshdir, modelPath, filePath).string();
+          mapOfMeshes.insert(std::make_pair(name, mesh));
+          return;
+        }
+
+        // Handle vertex-based mesh
+        auto vertex = el.get_optional<std::string>("<xmlattr>.vertex");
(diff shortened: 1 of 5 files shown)


src/parsers/mjcf/mjcf-graph.cpp
```diff
@@ -582,20 +582,49 @@ namespace pinocchio
...
+            updatePath(compilerInfo.strippath, compilerInfo.meshdir, modelPath, filePath).string();
+          mapOfMeshes.insert(std::make_pair(name, mesh));
+          return;
+        }
+
+        // Handle vertex-based mesh
+        auto vertex = el.get_optional<std::string>("<xmlattr>.vertex");
+        if (!vertex)
+        {
+          throw std::invalid_argument("Only meshes with files/vertices are supported");
```
jcarpent, line 603:
```suggestion
          PINOCCHIO_THROW_PRETTY(std::invalid_argument,"Only meshes with files/vertices are supported");
```

src/parsers/mjcf/mjcf-graph.cpp
```diff
@@ -582,20 +582,49 @@ namespace pinocchio
...
+        {
+          throw std::invalid_argument("Mesh with vertices without a name is not supported");
+        }
 
-        mapOfMeshes.insert(std::make_pair(name, mesh));
+        // Parse and validate vertices
+        Eigen::VectorXd meshVertices = internal::getUnknownSizeVectorFromStream(*vertex);
+        if (meshVertices.size() % 3 != 0)
+        {
+          throw std::invalid_argument("Number of vertices is not a multiple of 3");
```
jcarpent, line 616: Same here.

src/parsers/mjcf/mjcf-graph.cpp
```diff
@@ -582,20 +582,49 @@ namespace pinocchio
...
+        auto vertex = el.get_optional<std::string>("<xmlattr>.vertex");
+        if (!vertex)
+        {
+          throw std::invalid_argument("Only meshes with files/vertices are supported");
+        }
+
+        auto name = el.get_optional<std::string>("<xmlattr>.name");
+        if (!name)
+        {
+          throw std::invalid_argument("Mesh with vertices without a name is not supported");
```
jcarpent, line 609: Same here.

jcarpent (changes requested):
Thanks @JafarAbdi for the nice PR.
Could you use the Pinocchio macro `PINOCCHIO_THROW_PRETTY` instead of throw? 
I've provided a first suggestion.

Thanks in advance

src/parsers/mjcf/mjcf-graph.cpp
JafarAbdi: Thanks for the review! I addressed the comments

jcarpent (approved):

PR #2551, opened 2025-01-09
Use Eigen expression to compute spatial algebra

jorisv:
This PR aim to use Eigen expression to compute spatial algebra.

Our SE3 implementation (same for Motion, Force and Inertia) prevent to use the Eigen `noalias` fonction.
This prevent the compiler to optimize the code well.
Some benchmark had shown that we can accelerate the forward kinematics computation time by 1.5x by using this trick.

# TODO V1

- [x] Implement SE3ExprBase
- [ ] Implement SE3TplExpr
  - [x] act
  - [ ] actInv
- [ ] Add unit test
- [ ] Create spatial benchmark
- [ ] Use SE3Expr in kinematics (0 order)
- [ ] Benchmark old vs new kinematics

# TODO V2

- [ ] Find a way to mix SE3, Motion, Force and Inertia expression
- [ ] Use expression in Motion
- [ ] Use expression in Force
- [ ] Use expression in Inertia
- [ ] Use expression in kinematics
- [ ] Use expression in aba
- [ ] Use expression in crba


ManifoldFR:
I allowed myself to re-title the PR @jorisv.

ManifoldFR:
For the spatial benchmark, are we going to write it like we usually did for Pinocchio, or start switching to the more standard [google bench](https://github.com/google/benchmark/) ?

I vote for the latter option.

jorisv:
@ManifoldFR We will write it using google benchmark. We plan to port all benchmark to this library in the future.

ManifoldFR:
> @ManifoldFR We will write it using google benchmark. We plan to port all benchmark to this library in the future.

Excellent, I feel it will make our bench code much more professional. Is the port of all the benchmarks somewhere on the roadmap/issue tracker? We could work on that on the side.

jcarpent (changes requested):
I would suggest using `Expression` in full name for clarity and readibility.

jcarpent (changes requested):
Could you split benchmark refactorization and expression implementation to ease the review process?

include/pinocchio/spatial/se3-expr-base.hpp
```diff
@@ -0,0 +1,168 @@
...
+  // in context/cppadcg.hpp we can include se3.hpp without including spatial/fwd.hpp first.
+  template<typename Derived>
+  struct SE3ExprBase;
+  template<typename Derived>
+  struct SE3ExprNoalias;
+  template<typename RotProduct, typename TransProduct>
+  struct SE3ExprProduct;
+  template<typename RotProduct, typename TransProduct>
+  SE3ExprProduct<RotProduct, TransProduct>
+  make_se3_expr_product(RotProduct rot_prod, TransProduct trans_prod);
```
jcarpent, line 21: I would suggest a function `make_product` generic, that calls then a template specialization for readability.

Issue #2563, opened 2025-01-29
Test failure on crocoddyl on macos: `TypeError: No to_python (by-value) converter found for C++ type: boost::shared_ptr<pinocchio::ModelTpl<double, 0, pinocchio::JointCollectionDefaultTpl>>`


nim65s:
Hi,

crocoddyl v2.1.0 was building fine on macos with pinocchio v3.3.0, but some tests are failing with pinocchio v3.3.1:
```python
crocoddyl> The following tests FAILED:
crocoddyl>       29 - test_pybinds_actuations (Failed)
crocoddyl>       30 - test_pybinds_actions (Failed)
crocoddyl>       31 - test_pybinds_shooting (Failed)
crocoddyl>       33 - test_pybinds_costs (Failed)
crocoddyl>       34 - test_pybinds_contacts (Failed)
crocoddyl>       35 - test_pybinds_impulses (Failed)
crocoddyl>       37 - test_pybinds_copy (Failed)
```

with eg.
```python
crocoddyl> ======================================================================
crocoddyl> ERROR: test_calc (__main__.TalosArmShootingTest.test_calc)
crocoddyl> ----------------------------------------------------------------------
crocoddyl> Traceback (most recent call last):
crocoddyl>   File "…/unittest/bindings/test_shooting.py", line 30, in setUp
crocoddyl>     self.PROBLEM_DER = crocoddyl.ShootingProblem(
crocoddyl>                        ^^^^^^^^^^^^^^^^^^^^^^^^^^
crocoddyl>   File "…/unittest/bindings/factory.py", line 445, in createData
crocoddyl>     data = DifferentialFreeFwdDynamicsDataDerived(self)
crocoddyl>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
crocoddyl>   File "…/unittest/bindings/factory.py", line 459, in __init__
crocoddyl>     self.pinocchio = pinocchio.Model.createData(model.state.pinocchio)
crocoddyl>                                                 ^^^^^^^^^^^^^^^^^^^^^
crocoddyl> TypeError: No to_python (by-value) converter found for C++ type: boost::shared_ptr<pinocchio::ModelTpl<double, 0, pinocchio::JointCollectionDefaultTpl>>
```

[...]

PR #2566, opened 2025-01-29
python: register Model

nim65s:
Hi again,

This fix https://github.com/stack-of-tasks/pinocchio/issues/2563, when https://github.com/loco-3d/crocoddyl/pull/1339 is also applied.


jcarpent (changes requested):
I would also add it for Data, GeometryModel, and GeometryData to be complete.

jcarpent (approved):

PR #2567, opened 2025-01-29
hpp-fcl -> coal

nim65s:
Not sure why I started working on that. We probably just have to start somewhere to get rid of those warnings I guess.
Anyway, in the current state, this PR is probably very wrong in different places. Please don't start reviewing if you care about your sanity.
Also, ROS things I started with are not working, waiting for eigenpy & coal releases including https://github.com/jrl-umi3218/jrl-cmakemodules/pull/735


ManifoldFR:
> Please don't start reviewing if you care about your sanity.

That's perfect then, after my last few PRs on the jrl-cmakemodules my sanity is completely gone!

unittest/python/bindings_geometry_object.py
```diff
@@ -4,7 +4,7 @@
 import pinocchio as pin
 
 
-@unittest.skipUnless(pin.WITH_HPP_FCL, "Needs HPP-FCL")
+@unittest.skipUnless(pin.WITH_COAL, "Needs HPP-FCL")
```
jcarpent, line 7:
```suggestion
@unittest.skipUnless(pin.WITH_COAL, "Needs Coal")
```

unittest/parallel-geometry.cpp
```diff
@@ -260,8 +260,8 @@ BOOST_AUTO_TEST_CASE(test_talos)
       }
       else
       {
-        // Only run this part in coal, this issue doesn't happen in hpp-fcl
-#if HPP_FCL_VERSION_AT_LEAST(3, 0, 0)
+        // Only run this part in coal, this issue doesn't happen in coal
+#if COAL_VERSION_AT_LEAST(3, 0, 0)
```
jcarpent, line 264:
```suggestion
```

unittest/parallel-geometry.cpp
```diff
@@ -278,7 +278,7 @@ BOOST_AUTO_TEST_CASE(test_talos)
         BOOST_CHECK(contact_ref.pos != contact_ref.pos);
         BOOST_CHECK(contact_ref.nearest_points[0] != contact_ref.nearest_points[0]);
         BOOST_CHECK(contact_ref.nearest_points[1] != contact_ref.nearest_points[1]);
-#endif // hpp-fcl >= 3.0.0
+#endif // coal >= 3.0.0
```
jcarpent, line 281:
```suggestion
#endif // coal >= 3.0.0
```

jcarpent:
I would suggest waiting for important change currently happening on Pinocchio from Willow side. As it is mostly find&replace actions, it would be easier for the merging.

.github/workflows/ros-ci.yml
jcarpent, line 1: Should be moved to a dedicated PR.

jcarpent:
After some discussions with @nim65s, we will redo this work after the merging of ongoing developments from Willow side.

PR #2574, opened 2025-02-12
Add C++ visualization API and some macros for scalar contexts

### LINKED ISSUE #2573: Add a basic C++ visualization API / upstream `Simple-Robotics/pinocchio-visualizers`
Make sure your proposal checks out the following questions:

- [x] Is your feature request related to a problem? Please describe, *e.g.* "I'm always frustrated when..."
- [x] Describe concisely the solution you'd like.
- [x] Describe alternatives you've considered.
- [x] Do you have screenshots or sketches to illustrate your proposal?

----

A while back, there was discussion about providing basic tools for users to provide C++-based visualizers to Pinocchio, which would integrate with existing elements of the Pinocchio C++ API (computing frame and geometry placements, etc), and provide a C++ equivalent to the `pinocchio.visualizers.BaseVisualizer` Python class (although @jcarpent suggests this interface itself could/should be improved).

As a result I made the header-only library https://github.com/Simple-Robotics/pinocchio-visualizers, which also embarks a Boost.Python visitor class for exposing the C++ visualizer to Python. So far only two (very WIP) projects use it:
- [my attempt at a C++ Rerun visualizer](https://github.com/ManifoldFR/pinocchio-rerun/blob/main/src/rerun_visualizer.hpp)
- the new SDL3-based [proof-of-concept renderer](https://github.com/ManifoldFR/candlewick)

After discussion earlier today with @jcarpent we think it might be a good idea to upstream this into Pinocchio, perhaps as part of the extras CMake target.

ManifoldFR:
This pull request upstreams the contents of the single-header library [Simple-Robotics/pinocchio-visualizers](https://github.com/Simple-Robotics/pinocchio-visualizers), as discussed on issue #2573.
This API reproduces the behaviour of the existing Pinocchio Python bindings' `pinocchio.visualizers.BaseVisualizer` class (which is in Python).

* Add header `<pinocchio/visualizers/base-visualizer.hpp>`
* Add new CMake target `pinocchio_visualizers`
* Add namespace `pinocchio::visualizers`
* Add base class `pinocchio::visualizers::BaseVisualizer`, for the `context::Scalar` scalar type. Common functionality (updating geometry placements, handling frames, the `viz.play(qs)` function) is compiled into the new `pinocchio_visualizers` target.
* Add `pinocchio::python::VisualizerPythonVisitor` visitor for exposing visitor classes

Resolves #2573

## Lingering questions

* As a way of forcing implementers to implement the right functions, should we:
  * Keep using a virtual class with pure virtual members?
  * ~~Or just have the API throw a runtime error if unimplemented members get called?~~
  * ~~Or expose a CRTP class with the actual common logic implemented in a private base class?~~
* ~~Better name or mechanism for the protected virtual member function `BaseVisualizer::displayPrecall()` - the idea is to be able to inject a hook before the kinematics and geometry get updated, e.g. locking the `GeometryData` using a mutex.~~
* ~~Whether to name all of the unimplemented functions `XXXImpl()` or `XXX_impl()` and whether to make them protected.~~ 
* How to fuse this into the actual Python `BaseVisualizer` class, which we could significantly cut down.


jcarpent (dismissed):
Looks good to me.

ManifoldFR:
I forgot to update the README...

include/pinocchio/visualizers/base-visualizer.hpp
```diff
@@ -0,0 +1,227 @@
...
+      {
+      }
+      virtual void displayImpl() = 0;
+
+      void destroyData();
+    };
+  } // namespace visualizers
+} // namespace pinocchio
+
+namespace pinviz = ::pinocchio::visualizers; // NOLINT
```
jcarpent, line 225: I don't if it is a good idea to add it. If it is not used, maybe we should let the user define it in its own project I would say.

ManifoldFR: You're right, this was a holdover from the upstream `pinocchio-visualizers`.

ManifoldFR:
```suggestion
```

unittest/visualizer.cpp
```diff
@@ -0,0 +1,19 @@
...
+{
+  void loadViewerModel() override
+  {
+  }
+  void displayImpl() override
+  {
+  }
+};
+
+BOOST_AUTO_TEST_SUITE_END()
```
jcarpent, line 19: The test looks empty. No?

ManifoldFR: I was writing it up, it's pushed now

ManifoldFR:
I think this is ok to merge @jcarpent :)

jcarpent (approved):

PR #2576, opened 2025-02-12
[cmake] Change formatting of listfiles to use gersemi, remove cmake-format

ManifoldFR:
This PR removes the abandoned [cmake-format](https://github.com/cheshirekow/cmake_format) to use the maintained [gersemi](https://github.com/BlankSpruce/gersemi). It also reformats all the CMake listfiles, picking up case-sensitive definitions from the jrl-cmakemodules and other listfiles in Pinocchio.


jcarpent:
Should it possible to postpone this PR due to parallel developments?
It will be hard to do the proper merging with Pinocchio 4 in preparation.

jcarpent:
I will handle the merging ;)

jorisv:
Gersemi seem to well handle functions with optional arguments. The result is nice even without specifying the arguments in a config file (like we did in .cmake-format.yaml).

jorisv (approved):

ManifoldFR:
> Gersemi seem to well handle functions with optional arguments. The result is nice even without specifying the arguments in a config file (like we did in .cmake-format.yaml).

Yes, the author of gersemi BlankSpruce did this deliberately to simplify the parsing logic. Users just have to write their functions/macros which use `cmake_parse_arguments` in a standard way and fulfill the ["deal"](https://github.com/BlankSpruce/gersemi?tab=readme-ov-file#lets-make-a-deal) in the README.

I've already used this successfully in aligator to simplify some CMake functions. It works much better than what I could achieve with cmake-format.

@jcarpent we can wait for Pinocchio 4, sure 😃

ManifoldFR:
As discussed previously, we will be doing this for the upcoming Pinocchio 4 release, so I'm closing this PR for now.

PR #2607, opened 2025-03-05
Use google benchmark instead of our legacy benchmark system

jorisv:
This PR aim to switch to [Google benchmark](https://github.com/google/benchmark/blob/main/docs/user_guide.md)

Our legacy system doesn't support some Google benchmark nice features:
- Nice stdout output
- JSON output
- Benchmark filters
- Benchmark Warmup
- Benchmark parameters (to run the same function with different parameters)
- Custom CPU timer
- Function to prevent code optimization
- Benchmark iteration auto tuning
- Benchmark repetition (run multiple time a benchmark to measure mean/median and standard deviation)

## Avoid code optimization

A big issue in micro benchmarking is compiler optimization.
Compiler can over optimize the code:
- Result is always stored in the same variable: Loop is only run once.
- Parameters never change: Some code can be computed only once.

To avoid that, Google benchmark provide `benchmark::DoNotOptimize` and `benchmark::ClobberMemory()`.
These functions can be a little tricky to use, so I prefer, when possible, to wrap the benchmarked function into a `EIGEN_DONT_INLINE` function. This will prevent optimization since the function can't be inlined.
As a downside, function call cost will be added to the benchmark result. Since all our functions are benchmarked the same way, this is not an issue since they will all have the same penalty.

Example:

```cpp
EIGEN_DONT_INLINE static void rneaCall(
  const Model & model,
  Data & data,
  const Eigen::VectorXd & q,
  const Eigen::VectorXd & v,
  const Eigen::VectorXd & a)
{
  pinocchio::rnea(model, data, q, v, a);
}
```

## Google benchmark VS legacy timing

Here the comparaison between Google benchmark and legacy system timing.
It can be seen that timing are quit the same, implying no regression.

Only exception is for the EMPTY_FORWARD_PASS_XXX benchmarks.
Some of them were optimized by the compiler (XXX_NO_DATA), so I have added some operation in the benchmark to remove the optimization.
This give slower results on Google benchmark, but I suspect the legacy system suffer of the same unfixed issues since the timing are near 0.

|Algorithm | google benchmark (ns)| legacy benchmark (ns)
|--- |--- |---
|RNEA                                       |   5994 |   6078
|NLE                                        |   6011 |   6080
|NLE_via_RNEA                               |   6045 |   6097
|CRBA_LOCAL                                 |   3574 |   3761
|CRBA_WORLD                                 |   4963 |   5068
|COMPUTE_ALL_TERMS                          |  18449 |  18560
|CHOLESKY_DECOMPOSE                         |   1832 |   1906
|DENSE_CHOLESKY_DECOMPOSE                   |   5674 |   6110
|COMPUTE_JOINT_JACOBIANS                    |   2244 |   2282
|COMPUTE_JOINT_JACOBIANS_TIME_VARIATION     |   3852 |   3987
|JACOBIAN_CENTER_OF_MASS                    |   2527 |   2689
|CENTER_OF_MASS                             |   5265 |   5451
|FORWARD_KINEMATICS_Q                       |   1645 |   1806
|FORWARD_KINEMATICS_Q_V                     |   2429 |   2532
|FORWARD_KINEMATICS_Q_V_A                   |   4314 |   4342
|FRAME_FORWARD_KINEMATICS                   |   2403 |   2679
|UPDATE_FRAME_PLACEMENTS                    |    742 |    933
|CCRBA                                      |   3486 |   3659
|ABA_LOCAL                                  |  14843 |  14805
|ABA_WORLD                                  |  13235 |  13297
|COMPUTE_CORIOLIS_MATRIX                    |  15493 |  15662
|COMPUTE_M_INVERSE_Q                        |  12268 |  12105
|COMPUTE_M_INVERSE                          |   5431 |   5649
|EMPTY_FORWARD_PASS_UNARY_VISIT             |    203 |    140
|EMPTY_FORWARD_PASS_UNARY_VISIT_NO_DATA     |    129 |      0
|EMPTY_FORWARD_PASS_BINARY_VISIT            |    341 |    324
|EMPTY_FORWARD_PASS_BINARY_VISIT_NO_DATA    |    161 |      7

## TODO

- [x] Port all benchmark
  - [x] timings-cg.cpp
  - [x] timings-cholesky.cpp
  - [x] timings-constrained-dynamics-derivatives.cpp
  - [x] timings-contact-dynamics.cpp
  - [x] timings.cpp
  - [x] timings-cppad-jit.cpp
  - [x] timings-delassus-operations.cpp
  - [x] timings-derivatives.cpp
  - [x] timings-eigen.cpp
  - [x] timings-geometry.cpp
  - [x] timings-impulse-dynamics.cpp
  - [x] timings-impulse-dynamics-derivatives.cpp
  - [x] timings-jacobian.cpp
  - [x] timings-parallel.cpp
- [x] Add a README.md in benchmark directory to give some command line snippets
- [x] Add a script to plot benchmark results
- [x] Define PINOCCHIO_DONT_INLINE
- [x] prefix benchmark targets by pinocchio


ManifoldFR:
Magnificent work @jorisv. One more step in the direction of modernising our stack!

jcarpent (commented):
@jorisv Could you add a macro PINOCCHIO_DONT_INLINE for code consistency?
(by just a define over EIGEN_DONT_INLINE)

jorisv:
Here a plot example:

2025-03-19-155620_1907x763_scrot

jcarpent (approved):

jcarpent:
Very nice work @jorisv. This is a significant enhancement of the whole benchmark pipeline in Pinocchio.

PR #2644, opened 2025-04-08
[Model] Improve error message

florent-lamiraux:


ManifoldFR:
Thanks for this @florent-lamiraux !

We would want to drop the two commits from the @pre-commit-ci bot however, we're not ready to update the formatting rules/linters on Pinocchio until v4 as @jcarpent mentioned in #2643.

jcarpent (approved):

PR #2652, opened 2025-04-16
Update mjcf parser

valeriaskvo:
Fix: Add support for friction parameter in MJCF model parsing

I noticed that during the parsing of MJCF models, the friction parameter was not being passed through. According to the documentation for URDF models, [friction](https://wiki.ros.org/urdf/XML/joint#:~:text=friction%20(optional%2C%20defaults%20to%200)) refers to the static friction value.

In Mujoco, this same parameter is referred to as [frictionloss](https://mujoco.readthedocs.io/en/stable/XMLreference.html#body-joint-frictionloss), and it represents the same value. I have modified the code to ensure that the friction parameter is now properly passed through when parsing MJCF models.


hrp2-14:

lmontaut:
Hello @valeriaskvo, the `frictionloss` parameter is already parsed in Pinocchio 4, which is being prepared and will be released soon.

PR #2659, opened 2025-04-22
Fix custom scalar support

jorisv:
Some algorithm had been deactivated for some custom scalar.
This PR aim to add more support.

casadi:
- contraintDynamics

cppad:

cppadcg:


include/pinocchio/algorithm/cholesky.hxx
```diff
@@ -363,7 +363,7 @@ namespace pinocchio
 
           for (int k = model.nv - 1; k >= 0; --k)
           {
-            vout_[k] = M.row(k).segment(k, nvt[(size_t)k]) * vin.segment(k, nvt[(size_t)k]);
+            vout_[k] = M.row(k).segment(k, nvt[(size_t)k]).dot(vin.segment(k, nvt[(size_t)k]));
```
jcarpent, line 366: What was the issue here?

jorisv: When using CppAd, `vout_[k] =M.row(k).segment(k, nvt[(size_t)k]) * vin.segment(k, nvt[(size_t)k]);` call `CppAd::operator=` that try to assign an Eigen::Product to a double.
The workaround I found is to use the dot product that will not return an Eigen::Product.

jorisv: I can provide you the full build trace if you're interested

jcarpent (approved):

PR #2664, opened 2025-04-25
Introducing model graph

MegMll:
This pr aims at introducing a step between parsing and creating a pinocchio model. 

It will allow a new representation based on boost::graph, that should facilitate model manipulation (append two models, blocks some joints....) 

It's still in its very very early stages.

# TODO

Mandatory:
- [x] Manage generic root joint
- [x] Manage root joint position
- [x] Add possibility to change root joint name 
- [x] Add test to make sure inertia are parsed correctly
- [x] Manage all joints
  - [x] Fixed
  - [x] Revolute
   - reverse: inverse q
  - [x] Prismatic
   - reverse: inverse q
  - [x] Unbound
   - reverse: inverse q
  - [x] Mimic
   - reverse: raise exception
  - [x] Universal
   - reverse: change joint axis and negate them (no negating q)
  - [x] Planar
   - reverse: raise exception
  - [x] FreeFlyer
   - reverse: inverse q
  - [x] Translation
   - reverse: inverse q
  - [x] Spherical
   - reverse: inverse q
  - [x] Spherical ZYX
   - reverse: inverse q 
   - use eigen to reverse euler angles (from -x-y-z back to zyx)
  - [x] Helical
   - reverse: inverse q
  - [x] Composite
   - reverse: reverse all inside joints
- [x] Do a struct to have a configuration, velocity, acceleration, torque converter between two models with different root_body
- [x] Merge model algorithm (appendModel)
- [x] Joint type transformation (reduceModel)
  - q ref for moving joints
  - se3 for fixed joint 
  - deal first with q_ref for mujoco and then reduce model should be straightforward
- [x] Document Vertex/Edges frames
- [x] How to deal with differents frame type ?
  - vertex will now be a variant (OP_FRAME, BODY, SENSOR). It will allow for more flexibility and new types of frames could be created.
- [x] Documentation
  - [x] Doxygen
  - [x] Example
- [x] Organize code 
- [x] Python bindings
- [x] Hide visitors in cpp when possible or use internal
- [x] only use reverse or forward
  - [x] change in struct use a boolean forward instead of reverse  
- [x] Remove boost optional  
- [x] name for lock joints
- [x] q ref for free flyer
- [x] add a function for prefixing a graph
- [x] move buildModel outside of the graph (model-graph-algo.hpp)
- [x] Manage joint bias (mujoco reference)
- [x] Implement geometries for body and how they can enter the graph
  - [x] add them to the vertex (BodyGraph) as a variant itself
    - Mesh
    - Cylinder
    - Capsule
    - Box    
- [x] Joint limits  
   - [x] Find a way to check jointLimits (especially reverse) 
- [x] Split bindings in better way ? exposeModelGraphAlgo might be a bit too big and generic
- [x] Split better in the different files. Now with Geometries, it's a bit messy...
- [x] Fix qref call for jointfixed with offset
- [x] Fix as many compilations warnings as possible
- [x] Lock and Merge examples

Next:
- [ ] Optimize model order to manage more mimic configuration
- [ ] Use ModelGraph in parser
  - [ ]  Deprecate appendModel and buildReducedModel 
- [ ] Convert configuration/tangent vector from a model to a model that had been merged or locked (with more joints or less joints)
- [ ] Constraints ?


include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+#include <string>
+#include <iostream>
+#include <unordered_map>
+
+namespace pinocchio
+{
+  namespace internal
+  {
+    template<typename Derived>
+    struct JointBaseGraph
```
jorisv: I'm not sure we need CRTP here.

Also, I don't think we need inheritance here: the name is contains in ModelGraphEdge

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+            boost::get<pinocchio::internal::JointRevoluteGraph>(&edge.joint))
+        {
+          if (jrg->flipped)
+          {
+            pinocchio::SE3 transf = pinocchio::SE3(
+              Eigen::AngleAxisd(M_PI, jrg->flip_axis).toRotationMatrix(), Eigen::Vector3d::Zero());
+            joint_pose.rotation() *= transf.rotation();
+            body_pose = transf * body_pose;
+          }
+        }
```
jorisv: It's better to use a visitor here

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+      {
+        return joint;
+      }
+      ReturnType operator()(const JointFreeFlyerGraph & joint) const
+      {
+        return joint;
+      }
+    };
+
+    struct createJointModel : public boost::static_visitor<JointModel>
```
jorisv:
```suggestion
    struct CreateJointModel : public boost::static_visitor<JointModel>
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+    template<typename Derived>
+    struct JointBaseGraph
+    {
+      std::string name;
+
+      JointBaseGraph(const std::string & n)
+      : name(n) {};
+    };
+
+    struct JointFixedGraph : public JointBaseGraph<JointFixedGraph>
```
jorisv: Joints should not be in internal namespace since use have to provide the to ModelGraph::addJoint

jorisv (changes requested):
I will continue the review tomorrow.

Nice first draft !

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
+//
+// Copyright (c) 2024 INRIA
```
jorisv:
```suggestion
// Copyright (c) 2025 INRIA
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+
+    struct JointRevoluteGraph : public JointBaseGraph<JointRevoluteGraph>
+    {
+      // rotation axis
+      Eigen::Vector3d axis;
+      // axis to flip around if needed
+      Eigen::Vector3d flip_axis;
+      // if axis is unitary (X, Y, Z) and negative, then flipped is true, to remember to flip
+      // position when adding joint to model
+      bool flipped = false;
```
jorisv: I think this is better to manage this in `CreateJointModel` since this transformation allow to keep some pinocchio optimization.

`CreateJointModel` can return a static transform and a JointModel.

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+      std::string name;
+
+      JointBaseGraph(const std::string & n)
+      : name(n) {};
+    };
+
+    struct JointFixedGraph : public JointBaseGraph<JointFixedGraph>
+    {
+      JointFixedGraph()
+      : JointBaseGraph("") {};
```
jorisv:
```suggestion
      JointFixedGraph() = default;
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+          axis = Eigen::Vector3d::UnitY();
+          flip_axis = Eigen::Vector3d::UnitX();
+        }
+        else if (axis.isApprox(-Eigen::Vector3d::UnitZ()))
+        {
+          flipped = true;
+          axis = Eigen::Vector3d::UnitZ();
+          flip_axis = Eigen::Vector3d::UnitY();
+        }
+      };
```
jorisv:
```suggestion
      }
```
You don't need to put ; after a function/method definition. You can remove it in all the file.

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+      Eigen::Vector3d axis;
+      JointPrismaticGraph(const std::string & n, const Eigen::Vector3d & ax)
+      : JointBaseGraph(n)
+      , axis(ax) {};
+    };
+
+    struct JointFreeFlyerGraph : public JointBaseGraph<JointFreeFlyerGraph>
+    {
+      JointFreeFlyerGraph()
+      : JointBaseGraph("") {};
```
jorisv:
```suggestion
      JointFreeFlyerGraph() = default;
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+      : JointBaseGraph(n) {};
+    };
+
+    using JointGraphVariant =
+      boost::variant<JointFixedGraph, JointRevoluteGraph, JointPrismaticGraph, JointFreeFlyerGraph>;
+
+    // TODO move in another header
+    // TODO support all joints
+    // TODO tests
+    // TODO Redefine Rev{X,Y,Z} to use our options
```
jorisv:
```suggestion
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+      std::vector<EdgeDesc> * edges;
+    };
+
+  } // namespace internal
+
+  struct ModelGraphVertex
+  {
+    EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+
+    typedef std::unordered_map<std::string, pinocchio::SE3> com_pos_wrt_edges;
```
jorisv: Not used
```suggestion
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+    // transformation from body supporting to joint frame
+    SE3 out_to_joint;
+    // Transformation from joint to in vertex
+    // transformation from joint frame to body supported
+    SE3 joint_to_in;
+  };
+
+  struct ModelGraph
+  {
+    EIGEN_MAKE_ALIGNED_OPERATOR_NEW
```
jorisv:
```suggestion
```
Since `ModelGraph` doesn't hold an Eigen type by value (only by ptr through `Graph`) I think this definition is not useful.

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+    // Inertia at the CoM
+    // CoM position is defined wrt to a body frame
+    // Since we reverse, how do we keep this ? Many joints
+    // Inertia matrix also, no ?
+    Inertia inertia;
+  };
+
+  struct ModelGraphEdge
+  {
+    EIGEN_MAKE_ALIGNED_OPERATOR_NEW
```
jorisv:
```suggestion
```
SE3 doesn't need alignment (odd number)

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+      }
+
+      std::vector<EdgeDesc> * edges;
+    };
+
+  } // namespace internal
+
+  struct ModelGraphVertex
+  {
+    EIGEN_MAKE_ALIGNED_OPERATOR_NEW
```
jorisv:
```suggestion
```
`Inertia` doesn't need alignment

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+    void addBody(const std::string & vertex_name, const Inertia & inertia)
+    {
+      auto vertex_desc = boost::add_vertex(g);
+      ModelGraphVertex & vertex = g[vertex_desc];
+      vertex.name = vertex_name;
+      vertex.inertia = inertia;
+      if (!name_to_vertex.insert({vertex_name, vertex_desc}).second)
+      {
+        throw std::runtime_error("TODO");
+      }
```
jorisv: Check (`if (!name_to_vertex.insert({vertex_name, vertex_desc}).second)`) should be done before `boost::add_vertex` to avoid creating an invalid vertex.

You can use `name_to_vertex.find`.

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+      edge.out_to_joint = out_to_joint;
+      edge.joint_to_in = joint_to_in;
+
+      auto reverse_edge_desc = boost::add_edge(in_vertex->second, out_vertex->second, g);
+      if (!reverse_edge_desc.second)
+      {
+        throw std::runtime_error("TODO");
+      }
+      ModelGraphEdge & reverse_edge = g[reverse_edge_desc.first];
+      reverse_edge.name = joint_name + std::string("_reverse");
```
jorisv: We shouldn't add `_reverse` suffix to keep unique joint identifier.

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,370 @@
...
+        if (
+          const pinocchio::internal::JointRevoluteGraph * jrg =
+            boost::get<pinocchio::internal::JointRevoluteGraph>(&edge.joint))
+        {
+          if (jrg->flipped)
+          {
+            pinocchio::SE3 transf = pinocchio::SE3(
+              Eigen::AngleAxisd(M_PI, jrg->flip_axis).toRotationMatrix(), Eigen::Vector3d::Zero());
+            joint_pose.rotation() *= transf.rotation();
+            body_pose = transf * body_pose;
```
jorisv: When the joint is flipped I think the previous body to joint transformation should be changed.
But if I get it well, the joint to next body transformation is changed here.

unittest/model-graph.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+  ///////////////// Model
+  pinocchio::Model m = g.buildModel("body1", pinocchio::JointModelFreeFlyer());
+
+  BOOST_CHECK(m.jointPlacements[m.getJointId("body1_to_body2")].isApprox(
+    pinocchio::SE3(Eigen::Matrix3d::Identity(), Eigen::Vector3d(2, 0, 0))));
+  BOOST_CHECK(m.jointPlacements[m.getJointId("body2_to_body3")].isApprox(
+    pinocchio::SE3(Eigen::Matrix3d::Identity(), Eigen::Vector3d(4, -1, 0))));
+
+  pinocchio::Model m1 = g.buildModel("body3", pinocchio::JointModelFreeFlyer());
```
jorisv: We should check the jointPlacement also:

- Add comments to explains how jointPlacements are manually computed
- Use FK with manually done equivalent Q vector to test the equal X_0_i (with a m.X_0_i[b3] transformation)

A

unittest/model-graph.cpp
```diff
@@ -0,0 +1,189 @@
...
+  for (size_t i = 1; i < m.inertias.size(); i++)
+  {
+    BOOST_CHECK(m1.inertias[i].isApprox(m.inertias[m.inertias.size() - i]));
+  }
+}
+
+BOOST_AUTO_TEST_CASE(test_flipping_axis)
```
jorisv: You already tested axis flipping in the previous test since constructing the Model from body3 use the reversed joints

unittest/model-graph.cpp
```diff
@@ -0,0 +1,189 @@
...
+  g.addJoint(
+    "torso_to_left_leg",
+    pinocchio::internal::JointRevoluteGraph("torso_left_leg", Eigen::Vector3d::UnitX()), "torso",
+    pinocchio::SE3::Identity(), "left_leg", pinocchio::SE3::Identity());
+  g.addJoint(
+    "torso_to_right_leg",
+    pinocchio::internal::JointRevoluteGraph("torso_right_leg", Eigen::Vector3d::UnitX()), "torso",
+    pinocchio::SE3::Identity(), "right_leg", pinocchio::SE3::Identity());
+
+  g.buildModel("torso", pinocchio::JointModelFreeFlyer());
```
jorisv: To finish

include/pinocchio/multibody/model-graph.hpp
MegMll: Do you mean checking that the vertex was not already added to the graph ?

jorisv: Yes, here we add the vertex, then we check if the vertex is already in name_to_vertex.

include/pinocchio/multibody/model-graph.hpp
MegMll: put everything into the new visitor `addJointModel`

unittest/model-graph.cpp
MegMll: if you have an idea of what to test here, because, i'm not sure what would be the goal of this test

MegMll: Finished

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,794 @@
...
+    boost::recursive_wrapper<JointCompositeGraph>>;
+
+  struct JointCompositeGraph
+  {
+    std::vector<JointGraphVariant> joints;
+    std::vector<SE3> jointsPlacements;
+
+    JointCompositeGraph() = default;
+
+    JointCompositeGraph(const JointGraphVariant & j, const SE3 jPose)
```
jorisv:
```suggestion
    JointCompositeGraph(const JointGraphVariant & j, const SE3& jPose)
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,794 @@
...
+    {
+      joints.push_back(j);
+      jointsPlacements.push_back(jPose);
+    };
```
jorisv:
```suggestion
    }
```

Pas besoin de ; après une définition (à supprimer partout)

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,794 @@
...
+
+    JointCompositeGraph(const std::vector<JointGraphVariant> & js, const std::vector<SE3> jPoses)
```
jorisv:
```suggestion
    JointCompositeGraph(const std::vector<JointGraphVariant> & js, const std::vector<SE3>& jPoses)
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,794 @@
...
+    : joints(js)
+    , jointsPlacements(jPoses) {};
+
+    void addJoint(const JointGraphVariant & jm, const SE3 pose = SE3::Identity())
```
jorisv:
```suggestion
    void addJoint(const JointGraphVariant & jm, const SE3& pose = SE3::Identity())
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,794 @@
...
+
+    CreateJointModel(SE3 & joint_p, SE3 & body_p)
+    : joint_pose(joint_p)
+    , supported_pose(body_p)
+    {
+    }
+
+    ReturnType operator()(const JointFixedGraph & joint)
+    {
+      throw std::invalid_argument(
```
jorisv: Use `PINOCCHIO_THROW_PRETTY` or `PINOCCHIO_THROW`.

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,794 @@
...
+    typedef boost::
+      adjacency_list<boost::vecS, boost::vecS, boost::directedS, ModelGraphVertex, ModelGraphEdge>
+        Graph;
+    typedef typename boost::graph_traits<Graph>::vertex_descriptor VertexDesc;
+    typedef typename boost::graph_traits<Graph>::edge_descriptor EdgeDesc;
+
+    void addBody(const std::string & vertex_name, const Inertia & inertia)
+    {
+      if (name_to_vertex.find(vertex_name) != name_to_vertex.end())
+        throw std::runtime_error("Graph - vertex already in graph");
```
jorisv: Use PINOCCHIO_THROW_PRETTY or PINOCCHIO_THROW.

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,794 @@
...
+      const JointGraphVariant & joint,
+      const std::string & out_body,
+      const SE3 & out_to_joint,
+      const std::string & in_body,
+      const SE3 & joint_to_in)
+    {
+      auto out_vertex = name_to_vertex.find(out_body);
+      if (out_vertex == name_to_vertex.end())
+      {
+        throw std::runtime_error("Graph - out vertex does not exist");
```
jorisv: Use PINOCCHIO_THROW_PRETTY or PINOCCHIO_THROW.

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,794 @@
...
+      }
+      auto in_vertex = name_to_vertex.find(in_body);
+      if (in_vertex == name_to_vertex.end())
+      {
+        throw std::runtime_error("Graph - in vertex does not exist");
```
jorisv: Use PINOCCHIO_THROW_PRETTY or PINOCCHIO_THROW.

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,794 @@
...
+      }
+      auto edge_desc = boost::add_edge(out_vertex->second, in_vertex->second, g);
+
+      //
+      if (!edge_desc.second)
+      {
+        throw std::runtime_error("Graph - Edge cannot be added between these two vertexes");
```
jorisv: Use PINOCCHIO_THROW_PRETTY or PINOCCHIO_THROW.

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,794 @@
...
+      ModelGraphEdge & edge = g[edge_desc.first];
+      edge.name = joint_name;
+      edge.joint = joint;
+      edge.out_to_joint = out_to_joint;
+      edge.joint_to_in = joint_to_in;
+
+      auto reverse_edge_desc = boost::add_edge(in_vertex->second, out_vertex->second, g);
+      if (!reverse_edge_desc.second)
+      {
+        throw std::runtime_error("Graph - reverse edge cannot be added");
```
jorisv: Use PINOCCHIO_THROW_PRETTY or PINOCCHIO_THROW.

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,794 @@
...
+    Model buildModel(
+      const std::string & root_body,
+      const pinocchio::SE3 root_position,
+      const JointModel & root_joint,
+      const std::string & root_joint_name = "root_joint") const
+    {
+      auto root_vertex = name_to_vertex.find(root_body);
+      if (root_vertex == name_to_vertex.end())
+      {
+        throw std::runtime_error("Graph - root body does not exist in the graph");
```
jorisv: Use PINOCCHIO_THROW_PRETTY or PINOCCHIO_THROW.

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,794 @@
...
+        const ModelGraphVertex & target_vertex = g[target_vertex_desc];
+
+        AddJointModel visitor(source_vertex, target_vertex, edge, model);
+        boost::apply_visitor(visitor, edge.joint);
+      }
+      return model;
+    }
+
+    // When root is a fixed joint
+    Model buildModel(const std::string & root_body, const pinocchio::SE3 root_position) const
```
jorisv: We should find a way to not duplicate all this code.

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,794 @@
...
+      ModelGraphEdge & reverse_edge = g[reverse_edge_desc.first];
+      reverse_edge.name = joint_name;
+      reverse_edge.joint = boost::apply_visitor(ReverseJointVisitor(), joint);
+      reverse_edge.out_to_joint = joint_to_in.inverse();
+      reverse_edge.joint_to_in = out_to_joint.inverse();
+    }
+
+    Model buildModel(
+      const std::string & root_body,
+      const pinocchio::SE3 root_position,
```
jorisv:
```suggestion
      const pinocchio::SE3& root_position,
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,766 @@
...
+    }
+    ReturnType operator()(const JointMimicGraph & joint) const
+    {
+      return joint;
+    }
+    ReturnType operator()(const JointCompositeGraph & joint) const
+    {
+      JointCompositeGraph jReturn;
+      // Reverse joints
+      for (size_t i = joint.joints.size(); i-- > 0;)
```
jorisv: You decrement `i` in the control part of the loop. It's really uncommon and hard to read.

```cpp
for (int i = static_cast<int>(joint.joints.size()); i > 0; --i)
{
...
}
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,766 @@
...
+  {
+    JointFixedGraph() = default;
+  };
+
+  struct JointRevoluteGraph
+  {
+    // rotation axis
+    Eigen::Vector3d axis;
+
+    JointRevoluteGraph(const Eigen::Vector3d & ax)
```
jorisv:
```suggestion
    explicit JointRevoluteGraph(const Eigen::Vector3d & ax)
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,766 @@
...
+    : axis(ax)
+    {
+    }
+  };
+
+  struct JointRevoluteUnboundedGraph
+  {
+    Eigen::Vector3d axis;
+
+    JointRevoluteUnboundedGraph(const Eigen::Vector3d & ax)
```
jorisv:
```suggestion
    explicit JointRevoluteUnboundedGraph(const Eigen::Vector3d & ax)
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,766 @@
...
+    : axis(ax)
+    {
+    }
+  };
+
+  struct JointPrismaticGraph
+  {
+    Eigen::Vector3d axis;
+    JointPrismaticGraph(const Eigen::Vector3d & ax)
```
jorisv:
```suggestion
    explicit JointPrismaticGraph(const Eigen::Vector3d & ax)
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,766 @@
...
+  {
+    JointPlanarGraph() = default;
+  };
+
+  struct JointHelicalGraph
+  {
+    Eigen::Vector3d axis;
+    double pitch;
+
+    JointHelicalGraph(const Eigen::Vector3d & ax, const double & p)
```
jorisv:
```suggestion
    JointHelicalGraph(const Eigen::Vector3d & ax, double p)
```
Pas besoin de ref pour un type primitif.

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,766 @@
...
+
+    JointMimicGraph() = default;
+
+    JointMimicGraph(
+      const JointGraphVariant & jmodel_secondary,
+      const std::string & name_primary,
+      const double & scaling_,
+      const double & offset_)
+    : secondary_joint(jmodel_secondary)
+    , primary_name(name_primary)
```
jorisv:
```suggestion
    : primary_name(name_primary)
    , secondary_joint(jmodel_secondary)
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,766 @@
...
+    double scaling;
+    double offset;
+
+    JointMimicGraph() = default;
+
+    JointMimicGraph(
+      const JointGraphVariant & jmodel_secondary,
+      const std::string & name_primary,
+      const double & scaling_,
+      const double & offset_)
```
jorisv:
```suggestion
      double scaling_,
      double offset_)
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,766 @@
...
+  };
+
+  struct JointUniversalGraph
+  {
+    Eigen::Vector3d axis1;
+    Eigen::Vector3d axis2;
+
+    JointUniversalGraph(const Eigen::Vector3d & ax1, const Eigen::Vector3d & ax2)
+    : axis1(ax1)
+    , axis2(ax2) {};
```
jorisv:
```suggestion
    , axis2(ax2) {}
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,766 @@
...
+  };
+
+  struct JointHelicalGraph
+  {
+    Eigen::Vector3d axis;
+    double pitch;
+
+    JointHelicalGraph(const Eigen::Vector3d & ax, const double & p)
+    : axis(ax)
+    , pitch(p) {};
```
jorisv:
```suggestion
    , pitch(p) {}
```

include/pinocchio/multibody/model-graph.hpp
```diff
@@ -0,0 +1,766 @@
...
+#include <vector>
+#include <string>
+#include <iostream>
+#include <unordered_map>
+
+namespace pinocchio
+{
+  struct JointFixedGraph
+  {
+    JointFixedGraph() = default;
```
jorisv: Maybe `JointFixedGraph` should hold a static transformation.

This allow more configuration from the user and simplify the reduceModel algorithm.
Or we wait to manage the joint offset in a generic way (mujoco ref).

unittest/model-graph.cpp
```diff
@@ -0,0 +1,788 @@
...
+  g.addBody("body2", pinocchio::Inertia::Identity());
+
+  /////////////////////////////////////// Joints
+  g.addJoint(
+    "body1_to_body2", pinocchio::JointRevoluteGraph(Eigen::Vector3d::UnitZ()), "body1",
+    pinocchio::SE3(Eigen::Matrix3d::Identity(), Eigen::Vector3d(2., 0., 0.)), "body2",
+    pinocchio::SE3(Eigen::Matrix3d::Identity(), Eigen::Vector3d(2., 2., 0.)));
+
+  VertexDesc v_out = g.name_to_vertex["body1"];
+  VertexDesc v_in = g.name_to_vertex["body2"];
```
jorisv:
```suggestion
  auto v_out = g.name_to_vertex["body1"];
  auto v_in = g.name_to_vertex["body2"];
```

unittest/model-graph.cpp
```diff
@@ -0,0 +1,788 @@
...
+}
+
+/// @brief Test if edges and their reverse are added correctly to the graph
+BOOST_AUTO_TEST_CASE(test_add_edges)
+{
+  typedef boost::adjacency_list<
+    boost::vecS, boost::vecS, boost::directedS, pinocchio::ModelGraphVertex,
+    pinocchio::ModelGraphEdge>
+    Graph;
+  typedef typename boost::graph_traits<Graph>::vertex_descriptor VertexDesc;
```
jorisv:
```suggestion
```

unittest/model-graph.cpp
```diff
@@ -0,0 +1,788 @@
...
+    "body1", pinocchio::SE3::Identity(),
+    pinocchio::JointGraphVariant(pinocchio::JointFreeFlyerGraph()));
+
+  BOOST_CHECK(m.njoints == 3);
+  BOOST_CHECK(m.frames[m.getFrameId("body2_to_body3", pinocchio::FIXED_JOINT)].placement.isApprox(
+    pinocchio::SE3(Eigen::Matrix3d::Identity(), Eigen::Vector3d(4., -1., 0.))));
+}
+
+/// @brief test out reverse joint for revolute
+BOOST_AUTO_TEST_CASE(test_reverse_revolute)
```
jorisv: I'm a little lost with this test.
You construct two ModelGraph with a reversed joint, inversed static transform and reversed q.

I propose the following alternative:
```cpp
  pinocchio::ModelGraph g;
  //////////////////////////////////////// Bodies
  g.addBody("body1", pinocchio::Inertia::Identity());
  g.addBody(
    "body2",
    pinocchio::Inertia(4., pinocchio::Inertia::Vector3(0., 2., 0.), pinocchio::Symmetric3::Zero()));

  /////////////////////////////////////// Joints
  pinocchio::SE3 poseBody1 =
    pinocchio::SE3(Eigen::Matrix3d::Identity(), Eigen::Vector3d(2., 0., 0.));
  pinocchio::SE3 poseBody2 =
    pinocchio::SE3(Eigen::Matrix3d::Identity(), Eigen::Vector3d(0., 3., 0.));
  g.addJoint(
    "body1_to_body2", pinocchio::JointRevoluteGraph(Eigen::Vector3d::UnitX()), "body1", poseBody1,
    "body2", poseBody2);

  //////////////////////////////////// Forward model
  pinocchio::Model m_equi = g.buildModel("body1", pinocchio::SE3::Identity());
  pinocchio::Data d_equi(m_equi);
  Eigen::VectorXd q = Eigen::VectorXd::Zero(m_equi.nq);
  q[0] = M_PI / 2;
  pinocchio::framesForwardKinematics(m_equi, d_equi, q);

  //////////////////////////////////// Reverse model
  pinocchio::Model m_reverse =
    g.buildModel("body2", d_equi.oMf[m_equi.getFrameId("body2", pinocchio::BODY)]);
  pinocchio::Data d_reverse(m_reverse);
  pinocchio::framesForwardKinematics(m_reverse, d_reverse, -q);

  //////////////////////////////////// All bodies should be at the same configuration
  BOOST_CHECK(d_reverse.oMf[m_reverse.getFrameId("body1", pinocchio::BODY)].isApprox(
    d_equi.oMf[m_equi.getFrameId("body1", pinocchio::BODY)]));
  BOOST_CHECK(d_reverse.oMf[m_reverse.getFrameId("body2", pinocchio::BODY)].isApprox(
    d_equi.oMf[m_equi.getFrameId("body2", pinocchio::BODY)]));
```

Here we only reverse q. It's working for Revolute and Prismatic.
I first compute body2 frame and create a reversed model with root at this frame. Then, I have well reversed the joint I should body1 frame should be the same.

You can create a function to factorize the model creation code:

```cpp
ModelGraph buildReversableModelGraph(const JointGraphVariant & joint)
{
  pinocchio::ModelGraph g;
  //////////////////////////////////////// Bodies
  g.addBody("body1", pinocchio::Inertia::Identity());
  g.addBody(
    "body2",
    pinocchio::Inertia(4., pinocchio::Inertia::Vector3(0., 2., 0.), pinocchio::Symmetric3::Zero()));

  /////////////////////////////////////// Joints
  pinocchio::SE3 poseBody1 =
    pinocchio::SE3(Eigen::Matrix3d::Identity(), Eigen::Vector3d(2., 0., 0.));
  pinocchio::SE3 poseBody2 =
    pinocchio::SE3(Eigen::Matrix3d::Identity(), Eigen::Vector3d(0., 3., 0.));
  g.addJoint(
    "body1_to_body2", joint, "body1", poseBody1,
    "body2", poseBody2);
}
```

unittest/model-graph.cpp
```diff
@@ -0,0 +1,788 @@
...
+  pinocchio::Model m_backward = g.buildModel("body2", pinocchio::SE3::Identity());
+  pinocchio::Data d_b(m_backward);
+  ///////////////// Model
+  pinocchio::Model m_equi = g1.buildModel("body1", pinocchio::SE3::Identity());
+  pinocchio::Data d1(m_equi);
+
+  Eigen::VectorXd q = Eigen::VectorXd::Zero(m_backward.nq);
+  q[0] = 0.6;
+  q[1] = 0.1;
+  q[2] = -0.2;
```
jorisv: q2 and q0 should be swapped

include/pinocchio/multibody/model-graph.hpp
jorisv: This static transform can be a way to manage ref:

Model without ref:
```
J0       J1          J2
 x---------x----------x
```

Model with ref as a virtual FixedJoint transform.
```
J0       FixedJoint    J1      J2
 x-----------x----------x-------x
```

include/pinocchio/parsers/graph/frames.hpp
```diff
@@ -0,0 +1,48 @@
+//
+// Copyright (c) 2025 INRIA
+//
+
+#ifndef __pinocchio_frame_graph_hpp__
+#define __pinocchio_frame_graph_hpp__
```
jorisv: We must change all our guards since we move in parsers module:

`__pinocchio_parsers_graph_frames_hpp__`

include/pinocchio/parsers/graph/frames.hpp
```diff
@@ -0,0 +1,48 @@
...
+  };
+
+  struct OpFrameGraph
+  {
+    pinocchio::FrameType f_type = OP_FRAME;
+
+    OpFrameGraph() = default;
+  };
+
+  using FrameGraphVariant = boost::variant<BodyFrameGraph, SensorFrameGraph, OpFrameGraph>;
```
jorisv:
```suggestion
  typedef  boost::variant<BodyFrameGraph, SensorFrameGraph, OpFrameGraph> FrameGraphVariant;
```

I prefer using, but in pinocchio, typedef is used.

include/pinocchio/parsers/graph/graph-visitor.hpp
```diff
@@ -0,0 +1,408 @@
+//
+// Copyright (c) 2025 INRIA
+//
+
+#ifndef __pinocchio_graph_visitor_hpp__
```
jorisv: `__pinocchio_parsers_graph_graph_visitor_hpp__`

include/pinocchio/parsers/graph/graph-visitor.hpp
jorisv, line 1: If a visitor is only used in model-graph.cpp, you can move it to the .cpp file.
If a visitor can be used elsewhere, I think you can put it in the internal namespace.

include/pinocchio/parsers/graph/joints.hpp
```diff
@@ -0,0 +1,239 @@
+//
+// Copyright (c) 2025 INRIA
+//
+
+#ifndef __pinocchio_joint_graph_hpp__
```
jorisv: Update guard

include/pinocchio/parsers/graph/model-graph.hpp
```diff
@@ -0,0 +1,157 @@
+//
+// Copyright (c) 2025 INRIA
+//
+
+#ifndef __pinocchio_multibody_model_graph_hpp__
```
jorisv: update guard

include/pinocchio/parsers/graph/graph-visitor.hpp
```diff
@@ -0,0 +1,408 @@
...
+        previous_body.placement * joint_pose, edge.name);
+
+      model.addJointFrame(j_id);
+      model.appendBodyToJoint(j_id, b_f.inertia); // check this
+      model.addBodyFrame(target_vertex.name, j_id, body_pose);
+    }
+
+    void operator()(const JointFixedGraph & /*joint*/, const BodyFrameGraph & b_f)
+    {
+      if (source_vertex.frame.which() != 0) // body frame is index 0 in variant
```
jorisv: You can use boost::get to have something more robust:

```cpp
if(boost::get<BodyFrameGraph>(&source_vertex.frame) == nullptr)
```

jorisv: Also add a comment to explain this special case

include/pinocchio/parsers/graph/graph-visitor.hpp
```diff
@@ -0,0 +1,408 @@
...
+    , edge(edge_)
+    , model(model_)
+    {
+    }
+
+    template<typename JointGraph, typename FrameGraph>
+    void operator()(const JointGraph & /*joint*/, const FrameGraph & /*f_*/)
+    {
+      PINOCCHIO_THROW_PRETTY(
+        std::runtime_error, "Graph - Joint linking frame to model is not possible ?");
```
jorisv: Find a better message (at least, remove the `?`)

include/pinocchio/parsers/graph/graph-visitor.hpp
```diff
@@ -0,0 +1,408 @@
...
+        target_vertex.name, previous_body.parentJoint,
+        previous_body.placement * edge.out_to_joint * edge.joint_to_in, f_.f_type));
+    }
+
+    void operator()(const JointMimicGraph & joint, const BodyFrameGraph & b_f)
+    {
+      if (edge.reverse)
+        PINOCCHIO_THROW_PRETTY(std::runtime_error, "Graph - JointMimic cannot be reversed.");
+
+      if (source_vertex.frame.which() != 0) // body frame is index 0 in variant
```
jorisv: Use `boost::get` instead of which

src/parsers/graph/model-graph.cpp
```diff
@@ -0,0 +1,185 @@
+//
+// Copyright (c) 2025 INRIA
+//
+
+#include "pinocchio/parsers/graph/model-graph.hpp"
+#include "pinocchio/parsers/graph/graph-visitor.hpp"
+
+namespace pinocchio
+{
+  void ModelGraph::addFrame(const std::string & vertex_name, const FrameGraphVariant & frame)
```
jorisv: You can add the `addBody` helper function (with Inertia as parameter instead of the variant)

jorisv (changes requested):

include/pinocchio/parsers/graph/model-graph.hpp
```diff
@@ -0,0 +1,391 @@
...
+      /// \note Since it's a bidirectional graph, two edges are added to the graph.
+      /// Joints and transformation are inverted, to create reverse edge.
+      void addJoint(
+        const std::string & joint_name,
+        const JointVariant & joint,
+        const std::string & source_body,
+        const SE3 & source_to_joint,
+        const std::string & target_body,
+        const SE3 & joint_to_target,
+        const boost::optional<Eigen::VectorXd> & q_ref = boost::none);
```
jorisv:
```suggestion
      void addJoint(
        const std::string & joint_name,
        const JointVariant & joint,
        const std::string & source_body,
        const SE3 & source_to_joint,
        const std::string & target_body,
        const SE3 & joint_to_target);
```

On this version, only use the minimal parameter set.

include/pinocchio/parsers/graph/model-graph.hpp
```diff
@@ -0,0 +1,391 @@
...
+      ///
+      /// \param[in] params Structure that holds all of the joint parameters
+      ///
+      /// \note Since it's a bidirectional graph, two edges are added to the graph.
+      /// Joints and transformation are inverted, to create reverse edge.
+      void addJoint(const EdgeParameters & params);
+
+      /// \brief Create an EdgeBuilde. This will allow to use EdgeBuilder interface to have a more
+      /// flexible edge configuration.
+      EdgeBuilder useEdgeBuilder();
```
jorisv:
```suggestion
      EdgeBuilder edgeBuilder();
```

include/pinocchio/parsers/graph/model-graph.hpp
```diff
@@ -0,0 +1,391 @@
...
+      /// \param[in] frame which type of frame will be added to the model (op_frame, sensor, body)
+      void addFrame(const std::string & vertex_name, const FrameVariant & frame);
+
+      /// \brief Add a new body to the graph
+      ///
+      /// \param[in] vertex_name Name of the vertex
+      /// \param[in] inert inertia of the body
+      void addBody(const std::string & vertex_name, const Inertia & inert);
+
+      GeometryBuilder useGeometryBuilder();
```
jorisv:
```suggestion
      GeometryBuilder geometryBuilder();
```

src/parsers/graph/model-graph.cpp
```diff
@@ -0,0 +1,277 @@
...
+      friction.tail(nv) = range.friction;
+
+      armature.conservativeResize(armature.size() + nv);
+      armature.tail(nv) = range.armature;
+    }
+
+    EdgeBuilder & EdgeBuilder::withJointType(const JointVariant & jtype)
+    {
+      param.joint = jtype;
+      param.jlimit = boost::apply_visitor(internal::MakeJointLimitsDefaultVisitor(), jtype);
```
jorisv: You should do that in build:
- If you call `withMinConfig` and then `withJointType` you will erase the withMinConfig call.

include/pinocchio/parsers/graph/model-graph.hpp
```diff
@@ -0,0 +1,391 @@
...
+      ///
+      /// \param[in] vertex_name Name of the vertex
+      /// \param[in] frame which type of frame will be added to the model (op_frame, sensor, body)
+      void addFrame(const std::string & vertex_name, const FrameVariant & frame);
+
+      /// \brief Add a new body to the graph
+      ///
+      /// \param[in] vertex_name Name of the vertex
+      /// \param[in] inert inertia of the body
+      void addBody(const std::string & vertex_name, const Inertia & inert);
```
jorisv, line 110: Both method can return a `ModelGraphVertex`. So user can directly add Geometry inside the vertex.

jorisv (changes requested):

include/pinocchio/parsers/graph/geometries.hpp
```diff
@@ -0,0 +1,128 @@
...
+#define __pinocchio_parsers_geometries_graph_hpp__
+
+#include "pinocchio/spatial/se3.hpp"
+
+namespace pinocchio
+{
+  namespace graph
+  {
+    // Kind of possible geometry
+    enum GEOM_TYPE
```
jorisv:
```suggestion
    enum struct GeomType
```
Enum are in CamelCase.
We can use enum struct to avoid enum to leaking inside the namespace (we will have to allway do `GeomType::VISUAL`)

include/pinocchio/parsers/graph/geometries.hpp
```diff
@@ -0,0 +1,128 @@
...
+      placement(placement),
+      scale(scale),
+      color(color),
+      type(type),
+      geometry(geom)
+      {}
+    };
+  
+
+    struct GeometryParameters
```
jorisv: This structure is only useful if a method take it as an argument.
Since we copy `Geometry` inside `ModelGraphVertex` without any modification, I think this structure use useless. We can directly store a Geometry inside `GeometryBuilder`.

include/pinocchio/parsers/graph/model-graph.hpp
MegMll: if we want the changes to be applied to the graph, we need to return `ModelGraphVertex &`. 

I don't know if it's very safe, but if that's fine by you, i can make the changes.

MegMll: It would also mean exposing `ModelGraphVertex` in python which we are not currently doing

jorisv: You right, let's forget this change request

include/pinocchio/parsers/graph/graph-visitor.hpp
```diff
@@ -0,0 +1,901 @@
...
+#include <utility>
+#include <stdexcept>
+
+namespace pinocchio
+{
+  namespace graph
+  {
+    namespace internal
+    {
+      struct MakeJointLimitsDefaultVisitor : public boost::static_visitor<JointLimits>
```
jorisv: I think some Visitor can be moved to model-graph.cpp or model-graph-algo.cpp.

When you move it, no need to put them in pinocchio::graph::internal namespace, you can use pinocchio::graph::{anonymous} instead.

include/pinocchio/parsers/graph/graph-visitor.hpp
```diff
@@ -0,0 +1,901 @@
...
+        JointLimits operator()(const Joint &) const
+        {
+          JointLimits jlimit_return = jlimit;
+          jlimit_return.maxConfig = -jlimit.minConfig;
+          jlimit_return.minConfig = -jlimit.maxConfig;
+
+          return jlimit_return;
+        }
+
+        // For freeFlyer = no changes, except for translation limits.
```
jorisv: Is this comment still relevant ?

By looking into `joint-free-flyer.hpp` I saw the following code:

```cpp
    const std::vector<bool> hasConfigurationLimit() const
    {
      return {true, true, true, false, false, false, false};
    }

    const std::vector<bool> hasConfigurationLimitInTangent() const
    {
      return {true, true, true, false, false, false};
    }
```

include/pinocchio/parsers/graph/graph-visitor.hpp
```diff
@@ -0,0 +1,901 @@
...
+          for (int i = 0; i < j.nv; i++)
+          {
+            jlimit_return.maxEffort[i] = jlimit.maxEffort[j.nv - 1 - i];
+            jlimit_return.maxVel[i] = jlimit.maxVel[j.nv - 1 - i];
+
+            jlimit_return.friction[i] = jlimit.friction[j.nv - 1 - i];
+            jlimit_return.damping[i] = jlimit.damping[j.nv - 1 - i];
+
+            jlimit_return.armature[i] = jlimit.armature[j.nv - 1 - i];
+          }
```
jorisv: The comment explain well what you're doing. But a think using std::swap instead of doing this way is even clearer.

include/pinocchio/parsers/graph/model-graph.hpp
```diff
@@ -0,0 +1,395 @@
...
+      JointLimits jlimit;
+
+      boost::optional<Eigen::VectorXd> minConfig;
+      boost::optional<Eigen::VectorXd> maxConfig;
+      boost::optional<Eigen::VectorXd> maxVel;
+      boost::optional<Eigen::VectorXd> maxEffort;
+      boost::optional<Eigen::VectorXd> armature;
+      boost::optional<Eigen::VectorXd> friction;
+      boost::optional<Eigen::VectorXd> damping;
+      double frictionLoss = 0;
```
jorisv: This is only modified in the EdgeBuilder so it should be stored in EdgeBuilder.

EdgeParameters should only contains data used by ModelGraph::addJoint.

src/parsers/graph/model-graph.cpp
```diff
@@ -0,0 +1,308 @@
...
+        boost::apply_visitor(internal::ReverseJointGraphVisitor(), params.joint);
+      reverse_edge.joint = reversed_joint.first;
+      if (params.q_ref)
+      {
+        const Eigen::VectorXd q_ref_reverse =
+          boost::apply_visitor(internal::ReverseQVisitor(*params.q_ref), params.joint);
+        reverse_edge.source_to_joint =
+          params.joint_to_target.inverse()
+          * boost::apply_visitor(
+            internal::UpdateJointGraphPoseVisitor(q_ref_reverse), reverse_edge.joint);
```
jorisv: Are the following calls equivalent ?

```cpp
boost::apply_visitor(internal::UpdateJointGraphPoseVisitor(q_ref_reverse), reverse_edge.joint)
```
```cpp
boost::apply_visitor(internal::UpdateJointGraphPoseVisitor(*params.q_ref), edge.joint).inverse()
```

src/parsers/graph/model-graph-algo-geometry.cpp
```diff
@@ -0,0 +1,122 @@
...
+        GeometryObject geometry_object(
+          geom.name, body_frame.parentJoint, f_id, body_frame.placement * geom.placement, g, "",
+          geom.scale, true, geom.color);
+
+        model.addGeometryObject(geometry_object);
+      }
+
+      void operator()(const Sphere & s)
+      {
+        addGeometry(std::shared_ptr<fcl::CollisionGeometry>(new fcl::Sphere(s.radius)));
```
jorisv: use `std::make_shared<fcl::Sphere>(s.radius)`

Do the same for all other shared_ptr construction.

jorisv (changes requested):

src/parsers/graph/model-graph.cpp
MegMll: They are except for jointComposite, where joint positions are modified internally when it's reversed

include/pinocchio/parsers/graph/graph-visitor.hpp
MegMll: Yes sorry, i put the comment and then forgot to change the function.

include/pinocchio/parsers/graph/graph-visitor.hpp
jorisv: Can you check if we can move:
ReverseJointGraphVisitor => model-graph.cpp
CreateJointModelVisitor => model-graph-algo.cpp
AddJointModelVisitor => model-graph-algo.cpp
RecordTreeEdgeVisitor => model-graph-algo.cpp

bindings/python/parsers/graph/expose-geometries.cpp
```diff
@@ -0,0 +1,137 @@
...
+      bp::class_<pinocchio::graph::Sphere>("Sphere", "Represents a sphere.", bp::init<>())
+        .def(bp::init<const double>((bp::arg("radius")), "Constructor from radius."))
+        .def_readwrite("radius", &pinocchio::graph::Sphere::radius, "Radius of the sphere.");
+
+      bp::to_python_converter<pinocchio::graph::GeomVariant, VariantToPythonVisitor<GeomVariant>>();
+      bp::implicitly_convertible<pinocchio::graph::Mesh, pinocchio::graph::GeomVariant>();
+      bp::implicitly_convertible<pinocchio::graph::Box, pinocchio::graph::GeomVariant>();
+      bp::implicitly_convertible<pinocchio::graph::Cylinder, pinocchio::graph::GeomVariant>();
+      bp::implicitly_convertible<pinocchio::graph::Capsule, pinocchio::graph::GeomVariant>();
+      bp::implicitly_convertible<pinocchio::graph::Sphere, pinocchio::graph::GeomVariant>();
```
jorisv: You can use the [EigenPy VariantConverter](https://github.com/stack-of-tasks/eigenpy/blob/devel/include/eigenpy/variant.hpp#L337)

bindings/python/parsers/graph/expose-edges.cpp
```diff
@@ -0,0 +1,332 @@
...
+      bp::implicitly_convertible<JointPrismatic, JointVariant>();
+      bp::implicitly_convertible<JointFreeFlyer, JointVariant>();
+      bp::implicitly_convertible<JointSpherical, JointVariant>();
+      bp::implicitly_convertible<JointSphericalZYX, JointVariant>();
+      bp::implicitly_convertible<JointTranslation, JointVariant>();
+      bp::implicitly_convertible<JointPlanar, JointVariant>();
+      bp::implicitly_convertible<JointHelical, JointVariant>();
+      bp::implicitly_convertible<JointUniversal, JointVariant>();
+      bp::implicitly_convertible<JointComposite, JointVariant>();
+      bp::implicitly_convertible<JointMimic, JointVariant>();
```
jorisv: You can use the [EigenPy VariantConverter](https://github.com/stack-of-tasks/eigenpy/blob/devel/include/eigenpy/variant.hpp#L337)

bindings/python/parsers/graph/expose-frames.cpp
```diff
@@ -0,0 +1,68 @@
...
+        bp::init<>(bp::args("self"), "Default constructor."))
+        .def_readwrite(
+          "f_type", &OpFrame::f_type,
+          "Type of the frame (should be pinocchio.FrameType.OP_FRAME).");
+
+      bp::to_python_converter<FrameVariant, VariantToPythonVisitor<FrameVariant>>();
+
+      bp::implicitly_convertible<BodyFrame, FrameVariant>();
+      bp::implicitly_convertible<SensorFrame, FrameVariant>();
+      bp::implicitly_convertible<OpFrame, FrameVariant>();
```
jorisv: You can use the [EigenPy VariantConverter](https://github.com/stack-of-tasks/eigenpy/blob/devel/include/eigenpy/variant.hpp#L337)

sources.cmake
```diff
@@ -348,7 +351,18 @@ set(${PROJECT_NAME}_PARSERS_PUBLIC_HEADERS
     ${PROJECT_SOURCE_DIR}/include/pinocchio/parsers/mjcf/model.hxx
     ${PROJECT_SOURCE_DIR}/include/pinocchio/parsers/mjcf/geometry.hxx
     ${PROJECT_SOURCE_DIR}/include/pinocchio/parsers/mjcf/mjcf-graph.hpp
-    ${PROJECT_SOURCE_DIR}/include/pinocchio/parsers/sample-models.hpp)
+    ${PROJECT_SOURCE_DIR}/include/pinocchio/parsers/sample-models.hpp
+    ${PROJECT_SOURCE_DIR}/include/pinocchio/parsers/graph/model-graph.hpp
```
jorisv, line 356:
```suggestion
    ${PROJECT_SOURCE_DIR}/include/pinocchio/parsers/graph/fwd.hpp
    ${PROJECT_SOURCE_DIR}/include/pinocchio/parsers/graph/model-graph.hpp
```

CHANGELOG.md
```diff
@@ -6,6 +6,21 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+### Added
+- Add `pinocchio::graph::ModelGraph` class
+  - Simple API to build a model and a geometryModel
+  - Manage joint offset
+  - Create a `pinocchio::Model` with any frame as root frame
+  - Create a `pinocchio::Model` with any joint as root joint
+  - Reliably merge two models with `pinocchio::graph::mergeGraphs` function
+  - Reliably fix joints at any configuration with `pinocchio::graph::fixJointsGraph`
+  - [C++ example](./examples/model-graph.cpp)
+  - [Python example](./examples/model-graph.py)
```
jorisv, line 23: Add model-graph-geometry examples

jorisv (dismissed):

jorisv (approved):

jorisv (approved):

jcarpent (approved):

PR #2666, opened 2025-04-29
Change default branch

jorisv:
In our past practices, we had two main branches:
- `devel`: work in progress
- `master`: stable release (default)

By default, user will clone `master`, the stable and tested version.

This workflow is no more mandatory:
- Our CI has improved, `devel` is more stable than before
- Most user doesn't compile from sources and use packages instead (pip, conda, nix, apt, ...)

Given these changes and the fact that several tools assume the default branch to be the work in progress one, we have decided to change the default branch to `devel`.

This means that when you clone our repository, you will now be working with the ongoing development version by default.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -6,25 +6,21 @@ on:
       - doc/**
       - .gitlab-ci.yml
       - .gitignore
-      - '*.md'
+      - '**.md'
       - CITATION.*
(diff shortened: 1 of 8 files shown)


hrp2-14:

ManifoldFR (approved):
Looks good to me. Might need someone else to take a second look and confirm 😃

PR #2680, opened 2025-05-20
Remove support of 20.04 in linux CI

jcarpent:
And update the cmake module to the latest version.


nim65s (dismissed):

jorisv (approved):

jorisv (changes requested):
There is some `# TODO: Remove quote when 20.04 is not supported` in the code.
I will remove them.

jorisv (dismissed):

jorisv (approved):

PR #2686, opened 2025-05-26
CMake: add COAL_DISABLE_HPP_FCL_WARNINGS

nim65s:
Because it is not linked to hpp-fcl::hpp-fcl.

Ref. https://github.com/coal-library/coal/pull/709

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -191,13 +191,12 @@ pinocchio_specific_type(default DEFAULT_SCOPE)
 # some are template instantiated, or some user can link only on pinocchio_default, we muste define
 # PINOCCHIO_WITH_HPP_FCL in pinocchio_default.
 if(BUILD_WITH_HPP_FCL_SUPPORT)
-  target_compile_definitions(pinocchio_default PUBLIC PINOCCHIO_WITH_HPP_FCL)
+  target_compile_definitions(
+    pinocchio_default
+    PUBLIC PINOCCHIO_WITH_HPP_FCL
+    PRIVATE COAL_DISABLE_HPP_FCL_WARNINGS)
   target_include_directories(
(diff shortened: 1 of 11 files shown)


src/CMakeLists.txt
```diff
@@ -194,6 +194,9 @@ if(BUILD_WITH_HPP_FCL_SUPPORT)
   target_compile_definitions(pinocchio_default PUBLIC PINOCCHIO_WITH_HPP_FCL)
   target_include_directories(
     pinocchio_default PUBLIC $<TARGET_PROPERTY:hpp-fcl::hpp-fcl,INTERFACE_INCLUDE_DIRECTORIES>)
+  if(COAL_DISABLE_HPP_FCL_WARNINGS)
```
jcarpent: Should we add an option within the CMake for `COAL_DISABLE_HPP_FCL_WARNINGS`?

nim65s: Our options include:
- add this as an option in pinocchio as you suggest
- use this PR as it
- remove the `if(COAL_DISABLE_HPP_FCL_WARNINGS)` condition and always disable this warning in pinocchio
- use `target_link_libraries(pinocchio_default PUBLIC hpp-fcl::hpp-fcl)`

I don't have a strong opinion on this.

jorisv: One issue I see is user that use hpp-flc and pinocchio will never see the warning since the define will be set globally.
It's maybe better to let end-user add the `COAL_DISABLE_HPP_FCL_WARNINGS` define instead of forwarding it in the target ? I also raise the same issue with the coal PR

nim65s: done

jorisv (dismissed):

jorisv (approved):

jcarpent (approved):

PR #2714, opened 2025-06-24
CMake:  add BUILD_STANDALONE_PYTHON_INTERFACE option

nim65s:
Follows https://github.com/coal-library/coal/pull/658

Still not tested enough, but it already seems good in multiple tests


examples/CMakeLists.txt
```diff
@@ -47,28 +47,30 @@ function(ADD_PINOCCHIO_CPP_EXAMPLE EXAMPLE)
...
-  add_pinocchio_cpp_example(build-reduced-model PARSERS)
-  add_pinocchio_cpp_example(geometry-models PARSERS)
-  add_pinocchio_cpp_example(kinematics-derivatives PARSERS)
-  add_pinocchio_cpp_example(forward-dynamics-derivatives PARSERS)
-  add_pinocchio_cpp_example(inverse-dynamics-derivatives PARSERS)
-  if(BUILD_ADVANCED_TESTING)
-    add_pinocchio_cpp_example(multiprecision PARSERS)
-  endif()
-endif()
+if(NOT BUILD_STANDALONE_PYTHON_INTERFACE)
```
jcarpent: I would suggest deactivaing the macro `add_pinocchio_cpp_example`, otherwise, from a logic viewpoint, it is hard to follow.

CMakeLists.txt
```diff
@@ -446,10 +450,14 @@ function(ADD_WINDOWS_DLL_PATH_TO_TEST TEST_NAME)
 endfunction()
 
 # --- MAIN LIBRARY -------------------------------------------------------------
-add_subdirectory(src)
+if(NOT BUILD_STANDALONE_PYTHON_INTERFACE)
```
jcarpent, line 471: I would add a variable called: BUILD_MAIN_PROJECT_AND_ASSOCIATED_OUPUTS

CMakeLists.txt
jorisv, line 1: Python public headers are installed the following way:

```cmake
if(BUILD_PYTHON_INTERFACE)
  list(APPEND ${PROJECT_NAME}_HEADERS ${${PROJECT_NAME}_BINDINGS_PYTHON_PUBLIC_HEADERS})
endif()
```

If we split the pinocchio packaging (c++ lib and python binding), bindings headers would be never installed (BUILD_PYTHON_INTERFACE=OFF in the c++ lib and `${PROJECT_NAME}_HEADERS` is not used when BUILD_STANDALONE_PYTHON_INTERFACE=ON).
I think we must install them in the c++ lib package, so we should drop the if.

jorisv (changes requested):

examples/CMakeLists.txt
jorisv: I agree, most of cpp unit test and examples are created from a macro. It would simplify the code a lot.

jorisv (changes requested):

CMakeLists.txt
nim65s: Do you mean something like this ?
```cmake
set(BUILD_MAIN_PROJECT_AND_ASSOCIATED_OUPUTS (NOT BUILD_STANDALONE_PYTHON_INTERFACE))
```

jcarpent: Yes

jorisv (dismissed):

jorisv (dismissed):

jorisv (dismissed):

jorisv (approved):

ManifoldFR:
finally merged, good work @nim65s @jorisv 🚀

PR #2718, opened 2025-06-29
Beta version of Viser visualizer

sea-bass:
I've been enjoying the [Viser](https://github.com/nerfstudio-project/viser) visualizer quite a lot, so I decided to make an "official" Pinocchio model visualizer for it.

It's like Meshcat, except it's more modern, currently maintained, and has a lot more built-in interactive widgets!

It's still in early stages, but wanted to get it in front of the maintainers!

https://github.com/user-attachments/assets/2dcb00ae-ef8a-4ab5-9e83-e960d2d91403

Usage wise, it should behave quite similarly to meshcat:

```python
from pinocchio.visualize import ViserVisualizer

# ... make your Pinocchio model

viz = ViserVisualizer(model, collision_model, visual_model, data=data)
viz.initViewer(open=True, loadModel=True)
viz.displayCollisions(True)
viz.displayVisuals(True)
viz.displayFrames(True)

viz.display(pinocchio.neutral(model))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,7 @@ Please note that we always advise including the `pinocchio/fwd.hpp` header as th
 -   [Meshcat](https://github.com/rdeits/meshcat): supporting visualization in Python and which can be embedded inside any browser.
 -   [Panda3d](https://github.com/ikalevatykh/panda3d_viewer): supporting visualization in Python and which can be embedded inside any browser.
 -   [RViz](https://github.com/ros-visualization/rviz): supporting visualization in Python and which can interact with other ROS packages.
+-   [Viser](https://github.com/nerfstudio-project/viser): supporting browser-based visualization in Python, with support for widgets such as sliders and interactive markers.
 
 Many external viewers can also be integrated. For more information, see the example [here](https://github.com/stack-of-tasks/pinocchio/blob/devel/bindings/python/pinocchio/visualize/base_visualizer.py).
 
@@ -275,7 +276,7 @@ In addition to the core dev team, the following people have also been involved i
 -   [Sarah El Kazdadi](https://github.com/sarah-ek) (Inria): multi-precision arithmetic support
 -   [Nicolas Torres Alberto](https://scholar.google.com/citations?user=gYNLhEIAAAAJ&hl=en) (Inria): features extension
 -   [Shubham Singh](https://github.com/shubhamsingh91) (UT Austin): second-order inverse dynamics derivatives
--   [Sebastian Castro](https://roboticseabass.com) (The AI Institute): MeshCat viewer feature extension
+-   [Sebastian Castro](https://roboticseabass.com) (RAI Institute): Viser visualizer and MeshCat visualizer feature extension
 -   [Lev Kozlov](https://github.com/lvjonok): Kinetic and potential energy regressors
 -   [Simeon Nedelchev](https://github.com/simeon-ned): Pseudo inertia and Log-Cholesky parametrization
 
--- a/bindings/python/pinocchio/visualize/viser_visualizer.py
+++ b/bindings/python/pinocchio/visualize/viser_visualizer.py
@@ -199,21 +199,55 @@ class ViserVisualizer(BaseVisualizer):
                 opacity=color_override[3],
             )
         elif isinstance(geom, MESH_TYPES):
-            mesh = trimesh.load(geometry_object.meshPath)
-            if color is None:
-                frame = self.viewer.scene.add_mesh_trimesh(name, mesh)
+            frame = self._add_mesh_from_path(name, geometry_object.meshPath, color_override)
+        elif isinstance(geom, hppfcl.Convex):
+            if len(geometry_object.meshPath) > 0:
+                frame = self._add_mesh_from_path(name, geometry_object.meshPath, color_override)
             else:
-                frame = self.viewer.scene.add_mesh_simple(
-                    name,
-                    mesh.vertices,
-                    mesh.faces,
-                    color=color_override[:3],
-                    opacity=color_override[3],
-                )
+                frame = self._add_mesh_from_convex(name, geom, color_override)
         else:
             raise RuntimeError(f"Unsupported geometry type for {name}: {type(geom)}")
 
         self.frames[name] = frame
+
+    def _add_mesh_from_path(self, name, mesh_path, color):
+        """
+        Load a mesh from a file.
+        """
+        mesh = trimesh.load(mesh_path)
+        if color is None:
+            return self.viewer.scene.add_mesh_trimesh(name, mesh)
+        else:
+            return self.viewer.scene.add_mesh_simple(
+                name,
+                mesh.vertices,
+                mesh.faces,
+                color=color[:3],
+                opacity=color[3],
+            )
+
+    def _add_mesh_from_convex(self, name, geom, color):
+        """
+        Load a mesh from triangles stored inside a hppfcl.Convex.
+        """
+        num_tris = geom.num_polygons
+        call_triangles = geom.polygons
+        call_vertices = geom.points
+
+        vertices = call_vertices()
+        vertices = vertices.astype(np.float32)
+        faces = np.empty((num_tris, 3), dtype=int)
+        for k in range(num_tris):
+            tri = call_triangles(k)
+            faces[k] = [tri[i] for i in range(3)]
+
+        return self.viewer.scene.add_mesh_simple(
+            name,
+            vertices,
+            faces,
+            color=color[:3],
+            opacity=color[3],
+        )
 
     def display(self, q=None):
         """
--- a/bindings/python/pinocchio/visualize/viser_visualizer.py
+++ b/bindings/python/pinocchio/visualize/viser_visualizer.py
@@ -326,8 +326,8 @@ class ViserVisualizer(BaseVisualizer):
         Capture an image from the Viser viewer and return an RGB array.
 
         Parameters:
-            w: The width of the captured image.
-            h: The height of the captured image.
+            w: The width of the captured image. If None, uses the actual camera width.
+            h: The height of the captured image. If None, uses the actual camera height.
             client_id: The ID of the Viser client handle.
                 If None, uses the first available client.
             transport_format: The transport format to use for the captured image.
@@ -346,7 +346,11 @@ class ViserVisualizer(BaseVisualizer):
         else:
             cli = clients[client_id]
 
-        return cli.get_render(height=h, width=w, transport_format=transport_format)
+        height = h or cli.camera.image_height
+        width = w or cli.camera.image_width
+        return cli.get_render(
+            height=height, width=width, transport_format=transport_format
+        )
 
     def setBackgroundColor(self, preset_name: str = "gray", col_top=None, col_bot=None):
         raise NotImplementedError("setBackgroundColor is not yet implemented.")
--- a/bindings/python/pinocchio/visualize/viser_visualizer.py
+++ b/bindings/python/pinocchio/visualize/viser_visualizer.py
@@ -1,0 +1,332 @@
+import time
+
+try:
+    import hppfcl
+except ImportError:
+    raise ImportError("hppfcl not found, but it is currently required by this viewer.")
+
+import numpy as np
+
+from .. import pinocchio_pywrap_default as pin
+from . import BaseVisualizer
+
+try:
+    import trimesh  # Required by viser
+    import viser
+except ImportError:
+    import_viser_succeed = False
+else:
+    import_viser_succeed = True
+
+
+MESH_TYPES = (hppfcl.BVHModelBase, hppfcl.HeightFieldOBBRSS, hppfcl.HeightFieldAABB)
+
+
+class ViserVisualizer(BaseVisualizer):
+    """A Pinocchio visualizer using Viser."""
+
+    def __init__(
+        self,
+        model=pin.Model(),
+        collision_model=None,
+        visual_model=None,
+        copy_models=False,
+        data=None,
+        collision_data=None,
+        visual_data=None,
+    ):
+        if not import_viser_succeed:
+            msg = (
+                "Error while importing the viewer client.\n"
+                "Check whether viser is properly installed "
+                "(pip install --user viser)."
+            )
+            raise ImportError(msg)
+
+        super().__init__(
+            model,
+            collision_model,
+            visual_model,
+            copy_models,
+            data,
+            collision_data,
+            visual_data,
+        )
+        self.static_objects = []
+
+    def initViewer(
+        self,
+        viewer=None,
+        open=False,
+        loadModel=False,
+        host="localhost",
+        port="8000",
+    ):
+        """
+        Start a new Viser server and client.
+        Note: the server can also be passed in through the `viewer` argument.
+
+        Parameters:
+            viewer: An existing ViserServer instance, or None.
+                If None, creates a new ViserServer in this visualizer.
+            open: If True, automatically opens a browser tab to the visualizer.
+            loadModel: If True, loads the Pinocchio models passed in.
+            host: If `viewer` is None, this will be the host URL.
+            port: If `viewer` is None, this will be the host port.
+        """
+        if (viewer is not None) and (not isinstance(viewer, viser.ViserServer)):
+            raise RuntimeError(
+                "'viewer' argument must be None or a valid ViserServer instance."
+            )
+
+        self.viewer = viewer or viser.ViserServer(host=host, server_port=port)
+        self.frames = {}
+
+        if open:
+            import webbrowser
+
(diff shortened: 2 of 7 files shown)


sea-bass:
@brentyi FYI -- in case this is also of interest to the Viser user community, thought it would be good to know this is in the works.

jcarpent (dismissed):
Fantastic contributions @sea-bass.
We do appreciate it.

sea-bass:
> Fantastic contributions @sea-bass. We do appreciate it.

Thanks! Let me do some more testing with other geometries/models this week and I will let you know when it's in good shape.

README.md
```diff
@@ -279,7 +279,7 @@ In addition to the core dev team, the following people have also been involved i
 -   [Sarah El Kazdadi](https://github.com/sarah-ek) (Inria): multi-precision arithmetic support
 -   [Nicolas Torres Alberto](https://scholar.google.com/citations?user=gYNLhEIAAAAJ&hl=en) (Inria): features extension
 -   [Shubham Singh](https://github.com/shubhamsingh91) (UT Austin): second-order inverse dynamics derivatives
--   [Sebastian Castro](https://roboticseabass.com) (The AI Institute): MeshCat viewer feature extension
+-   [Sebastian Castro](https://roboticseabass.com) (RAI Institute): Viser visualizer and MeshCat viewer feature extension
```
sea-bass: Note to self: update the README in the visualization section

sea-bass:
Ready for review now! Added Spheres and Cylinder geometries, updated README, and fixed one other bug.

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -0,0 +1,290 @@
...
+            # Get frame pose.
+            M = self.data.oMf[self.model.getFrameId(frame.name)]
+
+            # Update viewer configuration.
+            frame_name = self.framesRootNodeName + "/" + frame.name
+            frame = self.frames[frame_name]
+            frame.position = M.translation
+            frame.wxyz = pin.Quaternion(M.rotation).coeffs()[
+                [3, 0, 1, 2]
+            ]  # Pinocchio uses xyzw
```
ManifoldFR: Please do not name this variable `frame` as it is your for-loop variable (for code readability). I think Codacy might've flagged this but noting it here just in case.

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -0,0 +1,290 @@
...
+            frame.position = M.translation * geometry_object.meshScale
+            frame.wxyz = pin.Quaternion(M.rotation).coeffs()[
+                [3, 0, 1, 2]
+            ]  # Pinocchio uses xyzw
+
+    def updateFrames(self):
+        pin.updateFramePlacements(self.model, self.data)
+        for frame in self.model.frames:
+            # Get frame pose.
+            M = self.data.oMf[self.model.getFrameId(frame.name)]
```
ManifoldFR: Prefer to iterate as
```python
for i, frame in enumerate(self.model.frames):
```
to avoid querying the frame ID by its name and occurring a lookup for every frame for every call to this function.

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -0,0 +1,290 @@
...
+            frame_name = self.framesRootNodeName + "/" + frame.name
+            frame = self.frames[frame_name]
+            frame.position = M.translation
+            frame.wxyz = pin.Quaternion(M.rotation).coeffs()[
+                [3, 0, 1, 2]
+            ]  # Pinocchio uses xyzw
+
+    def captureImage(self, w=None, h=None):
+        """Capture an image from the Viser viewer and return an RGB array."""
+        raise NotImplementedError("captureImage is not yet implemented.")
```
ManifoldFR: Does Viser offer support for capturing images from the viewport?

I added that to meshcat a few years ago (it's still stuck on the master branch because there's been no release since then) and it's been rather helpful. If it's planned for Viser, we can put this on the informal roadmap for this visualizer.

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -0,0 +1,290 @@
...
+            geom_model = self.visual_model
+            geom_data = self.visual_data
+            prefix = self.viewerRootNodeName + "/visual"
+        else:
+            geom_model = self.collision_model
+            geom_data = self.collision_data
+            prefix = self.viewerRootNodeName + "/collision"
+
+        pin.updateGeometryPlacements(self.model, self.data, geom_model, geom_data)
+        for geometry_object in geom_model.geometryObjects:
```
ManifoldFR: Same remark as below regarding the iteration and avoiding a name lookup.

bindings/python/pinocchio/visualize/viser_visualizer.py
sea-bass: Seems like it does!

https://github.com/nerfstudio-project/viser/blob/06c0c34e25a0d593737ac0d60d5dcaa92fc07896/examples/03_interaction/02_get_renders.py

bindings/python/pinocchio/visualize/viser_visualizer.py
sea-bass: I think it's a good idea regardless, and will implement, but Pinocchio looks up frame IDs in constant time... right?

ManifoldFR: I'm not [sure](https://github.com/stack-of-tasks/pinocchio/blob/212311887a74d4042e7e2986c945938e0ca26d11/include/pinocchio/multibody/model.hxx#L470), depends on the complexity of `std::find_if`.

sea-bass: Oh wow, this whole time I was under the impression that there was an underlying map from frame name to ID 😰 -- thanks for sharing! I may have some other code to go fix.

I guess because there is an extra frame type argument, it may actually make more sense to have implemented it this way and incurring the search time.

bindings/python/pinocchio/visualize/viser_visualizer.py
sea-bass: @ManifoldFR I added this (one-line) change to this PR.

ManifoldFR (dismissed):
Excellent work @sea-bass ! Thanks for your contribution 😃

sea-bass:
> Excellent work @sea-bass ! Thanks for your contribution 😃

Sorry, found a small bug in the `captureImage` implementation. You should be able to approve now.

ManifoldFR:
No worries! You might have to check the Codacy static analysis results though, at least fix some of the medium-level issues.

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -0,0 +1,300 @@
...
+        Capture an image from the Viser viewer and return an RGB array.
+
+        Parameters:
+            w: The width of the captured image.
+            h: The height of the captured image.
+            client_id: The ID of the Viser client handle.
+            transport_format: The transport format to use for the captured image.
+                Can be "jpeg" (default) or "png".
+        """
+        cli = self.viewer._get_client_handle(client_id)
```
brentyi: If you're fine with any client (or assume one connected client at a time), maybe you could do `next(iter(viewer.get_clients().values()))`?

sea-bass: This is a good default. Maybe if the client ID is `None` it just grabs the first like this, otherwise it looks it up by ID? Thanks!

brentyi: The main issue here is that `client_id=0` will break if you just open the visualizer and refresh; the ID for the second connection would increment to 1

sea-bass: Done -- thanks for the suggestion!

ManifoldFR (dismissed):

sea-bass:
I addressed some of the Codacy issues, but I think the remaining ones may not be worth addressing right away.

Most of them (4/5) are actually more general issues with the base visualizer class / the way that Meshcat visualizer overrides it. which I "took inspiration" (read: copied) from. So I'd suggest later revisiting those more broadly to make static analysis happy for the entire visualization submodule.

The last one is just that Viser uses the placeholder `0.0.0.0` to mean `localhost`, whereas normally this would be a security risk for all ports open. But we know this is fine in this case as it's not the actual IP address used.

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -0,0 +1,323 @@
...
+
+        if open:
+            import webbrowser
+
+            host = self.viewer.get_host()
+            port = self.viewer.get_port()
+            if host == "0.0.0.0":
+                http_url = f"http://localhost:{port}"
+            else:
+                http_url = f"http://{host}:{port}"
```
ManifoldFR:
```suggestion
            if host == "0.0.0.0":
                host = "localhost"
            http_url = f"http://{host}:{port}"
```

ManifoldFR:
> I addressed some of the Codacy issues, but I think the remaining ones may not be worth addressing right away.
> 
> Most of them (4/5) are actually more general issues with the base visualizer class / the way that Meshcat visualizer overrides it. which I "took inspiration" (read: copied) from. So I'd suggest later revisiting those more broadly to make static analysis happy for the entire visualization submodule.
> 
> The last one is just that Viser uses the placeholder `0.0.0.0` to mean `localhost`, whereas normally this would be a security risk for all ports open. But we know this is fine in this case as it's not the actual IP address used.

Okay, sounds good to me then. To get around the host issue I'm going to suggest a tiny change to ensure it doesn't trip Codacy up on that. For the other issues I think we should go ahead and just overrule Codacy, would you agree @nim65s @jorisv ?

sea-bass:
That tiny change is also one less line of code so 👍🏻 -- thank you!

(edit: although it didn't seem to outsmart Codacy)

brentyi:
> The last one is just that Viser uses the placeholder `0.0.0.0` to mean `localhost`, whereas normally this would be a security risk for all ports open. But we know this is fine in this case as it's not the actual IP address used.

Are you sure this is a false positive? I'd expect `0.0.0.0` to mean "bind to all interfaces", which is different from `localhost`. The host here is passed directly to `websockets`, so I don't think there should be any special meaning.

Here's also a relevant note that might be causing confusion (sorry!): https://github.com/nerfstudio-project/viser/blob/c11782c706978836da3653666c700e74b095acae/src/viser/_viser.py#L758-L763

sea-bass:
Interesting. I think I had read that comment backwards, haha. 🙃 

So can we pass `host="localhost"` to our `ViserServer` instance to prevent this altogether?

EDIT: Tried it myself and it worked, I also added optional `host` and `port` args to this visualizer to pipe into `ViserServer`. The default is `localhost:8000`.

Lo and behold, now down to 4 Codacy issues due to inheriting from `BaseVisualizer`.

sea-bass:
Anything else needed before we merge this?

ManifoldFR:
We first need to fix CI on Wind*ws, @jorisv can you take a look?

Also can you rebase on devel in lieu of the merge commit?

sea-bass:
> We first need to fix CI on Wind*ws, @jorisv can you take a look?
> 
> Also can you rebase on devel in lieu of the merge commit?

Seems like the proper rebase cleared out the CI issue...

ManifoldFR:
> Seems like the proper rebase cleared out the CI issue...

The world is sometimes full of wonder indeed...

Only problem left is Codacy being a stickler. @jorisv do you think we could merge this anyway? I think we can so I'm putting my stamp of approval. (thanks a lot @sea-bass !)

ManifoldFR (dismissed):

jorisv (dismissed):

jorisv:
It's good to me. I plan to deactivate codacy in the future since it doesn't give good feedback in C++.

jorisv:
Hello @sea-bass,

I have added an example to test the viewer. The example is running fine but I have the following warning:

```
(viser) Connection opened (0, 1 total), 5 persistent messages
[Entering video recording context]
connection handler failed
  File "/home/jvaillan/dev/inria/pinocchio/pinocchio/.pixi/envs/collision/lib/python3.12/site-packages/websockets/asyncio/server.py", line 376, in conn_handler
    await self.handler(connection)
  File "/home/jvaillan/dev/inria/pinocchio/pinocchio/.pixi/envs/collision/lib/python3.12/site-packages/viser/infra/_infra.py", line 395, in ws_handler
    await asyncio.gather(
  File "/home/jvaillan/dev/inria/pinocchio/pinocchio/.pixi/envs/collision/lib/python3.12/site-packages/viser/infra/_infra.py", line 591, in _message_producer
    "messages": tuple(
                ^^^^^^
  File "/home/jvaillan/dev/inria/pinocchio/pinocchio/.pixi/envs/collision/lib/python3.12/site-packages/viser/infra/_infra.py", line 592, in <genexpr>
    message.as_serializable_dict() for message in outgoing
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jvaillan/dev/inria/pinocchio/pinocchio/.pixi/envs/collision/lib/python3.12/site-packages/viser/infra/_messages.py", line 115, in as_serializable_dict
    k: _prepare_for_serialization(v, hints[k]) for k, v in vars(self).items()
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jvaillan/dev/inria/pinocchio/pinocchio/.pixi/envs/collision/lib/python3.12/site-packages/viser/infra/_messages.py", line 58, in _prepare_for_serialization
    return int(value)
           ^^^^^^^^^^
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
(viser) Connection opened (1, 2 total), 298 persistent messages
```

Is it normal ?

sea-bass:
Nice, thank you!

I certainly wasn't seeing any warnings when I was testing, though I am on a different dev environment of course. Let me test your example later today...

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -0,0 +1,370 @@
...
+        if client_id is None:
+            cli = next(iter(clients.values()))
+        elif client_id not in clients:
+            raise RuntimeError(
+                f"Viser server does not have a client with ID '{client_id}'"
+            )
+        else:
+            cli = clients[client_id]
+
+        return cli.get_render(height=h, width=w, transport_format=transport_format)
```
brentyi: @sea-bass on the serialization error: `get_render()` expect integers for height and width, but it looks like this can end up passing `height=None` and `width=None`.

https://viser.studio/versions/1.0.0/api/handles/client_handles/#viser.ClientHandle.get_render

sea-bass: That would do it, thank you!

I guess should we make these required arguments? Default them to integers?

sea-bass: Update : I found that you can do `cli.camera.image_height` and `cli.camera.image_width`, so I've fallen back to that in the `None` case and it works for me.

@jorisv FYI, try again? Your example worked on my end with my latest commit.

---

BTW, it is probably not ideal that the `play()` method doesn't have a way to pass in the arguments to `captureImage()`, in case you wanted to use a nondefault resolution. But that's an issue for later.

jorisv (approved):

ManifoldFR:
Continuing the discussion here cos mattermost isn't working @jorisv

Did you install imageio-ffmpeg when testing?

sea-bass:
> Continuing the discussion here cos mattermost isn't working @jorisv
> 
> Did you install imageio-ffmpeg when testing?

If this is related to the error about TiffWriter not having an `fps` argument, then exactly this suggestion fixed that for me 🙂

sea-bass:
> > Continuing the discussion here cos mattermost isn't working @jorisv
> > Did you install imageio-ffmpeg when testing?
> 
> If this is related to the error about TiffWriter not having an `fps` argument, then exactly this suggestion fixed that for me 🙂

Wonder if this needs to be added as a dependency to viser?

ManifoldFR:
This makes sense, usually when you want to record video but don't have a video encoder you just dump an image sequence to a relatively lossless format. I remember that old PC games did that for demo recording.

What doesn't make sense is that imageio doesn't filter out irrelevant arguments like fps when falling back to the image sequence...

jorisv:
@ManifoldFR thank's your suggestion fix the issue.

Since our code is only working with ffmpeg plugin, did it make sense to change base_visualizer to force the use of ffmpeg format ?

But it's not related to this PR. So we can merge this nice work now.

ManifoldFR:
Thanks a lot @sea-bass !

sea-bass:
Thanks @ManifoldFR @jorisv @brentyi -- this is going to be a cool addition to the next Pinocchio release!

PR #2731, opened 2025-07-18
Fix missing argument in exposeDelassus() pybind definition

yvtungchou:
Hello,
I noticed that when you are binding the `computeDampedDelassusMatrixInverse()` function from C++ to Python, the `scaled = false` argument from the C++ function signature is not reflected in the Python binding. This results in an argument mismatch when calling the function from Python.

This pull request adds the missing argument to ensure the binding matches the C++ function signature and prevents runtime errors in downstream Python code.

Let me know if this fix the issue.


jcarpent (dismissed):

jcarpent:
@yvtungchou 
Thanks a lot for your nice fix.
Could you amend your PR with by completing the [changelog](https://github.com/stack-of-tasks/pinocchio/blob/devel/CHANGELOG.md)?

yvtungchou:
@jcarpent I've already add the change to the changelog, let me know if there are anything needs to be done!

jcarpent:
Perfect. Thanks a lot @yvtungchou.

jcarpent (approved):

PR #2743, opened 2025-08-11
Fix/cannot found casadi

dogged1021:
## Description

When building Pinocchio with CasADi support enabled (`-DBUILD_WITH_CASADI_SUPPORT=ON`), the build process fails at the linking stage. The linker cannot find the CasADi library because an incorrect CMake target name is used in `target_link_libraries`.

```bash
FAILED: unittest/casadi/pinocchio-test-cpp-casadi-basic 
: && /usr/bin/g++ -pedantic -Wno-long-long -Wall -Wextra -Wcast-align -Wcast-qual -Wformat -Wwrite-strings -Wconversion  -O3 -DNDEBUG  unittest/casadi/CMakeFiles/pinocchio-test-cpp-casadi-basic.dir/basic.cpp.o -o unittest/casadi/pinocchio-test-cpp-casadi-basic  -Wl,-rpath,${USER}/source_codes/pinocchio/build/src:${USER}/local/boost_1_85_0/lib  src/libpinocchio_default.so.3.7.0  ${USER}/local/boost_1_85_0/lib/libboost_unit_test_framework.so.1.85.0  -lcasadi  ${USER}/local/boost_1_85_0/lib/libboost_serialization.so.1.85.0 && :
/usr/bin/ld: cannot find -lcasadi: No such file or directory
collect2: error: ld returned 1 exit status
```
The typical symptom is a linker error reporting undefined symbols from the `casadi` library or that the library itself cannot be found.

### **Root Cause Analysis**

The root cause is the use of the plain target name `casadi` in `target_link_libraries`. Modern CMake best practices advocate for using "Imported Targets," especially for libraries found via `find_package`. These imported targets are often namespaced (e.g., `PackageName::TargetName`) and encapsulate all usage requirements, such as include directories, library paths, and link dependencies.

The `find_package(casadi ...)` command creates an imported target named `casadi::casadi`. When `target_link_libraries` uses `casadi`, CMake fails to resolve it to the correct library file and its dependencies, leading to the link failure.

### **Solution**
The correct approach is to link against the namespaced imported target `casadi::casadi`.

The solution is to update the `target_link_libraries` command in the affected `CMakeLists.txt` files to use the correct imported target name, `casadi::casadi`.

## Checklist

- [ ✓ ] I have performed a self-review of my own code


jorisv:
Hello @dogged1021,

Thank you for the contribution.
Unfortunately, we were using the `casadi` target because `casadi::casadi` is not included in casadi Ubuntu CMake module as shown in this [CI failure](https://github.com/stack-of-tasks/pinocchio/actions/runs/16871506314/job/47885042870?pr=2743).

You can modify you PR to handle this with this kind of code:

```cmake
if(BUILD_WITH_CASADI_SUPPORT)
  add_project_dependency(casadi 3.4.5 REQUIRED PKG_CONFIG_REQUIRES "casadi >= 3.4.5")
  if(NOT TARGET casadi::casadi)
    add_library(casadi::casadi ALIAS casadi)
  endif()
  list(APPEND LIBRARIES_DEPENDENCIES ${PROJECT_NAME}_casadi)
endif(BUILD_WITH_CASADI_SUPPORT)
```

I'm not sure this will work since CMake maybe can prevent creating a target in another namespace…
We can also store the target name in a variable.

By the way, what packaging system are you using ? We never experienced your issue with conda/nix/ubuntu packages

dogged1021:
Hello, @jorisv ,

OS: ubuntu 22.04
pinocchio-3.7.0
casadi-3.7.1
These are my os and the versions of the libraries used. I found that casadi-3.7.1’s CMake uses two different export mechanisms, one for the build tree and one for the install tree. They generate two sets of “targets files” with different target names: casadi for build  and casadi::casadi for install: 
```cmake
# Install the export set for use with the install-tree
file(RELATIVE_PATH REL_CMAKE_PREFIX "${CMAKE_INSTALL_PREFIX}" "${CMAKE_PREFIX}")
install(EXPORT casadi-targets NAMESPACE casadi:: DESTINATION ${REL_CMAKE_PREFIX})
```
https://github.com/casadi/casadi/blob/9763952a9aa742c5be3b53dc1a1d033a57235328/CMakeLists.txt#L2263
 
---

But in casadi-3.4.5 mentioned in pinocchio's CMake 
```cmake
  add_project_dependency(casadi 3.4.5 REQUIRED PKG_CONFIG_REQUIRES "casadi >= 3.4.5")
```
the two sets of “targets files” have the same target name: casadi
```cmake
install(FILES
  "${PROJECT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/casadi-config.cmake"
  "${PROJECT_BINARY_DIR}/casadi-config-version.cmake"
  DESTINATION "${CMAKE_PREFIX}" COMPONENT dev)

# Install the export set for use with the install-tree
install(EXPORT casadi-targets DESTINATION "${CMAKE_PREFIX}" COMPONENT dev)
```
https://github.com/casadi/casadi/blob/58aa427bc7ed85d7cacfaf7c123f4eaca13a2d35/CMakeLists.txt#L848

---

So perhaps this is the reason and we could modify CMakeLists.txt like this to be compatible with two link target names?
```cmake
if(TARGET casadi::casadi)
  set(CASADI_LINK casadi::casadi)
elseif(TARGET casadi)
  set(CASADI_LINK casadi)
elseif(DEFINED CASADI_LIBRARIES)
  set(CASADI_LINK ${CASADI_LIBRARIES})
elseif(DEFINED casadi_LIBRARIES)
  set(CASADI_LINK ${casadi_LIBRARIES})
else()
  message(FATAL_ERROR "CasADi found but no usable target or libraries variables exposed.")
endif()

target_link_libraries(${PROJECT_NAME}_casadi ${CASADI_SCOPE} ${CASADI_LINK})
```

Thank you for your efforts!

jorisv:
The APT CI is using casadi 3.6.7 and the pixi one is using 3.7.0.

I'm looking at the casadi CMake module in pixi and it doesn't export `casadi` target. So it only work because it will search for libcasadi in the pixi environment.

So, you can define a `CASADI_TARGET` variable in the main CMakeLists.txt, after the `add_project_dependency(casadi 3.4.5 REQUIRED PKG_CONFIG_REQUIRES "casadi >= 3.4.5")` call.
Then use this target instead of casadi.
Don't forget to comment why we are doing that (you can add a link to this discussion).

unittest/CMakeLists.txt
```diff
@@ -129,7 +129,16 @@ function(ADD_PINOCCHIO_UNIT_TEST name)
...
+    # Convert package names to proper CMake targets
+    set(CONVERTED_PKGS "")
+    foreach(PKG IN LISTS PKGS)
+      if(PKG STREQUAL "casadi")
+        list(APPEND CONVERTED_PKGS "casadi::casadi")
+      else()
+        list(APPEND CONVERTED_PKGS ${PKG})
+      endif()
+    endforeach()
+    target_link_libraries(${TEST_NAME} PRIVATE ${CONVERTED_PKGS})
```
jorisv: Instead of doing that, look for all `PACKAGES` argument and replace casadi by ${CASADI_TARGET} ([Like in this one](https://github.com/stack-of-tasks/pinocchio/blob/devel/unittest/casadi/CMakeLists.txt#L6))

This avoid adding more complexity in CMake.

jorisv (changes requested):

dogged1021:
Hello, @jorisv ,
I confirmed based on the [3.7.1-changelog](https://github.com/casadi/casadi/releases/tag/3.7.1) that CasADi will subsequently add namespaces for export targets.
> - **breaking** The casadi cmake target now has a namespace.

And I changed the CMakeLists.txt as you said in a new branch [dev/build-casadi-with-different-versions](https://github.com/dogged1021/pinocchio/tree/dev/build-casadi-with-different-versions) . What do you think?
(Sorry, I'm not familiar with these procedures, maybe I should update in the branch `dogged1021:fix/cannot-found-casadi`? )

jorisv:
Hello @dogged1021,

It's a better practice to push your new change on the same branch. Otherwise you will have to open a new PR and we will lost this discussion.

CMakeLists.txt
```diff
@@ -273,6 +273,33 @@ endif(BUILD_WITH_AUTODIFF_SUPPORT)
...
+  elseif(DEFINED CASADI_LIBRARIES)
+    set(CASADI_TARGET ${CASADI_LIBRARIES})
+  elseif(DEFINED casadi_LIBRARIES)
+    set(CASADI_TARGET ${casadi_LIBRARIES})
+  else()
+    message(FATAL_ERROR "CasADi found but no usable target or libraries variable exported.")
+  endif()
+
+  # Expose to subdirectories and cache for external use if needed.
+  set(CASADI_TARGET "${CASADI_TARGET}" CACHE STRING "CasADi target to link" FORCE)
```
jorisv: No need to set a variable in CACHE to allow sub-directories to access it.

```suggestion
```

CMakeLists.txt
```diff
@@ -273,6 +273,33 @@ endif(BUILD_WITH_AUTODIFF_SUPPORT)
...
+
+  # Optional: propagate include dirs only when not using imported targets.
+  if(NOT TARGET casadi::casadi AND NOT TARGET casadi)
+    if(DEFINED CASADI_INCLUDE_DIRS)
+      set(CASADI_INCLUDE_DIRS_NORMALIZED "${CASADI_INCLUDE_DIRS}" CACHE PATH "CasADi include dirs" FORCE)
+    elseif(DEFINED casadi_INCLUDE_DIRS)
+      set(CASADI_INCLUDE_DIRS_NORMALIZED "${casadi_INCLUDE_DIRS}" CACHE PATH "CasADi include dirs" FORCE)
+    endif()
+  endif()
+
```
jorisv: It's not clear what you want to achieve here. Especially, `CASADI_INCLUDE_DIRS_NORMALIZED` is not used anywhere in the PR.

```suggestion
```

CMakeLists.txt
```diff
@@ -273,6 +273,33 @@ endif(BUILD_WITH_AUTODIFF_SUPPORT)
...
+  # Normalize CasADi link target across versions.
+  # Prefer imported target (casadi::casadi), then legacy target (casadi), then library variables.
+  if(TARGET casadi::casadi)
+    set(CASADI_TARGET casadi::casadi)
+  elseif(TARGET casadi)
+    set(CASADI_TARGET casadi)
+  elseif(DEFINED CASADI_LIBRARIES)
+    set(CASADI_TARGET ${CASADI_LIBRARIES})
+  elseif(DEFINED casadi_LIBRARIES)
+    set(CASADI_TARGET ${casadi_LIBRARIES})
```
jorisv: During our discussion, we weren’t able to identify a CasADI CMake module that lacked a defined target. Could you explain why you’re aiming to manage this specific case?

```suggestion
```

jorisv (changes requested):

jorisv:
Hello @dogged1021,

To move forward with this PR, I force-pushed the`dev/build-casadi-with-different-versions` branch into `dogged1021:fix/cannot-found-casadi` I’ve reviewed your changes.
If you’re unable to continue working on this PR, I’m happy to finish it up.

jorisv (dismissed):

jorisv (dismissed):

CMakeLists.txt
```diff
@@ -274,6 +274,16 @@ endif(BUILD_WITH_AUTODIFF_SUPPORT)
 
 if(BUILD_WITH_CASADI_SUPPORT)
   add_project_dependency(casadi 3.4.5 REQUIRED PKG_CONFIG_REQUIRES "casadi >= 3.4.5")
+
+  # Normalize CasADi link target across versions. Prefer imported target (casadi::casadi), then
+  # legacy target (casadi)
+  if(TARGET casadi::casadi)
+    set(CASADI_TARGET casadi::casadi)
+  elseif(TARGET casadi)
+    set(CASADI_TARGET casadi)
```
nim65s:
```suggestion
    add_library(casadi::casadi ALIAS casadi)
```

(this is only a suggestion, the current solution would also be fine for me, but this seems easier to maintain on the long term)

jorisv: You're right.
I was thinking CMake would prohibit to create a library with a different namespace than our project. But a local test prove me wrong.

jorisv (approved):

nim65s (approved):

dogged1021:
@jorisv, I sincerely apologize for not being able to dedicate time to the relevant work recently due to certain circumstances. I deeply regret any inconvenience this may have caused you and truly appreciate your efforts and dedication.

jorisv:
Hello @dogged1021,
You don't have to apologize. You found an issue and provide a patch. Given my intention to create a new release I finalized some last points before your feedback.
Thank you to have help improve Pinocchio :)

PR #2748, opened 2025-08-22
Fix Viser visualizer loading DAE files as trimesh.Scene

JafarAbdi:
## Description

Fix Viser visualizer loading DAE files as trimesh.Scene. To reproduce run https://gist.github.com/JafarAbdi/4e33d485491c336d5d01f7cf6f2865bf with `uv run` (See _add_mesh_from_path function)

@sea-bass fyi

## Checklist

### DIFF
--- a/bindings/python/pinocchio/visualize/viser_visualizer.py
+++ b/bindings/python/pinocchio/visualize/viser_visualizer.py
@@ -218,7 +218,7 @@ class ViserVisualizer(BaseVisualizer):
         """
         Load a mesh from a file.
         """
-        mesh = trimesh.load(mesh_path)
+        mesh = trimesh.load_mesh(mesh_path)
         if color is None:
             return self.viewer.scene.add_mesh_trimesh(name, mesh)
         else:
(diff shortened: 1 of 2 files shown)


bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -218,7 +218,7 @@ def _add_mesh_from_path(self, name, mesh_path, color):
         """
         Load a mesh from a file.
         """
-        mesh = trimesh.load(mesh_path)
+        mesh = trimesh.load_mesh(mesh_path)
```
JafarAbdi, line 221: Or maybe?

```python
mesh = trimesh.load(mesh_path)
if isinstance(mesh, trimesh.Scene):
    mesh = mesh.to_mesh()
```

jcarpent (dismissed):

jcarpent:
@JafarAbdi Thanks a lot for the nice fix. Could you add a line in the Changelog to track this change?

jcarpent (approved):

AlexyLegrand:
Thank you for this nice project. I tried to load franka model in viser, but it seems that the color is not rendered. The same occurs for UR10. Do experience the same issue ?

bindings/python/pinocchio/visualize/viser_visualizer.py
sea-bass: @JafarAbdi I just ran into issues with this approach removing the color from DAE meshes that have complex textures. I believe we want something like this

https://github.com/open-planning/roboplan/commit/f93099f81d752e1c750b06b54a3870b391c0a6a5

EDIT: Ran your repro example with my change and it works too!

BTW see my whole adventures with the next PR that went in, which I believe should be reverted https://github.com/stack-of-tasks/pinocchio/issues/2829

PR #2784, opened 2025-10-10
Adding a spline joint to the joint collection

MegMll:
## Description

In this PR, we introduce the new spline joint to pinocchio. This implementation is based on the paper of Lee et al. available [here](https://web.cs.ucla.edu/~dt/papers/siggraph08/siggraph08.pdf)

It allows to define a number of SE3 frames, that will define the movement and dynamics of the joint. It's very useful for biomechanics, but can be used in other applications. 

Here are videos of two examples written by @lpuch
- Knee
[Knee Movement](https://github.com/user-attachments/assets/67384bce-1384-4123-8523-8907f4be8c24)

- Random Spline
[Random Spline](https://github.com/user-attachments/assets/587de626-cb73-4fd6-8e6e-04b1cec54a9d)

## Future 
- [ ] Introduce the [Spline Surface](https://web.mit.edu/hyperbook/Patrikalakis-Maekawa-Cho/node19.html)


include/pinocchio/src/multibody/joint/joint-collection.hxx
```diff
@@ -106,6 +109,7 @@ namespace pinocchio
       JointModelHz,
       JointModelHelicalUnaligned,
       JointModelUniversal,
+      JointModelSpline,
```
jorisv, line 114: Do we need it in the default joint collection ?

include/pinocchio/multibody/joint/joint-spline.hpp
```diff
@@ -0,0 +1,557 @@
...
+
+#include "pinocchio/macros.hpp"
+#include "pinocchio/spatial/inertia.hpp"
+#include "pinocchio/spatial/explog.hpp"
+#include "pinocchio/multibody/joint/joint-base.hpp"
+#include "pinocchio/multibody/joint-motion-subspace.hpp"
+#include "pinocchio/math/fwd.hpp"
+#include "pinocchio/math/quaternion.hpp"
+
+#include <iostream>
```
jorisv:
```suggestion
#include "pinocchio/eigen-macros.hpp"
#include "pinocchio/spatial/explog.hpp"
#include "pinocchio/multibody/joint/joint-base.hpp"
#include "pinocchio/multibody/joint-motion-subspace.hpp"

```

include/pinocchio/multibody/joint/joint-spline.hpp
```diff
@@ -0,0 +1,557 @@
...
+  }; // struct JointDataSplinerTpl
+
+  PINOCCHIO_JOINT_CAST_TYPE_SPECIALIZATION(JointModelSplineTpl);
+  template<typename _Scalar, int _Options>
+  struct JointModelSplineTpl : public JointModelBase<JointModelSplineTpl<_Scalar, _Options>>
+  {
+    EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+    typedef JointSplineTpl<_Scalar, _Options> JointDerived;
+    typedef SE3Tpl<_Scalar, _Options> SE3;
+    typedef MotionTpl<_Scalar, _Options> Motion;
```
jorisv: `Transformation_t` and `Motion_t` are already defined in `PINOCCHIO_JOINT_TYPEDEF_TEMPLATE`

include/pinocchio/multibody/joint/joint-spline.hpp
```diff
@@ -0,0 +1,557 @@
...
+      ctrlFrames.push_back(frame);
+      nbCtrlFrames++;
+    }
+
+    void buildJoint()
+    {
+      makeKnots();
+      computeRelativeMotions();
+    }
+
```
jorisv: This is a little error prone. `JointModelSplineTpl` instance will not work until you call this function.

Two way of fixing that:
1. Remove `addControlFrame`, add a `setControlFrames` function that will set the full vector and call buildJoint (now private). Do the same on the constructor.
2. Add a `JointModelSplineBuilderTpl` class that you can edit, then `JointModelSplineBuilderTpl::buildJoints` will return a well configured `JointModelSplineTpl`.

Option 1. is way simpler. I think, user can manage by themself to keep a ctrlFrames vector on their side.

include/pinocchio/multibody/joint/joint-spline.hpp
```diff
@@ -0,0 +1,557 @@
...
+    using Base::isEqual;
+    bool isEqual(const JointModelSplineTpl & other) const
+    {
+      return Base::isEqual(other) && other.degree == degree && other.nbCtrlFrames == nbCtrlFrames
+             && other.ctrlFrames == ctrlFrames;
+    }
+
+    template<typename ConfigVector>
+    void calc(JointDataDerived & data, const typename Eigen::MatrixBase<ConfigVector> & qs) const
```
jorisv:
```suggestion
    void calc(JointDataDerived & data, const Eigen::MatrixBase<ConfigVector> & qs) const
```

include/pinocchio/multibody/joint/joint-spline.hpp
```diff
@@ -0,0 +1,557 @@
...
+        data.S.matrix() = exp6(relativeMotions[i - 1] * phi_i).actInv(data.S)
+                          + relativeMotions[i - 1].toVector() * phi_dot_i;
+      }
+    }
+
+    template<typename ConfigVector, typename TangentVector>
+    void calc(
+      JointDataDerived & data,
+      const typename Eigen::MatrixBase<ConfigVector> & qs,
+      const typename Eigen::MatrixBase<TangentVector> & vs) const
```
jorisv:
```suggestion
      const Eigen::MatrixBase<ConfigVector> & qs,
      const Eigen::MatrixBase<TangentVector> & vs) const
```

include/pinocchio/multibody/joint/joint-spline.hpp
```diff
@@ -0,0 +1,557 @@
...
+
+      data.c = data.c * data.joint_v[0];
+      data.v = data.S * data.joint_v;
+    }
+
+    template<typename TangentVector>
+    void calc(
+      JointDataDerived & data,
+      const Blank not_used,
+      const typename Eigen::MatrixBase<TangentVector> & vs) const
```
jorisv:
```suggestion
      const Eigen::MatrixBase<TangentVector> & vs) const
```

include/pinocchio/src/multibody/joint/joint-spline.hxx
```diff
@@ -0,0 +1,557 @@
...
+    void calc(JointDataDerived & data, const typename Eigen::MatrixBase<ConfigVector> & qs) const
+    {
+      data.joint_q = qs.template segment<NQ>(idx_q());
+
+      SpanIndexes indexes = FindSpan<Scalar, Options>::run(qs, degree, nbCtrlFrames, knots);
+
+      // Basis functions and their derivatives
+      data.N.setZero();
+      data.N_der.setZero();
+      for (size_t i = indexes.start_idx; i < indexes.end_idx; i++)
```
jorisv, line 229: `start_idx` is an int where `size_t` is unsigned.
Since `start_idx` and `end_idx` can't be negative, I think it's better to turn them into size_t.

include/pinocchio/src/multibody/joint/joint-spline.hxx
```diff
@@ -0,0 +1,557 @@
...
+      res.setIndexes(id(), idx_q(), idx_v(), idx_vExtended());
+      return res;
+    }
+
+    void makeKnots()
+    {
+      const int n_knots = nbCtrlFrames + degree + 1;
+      knots.resize(n_knots);
+      knots.head(degree + 1).setZero();
+      const Scalar denominator = static_cast<Scalar>(nbCtrlFrames - degree + 1);
```
jorisv, line 398: Whe should trigger an exception if the user don't provide enough control point

include/pinocchio/multibody/joint/joint-spline.hpp
```diff
@@ -0,0 +1,557 @@
...
+      // joint subspace S
+      data.S.matrix().setZero();
+      for (size_t i = indexes.start_idx + 1; i < indexes.end_idx; i++)
+      {
+        const Scalar phi_i = data.N.segment(i, indexes.end_idx - i).sum();
+        const Scalar phi_dot_i = data.N_der.segment(i, indexes.end_idx - i).sum();
+
+        data.M = data.M * exp6(relativeMotions[i - 1] * phi_i);
+        data.S.matrix() = exp6(relativeMotions[i - 1] * phi_i).actInv(data.S)
+                          + relativeMotions[i - 1].toVector() * phi_dot_i;
```
jorisv: Compiler will probably optimize it, but you can store `exp6(relativeMotions[i - 1] * phi_i)` in a temporary variable.

include/pinocchio/multibody/joint/joint-spline.hpp
```diff
@@ -0,0 +1,557 @@
...
+        const Scalar phi_dot_i = data.N_der.segment(i, indexes.end_idx - i).sum();
+        const Scalar phi_ddot_i = data.N_der2.segment(i, indexes.end_idx - i).sum();
+
+        data.M = data.M * exp6(relativeMotions[i - 1] * phi_i);
+        data.c =
+          relativeMotions[i - 1] * phi_ddot_i
+          + exp6(relativeMotions[i - 1] * phi_i)
+              .actInv(data.c + Motion(data.S.matrix()).cross(relativeMotions[i - 1]) * phi_dot_i);
+        data.S.matrix() = exp6(relativeMotions[i - 1] * phi_i).actInv(data.S)
+                          + relativeMotions[i - 1].toVector() * phi_dot_i;
```
jorisv: Same here, `exp6(relativeMotions[i - 1] * phi_i)` can be stored somewhere. This will help readability.

include/pinocchio/multibody/joint/joint-spline.hpp
```diff
@@ -0,0 +1,557 @@
...
+        const Scalar phi_i = data.N.tail(nbCtrlFrames - (i + 1)).sum();
+        const Scalar phi_dot_i = data.N_der.tail(nbCtrlFrames - (i + 1)).sum();
+        const Scalar phi_ddot_i = data.N_der2.tail(nbCtrlFrames - (i + 1)).sum();
+
+        data.c =
+          relativeMotions[i] * phi_ddot_i
+          + exp6(relativeMotions[i] * phi_i)
+              .actInv(data.c + Motion(data.S.matrix()).cross(relativeMotions[i]) * phi_dot_i);
+        data.S.matrix() = exp6(relativeMotions[i] * phi_i).actInv(data.S)
+                          + relativeMotions[i].toVector() * phi_dot_i;
```
jorisv: Store `exp6(relativeMotions[i] * phi_i)` in a temporary.

include/pinocchio/src/multibody/joint/joint-spline.hxx
jorisv: Put some guards on degree size vs control points numbers

include/pinocchio/multibody/joint/joint-spline.hpp
```diff
@@ -0,0 +1,557 @@
...
+  private:
+    Scalar bsplineBasis(int i, int k, const Scalar x) const
+    {
+      if (k == 0)
+      {
+        return pinocchio::internal::if_then_else(
+          pinocchio::internal::LE, knots[i], x,
+          pinocchio::internal::if_then_else(
+            pinocchio::internal::LE, x, knots[i + 1], Scalar(1), Scalar(0)),
+          Scalar(0));
```
jorisv: Oh my god !

1.: You can do a `using internal::if_then_else;` at the beggining of the function ?
2.: Can you write the condition written normally in comment ? Like [here](https://github.com/stack-of-tasks/pinocchio/blob/devel/include/pinocchio/multibody/liegroup/special-orthogonal.hpp#L87-L93). You can use `// clang-format off` and `// clang-format on` comment around the commented code to keep a nice formatting.

include/pinocchio/multibody/joint/joint-spline.hpp
```diff
@@ -0,0 +1,557 @@
...
+      const int /*degree*/,
+      const int nbCtrlFrames,
+      const Eigen::MatrixBase<KnotsVector> & /*knots*/)
+    {
+      return {0, nbCtrlFrames};
+    }
+  };
+
+  template<int Options>
+  struct FindSpan<double, Options>
```
jorisv: double specialization will not work on float or other true scalar type.
Also, maybe some custom scalar can still extract true scalar and use the normal function.

So it's better to provide a default implementation and specialize for custom scalar in dedicated header as done [here](https://github.com/stack-of-tasks/pinocchio/blob/devel/include/pinocchio/autodiff/cppad/algorithm/aba.hpp)

include/pinocchio/src/serialization/joints-data.hxx
```diff
@@ -269,6 +269,16 @@ namespace boost
       fix::serialize(ar, static_cast<pinocchio::JointDataBase<JointType> &>(joint), version);
     }
 
+    template<class Archive, typename Scalar, int Options>
+    void serialize(
+      Archive & ar,
+      pinocchio::JointDataSplineTpl<Scalar, Options> & joint,
+      const unsigned int version)
+    {
+      typedef pinocchio::JointDataSplineTpl<Scalar, Options> JointType;
```
jorisv, line 298: You forget the bspline values:

```cpp
    Vector N;
    Vector N_der;
    Vector N_der2;
```

include/pinocchio/src/serialization/joints-model.hxx
```diff
@@ -311,6 +311,19 @@ namespace boost
...
+    template<class Archive, typename Scalar, int Options>
+    void serialize(
+      Archive & ar,
+      pinocchio::JointModelSplineTpl<Scalar, Options> & joint,
+      const unsigned int version)
+    {
+      typedef pinocchio::JointModelSplineTpl<Scalar, Options> JointType;
+      ar & make_nvp("ctrlFrames", joint.ctrlFrames);
+      ar & make_nvp("degree", joint.degree);
+      ar & make_nvp("nbCtrlFrames", joint.nbCtrlFrames);
```
jorisv, line 343: Did this will call `buildJoint` ?

jorisv (changes requested):

include/pinocchio/multibody/joint/joint-spline.hpp
MegMll: Yeah sorry, it's unreadable....

include/pinocchio/src/multibody/joint/joint-collection.hxx
MegMll: I'm almost certain it will be used with casadi and double for biomechanics application. But maybe we can put it somewhere else, if you think it's better

include/pinocchio/src/serialization/joints-model.hxx
MegMll: no indeed, so i should probably serialize knots vector and relativeMotions

include/pinocchio/multibody/joint/joint-spline.hpp
MegMll: only casadi seemed to cause problems here, so i only specialized it for casadi

include/pinocchio/algorithm/splines.hpp
```diff
@@ -0,0 +1,49 @@
...
+namespace pinocchio
+{
+  struct SpanIndexes
+  {
+    size_t start_idx;
+    size_t end_idx;
+  };
+
+  template<typename Scalar, int Options>
+  struct FindSpan
```
jorisv, line 26: Add a doc string to explain what are doing these structures.

include/pinocchio/src/parsers/graph/graph-visitor.hxx
```diff
@@ -129,6 +129,13 @@ namespace pinocchio
             std::invalid_argument,
             "Graph - Joint Mimic cannot have a q_ref. Please use the joint offset directly.");
         }
+
+        SE3 operator()(const JointSpline &) const
+        {
+          PINOCCHIO_THROW_PRETTY(
+            std::invalid_argument,
+            "Graph - Joint Spline cannot have a q_ref. Please use the joint offset directly.");
+        }
```
jorisv, line 130: Why ? It's strange but I don't get why it's not possible.

src/parsers/graph/model-graph-algo.cpp
```diff
@@ -338,6 +348,28 @@ namespace pinocchio
           model.addBodyFrame(target_vertex.name, j_id, body_pose);
         }
 
+        void operator()(const JointSpline & joint, const BodyFrame & b_f)
+        {
+          if (!edge.forward)
+            PINOCCHIO_THROW_PRETTY(
+              std::invalid_argument, "Graph - JointSpline cannot be reversed.");
```
jorisv: We have to think about that.

Right now I'm not very confident about the reversal of this joint.
I don't think creating a ctrlFrames vector with inverse transformation inside will give us the inverse transformation.

unittest/casadi/joints.cpp
```diff
@@ -285,6 +285,25 @@ struct init<pinocchio::JointModelMimicTpl<Scalar, Options, JointCollection>>
...
+  typedef pinocchio::SE3Tpl<Scalar> SE3;
+  typedef Eigen::Matrix<Scalar, 3, 3> Matrix3s;
+  typedef Eigen::Matrix<Scalar, 3, 1> Vector3s;
+
+  static JointModel run()
+  {
+    PINOCCHIO_ALIGNED_STD_VECTOR(SE3) ctrlFrames;
+    ctrlFrames.push_back(SE3::Identity());
+    ctrlFrames.push_back(SE3(Matrix3s::Identity(), Vector3s(Scalar(0.), Scalar(0.), Scalar(1.))));
+    JointModel jmodel(ctrlFrames, 1);
```
jorisv: We should create a degree 3 spline so `bsplineBasisDerivative` and `bsplineBasisDerivative2` are called.

unittest/casadi/joints.cpp
```diff
@@ -359,6 +378,24 @@ struct TestADOnJoints
...
+  {
+    typedef pinocchio::JointModelSplineTpl<Scalar, Options> JointModel;
+    typedef pinocchio::SE3Tpl<Scalar> SE3;
+    typedef Eigen::Matrix<Scalar, 3, 3> Matrix3s;
+    typedef Eigen::Matrix<Scalar, 3, 1> Vector3s;
+
+    PINOCCHIO_ALIGNED_STD_VECTOR(SE3) ctrlFrames;
+    ctrlFrames.push_back(SE3::Identity());
+    ctrlFrames.push_back(SE3(Matrix3s::Identity(), Vector3s(Scalar(0.), Scalar(0.), Scalar(1.))));
+    JointModel jmodel(ctrlFrames, 1);
```
jorisv: We should create a degree 3 spline so `bsplineBasisDerivative` and `bsplineBasisDerivative2` are called.

unittest/cppad/joints.cpp
```diff
@@ -189,6 +189,23 @@ struct TestADOnJoints
...
+  {
+    typedef pinocchio::JointModelSplineTpl<Scalar, Options> JointModel;
+    typedef pinocchio::SE3Tpl<Scalar> SE3;
+    typedef Eigen::Matrix<Scalar, 3, 3> Matrix3s;
+    typedef Eigen::Matrix<Scalar, 3, 1> Vector3s;
+
+    PINOCCHIO_ALIGNED_STD_VECTOR(SE3) ctrlFrames;
+    ctrlFrames.push_back(SE3::Identity());
+    ctrlFrames.push_back(SE3(Matrix3s::Identity(), Vector3s(Scalar(0.), Scalar(0.), Scalar(1.))));
+    JointModel jmodel(ctrlFrames, 1);
```
jorisv: We should create a degree 3 spline so `bsplineBasisDerivative` and `bsplineBasisDerivative2` are called.

unittest/finite-differences.cpp
```diff
@@ -212,6 +212,22 @@ struct init<pinocchio::JointModelMimicTpl<Scalar, Options, JointCollection>>
...
+struct init<pinocchio::JointModelSplineTpl<Scalar, Options>>
+{
+  typedef pinocchio::JointModelSplineTpl<Scalar, Options> JointModel;
+
+  static JointModel run()
+  {
+    PINOCCHIO_ALIGNED_STD_VECTOR(pinocchio::SE3) ctrlFrames;
+    ctrlFrames.push_back(pinocchio::SE3::Identity());
+    ctrlFrames.push_back(pinocchio::SE3(Eigen::Matrix3d::Identity(), Eigen::Vector3d(0., 0., 1.)));
+    JointModel jmodel(ctrlFrames, 1);
```
jorisv: We should create a degree 3 spline so `bsplineBasisDerivative` and `bsplineBasisDerivative2` are called.

jorisv (changes requested):

unittest/casadi/joints.cpp
```diff
@@ -390,6 +427,145 @@ struct TestADOnJoints
...
+    pinocchio::JointDataBase<JointData> & jdata_base = jdata;
+
+    JointModelAD jmodel_ad = jmodel.template cast<AD_double>();
+    JointDataAD jdata_ad(jmodel_ad.createData());
+    pinocchio::JointDataBase<JointDataAD> & jdata_ad_base = jdata_ad;
+
+    ConfigVector q(jmodel.nq());
+
+    ConfigVector lb(ConfigVector::Constant(jmodel.nq(), 0.));
+    ConfigVector ub(ConfigVector::Constant(jmodel.nq(), 1.));
```
jorisv: The only difference between generic `test` function and specialized one for `JointModelSplineTpl` is the configuration bound.
Is there way to specialize RandomConfigurationStep to handle that ?
If there is no way, can generate the bounds from `TestADOnJoints` visitor to avoid duplicating the code.

unittest/finite-differences.cpp
```diff
@@ -222,6 +238,46 @@ struct FiniteDiffJoint
...
+  {
+    typedef typename JointModelSpline::ConfigVector_t CV;
+    typedef typename JointModelSpline::TangentVector_t TV;
+    typedef typename LieGroup<JointModelSpline>::type LieGroupType;
+
+    JointModelSpline jmodel = init<JointModelSpline>::run();
+    std::cout << "name: " << jmodel.classname() << std::endl;
+    JointDataSpline jdata = jmodel.createData();
+
+    CV q = LieGroupType().randomConfiguration(CV::Zero(), CV::Ones());
```
jorisv: The only difference here is randomConfiguration generation. Is there a way to handle it in LieGroup ?
Otherwise, it could be nice to generate it from the visitor to avoid duplicating the code.

unittest/joint-generic.cpp
```diff
@@ -75,7 +75,122 @@ void test_joint_methods(
...
+  std::cout << "Testing Joint over " << jmodel.shortname() << std::endl;
+
+  Eigen::VectorXd q1, q2;
+  Eigen::VectorXd armature =
+    Eigen::VectorXd::Random(jdata.S().nv()) + Eigen::VectorXd::Ones(jdata.S().nv());
+
+  q1 = LieGroupType().randomConfiguration(
+    Eigen::VectorXd::Zero(jmodel.nq()), Eigen::VectorXd::Ones(jmodel.nq()));
+  q2 = LieGroupType().randomConfiguration(
+    Eigen::VectorXd::Zero(jmodel.nq()), Eigen::VectorXd::Ones(jmodel.nq()));
```
jorisv, line 153: Same here, randomConfiguration is the only difference between generic and spline joint.
If there is no way to specialize randomConfiguration, we should add a mechanism to this test to generate a valid randomConfiguration without duplicating the test.

unittest/joint-generic.cpp
```diff
@@ -279,6 +394,22 @@ struct init<pinocchio::JointModelHelicalUnalignedTpl<Scalar, Options>>
...
+struct init<pinocchio::JointModelSplineTpl<Scalar, Options>>
+{
+  typedef pinocchio::JointModelSplineTpl<Scalar, Options> JointModel;
+
+  static JointModel run()
+  {
+    PINOCCHIO_ALIGNED_STD_VECTOR(SE3) ctrlFrames;
+    ctrlFrames.push_back(SE3::Identity());
+    ctrlFrames.push_back(SE3(Eigen::Matrix3d::Identity(), Eigen::Vector3d(0., 0., 1.)));
+    JointModel jmodel(ctrlFrames, 1);
```
jorisv: We should create an order 3 splines to test all the joint.

unittest/joint-spline.cpp
```diff
@@ -0,0 +1,422 @@
...
+  ctrlFrames.push_back(SE3::Random());
+
+  JointModelSpline jmodel(ctrlFrames, degree);
+
+  // Check size
+  BOOST_CHECK(jmodel.knots.size() == (degree + ctrlFrames.size() + 1));
+
+  // Check Values
+  Eigen::VectorXd knots_expected(degree + ctrlFrames.size() + 1);
+  knots_expected << 0., 0., 0., 0., 1., 1., 1., 1.;
```
jorisv: Add more control frame to test intermediate knots construction.

unittest/joint-spline.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+  JointModelSpline jmodel(ctrlFrames, degree);
+
+  // Check size
+  BOOST_CHECK(jmodel.relativeMotions.size() == (ctrlFrames.size() - 1));
+
+  // check values
+  for (size_t i = 0; i < ctrlFrames.size() - 1; i++)
+    BOOST_CHECK(jmodel.relativeMotions[i].isApprox(
+      pinocchio::log6(ctrlFrames[i].inverse() * ctrlFrames[i + 1])));
```
jorisv: This test duplicate the `computeRelativeMotions` code.

To avoid that and have a more robust test we can generate ctrlFrames from random `Motion` and check that `jmodel.relativeMotions` contains the generated random `Motion`

include/pinocchio/src/multibody/joint/joint-spline.hxx
jorisv, line 1: Is there a way that `neutral` and `randomConfiguration` return us a valid spline configuration ?

jorisv, line 1: Do we want to manage spline with abritrary bounds ?

jorisv (changes requested):

unittest/casadi/joints.cpp
MegMll: we can specialize the RandomConfigurationStep, like it's done for CompositeJoint, indeed. I'll do it, and you let me know, what you think.

unittest/finite-differences.cpp
MegMll: i checked, and it requires a new lieGroup, from my understanding. So i will generate bounds from the visitors

include/pinocchio/src/multibody/joint/joint-spline.hxx
MegMll: neutral is working, because it returns 0. However if we want random, we need to redefine a lieGroup...

unittest/finite-differences.cpp
```diff
@@ -61,154 +61,231 @@ void filterValue(MatrixBase<Matrix> & mat, typename Matrix::Scalar value)
 }
 
 template<typename JointModel_>
-struct init;
+struct JmodelWithBounds
```
jorisv: Maybe we should use a more generic name (if with want to add more attribute in the future).
Maybe JointModelWithParameters.

jorisv (changes requested):

unittest/finite-differences.cpp
```diff
@@ -229,14 +306,18 @@ struct FiniteDiffJoint
     typedef typename JointModel::TangentVector_t TV;
     typedef typename LieGroup<JointModel>::type LieGroupType;
 
-    JointModel jmodel = init<JointModel>::run();
+    JmodelWithBounds jmodelWithBounds = init<JointModel>::run();
```
jorisv:
```suggestion
    auto jmodelWithBounds = init<JointModel>::run();
```
Some compiler didn't like the missing template argument in the return type.

unittest/joint-generic.cpp
jorisv, line 436: This test can fail because on bad range bound with spline joint (seen on OSX)

jorisv (dismissed):

include/pinocchio/src/parsers/graph/graph-visitor.hxx
MegMll: I looked into it, and for now the spline joint can not accept q_ref. if we put a q_ref, that means we need to redefine the spline joint, from this q_ref to the end, or make the q_ref the 0 of the joint, and compute position from both side of the q_ref (-0.5/0.5, for example).
So q_ref is for now, not possible, but could be integrated when the fit_spline algorithm is added or when the upper and lower bounds are implemented

src/parsers/graph/model-graph-algo.cpp
MegMll: I looked into it as well. Indeed, reversing the ctrlFrames vector will only work when degree of the spline is 1. 
So maybe it's better to leave it as it is, right now.

include/pinocchio/src/multibody/joint/joint-spline.hxx
MegMll: in the future, it would be nice

florent-lamiraux:
Please be aware that modifying the joint collection breaks the API.

jorisv:
@florent-lamiraux We are not considering adding a joint an API break.
If your code rely on JointCollection order or must handle all the joints inside it, please, create your own joint collection.

Ipuch:
@MegMll and I were talking about this PR. As Megan left WILLOW and she is full time on other projects now, it's hard for her to pursue the debugging that this PR requires, and looking for why the tests were not passing. Does anyone would like to help and/or review this PR?

Here is the focused help, that would be appreciated and could unlock our next developments.

- We don't know why window tests with python doesn't pass ?
- We don't know why it compiles with MacOs Arm, but the test doesn't pass when executed ?

Let us know in which time frame anyone can help, before @MegMll totally loose track of what she had undertaken.

jorisv:
Hello @Ipuch,

Now we released pinocchio 4, I will probably be able to work again on this PR next month.
I will also check if there is some other engineer in the team able to work on it.

MegMll:
Okay i won't be available in May, but if i need to make some changes myself, i guess it can wait June, anyway

Ipuch:
Thanks @MegMll for rebasing :+1:

MegMll:
### To do
- Finish the min/max knot vectors (still need to update constructor to add min/max limits)
- Add multiple knot vector computation (open, closed ...)
- Fit spline algorithm
- Fix ubuntu CI
-  Redo changelog
- precommit

Ipuch:
@MegMll and @LucasJoseph pinocchio/algorithm/splines.hpp also need to be added into the sources.cmake

include/pinocchio/src/multibody/joint/spline-utils.hxx
jorisv, line 1: This file should not be in algorithm: it's not an algorithm we want to directly distribute to the community.
We can put it in `include/pinocchio/src/multibody/joint/splines_utils.hxx` and put struct and functions in the internal namespace (pinocchio::internal).

The casadi specialization can be put in `include/pinocchio/src/autodiff/casadi/multibody/joint/splines_utils.hxx`

jorisv (changes requested):

examples/spline-joint.py
```diff
@@ -0,0 +1,94 @@
+import time
+
+import hppfcl
```
jorisv: Use coal instead

include/pinocchio/src/multibody/joint/joint-spline.hxx
```diff
@@ -0,0 +1,697 @@
...
+          PINOCCHIO_THROW_PRETTY(
+            std::invalid_argument, "JointSpline - Knot vector must be non-decreasing (knots must "
+                                   "satisfy knots[i] <= knots[i+1]).");
+      }
+
+      min_q = knotVector[0];
+      max_q = knotVector[knots.size() - 1];
+      knots = knotVector;
+
+      ctrlFrames = controlFrames;
```
jorisv: You can copy it in the initialization list

```
    JointModelSplineTpl(
      const std::vector<Transformation_t> & controlFrames,
      const Vector & knotVector,
      const size_t degree)
    : degree(degree)
    , knots(knotVector)
    , ctrlFrames(controlFrames)
```

include/pinocchio/src/multibody/joint/joint-spline.hxx
```diff
@@ -0,0 +1,697 @@
...
+             && other.ctrlFrames == ctrlFrames && other.knots == knots
+             && other.relativeMotions == relativeMotions;
+    }
+
+    template<typename ConfigVector>
+    void calc(JointDataDerived & data, const Eigen::MatrixBase<ConfigVector> & qs) const
+    {
+      assert(
+        check_expression_if_real<Scalar>(qs[0] >= knots[0] && qs[0] <= knots(knots.size() - 1))
+        && "Spline joint configuration (q) must be between 0 and 1. ");
```
jorisv:
```suggestion
      assert(
        check_expression_if_real<Scalar>(qs[0] >= min_q && qs[0] <= max_q)
        && "Spline joint configuration (q) must be between min_q and max_q.");
```

include/pinocchio/src/multibody/joint/joint-spline.hxx
```diff
@@ -0,0 +1,697 @@
...
+
+    template<typename ConfigVector, typename TangentVector>
+    void calc(
+      JointDataDerived & data,
+      const Eigen::MatrixBase<ConfigVector> & qs,
+      const Eigen::MatrixBase<TangentVector> & vs) const
+    {
+      assert(
+        check_expression_if_real<Scalar>(qs[0] >= knots[0] && qs[0] <= knots(knots.size() - 1))
+        && "Spline joint configuration (q) must be between min and max. ");
```
jorisv:
```suggestion
      assert(
        check_expression_if_real<Scalar>(qs[0] >= min_q && qs[0] <= max_q)
        && "Spline joint configuration (q) must be between min_q and max_q.");
```

include/pinocchio/src/multibody/joint/joint-spline.hxx
```diff
@@ -0,0 +1,697 @@
...
+
+      res.min_q = static_cast<NewScalar>(min_q);
+      res.max_q = static_cast<NewScalar>(max_q);
+      res.knots = knots.template cast<NewScalar>();
+
+      res.setIndexes(id(), idx_q(), idx_v(), idx_vExtended());
+      return res;
+    }
+
+    void computeRelativeMotions()
```
jorisv, line 362: Should be private

include/pinocchio/src/multibody/joint/joint-spline.hxx
```diff
@@ -0,0 +1,697 @@
...
+    {
+      for (size_t i = 0; i < nbCtrlFrames - 1; i++)
+        relativeMotions.push_back(log6(ctrlFrames[i].inverse() * ctrlFrames[i + 1]));
+    }
+
+    void computeBasisFunctions(
```
jorisv: Should be private

include/pinocchio/src/multibody/joint/joint-spline.hxx
```diff
@@ -0,0 +1,697 @@
...
+      for (size_t i = indexes.start_idx; i < indexes.end_idx; i++)
+      {
+        data.N[i] = bsplineBasis(i, degree, joint_q_val);
+        data.N_der[i] = bsplineBasisDerivative(i, degree, joint_q_val);
+        if (computeSecondDerivative)
+          data.N_der2[i] = bsplineBasisDerivative2(i, degree, joint_q_val);
+      }
+    }
+
+    void computeTransformations(
```
jorisv: Should be private

include/pinocchio/src/multibody/joint/joint-spline.hxx
```diff
@@ -0,0 +1,697 @@
...
+      }
+      if (computeVelocity)
+      {
+        // C = Sdot * qdot = (dS/dq * qdot) * dot
+        data.c = data.c * data.joint_v[0] * data.joint_v[0];
+        data.v = data.S * data.joint_v;
+      }
+    }
+
+    Scalar bsplineBasis(size_t i, size_t k, const Scalar x) const
```
jorisv: `bsplineBasis`, `bsplineBasisDerivative` and `bsplineBasisDerivative2` should be standalone function in `internal` namespace in `spline-utils.hxx`. So they are easily testable and not in public API.

include/pinocchio/src/multibody/joint/joint-spline.hxx
```diff
@@ -0,0 +1,697 @@
...
+    {
+      knots_.resize(knots.size());
+      for (size_t i = 0; i < knots.size(); ++i)
+        knots_[i] = knots[i];
+      knot_policy_ = KnotPolicy::Custom;
+
+      return *this;
+    }
+
+    JointModelSplineBuilderTpl & withRange(const Scalar min_q, const Scalar max_q)
```
jorisv: Range should be a parameter of `withOpenUniformKnots` and `withUniformKnots`. It's only make sens when we call these functions.

jorisv (changes requested):

src/parsers/graph/model-graph-algo.cpp
```diff
@@ -243,6 +246,14 @@ namespace pinocchio
         {
           return boost::apply_visitor(*this, joint.secondary_joint);
         }
+        ReturnType operator()(const JointSpline & joint) const
+        {
+          return JointModelSplineBuilder()
+            .withControlFrameVector(joint.ctrlFrames)
+            .withDegree(joint.degree)
+            .withRange(joint.min_q, joint.max_q)
+            .build();
```
jorisv: See `graph::JointSpline` discussion

include/pinocchio/src/parsers/graph/joints.hxx
```diff
@@ -248,6 +248,49 @@ namespace pinocchio
       }
     };
 
+    struct JointSpline
```
jorisv, line 251: Hmm, this one is a little hard with the builder…
We can create a `graph::JointSplineBuilder` with the same API than JointModelSplineBuilderTpl.
`graph::JointSplineBuilder` should store all parameters to directly build a `JointModelSplineTpl` with parametrized constructor.
In this case, we should factorize the knot vector computation. You can set the knot vector computation algorithm in `spline-utils.hxx"

examples/spline-joint.py
jorisv, line 1: This should be added to `examples/CMakeLists.txt`

jorisv (changes requested):

Ipuch:
**Current todo**

- [ ] UT findspan (Joris)
- [ ] UT test q negatif (Joris)
- [ ] UT crash outside [0,1], manage that case (Joris)
- [x] What to do at "q_max+1" from theoretical point of view, open an issue (@MegMll)
- [x] Put the video into the header of the PR (@MegMll)
- [x] Give what are the constraints of spline (nb of knotvector points vs degree/order of spline) (@MegMll)
- [x] Give what is the maximum repetition in the knotvector, (verify it's degree +1) (@MegMll)
- [x] Verify the threshold of finitie differences, it's too low for now (@MegMll)

jorisv (dismissed):

jorisv (dismissed):

jorisv (approved):

PR #2786, opened 2025-10-13
Add names for joints inside a composite joint

MegMll:
## Description

This pr adds a list of names that correspond to the joints inside a composite joints.
It's especially useful for people in biomechanics to track a joint evolution, even when it's inside a composite.
## Checklist

### DIFF
--- a/include/pinocchio/bindings/python/multibody/joint/joints-models.hpp
+++ b/include/pinocchio/bindings/python/multibody/joint/joints-models.hpp
@@ -329,14 +329,14 @@ namespace pinocchio
     static context::JointModelComposite * init_proxy1(const context::JointModel & jmodel)
     {
       return boost::apply_visitor(
-        JointModelCompositeConstructorVisitor(context::SE3::Identity(), "joint_0"), jmodel);
+        JointModelCompositeConstructorVisitor(context::SE3::Identity(), "joint_1"), jmodel);
     }
 
     static context::JointModelComposite *
     init_proxy2(const context::JointModel & jmodel, const context::SE3 & joint_placement)
     {
       return boost::apply_visitor(
-        JointModelCompositeConstructorVisitor(joint_placement, "joint_0"), jmodel);
+        JointModelCompositeConstructorVisitor(joint_placement, "joint_1"), jmodel);
     }
 
     static context::JointModelComposite * init_proxy3(
--- a/include/pinocchio/bindings/python/multibody/joint/joints-models.hpp
+++ b/include/pinocchio/bindings/python/multibody/joint/joints-models.hpp
@@ -276,57 +276,76 @@ namespace pinocchio
     {
       context::JointModelComposite & m_joint_composite;
       const context::SE3 & m_joint_placement;
+      const std::string & m_name;
 
       JointModelCompositeAddJointVisitor(
-        context::JointModelComposite & joint_composite, const context::SE3 & joint_placement)
+        context::JointModelComposite & joint_composite,
+        const context::SE3 & joint_placement,
+        const std::string & name)
       : m_joint_composite(joint_composite)
       , m_joint_placement(joint_placement)
+      , m_name(name)
       {
       }
 
       template<typename JointModelDerived>
       context::JointModelComposite & operator()(JointModelDerived & jmodel) const
       {
-        return m_joint_composite.addJoint(jmodel, m_joint_placement);
+        return m_joint_composite.addJoint(jmodel, m_joint_placement, m_name);
       }
     }; // struct JointModelCompositeAddJointVisitor
 
     static context::JointModelComposite & addJoint_proxy(
       context::JointModelComposite & joint_composite,
       const context::JointModel & jmodel,
-      const context::SE3 & joint_placement = context::SE3::Identity())
+      const context::SE3 & joint_placement = context::SE3::Identity(),
+      const std::string & name = "")
     {
       return boost::apply_visitor(
-        JointModelCompositeAddJointVisitor(joint_composite, joint_placement), jmodel.toVariant());
+        JointModelCompositeAddJointVisitor(joint_composite, joint_placement, name),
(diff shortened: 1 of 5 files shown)


include/pinocchio/bindings/python/multibody/joint/joints-models.hpp
```diff
@@ -276,57 +276,76 @@ namespace pinocchio
...
+        return m_joint_composite.addJoint(jmodel, m_joint_placement, m_name);
       }
     }; // struct JointModelCompositeAddJointVisitor
 
     static context::JointModelComposite & addJoint_proxy(
       context::JointModelComposite & joint_composite,
       const context::JointModel & jmodel,
-      const context::SE3 & joint_placement = context::SE3::Identity())
+      const context::SE3 & joint_placement = context::SE3::Identity(),
+      const std::string & name = "")
```
jcarpent, line 302: Could we authorize empty names?

MegMll: Yes it's possible, so you would prefer that instead of having a default name "joint_" + index, that joints have an empty name ?

jcarpent: The question is that you should indeed have a unique identifier for each joint, otherwise, you won't be able to retrieve it. Having "joint_" + k looks good to me.
@jorisv What do you think?

include/pinocchio/multibody/joint/joint-composite.hpp
```diff
@@ -295,6 +306,13 @@ namespace pinocchio
       updateJointIndexes();
       njoints++;
 
+      std::string final_name = name;
+      if (name.empty())
+      {
+        final_name = "joint_" + std::to_string(njoints);
```
jorisv: Here the first joint will be called `joint_1` when the one created in the constructor on line 241 is called `joint_0`.
Also, if we use the constructor line 241, first joint will be called `joint_0`, and the next call of addJoint will create a `joint_2`.

include/pinocchio/multibody/joint/joint-composite.hpp
```diff
@@ -464,6 +485,16 @@ namespace pinocchio
       return res;
     }
 
+    int getJointIndex(const std::string & joint_name) const
```
jorisv:
```suggestion
    JointIndex getJointId(const std::string & joint_name) const
```
Use the same interface than `ModelTpl`. I don't like the name (getJointId to retrieve the index…) but it's the convention for now.

include/pinocchio/multibody/joint/joint-composite.hpp
```diff
@@ -464,6 +485,16 @@ namespace pinocchio
...
+    {
+      auto it = std::find(jointNames.begin(), jointNames.end(), joint_name);
+
+      if (it == jointNames.end())
+        return -1;
```
jorisv: `JointIndex` is unsigned. So this can't work. `ModelTpl::getJointId` throw an exception if the joint is not found.

unittest/joint-composite.cpp
```diff
@@ -344,4 +344,15 @@ BOOST_AUTO_TEST_CASE(test_kinematics)
...
+BOOST_AUTO_TEST_CASE(checkJointNames)
+{
+  JointModelComposite jmodel_composite;
+  jmodel_composite.addJoint(JointModelRZ())
+    .addJoint(JointModelRY(), SE3::Random(), "penultinum_joint")
+    .addJoint(JointModelRX(), SE3::Random(), "last_joint");
+
+  BOOST_CHECK(jmodel_composite.getJointIndex("last_joint") == 2);
+  BOOST_CHECK(jmodel_composite.getJointIndex("penultinum_joint") == 1);
+}
```
jorisv, line 357: Test when joint doesn't exists

jorisv (changes requested):

include/pinocchio/multibody/joint/joint-composite.hpp
MegMll: it doesn't throw when the joint is not found, it sends back the size of the vector of joints from the model. So either i do the same, or i throw. 
Let me know what you think

jorisv:
```cpp
    PINOCCHIO_CHECK_INPUT_ARGUMENT(
      (res < INT_MAX), "Id superior to int range. Should never happen.");
```
So this check is not working ? I mean, it was weird…

MegMll: i can't find INT_MAX definition.... but i tried it really quick in python with the sampleManipulator and a fake name, and the function return names.size() (ie 7)

jorisv: Ok, actually, `ModelTpl::getJointId` documentation state that the right behavior…
Anyway, I don't find it's a good one, so we can throw an exception instead.

jorisv (approved):

include/pinocchio/bindings/python/multibody/joint/joints-models.hpp
```diff
@@ -345,13 +364,19 @@ namespace pinocchio
...
           "Init JointModelComposite from a joint and a placement")
+        .def(
+          "__init__",
+          bp::make_constructor(
+            init_proxy3, bp::default_call_policies(),
+            bp::args("joint_model", "joint_placement", "name")),
+          "Init JointModelComposite from a joint, a placement and a name")
         .add_property("joints", &context::JointModelComposite::joints)
         .add_property("jointPlacements", &context::JointModelComposite::jointPlacements)
         .add_property("njoints", &context::JointModelComposite::njoints)
```
jorisv, line 375: You forgot to bind getJointId

jorisv (changes requested):

jorisv (approved):

jorisv (approved):

PR #2793, opened 2025-10-16
Add color support for robot meshes in Viser

### LINKED ISSUE #2792: [Feature]: Colors of robots are not fully applied in Viser
### 🚀 Feature

Currently, when loading robot meshes (especially .dae files), colors defined in the model are not consistently applied in the viewer.
This feature aims to ensure that all visual elements of the robot display their intended color information.

### Motivation

When visualizing robots in Viser, it’s important to distinguish each robot or component clearly, especially when working with multiple instances or analyzing visual vs collision geometry.
Proper color rendering improves debugging, scene readability, and visualization quality.

### Alternatives

Keep using default gray meshes which makes it harder to differentiate between robots and links.

Manually assign colors after loading which is inefficient and inconsistent.

### Additional context

A preliminary fix is being tested by modifying _add_mesh_from_path() in viser_visualizer.py to correctly handle effect.diffuse and pass the color data 

Before : 

Image

After : 

Image
Image

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/stack-of-tasks/pinocchio/issues)/[discussion](https://github.com/stack-of-tasks/pinocchio/discussions) in the repo.

AlexyLegrand:

## Description

This PR adds support for loading and displaying mesh colors in Viser, improving the visual fidelity of robot models.

## Checklist

### DIFF
--- a/bindings/python/pinocchio/visualize/viser_visualizer.py
+++ b/bindings/python/pinocchio/visualize/viser_visualizer.py
@@ -1,4 +1,5 @@
 import time
+from pathlib import Path
 
 try:
     import hppfcl
@@ -11,8 +12,10 @@ from .. import pinocchio_pywrap_default as pin
 from . import BaseVisualizer
 
 try:
+    import collada
     import trimesh  # Required by viser
     import viser
+
 except ImportError:
     import_viser_succeed = False
 else:
@@ -38,8 +41,8 @@ class ViserVisualizer(BaseVisualizer):
         if not import_viser_succeed:
             msg = (
                 "Error while importing the viewer client.\n"
-                "Check whether viser is properly installed "
-                "(pip install --user viser)."
+                "Check whether viser and its dependencies are properly installed.\n"
+                "Required packages: viser, trimesh, pycollada\n"
             )
             raise ImportError(msg)
 
@@ -214,20 +217,99 @@ class ViserVisualizer(BaseVisualizer):
 
         self.frames[name] = frame
 
-    def _add_mesh_from_path(self, name, mesh_path, color):
+    def _add_mesh_from_path(self, name, mesh_path, color=None):
         """
         Load a mesh from a file.
+        """
+        ext = Path(mesh_path).suffix.lower()
+
+        if ext == ".dae":
+            return self._load_collada_mesh(name, mesh_path, color)
+        else:
+            return self._load_standard_mesh(name, mesh_path, color)
+
+    def _load_collada_mesh(self, name, mesh_path, color):
+        """
+        Load a COLLADA mesh with color support.
+        """
+        try:
+            mesh_collada = collada.Collada(mesh_path)
+        except collada.DaeError:
+            return self._load_standard_mesh(name, mesh_path, color)
+
+        if len(mesh_collada.effects) < len(mesh_collada.geometries):
+            return self._load_standard_mesh(name, mesh_path, color)
+
+        frames = []
+        for i, (geometry, effect) in enumerate(
+            zip(mesh_collada.geometries, mesh_collada.effects)
+        ):
+            frame = self._process_collada_geometry(
+                name, i, geometry, effect, color, mesh_path
+            )
+            if frame:
+                frames.append(frame)
+
+        return frames[0] if frames else None
+
+    def _process_collada_geometry(
+        self, name, index, geometry, effect, fallback_color, mesh_path
+    ):
+        """
+        Process a single COLLADA geometry with its material.
+        """
+        try:
+            vertices, faces = self._extract_geometry_data(geometry)
+        except (AttributeError, IndexError, KeyError):
+            mesh = trimesh.load_mesh(mesh_path)
+            return self.viewer.scene.add_mesh_trimesh(f"{name}_{index}", mesh)
+
+        mesh_color = getattr(effect, "diffuse", None)
+
+        if mesh_color is not None:
+            return self.viewer.scene.add_mesh_simple(
+                f"{name}_{index}",
+                vertices,
+                faces,
+                color=mesh_color[:3],
+                opacity=mesh_color[3],
+            )
+        elif fallback_color is not None:
+            return self.viewer.scene.add_mesh_simple(
+                f"{name}_{index}",
+                vertices,
+                faces,
+                color=fallback_color[:3],
+                opacity=fallback_color[3],
+            )
+        else:
+            mesh = trimesh.load_mesh(mesh_path)
+            return self.viewer.scene.add_mesh_trimesh(f"{name}_{index}", mesh)
+
+    def _extract_geometry_data(self, geometry):
+        """
+        Extract vertices and faces from a COLLADA geometry.
+        """
+        vertices = geometry.primitives[0].sources["VERTEX"][0][4].data
+        indices = geometry.primitives[0].indices
+
+        if indices.ndim == 3:
+            faces = indices[:, :, 0]
+        else:
+            faces = indices.reshape(-1, 3)
+
+        return vertices, faces
+
+    def _load_standard_mesh(self, name, mesh_path, color):
+        """
+        Load a mesh using trimesh (STL and other formats).
         """
         mesh = trimesh.load_mesh(mesh_path)
         if color is None:
             return self.viewer.scene.add_mesh_trimesh(name, mesh)
         else:
             return self.viewer.scene.add_mesh_simple(
-                name,
-                mesh.vertices,
-                mesh.faces,
-                color=color[:3],
-                opacity=color[3],
+                name, mesh.vertices, mesh.faces, color=color[:3], opacity=color[3]
             )
 
     def _add_mesh_from_convex(self, name, geom, color):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - ROS: added jrl_cmakemodules dependency ([#2789](https://github.com/stack-of-tasks/pinocchio/pull/2789))
 - Add docker images ([#2776](https://github.com/stack-of-tasks/pinocchio/pull/2776))
 - Add names to joints that are inside a composite joint ([#2786](https://github.com/stack-of-tasks/pinocchio/pull/2786))
+- Add color support for robot meshes in Viser ([#2793](https://github.com/stack-of-tasks/pinocchio/pull/2793))
 
 ### Changed
 - Python version update ([#2802](https://github.com/stack-of-tasks/pinocchio/pull/2802)):
--- a/README.md
+++ b/README.md
@@ -287,6 +287,7 @@ In addition to the core dev team, the following people have also been involved i
 -   [Sebastian Castro](https://roboticseabass.com) (RAI Institute): Viser visualizer and MeshCat visualizer feature extension
 -   [Lev Kozlov](https://github.com/lvjonok): Kinetic and potential energy regressors
 -   [Simeon Nedelchev](https://github.com/simeon-ned): Pseudo inertia and Log-Cholesky parametrization
+-   [Alexy Legrand](https://www.linkedin.com/in/alexy-legrand-125889232/): Viser color bug fixes
 
 If you have participated in the development of **Pinocchio**, please add your name and contribution to this list.
 


bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -214,20 +218,106 @@ def loadViewerGeometryObject(self, geometry_object, prefix="", color=None):
 
         self.frames[name] = frame
 
-    def _add_mesh_from_path(self, name, mesh_path, color):
+    def _add_mesh_from_path(self, name, mesh_path, color=None):
         """
         Load a mesh from a file.
         """
+        ext = os.path.splitext(mesh_path)[1].lower()
```
nim65s: please use pathlib instead of os.path
ref. https://docs.python.org/3.14/library/pathlib.html

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -214,20 +218,106 @@ def loadViewerGeometryObject(self, geometry_object, prefix="", color=None):
...
+                        indices = geometry.primitives[0].indices
+                        
+                        if indices.ndim == 3:
+                            faces = indices[:, :, 0]
+                        elif indices.ndim == 2:
+                            faces = indices
+                        elif indices.ndim == 1:
+                            faces = indices.reshape(-1, 3)
+                        else:
+                            faces = indices.reshape(-1, 3)
```
nim65s: Are those intentionaly the same ? If so, couldn't they be merged ?

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -214,20 +218,106 @@ def loadViewerGeometryObject(self, geometry_object, prefix="", color=None):
...
+                        frames.append(frame)
+                        
+                    except Exception:
+                        try:
+                            mesh = trimesh.load_mesh(mesh_path)
+                            frame = self.viewer.scene.add_mesh_trimesh(f"{name}_{i}", mesh)
+                            if frame is not None:
+                                frames.append(frame)
+                        except Exception:
+                            pass
```
nim65s: could you add comments explaining in which cases we can get thoses exceptions ?

examples/gepetto-viewer.py
```diff
@@ -11,8 +11,8 @@
 # Load the URDF model.
 model_path = Path(os.environ.get("EXAMPLE_ROBOT_DATA_MODEL_DIR"))
 mesh_dir = model_path.parent.parent
-urdf_filename = "talos_reduced.urdf"
-urdf_model_path = model_path / "talos_data/robots" / urdf_filename
+urdf_filename = "panda.urdf"
+urdf_model_path = model_path / "panda_description/urdf" / urdf_filename
```
nim65s: I guess you only changed that to debug, but there is no need for this change in this PR

examples/viser-viewer.py
```diff
@@ -12,13 +11,14 @@
...
-model_path = Path(os.environ.get("EXAMPLE_ROBOT_DATA_MODEL_DIR"))
-mesh_dir = model_path.parent.parent
+model_path = pinocchio_model_dir / "example-robot-data/robots"
+mesh_dir = pinocchio_model_dir
 # urdf_filename = "talos_reduced.urdf"
 # urdf_model_path = join(join(model_path,"talos_data/robots"),urdf_filename)
-urdf_filename = "solo.urdf"
-urdf_model_path = model_path / "solo_description/robots" / urdf_filename
+urdf_filename = "tiago.urdf"
+urdf_model_path = model_path / "tiago_description/robots" / urdf_filename
```
nim65s: same

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -214,22 +219,111 @@ def loadViewerGeometryObject(self, geometry_object, prefix="", color=None):
 
         self.frames[name] = frame
 
-    def _add_mesh_from_path(self, name, mesh_path, color):
+    def _add_mesh_from_path(self, name, mesh_path, color=None):
         """
-        Load a mesh from a file.
+        Load a mesh from a file
         """
+        print("on passe la")
```
sea-bass: Remove this line

examples/viser-viewer.py
```diff
@@ -20,6 +20,7 @@
 urdf_filename = "solo.urdf"
 urdf_model_path = model_path / "solo_description/robots" / urdf_filename
 
+
```
sea-bass: Extra space not needed

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -214,22 +219,111 @@ def loadViewerGeometryObject(self, geometry_object, prefix="", color=None):
...
+        mesh_path = Path(mesh_path)
+        ext = mesh_path.suffix.lower()
+
+        if ext == ".dae":
+            try:
```
sea-bass: Consider factoring out into a helper function? This code has a lot of branching logic

sea-bass (commented):
Thanks for helping improve the minimal viser functionality I added -- some minor comments.

bindings/python/pinocchio/visualize/viser_visualizer.py
AlexyLegrand: Sorry, I forgot to delete it, careless mistake.

bindings/python/pinocchio/visualize/viser_visualizer.py
AlexyLegrand: Yes of course, I can make helper function to make the code clearer, is that right?

sea-bass: Yes, exactly!

AlexyLegrand: ok no problem, I will see and push something later

AlexyLegrand:
Hi, any update on this?

jorisv:
Hello @AlexyLegrand,
I will review your PR by the end of the week.

AlexyLegrand:
great, thank you !

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -214,20 +219,90 @@ def loadViewerGeometryObject(self, geometry_object, prefix="", color=None):
...
+                return self._load_standard_mesh(name, mesh_path, color)
+            
+            frames = []
+            for i, (geometry, effect) in enumerate(zip(mesh_collada.geometries, mesh_collada.effects)):
+                frame = self._process_collada_geometry(name, i, geometry, effect, color, mesh_path)
+                if frame:
+                    frames.append(frame)
+            
+            return frames[0] if frames else None
+        except Exception:
```
jorisv: This is really too general. Do we know which kind of exception can be raised by  `self._process_collada_geometry` ? Then we should only except on these exceptions

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -214,20 +219,90 @@ def loadViewerGeometryObject(self, geometry_object, prefix="", color=None):
...
+            return self._load_standard_mesh(name, mesh_path, color)
+
+    def _load_collada_mesh(self, name, mesh_path, color):
+        """
+        Load a COLLADA mesh with color support.
+        """
+        try:
+            mesh_collada = collada.Collada(mesh_path)
+            
+            if len(mesh_collada.effects) < len(mesh_collada.geometries):
```
jorisv: if there is more effects than geometrie, do the next loop work ?

leap.mp4
jorisv, line 1: You should delete this file

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -214,20 +219,90 @@ def loadViewerGeometryObject(self, geometry_object, prefix="", color=None):
...
+                    color=fallback_color[:3], opacity=fallback_color[3]
+                )
+            else:
+                mesh = trimesh.load_mesh(mesh_path)
+                return self.viewer.scene.add_mesh_trimesh(f"{name}_{index}", mesh)
+        except Exception:
+            try:
+                mesh = trimesh.load_mesh(mesh_path)
+                return self.viewer.scene.add_mesh_trimesh(f"{name}_{index}", mesh)
+            except Exception:
```
jorisv: This is really too general. You should target a more specialized exception.

jorisv (changes requested):

bindings/python/pinocchio/visualize/viser_visualizer.py
AlexyLegrand: The `zip()` on line 244 iterates over the minimum of both lists:

- If `effects < geometries`: some geometries won't be matched → that's why we fallback to trimesh here
- If `effects >= geometries`: all geometries get an effect → the loop processes them correctly

bindings/python/pinocchio/visualize/viser_visualizer.py
AlexyLegrand: Would catching specific exceptions like this work better?

```python
except (ValueError, KeyError, IndexError, AttributeError) as e:
```

These cover the typical cases when accessing geometry primitives, sources, or indices that may not exist or have unexpected shapes. Are there other specific exceptions you'd recommend catching here?

bindings/python/pinocchio/visualize/viser_visualizer.py
AlexyLegrand: Would it be better to catch specific exceptions like:
```python
except (IOError, ValueError) as e:
```

These cover file access issues and mesh parsing errors. Should I also include other specific exceptions here?

bindings/python/pinocchio/visualize/viser_visualizer.py
jorisv: All these exceptions can be triggered by some programming error. `AttributeError` can be raised because of a typo when calling `frames.append`. If you catch this exception, you will never catch the root error and you will end-up always executing the fallback.

I think you should remove this kind of exception handling.

I can accept something more granular like:
```python
try:
  mesh_collada = collada.Collada(mesh_path)
except collada.DaeError:
  # ...
```

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -7,12 +8,16 @@
 
 import numpy as np
 
+from pathlib import Path
+from trimesh.exchange import dae
```
jorisv:
```suggestion
```
Unused

bindings/python/pinocchio/visualize/viser_visualizer.py
jorisv, line 1: You can modify the error message line 45 because the import can also fail because of collada.

jorisv, line 1: I try to display go1 and nothing is shown. Can you provide me a working model ?

jorisv (changes requested):

bindings/python/pinocchio/visualize/viser_visualizer.py
AlexyLegrand: Can you try it with Talos?

bindings/python/pinocchio/visualize/viser_visualizer.py
AlexyLegrand: I've changed this; let me know if it's what you were expecting or not.

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -38,8 +42,8 @@ def __init__(
         if not import_viser_succeed:
             msg = (
                 "Error while importing the viewer client.\n"
-                "Check whether viser is properly installed "
-                "(pip install --user viser)."
+                "Check whether viser and its dependencies are properly installed.\n"
+                "Required packages: viser, trimesh, collada\n"
```
jorisv:
```suggestion
                "Required packages: viser, trimesh, pycollada\n"
```

bindings/python/pinocchio/visualize/viser_visualizer.py
jorisv: Your fix on `_load_collada_mesh` is good.
You can apply the same philosophy here: don't try/execpt on the full function but around some call that can throw a particular exception.

jorisv (changes requested):
Don't forget to update CHANGELOG.md and the Credits section in README.md

bindings/python/pinocchio/visualize/viser_visualizer.py
AlexyLegrand: I agree to change the CHANGELOG.md, but as long as my solution is not 100% approved, do I still have to modify the credits?

jorisv: Yes, you have to. If your solution is not approved, the MR will not be merged and credits will not be modified.

bindings/python/pinocchio/visualize/viser_visualizer.py
```diff
@@ -214,20 +218,91 @@ def loadViewerGeometryObject(self, geometry_object, prefix="", color=None):
...
+        elif fallback_color is not None:
+            return self.viewer.scene.add_mesh_simple(
+                f"{name}_{index}", vertices, faces,
+                color=fallback_color[:3], opacity=fallback_color[3]
+            )
+        else:
+            try:
+                mesh = trimesh.load_mesh(mesh_path)
+            except (IOError, ValueError):
+                return None
```
jorisv: Here you return None if `trimesh.load_mesh` fail. But function that `_add_mesh_from_path` didn't check the return type.

In my point of view, you should remove the try/except block here. If an error occur in `trimesh.load_mesh`, an exception will be raised, the program (that is not safety critical) will end and the exception will be displayed in the user terminal (and probably will show him that a mesh path didn't exists).

AlexyLegrand: Okay, no problem, I'll delete it.

jorisv:
Look good to me. I will rebase and probably squash your commit history since at some point you added an .mp4 file and I don't want ta pollute the git history with that.

AlexyLegrand:
Okay, thank you. If needed, I'm always available if there's a problem.

jorisv (approved):

PR #2800, opened 2025-11-03
Update SDF support include path in CMakeLists

roncapat:
JRL gets fetched but then its cmake file for sdformat is not used.

TODO:
- [x] check if it's ok for a ROS installation
  Note: this may require an if/else depending on the detection of a ROS 2 build (e.g. via colcon)?

## Description

source build (non-ROS workspace) - build fails as the `sdformat.cmake` is expected in an inexistent position in source folder.

## Checklist


jcarpent (dismissed):

jorisv:
Hello @roncapat,

Thank for the contribution.
You can add an entry in the changelog.md. Something like : Fix SDFormat support when jrl-cmakemodules are not used as a submodule.

jorisv (approved):

jcarpent (approved):

PR #2802, opened 2025-11-07
Update pixi.toml

jorisv:
## Description

- [x] Use python 3.10 as oldest python
- [x] Use python 3.14 as latest python version
- [x] Remove latest pixi warning
- [x] Remove some activation scripts
- [x] Improve ccache statistics
- [x] Improve language server integration
- [x] Setup pixi build
- [x] Use conda clang-cl on windows build

## Checklist


pixi.toml
```diff
@@ -30,6 +29,75 @@ matplotlib = ">=3.9.2"
...
+[package.build.config]
+extra-args = [
+  "-DGENERATE_PYTHON_STUBS=ON",
+  "-DBUILD_TESTING=OFF",
+  "-DBUILD_WITH_COLLISION_SUPPORT=ON",
+  "-DBUILD_WITH_ACCELERATE_SUPPORT=OFF",
+  "-DBUILD_WITH_CASADI_SUPPORT=OFF",
+  "-DBUILD_WITH_AUTODIFF_SUPPORT=OFF",
+  "-DBUILD_WITH_EXTRA_SUPPORT=OFF",
+  "-DBUILD_WITH_OPENMP_SUPPORT=OFF",
```
ManifoldFR, line 48: We turn this and casadi off in the pixi-build by default, because these are the defaults in our CMake listfile?

pixi.toml
```diff
@@ -30,6 +29,75 @@ matplotlib = ">=3.9.2"
...
+eigen = ">=3.4.0"
+numpy = ">=1.22.0"
+python = ">=3.10.0"
+eigenpy = ">=3.7.0"
+urdfdom = ">=4.0"
+hpp-fcl = ">=3"
+
+[package.run-dependencies]
+libboost-devel = ">=1.80.0"
+libboost-python-devel = ">=1.80.0"
```
ManifoldFR, line 89: do we have to include these as run dependencies?

ManifoldFR (approved):
I only have a couple of questions that should be answered here to have some kind of public trace of the answers but otherwise it's good to go imo

pixi.toml
jorisv: Because only want to build a « minimal » pinocchio package. I think collision and parsers are the bare minimum.

pixi.toml
jorisv: We should not but it's what we are doing in the conda-forge pinocchio package. Why ? Because consumer project except to be able to do a `find_package(pinocchio)`  without adding more dependencies.
If we don't add `libboost-devel` as a run dependencies, then the `find_package(pinocchio)` will not be able to find boost.

ManifoldFR: Makes sense.

jcarpent (approved):

PR #2807, opened 2025-11-20
Fix appendModel issue when a frame have non null transformation

### LINKED ISSUE #2805: modelAppend with URDF-Submodel with fixed Joints
Using Python (Pinocchio version: 3.7.0) 
I get unexpected behaviour with a trivial URDF that consists of just two Links that are connected via a fixed joint, that seperates the second link 1m in x-direction via fixed joint origin:

```xml
<?xml version="1.0"?>
<robot name="two_fixed_bodies">

  <!-- Link 1: 1000 kg cube -->
  <link name="link_1">
    <inertial>
      <origin xyz="0 0 0" rpy="0 0 0"/>
      <mass value="1000.0"/>
      <inertia
        ixx="166" ixy="0.0" ixz="0.0"
        iyy="166" iyz="0.0"
        izz="166"/>
    </inertial>
  </link>

  <!-- Link 2: 1000 kg cube -->
  <link name="link_2">
    <inertial>
      <origin xyz="0 0 0" rpy="0 0 0"/>
      <mass value="1000.0"/>
      <inertia
        ixx="166" ixy="0.0" ixz="0.0"
        iyy="166" iyz="0.0"
        izz="166"/>
    </inertial>
  </link>

  <!-- Fixed joint from link1 to link2 -->
  <joint name="urdf_joint1_fixed" type="fixed">
    <parent link="link_1"/>
    <child link="link_2"/>
    <origin xyz="1 0 0" rpy="0 0 0"/>
  </joint>

</robot>
```

When I attach this URDF to a free-flyer directly via
```python
model_urdf = pin.buildModelFromUrdf(r"URDF_FILE_PATH_HERE", pin.JointModelFreeFlyer())
```  
I get the expected Center-Of-Mass (inertia.lever) at the mid-position x = 0.5m.

I expected the same outcome when using the appendModel function like this:
```python
model_root = pin.Model()

model_combined = pin.appendModel(
    model_root,   # model A
    pin.buildModelFromUrdf(r"./fixed_joint_two_InertialTags_git.urdf", pin.JointModelFreeFlyer()),     # model B
    0, # index of the frame of modelA where to append modelB.
    pin.SE3.Identity() # pose of modelB universe joint (index 0) in frameInModelA.
)
```
But with it i get a COM of x = 1.0m.

Following script can be used for reproducing the behaviour: [debug_pin_fixedJoints_git.py](https://github.com/user-attachments/files/23527607/debug_pin_fixedJoints_git.py)

When the URDF Joint-Type is changed from "fixed" --> "continuous", both methods result in the same model.
**Is this comprehensible and I am missing something, or is this a bug?**

jorisv:
## Description

When a frame have inertia and a non null transformation, `appendModel` algorithm was doing some wrong computation to compute joint inertia of the new transform.

Close #2805 

## TODO
- [x] Fix `AppendJointOfModelAlgoTpl`
- [x] Fix  `appendUniverseToModel`
- [x] Check if there is an issue with eometry

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Remove pixi 0.57 warnings ([#2802](https://github.com/stack-of-tasks/pinocchio/pull/2802))
 - CMake: use `sdformat.cmake` from `jrl-cmakemodules` ([#2800](https://github.com/stack-of-tasks/pinocchio/pull/2800))
 - Fix doxygen document hierarchy ([#2811](https://github.com/stack-of-tasks/pinocchio/pull/2811))
-- Fix appendModel issue when a frame have non null transformation ([#2807](https://github.com/stack-of-tasks/pinocchio/pull/2807)
+- Fix `appendModel` issues ([#2807](https://github.com/stack-of-tasks/pinocchio/pull/2807)):
+  - Bad inertia transformation when multiple frame with inertia have non null placement
+  - Undefined behavior when more than two frame was attached to universe
 
 ### Removed
 - Remove CMake < 3.22 details ([#2790](https://github.com/stack-of-tasks/pinocchio/pull/2790))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Remove pixi 0.57 warnings ([#2802](https://github.com/stack-of-tasks/pinocchio/pull/2802))
 - CMake: use `sdformat.cmake` from `jrl-cmakemodules` ([#2800](https://github.com/stack-of-tasks/pinocchio/pull/2800))
 - Fix doxygen document hierarchy ([#2811](https://github.com/stack-of-tasks/pinocchio/pull/2811))
+- Fix appendModel issue when a frame have non null transformation ([#2807](https://github.com/stack-of-tasks/pinocchio/pull/2807)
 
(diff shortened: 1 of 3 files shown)


martisalcedo7:
Shouldn't this be changed as well in the `appendUniverseToModel()` method?

jorisv:
@martisalcedo7 I'm currently looking at it but frame with inertia attached to universe is kind of weird. Do you see an use case of it ?

martisalcedo7:
> @martisalcedo7 I'm currently looking at it but frame with inertia attached to universe is kind of weird. Do you see an use case of it ?

It is weird but it can be there. We have to make sure this inertia is not given to the body with addFrame(), therefore the argument append_inertia should be set to false.

jorisv:
Ok, after discussion it can model the inertia of the base of a manipulator.

martisalcedo7:
Can we also apply the fix here for the geometric bodies?

They are expressed with respect to the frames, not with respect to the joints, so this transformation is wrong. It was also shown in the issue.

In the function `appendUniverseToModel()`

`go.placement = (pframe.placement * pfMAB) * go.placement;`

jorisv:
As said in the issue I'm waiting to have an example to reproduce it.

martisalcedo7:
> As said in the issue I'm waiting to have an example to reproduce it.

I am working on that.

martisalcedo7:
It was a bug in our side. Please omit this geometry problem.

JTem:
i also confirm, ignore the geometry issue

jcarpent (dismissed):

EtienneAr (dismissed):

EtienneAr (approved):

PR #2809, opened 2025-11-20
Add Realtime Sanitizer (RTsan) support

ahoarau:
## Description

This PR adds initial support for the `Realtime Sanitizer (RTSan)` that appeared in clang 20. It allow to check for the "realtime-safeness", by detecting dynamic memory allocations and source of non-determinism (like mutex lock, systems calls etc).
Before going all-in and instrument all the code with `[[clang::nonblocking]]`, we start by writing a unittest that check for a the main  API functions. 

To test:
```
pixi run -e sanitizers test
```

Typical output might look like this:
image

https://clang.llvm.org/docs/RealtimeSanitizer.html
https://clang.llvm.org/docs/FunctionEffectAnalysis.html
https://github.com/realtime-sanitizer/rtsan

## Checklist


unittest/dynamic-allocations.cpp
```diff
@@ -0,0 +1,468 @@
...
+
+BOOST_AUTO_TEST_SUITE(BOOST_TEST_MODULE)
+
+BOOST_AUTO_TEST_CASE(dynamic_allocations)
+{
+  using namespace pinocchio;
+
+  // Build model
+  Model model;
+  buildModels::humanoidRandom(model, true);
```
jorisv: We should parametrize this test with different model.
Joint (mimic, composite) can have some hidden allocation.
Also, mimic only support the following algorithms:
- RNEA
- CRBA
- Forward Kinematics
- Jacobians and Frames
- Centroidal Algorithm (ccrba)

pixi.toml
```diff
@@ -349,3 +349,61 @@ new-version = { features = [
...
+  "-G",
+  "Ninja",
+  "-B",
+  "build",
+  "-S",
+  ".",
+  "-DCMAKE_BUILD_TYPE=Debug",
+  "-DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX",
+  "-DBUILD_TESTING=ON",
+  "-DBUILD_BENCHMARK=ON",
```
jorisv: Why did you build benchmark ?

jorisv (changes requested):

.github/workflows/macos-linux-windows-pixi.yml
jorisv, line 300: You can add you job in this list (and pinocchio-pixi-build and   pinocchio-python-standalone-pixi that I have forget)

.github/workflows/macos-linux-windows-pixi.yml
```diff
@@ -278,3 +278,20 @@ jobs:
         uses: re-actors/alls-green@release/v1
         with:
           jobs: ${{ toJSON(needs) }}
+
+  rtsan-build:
+    name: Realtime Sanitizer
+    runs-on: [ubuntu-latest, macos-latest]
+
```
jorisv, line 269:
```suggestion

    strategy:
      fail-fast: false
      
```

unittest/dynamic-allocations.cpp
ahoarau: Added tests with additional robots. Allocations detected !

.github/workflows/macos-linux-windows-pixi.yml
ahoarau: done 👍

ahoarau:
Added a warning annotation in the CI when RTSan unittest failed.

Squashed the commits.

warn

unittest/dynamic-allocations.cpp
jorisv: Mimic are only supported in few algorithm (listed in my previous comment).
Unsupported algo will throw and then biasing the test.

.github/workflows/macos-linux-windows-pixi.yml
```diff
@@ -264,12 +263,51 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            name: Ubuntu (x86_64)
+
+          - os: macos-latest
+            name: MacOS (ARM64)
```
jorisv, line 278:
```suggestion
        os: [ubuntu-latest, macos-latest]
```

`name` is never used.

jorisv (changes requested):

pixi.toml
ahoarau: turned to OFF 👍

.github/workflows/macos-linux-windows-pixi.yml
ahoarau: Using name instead 👍

unittest/dynamic-allocations.cpp
ahoarau: refactored the test to only execute certain tests if the model has composites or mimic joints. 👍 
I'm waiting for the green CI, then I'll squash some commits

jorisv (approved):

PR #2814, opened 2025-11-26
Change MimicChecker return value

ahoarau:
## Description

Changes condition in `MimicChecker`, where it did the exact opposite of what I think it was meant to do.

Before: MimicChecker returned false if the model has mimic joints.

After: MimicChecker checks that the models has mimic joints.

Alternative: rename it to NoMimicChecker

## Checklist


ahoarau:
@jorisv https://github.com/stack-of-tasks/pinocchio/blob/devel/unittest/algo-check.cpp#L58

This seem to have been designed like that, but the MimicChecker seem to mean "check that the model has mimic joints", but returns false if it has.. Go for updating the condition ?

ahoarau:
The `ParentCheck` seem to behave like the `MimicChecker`. Both would benefit from explicit renaming. In a next PR if useful.

Issue #2480, opened 2024-11-21
hpp-fcl to coal migration


v4hn:
hpp-fcl just finished its rebranding to [coal](https://github.com/coal-library/coal/tree/v3.0.0) with version 3.0.

Are there plans when and where to migrate to the new name/version?

jcarpent:
@v4hn We will soon update the CMake files of Pinocchio to link directly to coal. I appreciate your understanding.

AndreasKuhner:
Hi @jcarpent ,
when do you think you adapt the header files to include the `coal` parts instead of `fcl`? 

Or does `DCOAL_BACKWARD_COMPATIBILITY_WITH_HPP_FCL=ON` already handle warning you receive?

Our problem: We use pinocchio for its collision in the ROS 2 environment and it might be that it treats warnings as errors... 😓 

Cheers,
Andreas

jcarpent:
@jorisv Could you answer @AndreasKuhner questions?

jorisv:
Hello @AndreasKuhner,

>  when do you think you adapt the header files to include the coal parts instead of fcl?

I can't answer this question. We will switch to coal when some big changed will be merged into pinocchio. I don't know when these change will be ready.

> Our problem: We use pinocchio for its collision in the ROS 2 environment and it might be that it treats warnings as errors... 😓

The hpp-fcl/coal compatibility layer use `#pragma message`. This is not a warning and `-Werror` doesn't convert it into an error (I tested that on g++11).

Can you provide some logs ?

AndreasKuhner:
Sure, if you run something which is using Pinocchio via `cmake -DSTRICT=ON  ..`, you will get something like 

```
/opt/openrobots/include/hpp/fcl/coal.hpp:13:2: error: #warning "Please update your includes from 'hpp/fcl' to 'coal'" [-Werror=cpp]
   13 | #warning "Please update your includes from 'hpp/fcl' to 'coal'"
      |  ^~~~~~~
```

aka 'warning' as 'error'

nim65s:
My bad, I need to update robotpkg. I'll do thit asap

AndreasKuhner:
Thank you guys. This will resolve our issues for now and will be gone eventually after your next bigger release 🥳

jorisv:
Done in pinocchio 4

PR #2794, opened 2025-10-17
Update renamed dependency

christophfroehlich:
## Description

hpp-fcl was renamed to coal, see https://github.com/coal-library/coal/blob/devel/README.md
To confirm, the old git repo is redirected to the new org/repo https://github.com/humanoid-path-planner/hpp-fcl.git

As it was not removed from [rosdisto](https://github.com/ros/rosdistro/blob/484d839eeee81d081ea50749dc4ee9f983e021e2/rolling/distribution.yaml#L2907-L2916), rosdep installs the old dependency.

```
$ sudo apt list |grep hpp-fcl

ros-jazzy-hpp-fcl-dbgsym/noble 2.4.5-1noble.20251014.183513 amd64
ros-jazzy-hpp-fcl/noble 2.4.5-1noble.20251014.183513 amd64
ros-kilted-hpp-fcl-dbgsym/noble 2.4.5-2noble.20251014.170938 amd64
ros-kilted-hpp-fcl/noble 2.4.5-2noble.20251014.170938 amd64
ros-rolling-hpp-fcl-dbgsym/noble 2.4.5-1noble.20251014.170731 amd64
ros-rolling-hpp-fcl/noble 2.4.5-1noble.20251014.170731 amd64 
```

```
$ sudo apt list |grep coal

ros-jazzy-coal-dbgsym/noble 3.0.2-1noble.20251014.185600 amd64
ros-jazzy-coal/noble 3.0.2-1noble.20251014.185600 amd64
ros-kilted-coal-dbgsym/noble 3.0.2-1noble.20251015.151509 amd64
ros-kilted-coal/noble 3.0.2-1noble.20251015.151509 amd64
ros-rolling-coal-dbgsym/noble 3.0.2-1noble.20251014.185949 amd64
ros-rolling-coal/noble 3.0.2-1noble.20251014.185949 amd64
```

## Checklist

- [x] I have run `pre-commit run --all-files` or `pixi run lint`
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code where necessary
- [ ] I have made corresponding changes to the doxygen documentation
- [ ] I have added tests that prove my fix or feature works
- [x] I have updated the [CHANGELOG](https://github.com/stack-of-tasks/pinocchio/blob/devel/CHANGELOG.md) or added the "no changelog" label if it's a CI-related issue
- [ ] I have updated the [README credits section](https://github.com/stack-of-tasks/pinocchio?tab=readme-ov-file#credits)

PR #2822, opened 2025-12-04
Update `package-xml` to introduce colcon dependency over "coal" package

roncapat:
Maybe related to https://github.com/stack-of-tasks/pinocchio/issues/2480

When building `pinocchio` in a Colcon workspace (the ROS 2 way), a dependency on `hpp-fcl` is declared in `package.xml`. However, `hpp-fcl` is now `coal`.

`coal` needs to be built with retrocompatibility enabled. A specific CMake helper is installed only when `COAL_BACKWARD_COMPATIBILITY_WITH_HPP_FCL=ON`, thus the full command line has to be (assuming this patch is applied, to tell colcon that `pinocchio` package needs to wait for `coal` build):

```
colcon build --packages-select coal pinocchio --cmake-args -DCOAL_BACKWARD_COMPATIBILITY_WITH_HPP_FCL=ON -DBUILD_WITH_COLLISION_SUPPORT=ON
```

Wouldn't be better then also to update `find_package( )` directives to actively look for `coal` instead of `hpp-fcl`?

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -25,7 +25,7 @@
   <depend>eigen</depend>
   <depend>boost</depend>
   <depend>eigenpy</depend>
-  <depend>hpp-fcl</depend>
+  <depend>coal</depend>
 
   <!-- Compiling Pinocchio with clang reduces build time, required memory, and results in faster runtime. Note: This only specifies the build dependency, it does not auto-select the compiler - this needs to be specified manually -->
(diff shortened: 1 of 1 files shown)


nim65s:
Hi @roncapat,
Sorry for this, but we'll have to wait for pinocchio v4 for this.
ref.  #2794

PR #2831, opened 2025-12-31
Add RoboPlan, PlaCo, and CRISP controllers to open source projects

sea-bass:
## Description

Adds [RoboPlan](https://github.com/open-planning/roboplan) (which is my project) to the README of open-source projects, as well as 2 other cool projects I'm aware of that use Pinocchio but are not yet on this list.

No CHANGELOG needed.

## Checklist


jorisv (approved):

sea-bass:
Seems these checks that aren't running are because of the `if: always()` in here:

https://github.com/stack-of-tasks/pinocchio/blob/c9a03734449d754bccd67e40563c78d6aab0a238/.github/workflows/linux.yml#L220

https://github.com/stack-of-tasks/pinocchio/blob/c9a03734449d754bccd67e40563c78d6aab0a238/.github/workflows/macos-linux-windows-pixi.yml#L303

I guess there needs to be some other gating for when to properly run these jobs.

There are more complicated solutions, but an easy one is just removing the `if: always()` -- if the previous jobs fail, then this job won't even run which maybe is okay? I know it's a slightly different behavior...

PR #2832, opened 2025-12-31
Ignore non-source files in Nix CI job

sea-bass:
## Description

I noticed that a README change in https://github.com/stack-of-tasks/pinocchio/pull/2831 caused the Nix jobs to run even though other source-build ones were bypassed... so I just copy-pasted the pattern in other jobs here too, for example [this](https://github.com/stack-of-tasks/pinocchio/blob/4a0e2c40e6991173a40e587e1adff7f974172df3/.github/workflows/linux.yml#L4-L25).

No CHANGELOG needed here.

## Checklist

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -4,7 +4,25 @@ on:
   push:
     branches:
       - devel
+    paths-ignore:
+      - doc/**
+      - .gitlab-ci.yml
+      - .gitignore
+      - '**.md'
+      - CITATION.*
+      - COPYING.LESSER
+      - colcon.pkg
+      - .pre-commit-config.yaml
   pull_request:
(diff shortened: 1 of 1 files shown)


nim65s:
Hi @sea-bass,

This was on-purpose, because this feature is already handled by nix cache, which did work in that PR of yours, as nothing was actually compiled:

image

It should not arm to merge this, but earlier I told @jorisv not to implement this here because at some point we had an issue with it, but I can't remember exactly what.

Therefore, I'm fine with both merging it now or closing it.

sea-bass:
I'm OK closing this based on your description. Thank you!

nim65s:
Thanks !
Following discussions in internal meeting, we'll close it then

PR #2835, opened 2026-01-13
[ROS-O] Remove unused ROS_VERSION condition in package.xml

v4hn:
The ROS1 alternative was removed from the package.xml [here](https://github.com/stack-of-tasks/pinocchio/commit/8ca2ca37e2cbc3b830348d7e87216143daddd99f) which broke the dependency in ROS-O.
But the `urdfdom` dependency works great with ROS-O and ROS2 as well, so the conditional is not just obsolete, but actually harmful and breaks builds.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -21,7 +21,7 @@
   <depend>ros_environment</depend>
   <depend>python3</depend>
   <depend>python3-numpy</depend>
-  <depend condition="$ROS_VERSION == 2">urdfdom</depend>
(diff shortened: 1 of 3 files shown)


nim65s (dismissed):
Thanks !

jcarpent (dismissed):

jorisv (dismissed):

jorisv (approved):

PR #2836, opened 2026-01-14
Unconditionally depend on urdfdom

rhaschke:
`BUILD_WITH_URDF_SUPPORT` is ON by default.


traversaro:
Similar to https://github.com/stack-of-tasks/pinocchio/pull/2835/changes ?

rhaschke:
Sure, it's the same. Sorry, that I didn't check beforehand.

PR #2728, opened 2025-07-15
CMake: allow use of system example-robot-data

nim65s:
first step for #2575

PR #2865, opened 2026-03-25
examples: fix more use of system example-robot-data

nim65s:
ref. #2728

## Description

fix for system example-robot-data eg. 
```python
58/80 Test #65: pinocchio-example-py-contact-cholesky ........................***Failed    0.32 sec
Error:   File /build/source/models/example-robot-data/robots/anymal_b_simple_description/robots/anymal.urdf does not exist
          at line 56 in /build/source/urdf_parser/src/model.cpp
   File "/build/source/examples/contact-cholesky.py", line 10, in <module>
     model = pin.buildModelFromUrdf(urdf_filename)
ValueError: The file /build/source/models/example-robot-data/robots/anymal_b_simple_description/robots/anymal.urdf does not contain a valid URDF model.
```

## Checklist

### DIFF
--- a/examples/contact-cholesky.py
+++ b/examples/contact-cholesky.py
@@ -1,13 +1,12 @@
+import os
 from pathlib import Path
 
 import pinocchio as pin
 
-pinocchio_model_dir = Path(__file__).parent.parent / "models"
-urdf_filename = (
-    pinocchio_model_dir
-    / "example-robot-data/robots/anymal_b_simple_description/robots/anymal.urdf"
-)
-model = pin.buildModelFromUrdf(urdf_filename)
+model_path = Path(os.environ.get("EXAMPLE_ROBOT_DATA_MODEL_DIR"))
+urdf_filename = "anymal.urdf"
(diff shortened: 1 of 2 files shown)


examples/contact-cholesky.py
```diff
@@ -1,13 +1,12 @@
...
 
 import pinocchio as pin
 
-pinocchio_model_dir = Path(__file__).parent.parent / "models"
-urdf_filename = (
-    pinocchio_model_dir
-    / "example-robot-data/robots/anymal_b_simple_description/robots/anymal.urdf"
-)
-model = pin.buildModelFromUrdf(urdf_filename)
+model_path = Path(os.environ.get("EXAMPLE_ROBOT_DATA_MODEL_DIR"))
```
ahoarau, line 6: Maybe add an assert to check if model_path actually exists ?

nim65s: why not, but not in this PR, as the goal is to keep it minimal (before pinocchio 4 release and huge changes everywhere),
and this would have to be done in all tests, examples and benchmarks.

PR #2873, opened 2026-04-13
Upgrade Pinocchio Python bindings to nanobind

### LINKED ISSUE #2616: Annoucement - Switch to nanobind
We are in the process of switching from Boost.Python to [nanobind](https://github.com/wjakob/nanobind) in Pinocchio. The overall objective is, among other objectives, to:
- reduce the size of Python modules with similar or improved performances compared to Boost.Python
- simplify Python exposition
- facilitate Python documentation
- lower technical debt (wrt no longer maintained Boost.Python library)
- simplify CI with a unique use of stable Python API

ManifoldFR:
## Description

This pull request adds a new set of Python bindings using the modern [nanobind](https://github.com/wjakob/nanobind) Python bindings framework.

The new set of bindings is a full replacement of the Boost.Python bindings, which will be deprecated in the next major version of Pinocchio.

This PR is the sequel to https://github.com/coal-library/coal/pull/659 on the Coal library, and actually depends on the Coal nanobind bindings as well as nanoeigenpy.

## Checklist

## Task checklist

- [x] Math
  - [x] RPY (with submodule)
  - [x] Tridiagonal matrix
  - [x] Linalg (eigenvalue/eigenvector/inverse)
  - [x] Lanczos decomposition
  - [x] Gram-Schmidt
- [x] Spatial
    - [x] Inertia
        - [x] Inertia
        - [x] LogCholeskyParams
        - [x] PseudoInertia
    - [x] SE3
    - [x] Motion
    - [x] Symmetric
    - [x] explog
- [x] Multibody
  - [x] Model
  - [x] Data
  - [x] GeometryModel
  - [x] GeometryData
  - [x] Lie groups
  - [x] Joints
    - [x] Generic `JointModel`/`JointData`
    - [x] Variant + base visitor
    - [x] Derived classes (with specific add-ons)
    - [x] Joint Lie groups
  - [x] Pool (`class ModelPool`...)
- [x] Algorithm
  - [x] ABA
  - [x] ABA derivatives
  - [x] CAT (`computeAllTerms()`)
  - [x] Cholesky
  - [x] COM
  - [x] Centroidal
  - [x] Centroidal derivatives
  - [x] CRBA
  - [x] LCABA (added after https://github.com/Simple-Robotics/pinocchio/pull/32)
  - [x] Constrained dynamics
      - [x] Constraint Cholesky
      - [x] Delassus operators
  - [x] Constrained dynamics derivatives
  - [x] Constraint solvers
      - [x] ADMM
      - [x] PGS
  - [x] Contact dynamics
  - [x] Contact inverse dynamics
  - [x] Contact Jacobian
  - [x] Geometry (`updateGeometryPlacements()`...)
  - [x] Impulse dynamics
  - [x] Impulse dynamics derivatives
  - [x] Jacobian
  - [x] Delassus
  - [x] Kinematic regressor
  - [x] Kinematics
  - [x] Kinematics derivatives
  - [x] Regressor
  - [x] Model (`appendModel()`, `buildReducedModel()`, `findCommonAncestor()`, `transformJointIntoMimic()`)
  - [x] Energy
  - [x] Frames (`updateFramePlacement`, `framesForwardKinematics`...)
  - [x] Frames derivatives
  - [x] RNEA
  - [x] RNEA derivatives
  - [x] Parallel algorithms:
    - [x] Parallel ABA
    - [x] Parallel RNEA
    - [ ] bind `OpenMPException` exception **OPTIONAL**
    - [x] Util (`omp_get_max_threads()`)
- [ ] Collision
  - [x] Broadphase
  - [x] Collision (`expose-collision.cpp` in old bindings - `class ComputeCollision`, functions `computeCollision[s]()`...)
  - [ ] "Coal
    - [x] Conversion from coal.Transform3s to pin.SE3 (**REVERSE NOT POSSIBLE**)
    - [ ] Serialization
  - [x] Pool
    - [x] `GeometryModel`/`GeometryData` pool
    - [x] Broadphase manager pool
- [ ] Parsers
  - [x] URDF
    - [x] Model
    - [x] Geometry
  - [x] SDF
  - [x] SRDF
  - [x] MJCF
  - [x] Graph
- [x] Constraints 
  - [x] Generic
  - [x] Variant + base visitor
  - [x] Inheritance visitors (subclasses of `BinaryKinematicsConstraint{Model,Data}Base`, `PointConstraint{Model,Data}Base`)
  - [x] Derived classes
  - [x] Cones and sets (Coulomb, Dual Coulomb, Box, zero, full space, non-negative orthant)
- [ ] Utilities
  - [x] Type caster for `boost::variant` (done in `pinocchio/bindings/python-nb/utils/boost-variant.hpp`)
  - [x] Convert classes (mostly `pinocchio/spatial`) to numpy arrays using `__array__()` method, allows
    ```python
    f = pin.Force()
    f_arr = np.array(f, copy=...)
    ```
- [x] Custom call policy to disallow the use of `JointMimic`. [nanobind has call policies](https://nanobind.readthedocs.io/en/latest/api_core.html#_CPPv4I0EN8nanobind11call_policyE).

- [x] Serialization (pretty much everything missing)

- [x] Add Pixi feature and environment
- [x] Stubs (working)
 
### Known limitations

- This PR only adds nanobind bindings for the `double` scalar type (perhaps float).
- There is no equivalent to the `boost::python::convert::implicit` struct, which allows to define custom implicit converters between types. There are two avenues for creating implicit converters in nanobind:
  1. {constructor + declare `nb::implicitly_convertible`/use `nb::init_implicit`},
  2. creating a type caster.
  This allows defining the `coal.Transform3s` -> `pinocchio::SE3` conversion by registering an extra ctor, but **the reverse is not possible**.

### Other modified

- `pinocchio/visualize/meshcat_visualize.py` - remove import of `DeprecatedWarning`, replaced by Python built-in `DeprecationWarning`

### Other removed

- `pinocchio/deprecation.py` - only used in `meshcat_visualizer.py` through the `DeprecatedWarning`
- completely unused header `pinocchio/bindings/python/multibody/joint/joint.hpp`

### Upstream changes

- https://github.com/wjakob/nanobind/pull/1320 - Add support for `Eigen::Tensor` to nanobind :smile: 
- https://github.com/coal-library/coal/pull/833 - Fix problems accessing `geometry.convex` on some Coal BVH types

### Notes

- In some of the new Pinocchio 4 stuff we use `std::optional` whereas we use `boost::optional` elsewhere (for the bindings mostly):
  - some of the (member) function/ctor overloads in the Pinocchio C++ source currently exist because we _couldn't_ bind `boost::optional`:
  - we can now in Boost.Python/eigenpy since a while (I added it)
  - we should modernize the source code to use `std::optional`, and use it everywhere instead of `boost::optional`
  - nanobind 
- Similarly, we should start using `std::string_view` where appropriate
- The `__repr__` method nanobind implements when using `nb::bind_vector` to bind STL vectors uses the elements' `__repr__`, so a vector of `pinocchio.Model` has a **VERY** long representation.
    - We should distinguish `__repr__` from `__str__` on some classes.
- I tried to bind the `OpenMPException` exception in nanobind, inheriting from the mainline Python Exception class using:
  ```cpp
  nb::type<OpenMPException>(m, "OpenMPException", PyExc_Exception);
  ```
  but the exception class doesn't inherit from one of the STL exception base class and doesn't implement the member function `::what()`
- Conversion to NumPy arrays using `__array__` was factored into a new visitor (`ToNumpyArrayVisitor`). Properly supports copies and kwargs now.
- Nanobind is **stricter** than eigenpy was regarding conversion of nominally 2D (but actually 1D) `(n,1)` NumPy arrays to Eigen types: it's not allowed, which means code as follows:
  ```python
  model = ...
  qmax = np.ones((model.nq, 1))
  q0 = pin.randomConfiguration(model, -qmax, qmax)  # TypeError in NumPy → Eigen caster
  ```

Resolves #2616


ManifoldFR:
Thanks a lot for continuing my work on this @j-matheron ! 🙂

bindings/nanobind/expose-coal.cpp
```diff
@@ -18,6 +18,16 @@ void exposePoolCollision(nb::module_ m);
...
+  // Implicit conversion invokes the destination's Python constructor.
+  // Register these after both libraries have exposed their transform types.
+  nb::borrow<nb::class_<coal::Transform3s>>(nb::type<coal::Transform3s>())
+    .def("__init__", [](coal::Transform3s * self, const pinocchio::SE3 & placement) {
+      new (self) coal::Transform3s(placement.rotation(), placement.translation());
+    });
+  nb::borrow<nb::class_<pinocchio::SE3>>(nb::type<pinocchio::SE3>())
+    .def("__init__", [](pinocchio::SE3 * self, const coal::Transform3s & placement) {
+      new (self) pinocchio::SE3(placement.getRotation(), placement.getTranslation());
+    });
```
jorisv, line 32: SE3 -> Transform3s conversion is already defined in spatial/se3.hpp
Same thing for implicit conversion in this side.

Before calling nb::type<coal::Transform3s> I think we have to import coal or you have to check nb::type return a valid handle (is_valid()).

jorisv: @lmontaut

PR #2878, opened 2026-04-26
viser: apply <mesh scale> to vertices instead of translation

tingfan:

## Description

Fixes rendering of URDF meshes with a non-unit `<mesh scale="..."/>` attribute in the Viser visualizer.

**Before:** `ViserVisualizer._add_mesh_from_path` loaded mesh vertices at their native scale and ignored `GeometryObject.meshScale`. To compensate, `updatePlacements` multiplied the link translation by `meshScale` (`frame.position = M.translation * geometry_object.meshScale`). That is incorrect — the per-geometry scale is a *mesh* property, not a kinematic one. The net effect was: meshes rendered at the wrong size *and* wrong position, and different geometries on the same link were shifted apart from each other.

**After:** `meshScale` is baked into the mesh vertices at load time (matching how the `.dae` / `trimesh.Scene` branch works), and the link translation is passed through untouched. Viser scene nodes don't support a per-node non-uniform scale, so baking the scale into the vertices is the right place for it.

Added a regression test (`unittest/python/bindings_viser_visualizer.py`) that builds a one-link URDF with `<mesh scale="2 3 4"/>` at translation `(1, 2, 3)` and checks both:
- the scale is forwarded to `_add_mesh_from_path`, and
- after `display(q)`, the viser frame position equals `(1, 2, 3)` — not `(2, 6, 12)` as the old code produced.

No related issue on the tracker — happy to open one if preferred.

## Checklist

- [ ] I have made corresponding changes to the doxygen documentation *(N/A — Python-only internals, no doxygen affected)*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Fix Viser visualizer: apply URDF `<mesh scale>` to mesh vertices instead of scaling the link translation ([#2878](https://github.com/stack-of-tasks/pinocchio/pull/2878))
+
 - Fix build issue with g++ 12 ([#2890](https://github.com/stack-of-tasks/pinocchio/pull/2890))
 - Fix useless memory allocation in Python checkers ([#2897](https://github.com/stack-of-tasks/pinocchio/pull/2897))
 
 ### Changed
 
 - nix: switch to flakoboros ([#2882](https://github.com/stack-of-tasks/pinocchio/pull/2882))
+
 
 ## [4.0.0] - 2026-04-13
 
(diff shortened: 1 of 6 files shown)


j-matheron (changes requested):
Hello tingfan, thank you for the PR.
 
Would it be possible to change the function `_add_mesh_from_path` in order to dissociate the scaling handling from the mesh adding ? This atomic function could be named "handle_mesh_scaling" and be called by `_add_mesh_from_path`.
It would allow to test an atomic function without calling viser. Doing so the scaling could be testing instead of parameter forwarding. 
The test `test_scale_is_forwarded_to_mesh_loader` would be replaced by this function testing.

Thank you !

j-matheron:
Hello tingfan, 
will you have time to go through this PR ? Otherwise, I could take care of it

sea-bass:
For what it's worth, I think the requested change is too much abstraction for something as simple as this, and I would take it as-is, with one exception. Instead of doing the scaling manually, the `trimesh` scene objects already have an `apply_scale` method that can be called.

EDIT: Although I now am reading again that the abstraction is for the purpose of simplifying the unit test... hmm...

See my PR in https://github.com/stack-of-tasks/pinocchio/pull/2903#issuecomment-4616520500 and feel free to take from there and finish up this PR. This one has testing and more helpful comments than mine.

jorisv (approved):

j-matheron (approved):

PR #2779, opened 2025-10-03
Fix Eigen5 support on Pinocchio 3

jcarpent:
Thanks for taking the time to make and fill out this pull request!

## Checklist

- [x] I have run `pre-commit run --all-files` or `pixi run lint`
- [x] I have performed a self-review of my own code
- [ ] I have commented my code where necessary
- [ ] I have made corresponding changes to the doxygen documentation
- [x] I have added tests that prove my fix or feature works
- [x] I have updated the [CHANGELOG](https://github.com/stack-of-tasks/pinocchio/blob/devel/CHANGELOG.md) or added the "no changelog" label if it's a CI-related issue
- [ ] I have updated the [README credits section](https://github.com/stack-of-tasks/pinocchio?tab=readme-ov-file#credits)

PR #2886, opened 2026-05-12
Eigen 5 support

jorisv:
## Description

This PR take some elements from #2779.

## TODO

- [x] Casadi
- [x] cppad
- [x] MPFR
- [x] cppdacodegen
  - [x] Fix cppadcodegen
- [x] Fix test on all platforms

## Checklist


cmastalli:
It seems to me that this will be useful to fix issues occurring in https://github.com/loco-3d/crocoddyl/actions/runs/26946841274/job/79501813774?pr=1506.

Can you confirm this belief?

jorisv:
@cmastalli This PR only fix build issues with Eigen5 and some numerical precision issue in tests.

Eigen5 operations doesn't have the same numerical precision than Eigen3. So you will probably need to play with precision to make your tests pass.

cmastalli:
> @cmastalli This PR only fix build issues with Eigen5 and some numerical precision issue in tests.
> 
> Eigen5 operations doesn't have the same numerical precision than Eigen3. So you will probably need to play with precision to make your tests pass.

This is indeed what I'm experiencing. Thanks for the info!

jorisv:
Windows with MSVC compiler and one OSX build is no more passing because of too long compilation time.
This only happen when activating all pinocchio feature.
It tested locally for Windows and all tests are green.
On osx-intel, since other configuration are building well, I think this is not hiding any more issue.

include/pinocchio/src/algorithm/constrained-dynamics.hxx
```diff
@@ -440,35 +440,45 @@ namespace pinocchio
...
-      const bool convergence_criteria_reached =
-        check_expression_if_real<Scalar, false>(
-          settings.absolute_residual <= settings.absolute_accuracy)
-        || check_expression_if_real<Scalar, false>(
-          settings.relative_residual <= settings.relative_accuracy);
-      if (convergence_criteria_reached) // In the case where Scalar is not double, this will iterate
-                                        // for max_it.
-        break;
+      settings.absolute_residual = 0;
+      settings.relative_residual = 0;
```
j-matheron, line 481: Do we want theses variables equals to zero when there is no constraint ?

j-matheron (commented):
For the include removals, it is because of IWYU recommandations ?

include/pinocchio/src/algorithm/constrained-dynamics.hxx
jorisv: It's the value assign to them when there is no constraints

jorisv:
> For the include removals, it is because of IWYU recommandations ?
It's because compilation issue when removing the PCH file.
We didn't run IWYU on bindings.

j-matheron (approved):

PR #2889, opened 2026-05-18
Add Homebrew macOS CI

ahoarau:

## Description

This PR adds a CI for macOS, where we get the dependencies via Homebrew and run plain CMake.
This can serve as instructions for people wanting to build `pinocchio` without `conda/pixi`.

## Checklist

### DIFF
--- a/.github/workflows/macos-homebrew.yml
+++ b/.github/workflows/macos-homebrew.yml
@@ -1,0 +1,99 @@
+name: CI - macOS via Homebrew
+
+on:
+  push:
+    branches:
+      - devel
+    paths-ignore:
+      - doc/**
+      - .gitlab-ci.yml
+      - .gitignore
+      - "**.md"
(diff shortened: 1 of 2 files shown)


jorisv:
Hello,
Right now we doesn't want to support another package manager in our CI.
Homebrew has his own CI/build system, no need to test our compatibility with it here.

ahoarau:
Got it. If `brew` is officially supported, would the instructions be useful at least ?

jorisv:
It's not about being officially supported. It's about being used by core developers :
- conda is used by pinocchio core developers
- nix is used by pinocchio core developers and the CI is filling cachix
- Ubuntu test if we break compatibility with LTS (minimal supported tools and library version are defined by LTS)
- ROS is used by pinocchio core developers (also, providing a patch to ROS CI after release is a pain, so better catch it early)

PR #2900, opened 2026-06-02
fix: use portable Windows detection macro for compilation on MSVC

onurtuncer:
Replace bare `#ifdef WIN32` / `#if WIN32` guards with `#if defined(WIN32) || defined(_WIN32)` in macros.hxx and alloca.hpp.

MSVC defines `_WIN32` (with leading underscore) but not necessarily the unadorned `WIN32`, so the old guards could silently fall through to the GCC/Clang paths (`__PRETTY_FUNCTION__`, alloca.h) on a plain MSVC build, causing compilation failures on Windows.

## Checklist


jorisv:
Hello @onurtuncer,

Thanks for the PR.

Pinocchio 3 was including `windows.h` which could defined `WIN32` internally.
We doesn't include this header anymore, that's maybe why this kind of issue appear.
On the behalf of you knowledge, did we have to keep WIN32 define or could we only use _WIN32 ?

I also found `WIN32` in the following files:
- include/pinocchio/src/macros.hxx (2 remaining)
- include/pinocchio/src/geometry/coal.hxx

jorisv:
Hello @onurtuncer,

After some discussion we will remove all `WIN32` to replace them by `_WIN32`. Do you want to apply the change in the whole project or can I take over your PR ?

onurtuncer:
@jorisv you can take over and apply accross the whole project. Thank you...

include/pinocchio/src/macros.hxx
```diff
@@ -151,7 +151,7 @@
       _Pragma("GCC diagnostic ignored \"-Wmaybe-uninitialized\"")
     #define PINOCCHIO_COMPILER_DIAGNOSTIC_IGNORED_SELF_ASSIGN_OVERLOADED
   #endif
-#elif defined(WIN32)
+#elif defined(WIN_32)
```
jorisv:
```suggestion
#elif defined(_WIN32)
```

jorisv (changes requested):

jorisv (dismissed):

jorisv (dismissed):

jorisv (approved):

PR #2915, opened 2026-06-29
CMake: dont require mathjax

nim65s:
## Description

Mathjax is nice, but I think it should not be activated *unconditionally*. 

I believe it is the responsibility of the package maintainer to decide if docs should be build with nice latex svgs, or if mathjax should be bundled.

Currently, even with `-DDOXYGEN_USE_MATHJAX=OFF`, I get 23MB in the `share/doc/pinocchio/doxygen-html/MathJax`, so it won't even be re-used by other packages and would have to be duplicated.

For this reason, I think it should be OFF (or not set, as in the current state of this PR) by default.

Actually, I do not see any reason to turn this ON unless the user is not able to use a proper package manager, which is hopefuly a very rare case nowadays.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ set(CMAKE_VERBOSE_MAKEFILE True)
 # Need to be set before including base.cmake
 # ----------------------------------------------------
 option(INSTALL_DOCUMENTATION "Generate and install the documentation" OFF)
+option(PINOCCHIO_DOXYGEN_USE_MATHJAX "Avoid latex at build time" ON)
 
 # Check if the submodule cmake have been initialized
 set(JRL_CMAKE_MODULES "${CMAKE_CURRENT_LIST_DIR}/cmake")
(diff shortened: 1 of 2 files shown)


nim65s:
pixi CI was not happy about this, so I turned this back on there:
```cmake
ninja: error: loading 'build.ninja': No such file or directory
CMake Error at build/cmake_install.cmake:70 (file):
  file INSTALL cannot find
  "/home/runner/work/pinocchio/pinocchio/build/doc/doxygen-html": No such
  file or directory.
```

nim65s:
Hum, I guess we'll have to baby-steps this :(

CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ else()
     endif()
 endif()
 
-set(DOXYGEN_USE_MATHJAX YES)
+option(DOXYGEN_USE_MATHJAX "Avoid latex at build time" ON)
```
jorisv:
```suggestion
option(PINOCCHIO_DOXYGEN_USE_MATHJAX "Avoid latex at build time" ON)
```
We should prefix new option name by project name to avoid option name clash.

jorisv (changes requested):

jorisv (approved):

PR #2932, opened 2026-08-11
bindings/python: add float32 bindings

Sergim96:
## Description

This PR adds core single-precision Python bindings exposed through the new
`pinocchio.float32` module.

The change:

- adds a Python binding context using `float` as its scalar type;
- exposes `numpy.float32` as `pinocchio.float32.ScalarType`;
- makes sample-model factories and affected model/frame algorithms use the
  active binding scalar instead of hard-coded double types;
- adds float32-specific `exp`, `log`, and `utils` helpers;
- registers native submodules under `pinocchio.float32.*`;
- fixes the scalar-generic model, frame, LCABA, and ADMM implementations needed
  by the bindings;
- integrates the extension into the CMake build and installation, controlled by
  `BUILD_PYTHON_BINDINGS_WITH_FLOAT32_SUPPORT`;
- adds the extension to the stub-generation dependencies;
- adds regression coverage for RNEA, sample models, model and frame algorithms,
  helper dtypes, submodule imports, LCABA, ADMM, and collision-enabled sample
  geometry.

This allows Pinocchio's core kinematics and dynamics APIs to operate directly
on single-precision NumPy arrays.

Geometry and collision objects retain their existing double-backed
representation. File parsers, parallel algorithms, and reachable-workspace
helpers remain available only through the main module.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 
 ### Added
+- Add core single-precision Python bindings through the `pinocchio.float32` module
 - Add `DataTpl::lastChild` deprecation notice in Python binding
 - Fix `addFrame` to ignore frame without inertial to preserse parent body's CoM
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,6 +188,13 @@ cmake_dependent_option(
     OFF
 )
 cmake_dependent_option(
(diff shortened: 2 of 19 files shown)


jcarpent (changes requested):
Thanks @Sergim96 for this PR.
Could you split it into two parts: 
- a first PR where you have fixed the bindings (adding context, Scalar, etc.).
- a second one for support of float32?

Thanks in advance,
Best,
Justin

Sergim96:
Hi @jcarpent, 

I have created https://github.com/stack-of-tasks/pinocchio/pull/2939 I will open the second one once the first one is merged if you agree.

We can close this one

Best,
Sergi

PR #2935, opened 2026-08-20
example: fixing ellipsoid-joint python example

Ipuch:
The  ellipsoid joint python example wasn't running anymore.

pin.GeometryObject in pinocchio 4 takes (name, parent_joint, placement, collision_geometry) as arguments, whereas the pinocchio 3 call sites passed geometry and placement in the opposite order, which raised a Boost.Python.ArgumentError.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add `DataTpl::lastChild` deprecation notice in Python binding
 - Fix `addFrame` to ignore frame without inertial to preserse parent body's CoM
 
+### Fixed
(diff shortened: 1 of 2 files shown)


jcarpent (dismissed):
Have you run pre-commit locally?

Ipuch:
Indeed, my python black linter interfered. 

I've run pre-commit again despite some weird line breaks in the python scripts, for example:
```python
q_composite[:3] = data.oMi[
        1
    ].translation  # Set translation to current position on ellipsoid
```

jorisv:
Hello,

We didn't catch this regression because visualization code is between a big try/except block.
Do you know why this is done this way ? If there is no particular reason, I think it's best to delete it.

jorisv (dismissed):

jorisv (approved):

PR #2943, opened 2026-08-28
Adding a constant length constraint to the constraint collection

Ipuch:
A new constraint introduced to pinocchio: `ConstantLengthConstraintModel`, which enforces a constant distance between two points. The first point is rigidly attached to `joint1`, the second one to `joint2`, and denoting by $x(q)$ the position of the second point expressed **in the frame of the first one**, the residual:

$$\varphi(q) = \| x(q) \| - L - \delta$$

with $L$ the constant length and $\delta$ the `desired_constraint_offset` inherited from the binary kinematics constraints. This is what I need to model a cable, a tendon, a rigid rod, or the coupler bar of a closed loop. It is useful for biomechanics for ligament of constant length, or clavicle bone modelling.

The residual is a length (in meters), not a squared length: on the manifold $\|\nabla\varphi\| = 1$, so the error is the signed distance to the sphere.

A few points worth mentioning for the review:

* This is the **first scalar constraint** (`Size == 1`). Eigen refused a column major  `Matrix<Scalar, 1, N>` with `N != 1`, so `MatrixSize6` in  `binary-kinematics-constraint-model-base.hxx` now switches its storage order to `Eigen::RowMajor`  when `Size == 1`. This is the only modification made to existing code. It may be interesting for upcoming constraints I would like to implement such as: point on a Ellipsoid.
* The constraint derives from `BinaryKinematicsConstraintModelBase` rather than from `PointConstraintModelBase`, since the residual is a scalar function of the relative position:  $\varphi = f(x)$ with $f(x) = \|x\| - L$.
* The constraint is singular when the two points are coincident ($\|x\| = 0$). This is guarded by  an `assert(check_expression_if_real<Scalar>(...))` inside `calc`, not by a throw, so that the  CasADi and CppAD scalars keep compiling.
* The length is stored as a dedicated `m_length` with `getLength()`/`setLength()` accessors rather than being folded into `desired_constraint_offset`, so that both remain usable independently.

The constraint is registered in the default constraint collection (`ConstraintModelTpl`/`ConstraintDataTpl` variants), serialized, and exposed to Python.

## Examples

Two examples visualized with Viser come with the PR, we can keep only one of them if you prefer.

* `examples/constant-length-constraint-kinematics.py` — a **four-bar linkage** modelled as an open kinematic tree: a crank and a rocker pinned to the ground, and the coupler replaced by a single scalar constraint. 2 dof − 1 constraint = 1 dof.

https://github.com/user-attachments/assets/979273a8-d668-4de7-a50d-4a3950bb9bf1

* `examples/constant-length-constraint-3d.py` — a **spherical pendulum**: a ball carried by a translation joint, tethered to a fixed anchor by a cable of constant length. The constrained forward dynamics is formed as a KKT system, which uses the three quantities `calc()` stores in the constraint data. With a Baumgarte correction and `dt = 1e-3`, the cable violation stays below `1e-3 m` over 12 s of simulation.

https://github.com/user-attachments/assets/9a2ba0c4-b7b0-496c-9a9b-3fcfa86bfb95

  
## Tests

`unittest/constant-length-constraint.cpp` is modelled on `unittest/point-anchor-constraint.cpp`. Everything runs on `humanoidRandom` with a free flyer root, both points being carried by moving bodies, and in two configurations: one anchored on the universe, one closing a loop between the two legs. I've tested, Residual, Projectors, Jacobian, Velocity and acceleration errors, Force and motion mappings.

## Checklist


CHANGELOG.md
```diff
@@ -11,7 +11,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 
 ### Added
-- Add `DataTpl::lastChild` deprecation notice in Python binding ([#2928](https://github.com/stack-of-tasks/pinocchio/pull/2927))
+- Add `ConstantLengthConstraintModelTpl`/`ConstantLengthConstraintDataTpl`, a scalar constraint enforcing a
+  constant distance between two material points. It is part of the default constraint collection
+  (`ConstraintModelTpl`/`ConstraintDataTpl` variants) and is exposed to Python, along with two
+  examples visualized with Viser: `examples/constant-length-constraint-kinematics.py` (constrained
+  forward kinematics of a four-bar linkage) and `examples/constant-length-constraint-3d.py`
+  (constrained forward dynamics of a spherical pendulum)
+- Add `DataTpl::lastChild` deprecation notice in Python binding
```
jorisv: Put back the PR link

CHANGELOG.md
```diff
@@ -11,7 +11,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 
 ### Added
-- Add `DataTpl::lastChild` deprecation notice in Python binding ([#2928](https://github.com/stack-of-tasks/pinocchio/pull/2927))
+- Add `ConstantLengthConstraintModelTpl`/`ConstantLengthConstraintDataTpl`, a scalar constraint enforcing a
+  constant distance between two material points. It is part of the default constraint collection
+  (`ConstraintModelTpl`/`ConstraintDataTpl` variants) and is exposed to Python, along with two
+  examples visualized with Viser: `examples/constant-length-constraint-kinematics.py` (constrained
+  forward kinematics of a four-bar linkage) and `examples/constant-length-constraint-3d.py`
+  (constrained forward dynamics of a spherical pendulum)
```
jorisv: Add a link to the PR.

unittest/constant-length-constraint.cpp
```diff
@@ -0,0 +1,556 @@
...
+  {
+    // check set compliance
+    const Eigen::VectorXd compliance_ref =
+      Eigen::VectorXd::Random(cmodel.residualSize()).cwiseAbs();
+    cmodel.setCompliance(compliance_ref);
+    Eigen::VectorXd compliance(cmodel.residualSize());
+    cmodel.retrieveCompliance(compliance);
+    BOOST_CHECK(compliance == compliance_ref);
+  }
+}
```
jorisv: m_compliance is managed by parent class. Why testing it ?

unittest/constant-length-constraint.cpp
```diff
@@ -0,0 +1,556 @@
...
+
+BOOST_AUTO_TEST_CASE(variant)
```
jorisv: I don't think it worth to test if boost::visitor is working fine.

unittest/constant-length-constraint.cpp
```diff
@@ -0,0 +1,556 @@
...
+
+  Data::MatrixXs J(1, model.nv), J_(1, model.nv);
+  J.setZero();
+  J_.setZero();
+  cmodel.jacobian(model, data, cdata, J);
+  cmodel_.jacobian(model, data, cdata_, J_);
+  BOOST_CHECK(J.isApprox(J_));
+}
+
+BOOST_AUTO_TEST_CASE(cholesky)
```
jorisv: If we already tested jacobian and getRowSparsityPattern, what the purpose of this test ?

unittest/constant-length-constraint.cpp
jorisv: Some test look a little redundant.
- `rigid_displacement_invariance` test a property but it's really general.
- Do checking acceleration error is affine worth it after testing with finite difference ?

jorisv:
Hello @Ipuch,

When reading this PR it look that constrant-length and point constraints are really similar but it's not possible to inherit from `PointAnchorConstraintModelBase` to factorize the code.

An other way is to use composition by storing a PointAnchorConstraintModel inside a ConstantLengthConstraintModel and to post process output. Like done in the unit test.
It can be less efficient and will store some extra memory (we will need to store a 3 x dof jacobian in the data).

An other way is to create dedicated functions that could be called by both model with a functor as argument to adapt the function to the model.

include/pinocchio/src/constraints/constant-length-constraint.hxx
```diff
@@ -0,0 +1,1173 @@
...
+        ResidualVectorType cinertia;
+        constraint_inertia.diagonal(cinertia);
+        appendConstraintInertias(model, data, cdata, cinertia[0], reference_frame);
+        break;
+      }
+      case internal::MatrixBlockType::Plain: {
+        MatrixSize cinertia;
+        constraint_inertia.matrix(cinertia);
+        appendConstraintInertias(model, data, cdata, cinertia(0, 0), reference_frame);
+        break;
```
jorisv, line 861: Since appendConstraintInertias take a scalar, I think Diagonal and Plain are equivalent to 
```
        appendConstraintInertias(
          model, data, cdata, constraint_inertia.container()(0, 0), reference_frame);
```

unittest/constant-length-constraint.cpp
```diff
@@ -0,0 +1,556 @@
...
+
+  // The generic offset shifts the residual, exactly as the length does
+  cmodel.setLength(distance_ref);
+  cmodel.desired_constraint_offset[0] = 0.01;
+  cmodel.calc(model, data, cdata);
+  BOOST_CHECK_CLOSE(cdata.constraint_position_error[0], -0.01, 1e-8);
+}
+
+BOOST_AUTO_TEST_CASE(rigid_displacement_invariance)
+{
```
jorisv, line 179: It's nice to test property, but I don't think this test cover case missed by previous one.

unittest/constant-length-constraint.cpp
```diff
@@ -0,0 +1,556 @@
...
+  {
+    Data data_zero_acc(model);
+    forwardKinematics(model, data_zero_acc, q, v, VectorXd::Zero(model.nv));
+
+    ConstantLengthConstraintData cdata_zero_acc(cmodel);
+    cmodel.calc(model, data_zero_acc, cdata_zero_acc);
+
+    BOOST_CHECK((J * a + cdata_zero_acc.constraint_acceleration_error)
+                  .isApprox(cdata.constraint_acceleration_error));
+  }
```
jorisv, line 380: I think FD is already covering the correctness of constraint_acceleration_error.

unittest/constant-length-constraint.cpp
```diff
@@ -0,0 +1,556 @@
...
+
+  const auto cmodel_cast_double = cmodel.cast<double>();
+  BOOST_CHECK(cmodel_cast_double == cmodel);
+
+  const auto cmodel_cast_long_double = cmodel.cast<long double>();
+  BOOST_CHECK(cmodel_cast_long_double.getLength() == 0.75L);
+  BOOST_CHECK(cmodel_cast_long_double.cast<double>() == cmodel);
+}
+
+BOOST_AUTO_TEST_CASE(compliance)
```
jorisv, line 454: I don't get the purpose of this test. Compliance related methods are implemented by `ConstraintModelCommonParameters` that `ConstantLengthConstraint` inherit by `BinaryKinematicsConstraintModelBase`.

unittest/constant-length-constraint.cpp
jorisv, line 479: This test check if ConstantLengthConstraintModel can be converted to a generic ConstraintModel. This test we doesn't forget to add ConstantLengthConstraintModel to the join collection.
I think this can be removed.

unittest/constant-length-constraint.cpp
jorisv, line 511: What the purpose of this test ? Checking all method have been implemented ?

jorisv (changes requested):

jorisv:
@Ipuch After discussing with others maintainers, no need to address the code factorization issue.
It will change too much the purpose of this PR. We will address it ourself if the number of kinematics constraints increase.

unittest/constant-length-constraint.cpp
Ipuch: Removed.

unittest/constant-length-constraint.cpp
Ipuch: Removed.

unittest/constant-length-constraint.cpp
Ipuch: Removed. because finite-difference tests do the job.

unittest/constant-length-constraint.cpp
Ipuch: Removed.

include/pinocchio/src/constraints/constant-length-constraint.hxx
Ipuch: Modified.

unittest/constant-length-constraint.cpp
Ipuch: Removed anyway. It checked that the constraint assembles correctly in ConstraintCholeskyDecomposition
against a hand-built KKT matrix. Everything it relies on (Jacobian, sparsity, products) is already
tested directly, so it added nothing specific to this constraint.

CHANGELOG.md
```diff
@@ -15,6 +15,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add `DataTpl::lastChild` deprecation notice in Python binding ([#2928](https://github.com/stack-of-tasks/pinocchio/pull/2927))
 - Add spline joint (JointModelSplineTpl) to default joint collection ([#2784](https://github.com/stack-of-tasks/pinocchio/pull/2784))
   - Python example [here](./examples/spline-joint.py) and [here](./examples/spline-joint-knee.py)
+- Add `ConstantLengthConstraintModelTpl`/`ConstantLengthConstraintDataTpl`, a scalar constraint enforcing a
+  constant distance between two material points. It is part of the default constraint collection
+  (`ConstraintModelTpl`/`ConstraintDataTpl` variants) and is exposed to Python, along with two
+  examples visualized with Viser: `examples/constant-length-constraint-kinematics.py` (constrained
+  forward kinematics of a four-bar linkage) and `examples/constant-length-constraint-3d.py`
+  (constrained forward dynamics of a spherical pendulum)
+- Add `DataTpl::lastChild` deprecation notice in Python binding
```
jorisv:
```suggestion
```
Look like a rebase conflict has been miss handled

examples/constant-length-constraint-3d.py
```diff
@@ -0,0 +1,199 @@
...
+data = model.createData()
+
+# Set the constraint: the anchor belongs to the universe (joint 0), the other point is
+# the centre of the ball.
+placement_ball = pin.SE3.Identity()
+constraint_model = pin.ConstantLengthConstraintModel(
+    model, 0, placement_anchor, ball_id, placement_ball, length_cable
+)
+constraint_model.name = "cable"
+constraint_model.baumgarte_corrector_parameters.Kp = 1e4
```
jorisv: This look big, did you try with inferior value ?

examples/constant-length-constraint-3d.py
```diff
@@ -0,0 +1,199 @@
...
+constraint_model.baumgarte_corrector_parameters.Kd = 2.0 * np.sqrt(1e4)
+constraint_data = constraint_model.createData()
+
+Kp = constraint_model.baumgarte_corrector_parameters.Kp
+Kd = constraint_model.baumgarte_corrector_parameters.Kd
```
jorisv:
```suggestion
Kp = 1e4
Kd = 2.0 * np.sqrt(Kp)

constraint_model.baumgarte_corrector_parameters.Kp = Kp
constraint_model.baumgarte_corrector_parameters.Kd = Kd
constraint_data = constraint_model.createData()
```

examples/constant-length-constraint-3d.py
```diff
@@ -0,0 +1,199 @@
...
+        print(f"{k * dt:8.2f}   {abs(phi):9.2e}   {tension:11.3f}   {energy:10.4f}")
+
+pin.forwardKinematics(model, data, q, v)
+energy = pin.computeKineticEnergy(model, data, q, v) + pin.computePotentialEnergy(
+    model, data, q
+)
+
+print(f"\nworst cable violation over {n_steps * dt:.0f} s : {worst_violation:.3e} m")
+print(f"energy drift (semi-implicit Euler): {energy - energy0:+.3e} J")
+assert worst_violation < 1e-3, "the cable drifted away"
```
jorisv:
```suggestion
```

It's not a TU

examples/constant-length-constraint-kinematics.py
```diff
@@ -0,0 +1,214 @@
...
+    worst_residual = max(worst_residual, abs(residual))
+
+    if len(configurations) % (n_steps // 8) == 1:
+        print(
+            f"{np.rad2deg(theta):9.1f}   {np.rad2deg(q[rocker_idx_v]):12.1f}"
+            f"   {it:9d}   {abs(residual):.2e}"
+        )
+
+print(f"\nworst residual over the sweep: {worst_residual:.3e} m")
+assert worst_residual < 1e-10, "the loop did not close"
```
jorisv:
```suggestion
```

jorisv (changes requested):

jorisv:
When last remark are addressed it's all good to me.
@lmontaut @ymontmarin I wait your review before merging.

examples/constant-length-constraint-3d.py
Ipuch: I've tried with zero and I get a maximum cable violation of 2.823e-02 m. with the current setting it's 1.187e-04 m. That's why I kept it.

jorisv: Usually we setup Kp to 10 or 20. A too large value can create some discontinuity in the acceleration vector computed as the baumgarte will be really stiff. It's not an issue here because there is only one constraint, but it can be on more complicated setting.
With a Kp of 20 I obtain a maximum violation of 1.720e-03 m.

Ipuch: Understood, I can change it, it's just an example to me.

jorisv: If it's fine you can setup it to 20.

Ipuch: Done.

examples/constant-length-constraint-3d.py
```diff
@@ -0,0 +1,199 @@
...
+    # constrained points, and redrawn at every frame.
+    p_anchor = placement_anchor.translation
+    p_ball = (viz.data.oMi[ball_id] * placement_ball).translation
+    viz.viewer["cable"].set_object(
+        mg.LineSegments(
+            mg.PointsGeometry(np.stack([p_anchor, p_ball]).astype(np.float32).T),
+            mg.LineBasicMaterial(color=CABLE_COLOR, linewidth=4.0),
+        )
+    )
+    time.sleep(*n_display)
```
jorisv: it was dt*n_display before. It's normal you removed it ?

Ipuch: no, yes, i've edited the script through github interface in commit "example(constant-length): fixing Kp to 20". I must have deleted it with an uncontrolled "Ctrl+F". Sorry. I'll squash these commits.

jorisv (dismissed):

lmontaut:
Hey @Ipuch, thanks for the PR!
One thing you could do is modify `unittest/utils.hpp` to create a `ConstantLengthConstraintModel` under a `has_constant_length` boolean. This would allow the `delassus-operations` test to automatically verify that your constraint is working well with Pinocchio's delassus operators. Notably, your constraint's `appendCouplingConstraintInertias` would now be tested (typically needed when running a constraint solver + rigid body delassus backend in C++).

Otherwise it lgtm :)

jorisv (approved):

lmontaut (approved):

Ipuch:
can we rerun the failed job ?

